PR #7, opened 2025-04-09
docs: update ToolboxTool docstring

jackwotherspoon:
Update port for `ToolboxTool` to match default port used for Toolbox server of `5000`.

<s>`ToolboxTool` is missing from `tools.__init__.py`:

https://github.com/google/adk-python/blob/9002ace7e58d8a1e19e8b60f7119167baa6ce4d0/src/google/adk/tools/__init__.py#L34-L51

This means users need to use the full import path for `ToolboxTool` .

```python
from google.adk.tools.toolbox_tool import ToolboxTool
```

This also means `ToolboxTool` will be excluded from being imported when users import all tools.

```python
from google.adk.tools import *
```

Adding `ToolboxTool` to `tools.__init__.py` makes it easier for customers to leverage MCP Toolbox for databases.

```python
from google.adk.tools import ToolboxTool
```

Once this change is released we ca update the `ToolboxTool` imports in the docs back to their [original form](https://github.com/google/adk-docs/pull/22/files#diff-5528892b5f4abe5e07be11dcb0e00d04c5740fb9ae047aff7e50ebfe3df2d8eeL275).
</s>

### DIFF
--- a/src/google/adk/cli/utils/evals.py
+++ b/src/google/adk/cli/utils/evals.py
@@ -66,7 +66,7 @@ def convert_session_to_eval_format(session: Session) -> list[dict[str, Any]]:
                 'tool_input': tool_input,
             })
           elif subsequent_part.text:
-            # Also keep track of all the natural langauge responses that
+            # Also keep track of all the natural language responses that
             # agent (or sub agents) generated.
             intermediate_agent_responses.append(
                 {'author': event_author, 'text': subsequent_part.text}
@@ -75,7 +75,7 @@ def convert_session_to_eval_format(session: Session) -> list[dict[str, Any]]:
       # If we are here then either we are done reading all the events or we
       # encountered an event that had content authored by the end-user.
       # This, basically means an end of turn.
-      # We assume that the last natural langauge intermediate response is the
+      # We assume that the last natural language intermediate response is the
       # final response from the agent/model. We treat that as a reference.
       eval_case.append({
           'query': query,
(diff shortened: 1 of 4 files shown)


src/google/adk/tools/toolbox_tool.py
```diff
@@ -23,7 +23,7 @@ class ToolboxTool:
 
   Example:
   ```python
-  toolbox = ToolboxTool("http://127.0.0.1:8080")
+  toolbox = ToolboxTool("http://127.0.0.1:5000")
```
jackwotherspoon, line 26: toolbox server default port is `5000` so updating it to match

https://github.com/googleapis/genai-toolbox/blob/0a7d3ff06b88051c752b6d53bc964ed6e6be400e/cmd/root.go#L106

pentium10:
how do you start this toolbox?

jackwotherspoon:
> how do you start this toolbox?

@pentium10 `ToolboxTool` is for loading tools for [MCP Toolbox for Databases](https://googleapis.github.io/genai-toolbox/getting-started/introduction/). You can find a walkthrough in the ADK docs [here](https://google.github.io/adk-docs/tools/google-cloud-tools/#3-toolbox-tools-for-databases).

You start it by [installing and running the toolbox server](https://googleapis.github.io/genai-toolbox/getting-started/introduction/#installing-the-server).

fhinkel:
Unfortunately, that would add toolbox_langchain as a dependency. But I like the fix for the port!

jackwotherspoon:
> Unfortunately, that would add toolbox_langchain as a dependency. But I like the fix for the port!

Ahh great point. I had not noticed `toolbox_langchain` was not a core dep 🙃

Jacksunwei (approved):

josegonzalesjjg (approved):

josegonzalesjjg (approved):

PR #22, opened 2025-04-10
Fix: reponse -> response

Howe829:
image
Hi there! The reason why I changed the main-SLIAU2JL.js file is it has a typo as the pic shows above.


Jacksunwei (approved):
LGTM

src/google/adk/cli/browser/main-SLIAU2JL.js
Jacksunwei, line 1: Can we revert this files, since it's compiled version? We don't manually change this file.

Jacksunwei (changes requested):

src/google/adk/cli/browser/main-SLIAU2JL.js
hangfei: other changes are good. but let's revert this file. we will fix it. thanks.

Howe829: ok, I've reverted it.

PR #38, opened 2025-04-10
Fix README Getting Started

Fuwad13:
Fix a small error in Getting Started section of README.md

The error: In `my_agent/agent.py` the Agent instance is named **root_agent** but in `my_agent/__init__.py` the import is wrong.


hangfei:
that imports the agent.py file. 

If you test with your change, it will fail.

PR #59, opened 2025-04-11
Fix agent validations to occur at startup

### LINKED ISSUE #57: Agent validations should happen at startup not when the user interacts with the agent for the first time
This should be checked earlier at startup not just at interacting with the agent.

  Value error, Found invalid agent name: `chat agent`. Agent name must be a valid identifier. It should start with a letter (a-z, A-Z) or an underscore (_), and can only contain letters, digits (0-9), and underscores. [type=value_error, input_value='c

It shouldn't just use the verbatim `agent_name` in a human interaction, there should be a `display_name` or so.

`You can call me chat_agent. I'm here to chat and help you out!`

also `adk web` 

should auto refresh on changes

ShyamSathish005:
Fixes #57

Added a display_name attribute to the BaseAgent class in src/google/adk/agents/base_agent.py for more readable agent identification.

Updated the model_post_init method to call __validate_name, enabling agent name validation during startup.

Added a test case in src/google/adk/tests/unittests/agents/test_base_agent.py to verify that display_name is set correctly.

Added a test case to ensure a validation error is raised if the agent name is invalid at startup.

### DIFF
--- a/src/google/adk/agents/base_agent.py
+++ b/src/google/adk/agents/base_agent.py
@@ -82,6 +82,9 @@ class BaseAgent(BaseModel):
   One-line description is enough and preferred.
   """
 
+  display_name: str = ''
+  """The display name of the agent for human interaction."""
+
   parent_agent: Optional[BaseAgent] = Field(default=None, init=False)
   """The parent agent of this agent.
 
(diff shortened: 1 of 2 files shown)


boyangsvl:
Thanks for the PR, ShyamSathish005. We do not add display_name because it's something that you can include in the agent instruction. For the name validation, self.__validate_name is called automatically by Pydantic. The issue you are seeing is because the agent object is only constructed when you send the first message.

ShyamSathish005:
Oh fine. Got it. Thanks

PR #64, opened 2025-04-11
Update issue templates

hangfei:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
```
Jacksunwei, line 30: would be better to python version (python -V), adk version (pip show google-adk)

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,23 @@
...
+labels: ''
+assignees: ''
+
+---
+
+** Please make sure you read the contribution guide and file the issues in the rigth place. **
+[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
Jacksunwei, line 14: Or an issue number or discussion number?

Jacksunwei, line 14: s/I'm always frustrated/

When [...], I want to [...]

Jacksunwei (approved):

.github/ISSUE_TEMPLATE/feature_request.md
hangfei: I think this might be optional. Don't want to provide too much restriction. We can add later if it's a common pattern.

hangfei: this is the default template. let's see how it goes and we can fine-tune later.

PR #77, opened 2025-04-11
Updated gemini-2.0-flash-exp(deprecated) with gemini-2.0-flash

blueviggen:
"gemini-2.0-flash-exp" (experimental model) is no longer valid as a model version. "gemini-2.0-flash" has replaced "gemini-2.0-flash-exp" (https://ai.google.dev/gemini-api/docs/models)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ from google.adk.tools import google_search
 
 root_agent = Agent(
     name="search_assistant",
-    model="gemini-2.0-flash-exp", # Or your preferred Gemini model
+    model="gemini-2.0-flash", # Or your preferred Gemini model
     instruction="You are a helpful assistant. Answer user questions using Google Search when needed.",
     description="An assistant that can search the web.",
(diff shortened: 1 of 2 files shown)


blueviggen:
fix #78

hangfei:
@polong-lin could you review this change?

polong-lin (approved):

polong-lin:
It's OK but I might recommend using `gemini-2.0-flash-live-preview-04-09` instead, which will support Live API and might provide a better user experience.

For text-only, then `gemini-2.0-flash` is OK, so will leave the decision up to @hangfei

hangfei:
Thanks @polong-lin .

Let's do this: 
# For live(streaming), use gemini-2.0-flash-live-preview-04-09.
model="gemini-2.0-flash-exp",

hangfei:
it's fixed in another CL. Closing this for now.

PR #88, opened 2025-04-11
Parse description field if return type of a function is a pydantic.BaseModel 

lienchen0526:
Parse description field if return type of a function is a pydantic.BaseModel

The usage will be like:

```python
from google.adk.tools._automatic_function_calling_util import build_function_declaration

class M(BaseModel):
  foo_field: int = Field(description = "This is a tested description")

def foo() -> M:
  pass

print(build_function_declaration(foo))
```

TESTED=manual


boyangsvl:
Thanks for your contribution, @lienchen0526 ! Could you please add some test cases in tests/unittests/tools/test_build_function_declaration.py?

hangfei:
closing due to inactivity.

Issue #49, opened 2025-04-10
Ollama support


curiousily:
I'm trying to use the `LiteLLM` (LLM?) with Ollama support like this:

```py
from google.adk.agents import Agent
from google.adk.models.lite_llm import LiteLlm

MODEL = "qwen2.5"


def get_weather(city: str) -> str:
    """Retrieves the current weather report for a specified city.

    Args:
        city (str): The name of the city (e.g., "New York", "London", "Tokyo").

    Returns:
        str: A report on the current weather in the specified city.
    """

    return f"It is always sunny in {city}!"


root_agent = Agent(
    name="weather_reporter",
    model=LiteLlm(model=f"ollama_chat/{MODEL}"),
    instruction="You are a helpful assistant. You always answer weather questions.",
    description="A helpful assistant that provides weather information.",
    tools=[get_weather],
)
```

When I send a message to the agent I get a response, but when tool call is required (see the screenshot) I got:

```
litellm.exceptions.APIConnectionError: litellm.APIConnectionError: Ollama_chatException - {"error":"json: cannot unmarshal array into Go struct field ChatRequest.messages.content of type string"}
```

Tried also with Llama3.2, Gemma3, DeepSeekR1-14B - they can't even respond to a normal message.

Environment:

```
MacOS 15.4
Ollama 0.6.5
adk-python 0.1.0
litellm 1.65.5
```

Is there a better way to call Ollama models?
Maybe support the OpenAI SDK (also can be used with Ollama) or provide native Ollama support?

Thank you!

![Image](https://github.com/user-attachments/assets/83fcfe98-589f-4d8a-9790-66a47cd56e30)

codefromthecrypt:
FWIW you can get it to work by using the LiteLlm openai variant until there's an openai one (#27)

```python
# TODO: native openai https://github.com/google/adk-python/issues/27
os.environ["OPENAI_API_BASE"] = os.getenv("OPENAI_BASE_URL")

--snip--
    model = os.getenv("CHAT_MODEL", "gpt-4o-mini")
    agent = Agent(name=app_name, model=LiteLlm(model="openai/" + model), tools=[your_tools_here])
```

Then, use the normal openai ENV like this:
```
# Override default ENV variables for Ollama
OPENAI_BASE_URL=http://localhost:11434/v1
OPENAI_API_KEY=unused
# sizes <3b tend to miss tool calls
CHAT_MODEL=qwen2.5:3b
```

fesnault:
[...]

PR #102, opened 2025-04-12
Fix empty list in content on tool calls causing ollama parse error

selcukgun:
This is related to google/adk-python#49

```
litellm.exceptions.APIConnectionError: litellm.APIConnectionError: Ollama_chatException - {"error":"json: cannot unmarshal array into Go struct field ChatRequest.messages.content of type string"}
```


hangfei:
Thanks for the super quick fix!

PR #111, opened 2025-04-12
Fixed typo in README.md

Maxiron:
Coordinator was incorrectly spelt as cooridnator.


hangfei:
fixed by another CL. thanks though.

PR #125, opened 2025-04-12
Add multi-part content support to preload_memory_tool

sezerogras:
This PR adds support for multi-part content in the PreloadMemoryTool. Previously, the tool only supported text content, but now it can handle:

- Text parts
- Function calls
- Function responses
- Inline data (like images)

Each content type is formatted in a human-readable way in the memory context.

Resolves the TODO: "support multi-part content" in preload_memory_tool.py


sezerogras:
bug fix

PR #139, opened 2025-04-13
Add ShellCommandExecutor for running shell commands directly

l33tdawg:
This PR adds a new `ShellCommandExecutor` class that provides a safe, efficient way to execute shell commands directly from ADK applications. This executor follows the established ADK executor pattern while providing specific functionality for shell commands.

## Motivation

When building AI assistants that need to interact with the system, having a dedicated executor for shell commands makes the code cleaner and safer than building custom solutions. The current `UnsafeLocalCodeExecutor` is designed for running Python code, but not optimized for shell commands.

## Implementation Details

The implementation:

- Inherits from `BaseCodeExecutor` to maintain consistent API
- Properly handles command timeouts with graceful termination
- Provides detailed error reporting and exit code handling
- Handles multi-line commands
- Safely parses shell commands with proper argument handling
- Returns a standardized `CodeExecutionResult` object

## Usage Example

```python
from google.adk.code_executors import ShellCommandExecutor

# Initialize the executor

executor = ShellCommandExecutor(timeout=30)

# Execute a command

result = await executor.execute_code(code="ls -la")

# Check execution result

if result.success:
    print(f"Command output: {result.output}")
else:
    print(f"Command failed: {result.error}")

# Alternative direct execution

stdout, stderr, returncode = await executor.execute("ls -la")
```

## Testing Done

The executor has been tested with:
- Simple commands like `ls`, `echo`, `cat`
- Commands with complex arguments and quoting
- Commands that time out
- Commands that produce errors

## Related Issue

This implementation addresses the need for a standardized way to execute shell commands within the ADK framework. %

### DIFF
--- a/src/google/adk/code_executors/__init__.py
+++ b/src/google/adk/code_executors/__init__.py
@@ -17,6 +17,7 @@ import logging
 from .base_code_executor import BaseCodeExecutor
 from .code_executor_context import CodeExecutorContext
 from .unsafe_local_code_executor import UnsafeLocalCodeExecutor
+from .shell_command_executor import ShellCommandExecutor
 
 logger = logging.getLogger(__name__)
 
@@ -24,6 +25,7 @@ __all__ = [
     'BaseCodeExecutor',
     'CodeExecutorContext',
     'UnsafeLocalCodeExecutor',
+    'ShellCommandExecutor',
 ]
 
 try:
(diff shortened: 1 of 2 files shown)


hangfei:
Our philosophy for tools is to leverage what the community has built. For shell command executor, it's available in MCP an quite a few other tool repos so i think rebuilding it in ADK won't provide additional value.

PR #154, opened 2025-04-13
Allow specification of maximum event contents in prompt to mitigate error compounding

Praznat:
This PR allows specification of max_contents in an LlmAgent. Currently, the system prompt given to an agent will include its entire event history so that it can reference past content from the conversation, tool calls, etc. The problem is that if occasionally the agent hallucinates or even the user introduces bad content, this will remain in history and potentially lead to compounding error. For some applications where interactions with the agent are expected to be of a certain reasonable range of length, it can be beneficial to set max_contents accordingly, but this PR leaves it None (unlimited) by default. This configuration is related to include_contents which can be set to 'none' to disable historical content in the prompt altogether.


hangfei:
Hi, we plan to fix this in a more systematic way. 

See : 
Add options to limit history content count included in LLM call
https://github.com/google/adk-python/issues/752
Enhanced Content Management for ADK LLM Agent
https://github.com/google/adk-python/issues/614


Closing this PR for now.

PR #155, opened 2025-04-13
Allow malformed function calls to be fixed on the spot

Praznat:
Occasionally I have seen Gemini return errors like this:

{"candidates":[{"content":{},"finish_message":"Malformed function call: transfer_to_agent(agent_name=\"OtherAgent\")\n","finish_reason":"MALFORMED_FUNCTION_CALL"}], ...

This PR introduces code to grab the function string from responses like this and convert it into a healthy FunctionCall on the spot. In the long run, this may not be necessary as updates to Gemini are made to correct this, but putting it out here for anyone running into this issue now.

### DIFF
--- a/src/google/adk/models/google_llm.py
+++ b/src/google/adk/models/google_llm.py
@@ -40,6 +40,46 @@ _NEW_LINE = '\n'
 _EXCLUDED_PART_FIELD = {'inline_data': {'data'}}
 
 
+def _extract_function_call(input_string):
+    """ Use regex to convert a function call string into a FunctionCall object. """
+    pattern = r"Malformed function call:\s*(\w+)\((.*)\)"
+    match = re.search(pattern, input_string)
+    if match:
+        func_name = match.group(1)
+        args_str = match.group(2).strip()
+
+        # Create a dummy function call with the captured arguments.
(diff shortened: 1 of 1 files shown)


hangfei:
@seanzhou1023 could you help take a look?

src/google/adk/models/google_llm.py
```diff
@@ -42,6 +44,46 @@
 _EXCLUDED_PART_FIELD = {'inline_data': {'data'}}
 
 
+def _extract_function_call(input_string):
+    """ Use regex to convert a function call string into a FunctionCall object. """
+    pattern = r"Malformed function call:\s*(\w+)\((.*)\)"
+    match = re.search(pattern, input_string)
+    if match:
+        func_name = match.group(1)
```
seanzhou1023, line 57: Are we alway able to extract a valid func_name or args_str in any case ? Give we don't have any contract with Model API for the Malformated case. Say API return something hallucinated after `Malformed function call:` but match your patten ?

hangfei:
This should be fixed from the model side. The fix is under way. Will update the thread once it's ready.

hangfei:
This should be fixed now in latest wheel. Please retry. Feel free to re-open if the issue persists.

PR #156, opened 2025-04-13
Instructions for agent to answer what it can before transferring

Praznat:
Adding some specific instructions for multi-agent transfer, when a user's request involves some part that one agent can answer and other parts that require transfer to another agent. We have found that a multi-agent system can get stuck on complex multi-step requests -- saying "I can't answer that" then sending it to another agent who also says the same, especially if there are contexts or artifacts that need to be created by earlier agents in order to be used by later ones. This simple additional instruction seems to help quite a bit with this problem, allowing the current agent to do what work it can before transferring.


hangfei:
The current instruction has gone through a series of tuning and evaluation. Could you file an issue first with the problem you are facing, and some rounds of evaluation so we know the severity of the problem ? 

Will close the PR for now.

PR #162, opened 2025-04-14
Rename 'ctx' parameter in documentation label to 'tool_context', for consistency

gsarthakdev:
All other methods that have `tool_context` as an argument, refer to it as "tool_context" in the docstring under Args. However, `run_async` didn't, it used 'ctx', so I have updated it.


hangfei:
Thanks for the fix!

gsarthakdev:
Welcome! Thank you for the review and successful merge @hangfei!

PR #163, opened 2025-04-14
Format error using openai with multimodal

KatsuraUkyo:
When inputting images into OpenAI from the GUI, the following error occurs, likely because an outdated format is being used.

`litellm.BadRequestError: AzureException BadRequestError - Invalid type for 'messages[1].content[1].image_url': expected an object, but got a string instead.`


gsarthakdev:
@KatsuraUkyo To be able to move forward with reviewing the change and merging it, please sign the CLA [here](https://cla.developers.google.com/).

More info about the CLA [here](https://github.com/google/adk-python/pull/163/checks?check_run_id=40480225332)

KatsuraUkyo:
@gsarthakdev I have signed the CLA.

hangfei:
@selcukgun please help review the changes.

hangfei:
Could you fix other errors?

seanzhou1023:
thank you for providing the fix. However the fix seems not correct. mainly here data_uri is the base64 encoded data instead of a url, the problem with the codes here is that it's using ChatCompletionImageUrlObject which doesn't have a property image_url

PR #166, opened 2025-04-14
Fix LiteLLM infinite loop issue with Ollama/Gemma3 models

LovepreetSinghVerma:
**Problem**
When using ADK with LiteLLM and certain models (particularly Ollama/Gemma3), the system can get stuck in an infinite loop under specific conditions:

- The model makes a function call with arguments
- The function executes and returns a result
- The model tries to make another call to the same function, but with malformed JSON in the arguments
- Due to parsing failures, the system gets stuck repeating the same function call indefinitely

This creates a poor user experience and wastes resources as the model repeatedly attempts the same operation without making progress.

**Solution**
This PR implements two key fixes to address the issue:

1. Robust JSON Parsing: Enhanced the response parsing logic to handle malformed JSON in function call arguments, with multiple fallback strategies.
2. Loop Detection Mechanism: Added functionality to detect and break potential infinite loops when the same function is called consecutively more than a configurable threshold (default: 5 times).

**Implementation Details**

1. Improved _model_response_to_generate_content_response with comprehensive validation and error handling
2. Added loop tracking state to the LiteLlm class
3. Enhanced generate_content_async to monitor consecutive function calls and provide helpful responses when loops are detected
4. Maintained backward compatibility with all existing functionality

**Testing**
The fix includes:

1. Unit tests for robust JSON parsing (tests/litellm/test_litellm_patch.py)
2. System tests for loop detection with simulated conversations (tests/litellm/system_test_litellm.py)
3. Manual testing with various edge cases

**Documentation**

1. Added detailed code documentation
2. Created comprehensive developer guide entry (docs/developer_guide.md)
3. Added specific documentation for the fix (docs/litellm_loop_fix.md)
4. Updated the changelog

**Impact and Considerations**
This change:

1. Does not modify any external APIs or interfaces
2. Has minimal performance impact (<5ms per request)
3. May interrupt legitimate use cases where a function is called more than 5 times consecutively (configurable via _loop_threshold)

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -376,7 +376,20 @@ def _model_response_to_chunk(
 def _model_response_to_generate_content_response(
     response: ModelResponse,
 ) -> LlmResponse:
-  """Converts a litellm response to LlmResponse.
+  """Converts a litellm response to LlmResponse with robust error handling.
+
+  This enhanced version:
+  1. Adds validation for required fields with proper defaults
+  2. Improves JSON parsing to handle both standard JSON and Python-style strings
+  3. Implements comprehensive error handling to prevent crashes
+  4. Maintains compatibility with all model formats
+
+  Implementation note:
+  This function is part of the fix for an infinite loop issue that occurs when using
+  Ollama/Gemma3 models with LiteLLM. These models sometimes return malformed JSON in 
+  function call arguments, which can cause the system to get stuck in a loop.
+  The robust parsing ensures that even with malformed JSON, we can still extract
+  valid arguments and prevent failures.
 
   Args:
     response: The model response.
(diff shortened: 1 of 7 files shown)


src/google/adk/models/lite_llm.py
```diff
@@ -376,22 +376,137 @@ def _model_response_to_chunk(
...
+  2. Improves JSON parsing to handle both standard JSON and Python-style strings
+  3. Implements comprehensive error handling to prevent crashes
+  4. Maintains compatibility with all model formats
+
+  Implementation note:
+  This function is part of the fix for an infinite loop issue that occurs when using
+  Ollama/Gemma3 models with LiteLLM. These models sometimes return malformed JSON in 
+  function call arguments, which can cause the system to get stuck in a loop.
+  The robust parsing ensures that even with malformed JSON, we can still extract
+  valid arguments and prevent failures.
```
gsarthakdev, line 392: While looking at the semantics/style of the rest of the codebase and the fact that you have already provided a separate file documenting these changes, I believe this portion of the function docstring is not necessary.

gsarthakdev, line 392: While looking at the semantics/style of the rest of the codebase and the fact that you have already provided a separate file documenting these changes, I believe this portion of the function docstring is not necessary.

src/google/adk/models/lite_llm.py
```diff
@@ -582,11 +705,36 @@ def __init__(self, model: str, **kwargs):
...
+    - The model receives malformed JSON responses it cannot parse
+    - The model gets into a repetitive pattern of behavior
+    - The model misunderstands function results and keeps trying the same approach
+    
+    When the same function is called consecutively more than the threshold number of times
+    (default: 5), the loop detection mechanism interrupts the loop and provides a helpful
+    response to the user instead of continuing to call the model.
+    
+    This prevents wasted resources and improves user experience by avoiding situations
+    where the system would otherwise become unresponsive.
```
gsarthakdev, line 737: While looking at the semantics/style of the rest of the codebase and the fact that you have already provided a separate file documenting these changes, I believe this portion of the function docstring is not necessary.

gsarthakdev (changes requested):
Docstring updates

hangfei:
I believe this is already fixed in the latest wheel. Please update the wheel version and try again. Feel free to re-open if you still see this issue.

PR #231, opened 2025-04-17
fix(litellm): prevent 400 error during parallel tool calls

yuri12344:
API calls via the LiteLLM adapter failed with a 400 Bad Request when models like OpenAI returned multiple tool_calls (requesting parallel execution). The adapter incorrectly sent only one tool response message back, while the API requires a separate tool message for each requested tool_call_id.

## Cause:
The _get_completion_inputs function in lite_llm.py only processed the first function_response part from the ADK's aggregated tool response event, ignoring subsequent ones.
## Solution:

Modified _get_completion_inputs to iterate through all function_response parts in the aggregated event. It now correctly generates a distinct ChatCompletionToolMessage (role: 'tool') for each tool_call_id.
## Impact:

Resolves the 400 Bad Request error. Ensures correct message formatting for the API when handling parallel tool calls, enabling agent flows with multiple simultaneous tool executions.

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -503,11 +503,26 @@ def _get_completion_inputs(
   # 1. Construct messages
   messages: List[Message] = []
   for content in llm_request.contents or []:
-    message_param_or_list = _content_to_message_param(content)
-    if isinstance(message_param_or_list, list):
-      messages.extend(message_param_or_list)
-    elif message_param_or_list:  # Ensure it's not None before appending
-      messages.append(message_param_or_list)
+    is_tool_response_content = (
+        content.role == 'user'
+        and content.parts
+        and any(part.function_response for part in content.parts)
+    )
+    if is_tool_response_content:
+      for part in content.parts:
+        if part.function_response:
+          tool_message = ChatCompletionToolMessage(
+              role="tool",
+              tool_call_id=part.function_response.id,
+              content=_safe_json_serialize(part.function_response.response),
+          )
+          messages.append(tool_message)
+    else:
+      message_param_or_list = _content_to_message_param(content)
+      if isinstance(message_param_or_list, list):
+        messages.extend(message_param_or_list)
+      elif message_param_or_list:  # Ensure it's not None before appending
+        messages.append(message_param_or_list)
 
   if llm_request.config.system_instruction:
     messages.insert(


yuri12344:
@hangfei can you take a look? I get this problem when the AI call multiple tools and it was getting error because it do not pass the right arguments to openai, but I fixed it

To replicate just add a bunch of tools for ai, and make it call all of then in same time.

miko-garrido:
We currently have this implemented in our code and works to solve the LiteLLM parallel tool call problem. Is this no longer an issue with the present version?

hangfei:
@selcukgun Could you help review this?

miko-garrido:
Hey @hangfei - curious if parallel tool calls have been resolved in the most recent version?

yuri12344:
I mean, I made this PR after one days of release, and it took one month to be reviwed, then I switch back to langchain, idknw how the things are rn, Im sorry

miko-garrido:
I'm using your code in production @yuri12344 - honestly a lifesaver 🙏🏽

yuri12344:
@miko-garrido Hi Miko, that is nice, Im glad it helped you.

vrvlive:
> Hey @hangfei - curious if parallel tool calls have been resolved in the most recent version?

Still experiencing the issue in 1.5.0

hangfei:
Could you resolve the conflicts?

yuri12344:
@hangfei finally I can, I was waiting soo long for someone to reply, just small quick fix and I neeedd that soo much in the begining, okay I will fix the conmflicts

yuri12344:
@hangfei  Thanks for the guidance. Conflicts are resolved, and I've squashed the history into a single clean commit. It's ready for review whenever you have a moment.

seanzhou1023:
Thanks so much for the PR :)

However I think we already support parallel tool call even before this PR : https://github.com/google/adk-python/pull/172/files

seanzhou1023:
Please check the latest logic here :
https://github.com/google/adk-python/blob/48971d43d02d9c4054045bf6ae3dc62c288e6039/src/google/adk/models/lite_llm.py#L167

feel free to reopen if it's not true.

miko-garrido:
@seanzhou1023 Before patching in the code in this PR, I can confirm that parallel tool calls did not work with openai models via litellm.

miko-garrido:
What I can say is that I haven't tested the main src since I patched in this pr from @yuri12344 back in early June. I can confirm parallel tool calls weren't working before that point.

seanzhou1023:
@miko-garrido  are you running on latest codes ? could you please provide a sample agents codes that can reproduce the problem ?

@yuri12344 could you please fix the logic in _content_to_message_param method above if you want instead of adding an additional if block ?  feel free to reopen if you can spot some bug there and fix it. thanks :)

seanzhou1023:
> What I can say is that I haven't tested the main src since I patched in this pr from @yuri12344 back in early June. I can confirm parallel tool calls weren't working before that point.

I quickly compare the changes in this PR and logic here : https://github.com/google/adk-python/blob/48971d43d02d9c4054045bf6ae3dc62c288e6039/src/google/adk/models/lite_llm.py#L167-L178

don't see much difference....

yuri12344:
@seanzhou1023 hi ofc, I will take a look in the things you said and provide more information

seanzhou1023:
> @seanzhou1023 hi ofc, I will take a look in the things you said and provide more information

Thank you so much :) Do appreciate it :)

PR #256, opened 2025-04-17
Fixed Parameter required field and description

sumant-pangotra:
- If a parameter is set at required : false in OpenAPI spec file . It will not be included in required list generated for function declaration for LLM call
- If description is not present in parameter schema then use parameter description


seanzhou1023 (approved):

Jacksunwei:
@sumant-pangotra could you check the failed conventional commit action and fix it, so that the CL can be generated?

Jacksunwei:
Merged in https://github.com/google/adk-python/commit/e8fb4ed3fcba1f455507c38275ee17218017a409

PR #289, opened 2025-04-19
fix: do not use mutable objects for function argument defaults

n1lanjan:
This PR fixes usage of mutable defaults in a some methods. The usage of mutable defaults does not have any adverse impact currently since the object `initial_session` is not updated in the methods as of now, but any change in the future may introduce bugs due to this and hence a good practice not to use mutable defaults and instead initialise the mutable object inside the function body.

### Why It’s Dangerous
- Unintended Persistence: Changes to the default argument persist between calls, violating the expectation of a "fresh" default.
- Subtle Bugs: The behaviour is counterintuitive and can lead to hard-to-debug issues.
- Scope Risks: Even if unused, mutable defaults risk accidental mutation in complex functions.

For eg.
```python
def f(a, L=[]):
    L.append(a)
    return L

print(f(1))
print(f(2))
print(f(3))
```
will print
```console
[1]
[1, 2]
[1, 2, 3]
```
To further see why it is bad, please see 
- [mutable-argument-default (B006)](https://docs.astral.sh/ruff/rules/mutable-argument-default/#mutable-argument-default-b006)
- [Python documentation: Default Argument Values](https://docs.python.org/3/tutorial/controlflow.html#default-argument-values)

### DIFF
--- a/src/google/adk/evaluation/agent_evaluator.py
+++ b/src/google/adk/evaluation/agent_evaluator.py
@@ -17,6 +17,7 @@ import os
 from os import path
 from typing import Dict
 from typing import List
+from typing import Optional
 from typing import Union
 
 from .evaluation_generator import EvaluationGenerator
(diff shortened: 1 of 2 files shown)


hangfei:
Thanks for spotting this! I think this is already fixed recently. Please contribute more to this repo in the future!

PR #298, opened 2025-04-20
Fix: Pass Bearer Token to `tool_context.actions.requested_auth_configs`

### LINKED ISSUE #267: Google ADK Authentication Token Not Passed to Tool Context
# Bug Report: Google ADK Authentication Token Not Passed to Tool Context

## Describe the bug
When using Google ADK with HTTP Bearer authentication, the bearer token provided in the request is not properly passed to the `tool_context` in custom tools. The `tool_context.actions.requested_auth_configs` dictionary remains empty even when authentication information is properly provided in the request.

## To Reproduce
1. Create a custom tool that requires authentication using the ADK framework
2. Configure the tool to use HTTP Bearer authentication
3. Send a request with authentication information in one of these ways:
   - HTTP Authorization header with a Bearer token
   - In the request body under `actions.requested_auth_configs`
4. In the tool, attempt to access `tool_context.actions.requested_auth_configs`
5. See that the dictionary is empty and the token is not accessible

## Expected behavior
When a bearer token is provided in the request (either through headers or in the body), ADK should pass this token to the tool through `tool_context.actions.requested_auth_configs`, making it accessible using the standard ADK authentication flow with `tool_context.get_auth_response()`.

## Screenshots
N/A

## Desktop (please complete the following information):
- OS: macOS 24.4.0
- Python version: 3.13
- ADK version: The latest version (as installed by default in April 2025)

## Additional context
This appears to be an issue with how ADK's main runner processes authentication information. In the code of `runners.py`, the authentication information from the request body is not properly transferred to the `event_actions` object that's passed to the `tool_context`.

MagnusDot:
# This MR fixes issue #267 where HTTP Bearer tokens provided in ADK requests were not correctly propagated to `tool_context.actions.requested_auth_configs`.

Changes Made:
- `src/google/adk/agents/invocation_context.py`: Added a requested_auth_configs attribute to store authentication configurations requested by the client to pass to tools.

- `src/google/adk/runners.py`: Modified to support passing an authentication token during invocation via run_async and configuring it in the invocation context.

- `src/google/adk/tools/tool_context.py`: Enhanced the get_auth_response() method to first check if a Bearer token is available in requested_auth_configs before using the standard method of retrieving the token from the session state.

- `tests/unittests/auth/test_bearer_token_auth.py`: Added a new test file to validate the Bearer token authentication functionality, including tests to verify correct token configuration in the invocation context and proper retrieval by tools.

## Summary:
This change allows passing a Bearer authentication token directly during agent invocation, facilitating authentication to external APIs without requiring user interaction to obtain the token. The token is passed to the tool context and can be retrieved by tools via the get_auth_response() method.

Fixes #267


MagnusDot:
@boyangsvl can you look at it please 🙏

hangfei:
@seanzhou1023 PTAL.

MagnusDot:
@hangfei any news ? 🫣

seanzhou1023:
sorry we didn't provide enough documents. 

auth response is not passed in this way. It should be only passed as a function response to adk_request_credential, please see : https://github.com/google/adk-docs/blob/main/docs/tools/authentication.md

[ool_context.actions.requested_auth_configs is not something that is supposed to be touch out of adk.

seanzhou1023:
and this is only used when tools are requesting credential on demand. e.g. :
https://github.com/google/adk-python/blob/5d2a876ca8b68b038a81d0ae8d289bfdec945d41/contributing/samples/oauth_calendar_agent/agent.py#L54

and client send back the auth credential as response.

If client needs to actively pass in auth credential to tools, you can directly pass in the session.state, and if you want it to be removed after one invocation you can append the key with 'temp:' namespace, see https://google.github.io/adk-docs/tools/#state-management

PR #320, opened 2025-04-21
Add support for further telemetry collection with additional signal types and with additional initialization libraries.

michaelsafyan:
# Changes

 - Adds support for initializing additional telemetry subsystems:
   1. Open Telemetry requests instrumentation library - for collecting low-level HTTP requests to/from Vertext/Gemini
   2. Google Gen AI SDK Instrumentation -  OTel SemConv aligned model-level instrumentation of LLM calls

 - Adds support for generating logs and metrics in addition to traces

 - Adds the ability to use `ADK_CLOUD_O11Y=1` to cover logs, traces, and metrics (individual signal types can be covered with individual environment variables as well: `ADK_TRACE_TO_CLOUD`, `ADK_LOGS_TO_CLOUD`, `ADK_METRICS_TO_CLOUD`).

 - Adds support for more general OTel export; for example, users can set `OTEL_EXPORTER_OTLP_ENDPOINT` to an arbitrary OTLP endpoint to export logs, traces, and metrics (as an alternative to Google Cloud Observability, or to use the Open Telemetry Collector as part of their process for ingestion into Google Cloud).

# Not Included

This DOES NOT change the logic to deploy to Cloud Run or the command line interface. The commandline currently has a `--trace_to_cloud` environment variables that does not cover the full feature set here. A separate change would need to be made to expose these additional capabilities in `adk deploy` (and in other commands where not using environment variables manipulation to supply these options).


michaelsafyan:
It's clear that the size of this is too large to be reasonable to commit all at once.

I split out a first piece of this here:

https://github.com/google/adk-python/pull/569

@selcukgun @boyangsvl wondering if you might be able to review #569

michaelsafyan:
Closing out because related work was handed off.

PR #343, opened 2025-04-22
Use async on session service

thiagosalvatore:
The current implementation of  `BaseSessionService` is not async. It has a ton of I/O operations that would block the main event loop. Instead, we should leverage async to run database operations on the session.

This PR updates the BaseSessionService as well as its implementations to use async, preventing the event loop from being blocked and allowing the usage of other database implementations like `motor` for MongoDB.

This PR does not updates the `DatabaseSessionService` to use the async version of sqlalchemy.


DeanChensj:
Hi @thiagosalvatore , thanks for the great work! Due to some internal dependencies, we have to make this change on our own, because otherwise it will break our internal team. Sorry we can't merge this PR directly, really appreciate it!

hangfei:
> Hi @thiagosalvatore , thanks for the great work! Due to some internal dependencies, we have to make this change on our own, because otherwise it will break our internal team. Sorry we can't merge this PR directly, really appreciate it!

@thiagosalvatore thanks for your work! Looking forward to your future contributions. Closing this PR for now.

PR #364, opened 2025-04-23
Prevent MCP ClientSession hang

njbrake:
I expect that we want to make this a configurable parameter that we elevate somewhere, but this seems like the most efficient way to start the conversation 👍 

Per https://modelcontextprotocol.io/specification/draft/basic/lifecycle#timeouts

"Implementations SHOULD establish timeouts for all sent requests, to prevent hung connections and resource exhaustion. When the request has not received a success or error response within the timeout period, the sender SHOULD issue a cancellation notification for that request and stop waiting for a response.

SDKs and other middleware SHOULD allow these timeouts to be configured on a per-request basis."

I picked 5 seconds since that's the default for SSE

### DIFF
--- a/src/google/adk/tools/mcp_tool/mcp_session_manager.py
+++ b/src/google/adk/tools/mcp_tool/mcp_session_manager.py
@@ -2,6 +2,7 @@ from contextlib import AsyncExitStack
 import functools
 import sys
 from typing import Any, TextIO
+from datetime import timedelta
 import anyio
 from pydantic import BaseModel
 
@@ -133,6 +134,7 @@ class MCPSessionManager:
         connection_params=self.connection_params,
         exit_stack=self.exit_stack,
         errlog=self.errlog,
+
     )
 
   @classmethod
@@ -142,6 +144,7 @@ class MCPSessionManager:
       connection_params: StdioServerParameters | SseServerParams,
       exit_stack: AsyncExitStack,
       errlog: TextIO = sys.stderr,
+      timeout: int = 5,
   ) -> ClientSession:
     """Initializes an MCP client session.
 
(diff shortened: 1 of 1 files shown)


njbrake:
Hi @seanzhou1023 Do you have any updates about the status of this PR? Not sure if there's something else you'd like me to do. Thanks!

genquan9 (commented):
Thanks for your contributions!

The team fixed the issue in [45ef668](https://github.com/google/adk-python/commit/45ef6684352e3c8082958bece8610df60048f4a3).

Instead of adding a new parameter, we decided to add a default timeout for StdioServerParameters.
Note, SseServerParams and StreamableHTTPServerParams already provide default timeout.

njbrake:
Thanks @genquan9: if I understand that other PR correctly though, isn't the timeout now hardcoded to 5 seconds? What if I need to increase it to something like 15 seconds for my application?

genquan9:
Currently it is default to 5 seconds.
Do you have a high requirement for the customization of the timeout?

The team is still discussing for the proper interface for the timeout when use StdioServerParameters.

njbrake:
> Currently it is default to 5 seconds. Do you have a high requirement for the customization of the timeout?
> 
> The team is still discussing for the proper interface for the timeout when use StdioServerParameters.

Thanks! Yes we have a requirement for customization. Just today we were working with a developer who was using the Github MCP and it was timing out because it was a big query he was making that was taking longer than 5 seconds. Our framework supports Google ADK as well as a few other frameworks, and they have all added support for this customization of the timeout (you can see links to the PRs for all the other frameworks I made PRs for here https://github.com/mozilla-ai/any-agent/issues/129), since it complies with the requirement that is stated in the MCP protocol definition (stated in the description of this PR).

Issue #342, opened 2025-04-22
AttributeError: 'MCPTool' object has no attribute 'func'


bipulkarnani:
**Describe the bug**
`AttributeError: 'MCPTool' object has no attribute 'func'`
I am getting this error when I integrated airbnb mcp in adk streaming example along with the google search tool.

**To Reproduce**
In the agent if we add airbnb mcp example, I am getting this error when agent is trying to use the airbnb tool to execute task. For all the google search related queries, no issues were reported.

hangfei:
Live(https://google.github.io/adk-docs/get-started/quickstart-streaming/) is not working with MCPTool yet. 

We should add a error here when users try to use MCPTool for streaming. This is a good first issue(just adding the warning.)


We are working to support MCPTool with streaming/live. 

Could you share the example and steps to reproduce your error?

ghost:
Any timeline by when can we expect MCPTool support with streaming/live?

devxaitist:
Hi @hangfei  I'd like to work on this issue if it's still available. I understand that we need to add a warning/error message when users try to use MCPTool with streaming/live mode. I'll implement a check that detects when MCPTool is being used with streaming and provides a clear error message to guide users. If you're okay with me taking this on, I'll create a PR once I have a solution ready. Thanks!

hangfei:
@devxaitist please. thanks.

hangfei:
This is fixed by: https://github.com/google/adk-python/commit/f7b436a98519fda026409882f271c40b7375ab23. 

I have verified with : 

```
# ./adk_agent_samples/mcp_agent/agent.py
from google.adk.agents.llm_agent import LlmAgent
from google.adk.tools.mcp_tool.mcp_toolset import MCPToolset, StdioServerParameters

async def attach_everything_tools():
    tools, exit_stack = await MCPToolset.from_server(
        connection_params=StdioServerParameters(
            command='npx',
            args=[
                '-y',
                '@modelcontextprotocol/server-everything',
                '/tmp/mcp-test',
            ],
        )
    )
    # `tools` now contains the `echo`, `add`, etc. tools
    return tools, exit_stack

# async def create_agent():
#   """Gets tools from MCP Server."""
#   tools, exit_stack = await MCPToolset.from_server(
#       connection_params=StdioServerParameters(
#           command='npx',
#           args=["-y",    # Arguments for the command
#             "@modelcontextprotocol/server-filesystem",
[...]

PR #377, opened 2025-04-24
[Fix] MCPTool live mode error handle.

keli-wen:
### Description

Refer to https://github.com/google/adk-python/issues/342

- Attempted to add a handle for an MCPTool error under the live model.
- Supplemented a unittest at the corresponding location (might need to move more appropriate place?).


hangfei:
Thanks for the quick fix.


Could you provide a log for MCPTool in live vs regular tool in live to ensure correctness?

hangfei:
This is already fixed so this warning is no longer needed. Thanks though!

PR #403, opened 2025-04-26
Root agent incorrect definition resolution

BhadrakshB:
Currently, if I incorrectly define the root_agent, the error logs do not point to the exact reason why the error popped up or where the issue exists.

I added a check on the instance of self.agent, inside the ```_new_invocation_context``` of the ```Runner``` class. If ```self.agent``` of type ```BaseAgent```, the program runs through. If there is an error, a clear description is showing saying ```Your parent must be of type BaseAgent or a class that inherits from BaseAgent.```


BhadrakshB:
@hangfei Can you help me to review this PR? Thank you very much.

Issue #414, opened 2025-04-27
No module named 'enterprise'


nneverwei:
**Describe the bug**
Using **adk v0.3.0** + **litellm v1.67.4** (latest), no interaction with any agent can be performed in **adk web**. An exception is thrown: "**_No module named 'enterprise'_**".

_Last week should have been normal, but after updating adk and litellm to the latest version, it became abnormal._

**To Reproduce**
1. Direct new environment `pip install adk` `pip install litellm`
2. Follow the official example [weather_time_agent](https://google.github.io/adk-docs/get-started/quickstart/#agentpy)，Modify the model to connect to litellm:
```python
model=LiteLlm(
      model="openai/Qwen/Qwen2.5-72B-Instruct", 
      base_url="https://api.siliconflow.cn/v1", 
      api_key="***"
  ),
```
3. run `adk web`，Error occurred  "**_No module named 'enterprise'_**"

**Screenshots**
![Image](https://github.com/user-attachments/assets/d5ddfe40-5c51-43dd-a781-927efba55461)

**Exception stack**
```shell
2025-04-27 17:42:32,617 - INFO - envs.py:54 - No .env file found for mcp_file_v2_agent
2025-04-27 17:42:32,680 - ERROR - fast_api.py:637 - Error in event_generator: No module named 'enterprise'
Traceback (most recent call last):
  File "D:\tools\anaconda3\envs\test_adk_313\Lib\site-packages\google\adk\cli\fast_api.py", line 625, in event_generator    
    runner = await _get_runner_async(req.app_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\tools\anaconda3\envs\test_adk_313\Lib\site-packages\google\adk\cli\fast_api.py", line 796, in _get_runner_async  
    root_agent = await _get_root_agent_async(app_name)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\tools\anaconda3\envs\test_adk_313\Lib\site-packages\google\adk\cli\fast_api.py", line 773, in _get_root_agent_async
    agent_module = importlib.import_module(app_name)
  File "D:\tools\anaconda3\envs\test_adk_313\Lib\importlib\__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
[...]

Issue #387, opened 2025-04-25
A problem in "adk web" when running agent with mcp tools : RuntimeError: error getting root agent


YangZH1997:
I built an agent containing the mcp tools by referring to the example in the document, but a bug occurred when running it with adk web: RuntimeError: error getting root agent. 
////////////////////////////////////////////////////////////////////////////////////////////////////////////////
Operation Environment:
Windows 11 24H2  64bit
Python version 3.13.2
google-adk version 0.3.0

Package                            Version
---------------------------------- ------------
aiohappyeyeballs                   2.6.1
aiohttp                            3.11.18
aiosignal                          1.3.2
annotated-types                    0.7.0
anyio                              4.9.0
attrs                              25.3.0
authlib                            1.5.2
cachetools                         5.5.2
certifi                            2025.1.31
cffi                               1.17.1
charset-normalizer                 3.4.1
click                              8.1.8
colorama                           0.4.6
cryptography                       44.0.2
deprecated                         1.2.18
distro                             1.9.0
docstring-parser                   0.16
fastapi                            0.115.12
filelock                           3.18.0
frozenlist                         1.6.0
fsspec                             2025.3.2
google-adk                         0.3.0
google-api-core                    2.24.2
google-api-python-client           2.168.0
google-auth                        2.39.0
google-auth-httplib2               0.2.0
google-cloud-aiplatform            1.90.0
google-cloud-bigquery              3.31.0
google-cloud-core                  2.4.3
google-cloud-resource-manager      1.14.2
google-cloud-secret-manager        2.23.3
google-cloud-speech                2.32.0
google-cloud-storage               2.19.0
google-cloud-trace                 1.16.1
google-crc32c                      1.7.1
google-genai                       1.12.1
google-resumable-media             2.7.2
googleapis-common-protos           1.70.0
graphviz                           0.20.3
greenlet                           3.2.1
grpc-google-iam-v1                 0.14.2
grpcio                             1.71.0
grpcio-status                      1.71.0
h11                                0.16.0
httpcore                           1.0.9
httplib2                           0.22.0
httpx                              0.28.1
httpx-sse                          0.4.0
huggingface-hub                    0.30.2
idna                               3.10
importlib-metadata                 8.6.1
jinja2                             3.1.6
jiter                              0.9.0
jsonschema                         4.23.0
jsonschema-specifications          2024.10.1
litellm                            1.67.0.post1
markupsafe                         3.0.2
mcp                                1.6.0
multidict                          6.4.3
[...]

PR #415, opened 2025-04-27
fix(cli): Disable auto-reload feature on Windows system

nneverwei:
Fixed the issue caused by the auto-reload feature when running the CLI tool on Windows system. By detecting the operating system type, the auto-reload is disabled on Windows system to avoid potential errors: When mcp is asynchronously loaded, it will enter the _make_subprocess_transport NotImplementedError logic due to uvicorn reload=True in fastapi.

Although disabling the reload feature of uvicorn on Windows may result in some loss of functionality, as of now, MCP has large-scale usage scenarios, and it should be prioritized to ensure that MCP can be used normally

[issue-#414](https://github.com/google/adk-python/issues/414)
[issue-#387](https://github.com/google/adk-python/issues/387)

### DIFF
--- a/src/google/adk/cli/cli_tools_click.py
+++ b/src/google/adk/cli/cli_tools_click.py
@@ -356,7 +356,7 @@ def cli_web(
       app,
       host="0.0.0.0",
       port=port,
-      reload=True,
+      reload=False if os.name == 'nt' else True,
   )
 
   server = uvicorn.Server(config)
@@ -456,7 +456,7 @@ def cli_api_server(
       ),
       host="0.0.0.0",
       port=port,
-      reload=True,
+      reload=False if os.name == 'nt' else True,
   )
   server = uvicorn.Server(config)
   server.run()


hangfei:
Thanks for the fix @nneverwei ! 

I'd suggest we add an option in the CLI to enable or disable the reload feature. So users(esp. windows) users can disable this if they come across this  bug.

510myRday:
👍nice

BenMawnMahlauNBTC:
it might be useful to check if the platform is windows (os.name or sys.platform) and set the default for reload dynamically based on that.

hangfei:
Thanks for the addition!

hangfei:
Could someone help update the documentation(https://github.com/google/adk-docs) on not to reoload for windows?

nneverwei:
> Could someone help update the documentation(https://github.com/google/adk-docs) on not to reoload for windows?

https://github.com/google/adk-docs/pull/252

hangfei:
merged in https://github.com/google/adk-python/commit/85ccacbf2d2849fe668fe3c259db4ace72cd654d. thanks!

Issue #380, opened 2025-04-24
Support async callbacks


byt3bl33d3r:
It seems like all the callbacks don't support coroutines. This can be a problem with asynchronous applications as it will block the event loop.

 I have the following code:

```python
async def before_tool_cb(
    tool: BaseTool, args: Dict[str, Any], tool_context: ToolContext
) -> Optional[Dict]:
       print("called")

agent = LlmAgent(
  model=settings.gemini_model,
  name='MyAssistant',
  instruction=root_agent_prompt.PROMPT,
  tools=mcp_tools + agent_tools,
  before_tool_callback=before_tool_cb,
  after_tool_callback=after_tool_cb
)
```

which gives the following traceback (running it with `run_async()`):

```
 File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/runners.py", line 197, in run_async
    async for event in invocation_context.agent.run_async(invocation_context):
  File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/agents/base_agent.py", line 141, in run_async
    async for event in self._run_async_impl(ctx):
  File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/agents/llm_agent.py", line 232, in _run_async_impl
    async for event in self._llm_flow.run_async(ctx):
  File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/flows/llm_flows/base_llm_flow.py", line 231, in run_async
    async for event in self._run_one_step_async(invocation_context):
  File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/flows/llm_flows/base_llm_flow.py", line 261, in _run_one_step_async
    async for event in self._postprocess_async(
  File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/flows/llm_flows/base_llm_flow.py", line 329, in _postprocess_async
    async for event in self._postprocess_handle_function_calls_async(
  File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/flows/llm_flows/base_llm_flow.py", line 404, in _postprocess_handle_function_calls_async
    if function_response_event := await functions.handle_function_calls_async(
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/flows/llm_flows/functions.py", line 183, in handle_function_calls_async
[...]

PR #434, opened 2025-04-28
feat(llm_flows): support async before/after tool callbacks

AlankritVerma01:
**Fixes:** #380 

Previously, both `before_tool_callback` and `after_tool_callback` were invoked strictly as synchronous functions. If a user passed an `async def` coroutine, it would return a coroutine object that was never awaited—leading to runtime warnings (“coroutine was never awaited”) and Pydantic serialization failures when the coroutine object ended up in an event.

With this change, we:

- Invoke the user’s callback and inspect its return value.
- If the return value is awaitable, `await` it before proceeding.
- Apply this logic uniformly in both the async and live‐streaming code paths.

Users may now supply either synchronous or asynchronous callbacks transparently; existing synchronous callbacks continue to work unchanged.

## Test Plan
1. **New unit tests** in `test_async_tool_callbacks.py` verify that:
   - An `async def before_tool_callback(...)` short-circuits the tool call.
   - An `async def after_tool_callback(...)` can override the tool’s result.
2. Run the new tests:  
   ```bash
   pytest tests/unittests/flows/llm_flows/test_async_tool_callbacks.py
   ```
3. Run the full suite to ensure no regressions:  
   ```bash
   pytest
   ```

## Checklist

- [x] New tests added and passing  
- [x] Existing sync‐callback tests still pass  
- [x] Follows repository’s style and linting rules  
- [x] References issue #380 with “Fixes: #380”

### DIFF
--- a/src/google/adk/agents/llm_agent.py
+++ b/src/google/adk/agents/llm_agent.py
@@ -57,11 +57,11 @@ AfterModelCallback: TypeAlias = Callable[
 ]
 BeforeToolCallback: TypeAlias = Callable[
     [BaseTool, dict[str, Any], ToolContext],
-    Awaitable[Optional[dict]],
+    Union[Awaitable[Optional[dict]], Optional[dict]],
 ]
 AfterToolCallback: TypeAlias = Callable[
     [BaseTool, dict[str, Any], ToolContext, dict],
-    Awaitable[Optional[dict]],
+    Union[Awaitable[Optional[dict]], Optional[dict]],
 ]
 
 InstructionProvider: TypeAlias = Callable[[ReadonlyContext], str]
--- a/src/google/adk/agents/llm_agent.py
+++ b/src/google/adk/agents/llm_agent.py
@@ -15,12 +15,7 @@
 from __future__ import annotations
 
 import logging
-from typing import Any
-from typing import AsyncGenerator
-from typing import Callable
-from typing import Literal
-from typing import Optional
-from typing import Union
+from typing import Any, AsyncGenerator, Awaitable, Callable, Literal, Optional, Union
 
 from google.genai import types
 from pydantic import BaseModel
@@ -62,11 +57,11 @@ AfterModelCallback: TypeAlias = Callable[
 ]
 BeforeToolCallback: TypeAlias = Callable[
     [BaseTool, dict[str, Any], ToolContext],
-    Optional[dict],
+    Awaitable[Optional[dict]],
 ]
 AfterToolCallback: TypeAlias = Callable[
     [BaseTool, dict[str, Any], ToolContext, dict],
-    Optional[dict],
+    Awaitable[Optional[dict]],
 ]
 
 InstructionProvider: TypeAlias = Callable[[ReadonlyContext], str]
--- a/src/google/adk/flows/llm_flows/functions.py
+++ b/src/google/adk/flows/llm_flows/functions.py
@@ -237,12 +237,11 @@ async def handle_function_calls_live(
     #       tool, function_args, tool_context
     #   )
     if agent.before_tool_callback:
-      _maybe = agent.before_tool_callback(
-        tool=tool, args=function_args, tool_context=tool_context
-        )
-      if inspect.isawaitable(_maybe):
-        _maybe = await _maybe
-        function_response = _maybe
+      function_response = agent.before_tool_callback(
+          tool=tool, args=function_args, tool_context=tool_context
+      )
+      if inspect.isawaitable(function_response):
+        function_response = await function_response
 
     if not function_response:
       function_response = await _process_function_live_helper(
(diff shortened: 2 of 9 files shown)


DanMcInerney:
Can confirm this is working as expected.

src/google/adk/flows/llm_flows/functions.py
```diff
@@ -152,27 +152,35 @@ async def handle_function_calls_async(
...
-    # Calls the tool if before_tool_callback does not exist or returns None.
+    # # Calls the tool if before_tool_callback does not exist or returns None.
+    # if agent.before_tool_callback:
+    #   function_response = agent.before_tool_callback(
+    #       tool=tool, args=function_args, tool_context=tool_context
+    #   )
+    # Short-circuit via before_tool_callback (sync *or* async)
     if agent.before_tool_callback:
-      function_response = agent.before_tool_callback(
+      _maybe = agent.before_tool_callback(
```
Jacksunwei: nit: Let's just use `function_response` here.

src/google/adk/flows/llm_flows/functions.py
```diff
@@ -152,27 +152,35 @@ async def handle_function_calls_async(
...
       )
-
     # Calls after_tool_callback if it exists.
     if agent.after_tool_callback:
-      new_response = agent.after_tool_callback(
-          tool=tool,
-          args=function_args,
-          tool_context=tool_context,
-          tool_response=function_response,
+      _maybe2 = agent.after_tool_callback(
```
Jacksunwei: nit: let's use `altered_function_response`

src/google/adk/flows/llm_flows/functions.py
Jacksunwei, line 1: Could you also update the signature in [1] to include Awaitable[...] in result? 

https://github.com/google/adk-python/blob/main/src/google/adk/agents/llm_agent.py#L63-L70

Jacksunwei (commented):
Thanks for the PR!

src/google/adk/flows/llm_flows/functions.py
AlankritVerma01: Thanks for the feedback.
Its updated now.

src/google/adk/flows/llm_flows/functions.py
AlankritVerma01: Thanks for the feedback.
Its updated now.

src/google/adk/flows/llm_flows/functions.py
AlankritVerma01: Thanks for the feedback.
Its updated now.

src/google/adk/agents/llm_agent.py
```diff
@@ -62,11 +57,11 @@
 ]
 BeforeToolCallback: TypeAlias = Callable[
     [BaseTool, dict[str, Any], ToolContext],
-    Optional[dict],
+    Awaitable[Optional[dict]],
```
Jacksunwei: I think it should be `Union[Awaitable[Optional[dict]], Optional[dict]]`.

src/google/adk/agents/llm_agent.py
```diff
@@ -62,11 +57,11 @@
...
 ]
 AfterToolCallback: TypeAlias = Callable[
     [BaseTool, dict[str, Any], ToolContext, dict],
-    Optional[dict],
+    Awaitable[Optional[dict]],
```
Jacksunwei: ditto.

src/google/adk/flows/llm_flows/functions.py
Jacksunwei: I see you updated below handle_function_calls_live, but this one handle_function_calls_async is missed? 

Both methods need changing.

src/google/adk/flows/llm_flows/functions.py
Jacksunwei: also here.

tests/unittests/flows/llm_flows/test_async_tool_callbacks.py
Jacksunwei, line 1: Please use `pyink --config pyproject.toml tests/unittests/flows/llm_flows/test_async_tool_callbacks.py` to format this file, so that it's consistent with the rest of codebase.

AlankritVerma01: Thanks for sharing.
I think we should also include this somewhere in some documentation. Please correct me if its already there.

Jacksunwei (approved):
Thanks!

DanMcInerney:
Thank you @Jacksunwei @AlankritVerma01 for the speedy updates! This is an incredibly useful addition.

Jacksunwei:
Thanks for contributing to the PR! I'm coordinating the internal review.

Jacksunwei:
Merged in commit: https://github.com/google/adk-python/commit/27ce65ff50e284fa9700c1c37782ea58afd98901

PR #438, opened 2025-04-28
Fix: `config.after_timestamp` behavior in `InMemorySessionService.get_session()` and add a test

### LINKED ISSUE #437: 2 minor bugs in `config.after_timestamp` behavior of `InMemorySessionService.get_session()`
> _Adding a fix shortly.._
> _Combining 2 bugs due to simple & compatible logic_

**Describe the bug**
1. `config.after_timestamp` logic mistakenly includes one more index when doing the slicing.

2. I would expect `GetSessionConfig`'s `num_recent_events` and `after_timestamp` to be compatible with each other in `InMemorySessionService` (looks to be true in `DatabaseSessionService`'s implementation using SQL filter and then limit). Right now it checks `if config.num_recent_events` exists and only does the `config.after_timestamp` logic in the `elif` case. The conditionals can/should be independent. 

See screenshots below for steps to reproduce and expected behavior:
Image
Image

 - OS: tested on macOS 15.4.1 (local) and Debian GNU/Linux 11 (cloud)
 - Python version(python -V): tested on 3.9.6 & 3.12.10 (local) and 3.10.14 (cloud)
 - ADK version(pip show google-adk): 0.3.0

allengour:
Fixes #437

* Slice `events[i+1:]` instead of `events[i]` to avoid returning the incorrect index
* Change `elif config.after_timestamp` to independent `if` so both config filters can be applied
* Add unit test covering cases: 
  ```bash
  pytest tests/unittests/sessions/test_session_service.py -k test_get_session_with_config
  ```
  image

Updated behavior:
image


allengour:
Note, for the test, I did not include `SessionServiceType.DATABASE` in the `service_type` parameter because there appears to be a separate issue with sqlalchemy in the source code:
image

DeanChensj (approved):

DeanChensj:
Hi @allengour , could you update your branch so that we can merge? Thanks!

DeanChensj (approved):

hangfei:
It's in here: https://github.com/google/adk-python/commit/1393965720a8d364d8d6b979fc70bf1360696fed.

Issue #431, opened 2025-04-28
have "adk [command]" simply return the options for that command instead of the truncated error message


jsteinberg-rbi:
It's a fairly common convention to simply return the entire options of a particular CLI command if no options are given, assuming that _that_ particular command actually requires options. It would be much more convenient, for example, if when I execute something like `adk create` instead of getting the heavily truncated and redundant message below, I simply am returned all the options of `adk create` thus sparing me of having to do the extra `--help`. We can keep `--help` naturally, but making help a synonym of the command is better imo because `Try 'adk create --help' for help.` is redundant at this point in time regarding CLIs.

```
Usage: adk create [OPTIONS] APP_NAME
Try 'adk create --help' for help.

Error: Missing argument 'APP_NAME'.
```

**Describe the solution you'd like**
Make `--help` a synonym of `adk [command]`

**Describe alternatives you've considered**
There are none.

bl3ck:
Hi @hangfei, please I've already started working on this issue and have made some good progress (screenshot attached below). I'd love to collaborate with you to get it fully resolved. Would you be able to assign it to me so I can continue and submit a PR for your review?

### Before:
<img width="570" alt="Image" src="https://github.com/user-attachments/assets/dda9c2eb-855b-4405-8c91-e477b293f964" />

### After:
<img width="566" alt="Image" src="https://github.com/user-attachments/assets/c7317718-9754-4284-9fb1-143c5cf60a2c" />

bl3ck:
Hi @hangfei please any update on this?

hangfei:
@bl3ck thanks. I see you have committed the change and it's fixed.

PR #443, opened 2025-04-29
fix: Display full help text for adk create  CLI command when argument…

bl3ck:
fix: Display full help text for CLI commands when arguments are missing

Addresses issue #431 by implementing a custom Click Command class that shows the complete help text instead of just an error message when a required argument is missing. This improves user experience by following the common CLI convention where commands without options display their full help text.

- Added `HelpfulCommand` class to override default Click error behavior
- Applied the custom class to the "`create`" command
- Maintains the original error message at the end for clarity

This change makes the CLI more intuitive and saves users from having to
run an additional `--help` command to see available options.

How to test:
- Run `adk create` without the `APP_NAME`


hangfei:
Thanks! @bl3ck could you show before/after command line log?

hangfei (approved):

bl3ck:
> Thanks! @bl3ck could you show before/after command line log?

Hi @hangfei, Thanks for getting back to me on this. Here's the command line log:

### Before:
Image

### After:
Image

hangfei (approved):

hangfei:
@bl3ck The error message is lower case "app_name". Could we match to its original case("APP_NAME")?

bl3ck:
> @bl3ck The error message is lower case "app_name". Could we match to its original case("APP_NAME")?

@hangfei The update has been made. Sorry, for the delayed response.

bl3ck:
@hangfei The code was merged before without my last change. I've created a new [PR](https://github.com/google/adk-python/pull/724) which ensures the case is matched in the error message.

hangfei:
Thanks! Let's move the discussion to the other thread.

PR #448, opened 2025-04-29
feat: make new_message optional in Runner.run_async to support session resumption

idootop:
## Summary
This PR modifies the `run_async` method to make the `new_message` parameter optional,
allowing the agent execution flow to resume from an existing session when no new
message is provided.

图片

## Changes
- Changed `new_message` parameter type from required `types.Content` to `Optional[types.Content]`
- Added default value of `None` for `new_message` parameter

## Motivation
This change enables:
1. Session resumption capability by allowing the agent to continue processing
   without requiring a new message
2. More flexible usage patterns where message input might not always be available
3. Better support for long-running conversations that may be paused and resumed

## Impact
✅ Existing calls with explicit `new_message` will continue working as before.
✨ New calls can omit `new_message` to resume from session state.

### DIFF
--- a/src/google/adk/runners.py
+++ b/src/google/adk/runners.py
@@ -102,7 +102,7 @@ class Runner:
       *,
       user_id: str,
       session_id: str,
-      new_message: types.Content,
+      new_message: Optional[types.Content] = None,
       run_config: RunConfig = RunConfig(),
   ) -> Generator[Event, None, None]:
     """Runs the agent.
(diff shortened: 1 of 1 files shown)


boyangsvl:
Making session resumable is complicated, and we will need to handle edge cases like when the session was interrupted and not closed properly, etc. I think a lot more work is needed to make it work well. For now I will reject the PR but please feel free to reopen if you addressed all edge cases.

paward35:
Sad this was closed, it would make our evals a lot easier since we need to rerun old conversations and don't want to separate out the last user message

Issue #458, opened 2025-04-29
transfer_to_agent tool fails with unexpected kwarg


cschmidterNBTC:
**Describe the bug**
transfer_to_agent tool fails with unexpected kwarg.

The root agent called the transfer_to_agent tool with a keyword argument `query` which raised a TypeError.

**To Reproduce**
1. Use a multi-agent framework
2. Let the root agent call the transfer_to_agent tool
3. The agent may pass in an unexpected keyword argument to the tool call.

**Expected behavior**
Ideally, an agent should never call a tool with unexpected keyword args.
A bandage solution would be to pass kwargs through the function:
```python
def transfer_to_agent(agent_name: str, tool_context: ToolContext, **kwargs):
  """Transfer the question to another agent."""
  tool_context.actions.transfer_to_agent = agent_name
```

**Desktop:**
 - OS: Windows 11
 - Python version: 3.13
 - ADK version: 0.3.0

**Additional context**
This error was encountered using gemini-2.5-pro-preview-03-25 and gemini-2.0-flash-001.

AlankritVerma01:
@Jacksunwei @hangfei 
Can you please review the issue + PR and assign me the issue?
Additionally if the PR is fine, start the process to merge it?
Thanks!

Jacksunwei:
As I mentioned in the PR. 

I think the better approach could be improving the docstring, which will be in the prompt.

I guess the `question` in the docstring might cause model to fill query as arg. 

Could you try adding docstring for agent_name argument will work?

AlankritVerma01:
Hey @Jacksunwei 
Thanks for the feedback!
i have updated the docstring as suggested.
Please let me know if its fine now!
Thank you!

PR #466, opened 2025-04-30
Tools/transfer accept kwargs

AlankritVerma01:
Fixes: #458

**Background**  
When an LLM‐driven agent called `transfer_to_agent` with extra parameters (e.g. `query`, `temperature`), the tool raised:

```text
TypeError: transfer_to_agent() got an unexpected keyword argument 'query'
```

because its signature only accepted `agent_name` and `tool_context`.

**With this change, we:**
- Update `src/google/adk/tools/transfer_to_agent_tool.py` to:
  ```python
  def transfer_to_agent(
      agent_name: str,
      tool_context: ToolContext,
      **kwargs,
  ):
      """Transfer the question to another agent."""
      tool_context.actions.transfer_to_agent = agent_name
  ```
  so that any unexpected `kwargs` are silently ignored.
- Expand `tests/integration/tools/test_transfer.py` to include:
  1. A call with a **single** extra kwarg (`query`)  
  2. A call with **multiple** extra kwargs (`query`, `temperature`, `max_tokens`)  
  Each test is parametrized to run against both the GOOGLE_AI and VERTEX backends.

---

### Test Plan
1. Run the targeted integration tests:
   ```bash
   pytest tests/integration/tools/test_transfer.py
   ```
2. Run the full suite to ensure no regressions:
   ```bash
   pytest
   ```

---

### Checklist
- [x] New integration tests added and passing  
- [x] Existing behavior unchanged for calls without extra kwargs  
- [x] Code follows repository style and linting rules  
- [x] References issue with **Fixes: #458**  

---

**Next Steps / TODO**  
Per the existing TODO comment in `transfer_to_agent_tool.py`, we could further internalize this tool (e.g. rename to `_transfer_to_agent` or move into an `internal/` submodule) so it doesn’t leak into the public API.

### DIFF
--- a/src/google/adk/tools/transfer_to_agent_tool.py
+++ b/src/google/adk/tools/transfer_to_agent_tool.py
@@ -16,6 +16,6 @@ from .tool_context import ToolContext
 
 
 # TODO: make this internal, since user doesn't need to use this tool directly.
-def transfer_to_agent(agent_name: str, tool_context: ToolContext):
+def transfer_to_agent(agent_name: str, tool_context: ToolContext, **kwargs):
   """Transfer the question to another agent."""
   tool_context.actions.transfer_to_agent = agent_name
--- a/tests/integration/tools/test_transfer.py
+++ b/tests/integration/tools/test_transfer.py
@@ -1,0 +1,31 @@
+import pytest
+from google.adk.tools import transfer_to_agent
+
+class DummyActions:
+    def __init__(self):
+        self.transfer_to_agent = None
+
+class DummyToolContext:
+    def __init__(self):
+        self.actions = DummyActions()
+
+def test_transfer_to_agent_sets_correct_agent():
+    ctx = DummyToolContext()
+    transfer_to_agent(agent_name="math_agent", tool_context=ctx)
+    assert ctx.actions.transfer_to_agent == "math_agent"
+
+def test_transfer_to_agent_ignores_single_extra_kwarg():
+    ctx = DummyToolContext()
+    transfer_to_agent(agent_name="history_agent", tool_context=ctx, query="When was WWII?")
+    assert ctx.actions.transfer_to_agent == "history_agent"
+
+def test_transfer_to_agent_ignores_multiple_extra_kwargs():
+    ctx = DummyToolContext()
+    transfer_to_agent(
+        agent_name="code_agent",
+        tool_context=ctx,
+        query="foo",
+        temperature=0.5,
+        max_tokens=100
+    )
+    assert ctx.actions.transfer_to_agent == "code_agent"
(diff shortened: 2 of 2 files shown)


Jacksunwei:
Thanks for the PR! 

But I think this is not the way to go. Model shouldn't be passing `query` to this tool. We need inspect the prompt and see why. 

A better approach could be improving the docstring, which will be in the prompt.

AlankritVerma01:
Thanks @Jacksunwei  for the feedback! I’ve removed the `**kwargs` change and updated the docstring to clarify that only `agent_name` and `tool_context` are accepted and that parameters like `query` belong in the LLM prompt. Let me know if you’d like any further tweaks.

src/google/adk/tools/transfer_to_agent_tool.py
```diff
@@ -17,5 +17,18 @@
...
 def transfer_to_agent(agent_name: str, tool_context: ToolContext):
-  """Transfer the question to another agent."""
+  """Transfer the current request to another agent.
+
+  This tool hands off control to a different named agent.  Any additional
+  context (e.g. user query, temperature settings) should be carried in the
+  LLM prompt or tool_context, not passed as extra parameters here.
+
+  Args:
+    agent_name: The identifier of the agent to transfer to (e.g. "math_agent").
```
Jacksunwei: Let's remove math_agent to avoid confusing LLM.

src/google/adk/tools/transfer_to_agent_tool.py
```diff
@@ -17,5 +17,18 @@
...
+    tool_context: The current ToolContext, whose `actions.transfer_to_agent`
```
Jacksunwei: tool_context is not for LLM use, so we don't want to mention it.

src/google/adk/tools/transfer_to_agent_tool.py
Jacksunwei: s/identifier/name.

src/google/adk/tools/transfer_to_agent_tool.py
```diff
@@ -17,5 +17,18 @@
...
+      field will be set.
+
+  Returns:
```
Jacksunwei: We can just remove returns section, since there is none.

src/google/adk/tools/transfer_to_agent_tool.py
```diff
@@ -17,5 +17,18 @@
 
 # TODO: make this internal, since user doesn't need to use this tool directly.
 def transfer_to_agent(agent_name: str, tool_context: ToolContext):
-  """Transfer the question to another agent."""
+  """Transfer the current request to another agent.
+
+  This tool hands off control to a different named agent.  Any additional
```
Jacksunwei: These are likely not useful to LLM, for prompt, it's better to be direct and concise.

Jacksunwei (approved):

hangfei:
merged in https://github.com/google/adk-python/commit/854a5440614590c2a3466cf652688ba57d637205. Thanks!

PR #483, opened 2025-05-01
fix: Solved#402

BhadrakshB:
I added a check on the instance of self.agent. If it is of type BaseAgent, the program runs through. If there is an error, a clear description is showing saying there is an error the the root_agent definition

### DIFF
--- a/src/google/adk/runners.py
+++ b/src/google/adk/runners.py
@@ -395,9 +395,7 @@ class Runner:
         self.agent.tools.append(built_in_code_execution)
 
     if not isinstance(self.agent, BaseAgent):
-      raise ValueError(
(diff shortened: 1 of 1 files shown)


knileshh:
Hey @BhadrakshB, What are your thoughts on this text for the error?
"Invalid parent agent: The provided 'agent' must be an instance of 'BaseAgent' or a subclass of 'BaseAgent'. "

hangfei:
Thanks for your contribution. Please create an associated issue first(per https://github.com/google/adk-python/blob/main/CONTRIBUTING.md)

BhadrakshB:
I did create one. Issue#402

hangfei:
@BhadrakshB 
I think Pydantic already checks it? Could you share the error message without your change?

PR #520, opened 2025-05-03
Fix install command for Zsh compatibility.

Imfuyuwei:
Fix install command for Zsh compatibility. Wrapped extras list in quotes to prevent Zsh from expanding it as a glob pattern.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,7 +41,7 @@ This project follows
 3.  **Install dependencies:**
 
     ```shell
-    uv pip install -e .[dev,test,extensions,eval]
+    uv pip install -e ".[dev,test,extensions,eval]"
     ```
 4.  **Run unit tests:**
 


hangfei:
Could you fix the workflow errors? @Imfuyuwei

Imfuyuwei:
Thank you Hangfei @hangfei. Is the next step that you add the label to decide whether to merge the PR? If so, I'll leave it to you :)

BTW, not sure why the unit test checks are failing.

hangfei:
Yes. I will merge later.

There is test failure on head and need to fix that first. (fix here: https://github.com/google/adk-python/pull/538)

CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@ This project follows
 3.  **Install dependencies:**
 
     ```shell
-    uv pip install -e .[dev,test,extensions,eval]
+    uv pip install -e ".[dev,test,extensions,eval]"
```
Jacksunwei: `uv sync --all-extras` is better. For testing with specific extras. E.g. `uv sync --extra test --extra eval`

Imfuyuwei: Changed to  `uv sync --all-extras` instead

Jacksunwei (approved):
Thank you, Yuwei!

Issue #526, opened 2025-05-03
How to exit a loop agent correctly if the loop agent is a sub agent of another loop agent


TequilaSunset2023:
Hi, adk team,

I have a loop agent as a sub agent of a loop agent, when I call escalate in the sub loop, it breaks both the sub loop and  parent loop, is this a bug or by design? How can I work around it?

nx90:
I created a PR to fix this issue, https://github.com/google/adk-python/pull/535

ErZhouEr:
Same problem, seems to be unresolved yet

klateefa:
Hi @TequilaSunset2023  and thanks to everyone who participated in this thread.

We're currently doing some housekeeping on our issue tracker. Given the rapid pace of development and the number of updates since this was last active, there's a good chance this issue has been resolved in a newer version. To help us keep our backlog focused on current issues, we are closing this as stale.

If you're still facing this problem, could you please update to the very latest version of the library and confirm the issue still exists? If it does, please comment here, and we will be happy to reopen this issue and investigate further.

Thanks for your contribution and understanding!”

PR #535, opened 2025-05-03
Fix nested loop agents escalation issue

nx90:
In nested loop agents scenario, the event.action.escalate will break all loop agents, the expected behavior should be only break one level of the loop agents.

Issue link: https://github.com/google/adk-python/issues/526

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@
 ### ⚠ BREAKING CHANGES
 
 * Auth: expose `access_token` and `refresh_token` at top level of auth
-  credentails, instead of a `dict`
+  credentials, instead of a `dict`
   ([commit](https://github.com/google/adk-python/commit/956fb912e8851b139668b1ccb8db10fd252a6990)).
 
 ### Features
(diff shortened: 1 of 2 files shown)


Jacksunwei (approved):

Jacksunwei:
Actually I think the action should not be changed. We should check whether there is no another loop agent between the current agent and the event author.

src/google/adk/agents/loop_agent.py
```diff
@@ -48,9 +48,12 @@ async def _run_async_impl(
     while not self.max_iterations or times_looped < self.max_iterations:
       for sub_agent in self.sub_agents:
         async for event in sub_agent.run_async(ctx):
-          yield event
           if event.actions.escalate:
+            # Reset the escalate status to make sure this escalation only exits one level of the loop agent, not the entire nested loop agents structure.
+            event.actions.escalate = False
```
Jacksunwei, line 59: We should instead check whether this is no more loop agent between this agent and the event author agent. 

---

`event.actions.escalate` is the conversation history and we shouldn't alter it.

Jacksunwei:
@selcukgun could you take over and finish this PR?

rohityan:
Hi @nx90 and thanks to everyone who participated in this thread.

Issue #537, opened 2025-05-03
Error when manually updating events with DatabaseSessionService


byt3bl33d3r:
I'm using `DatabaseSessionService` and I have the following code that creates an event with `EventActions.state_delta` which seems to be the correct way to update the state manually (according to [the docs](https://google.github.io/adk-docs/sessions/state/#how-state-is-updated-recommended-methods))

```python

system_event = Event(
    invocation_id="user_tool_denial",
    author="system",
    actions=EventActions(state_delta={
        "user:tool_call_denied": func_approval_id
    })
)
session_service.append_event(session, system_event)
```

however I'm getting the following error message when calling `runner.run_async()`:

```python
  File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/runners.py", line 199, in run_async
    self.session_service.append_event(session=session, event=event)
  File "/workspaces/agents/.venv/lib/python3.12/site-packages/google/adk/sessions/database_session_service.py", line 485, in append_event
    raise ValueError(
ValueError: Session last_update_time 1746305857.067034 is later than the upate_time in storage 2025-05-03 20:57:48.274444
```

The culprit seems to be this check here:

https://github.com/google/adk-python/blob/main/src/google/adk/sessions/database_session_service.py#L485C1-L489C10

When commenting these lines out, everything works as expected.

This seems like a bug as the storage time will always be later then the session.last_update time?

harikrishnang33:
This issue seems to be because of our local session not getting updated with the remote session.
All our events will be appended and we can see those in the respective database that we use.

harikrishnang33:
@hangfei  This issue seems to be because of our session not getting updated with the remote storage session. There is an explicit [statement](https://github.com/google/adk-python/blob/e4317c9eb7b85981b49d8179151ac135cf87f286/src/google/adk/sessions/database_session_service.py#L552) to update the session last update time with the storage session update time. But still it is not getting updated which is an issue along with the wrong check if statement. Ideally both the timestamps should be equal IMO.

DeanChensj:
Hi @byt3bl33d3r, this if statement meant to check if the in-memory session is stale, where DB session already got new updates, while the in memory one is out-of-date. It is likely that you are using an outdated session to append event.

harikrishnang33:
@DeanChensj Just curious and thinking out loud, should the if condition be checking 
```
if storage_session.update_time.timestamp() != session.last_update_time:
```
instead of 
```
[...]

PR #571, opened 2025-05-05
Fixes DatabaseSessionService Storage Time timestamp comparison.

byt3bl33d3r:
Addresses #537 

Based on the text of the exception , I'm pretty sure this was the intended check.


hangfei:
@DeanChensj Could you help review? 

@byt3bl33d3r could you fix other errors first?

DeanChensj:
Hi @byt3bl33d3r, this if statement meant to  check if the in-memory session is stale, where DB session already got new updates, while the in memory one is out-of-date.  Thanks for the PR, we'll fix the confusing error message.

PR #579, opened 2025-05-06
fix: format by pyink on database_session_service.py

anencore94:
Fix failed pyink format on current main branch by `uv run pyink  --config pyproject.toml ./src`

### DIFF
--- a/.github/workflows/python-unit-tests.yml
+++ b/.github/workflows/python-unit-tests.yml
@@ -2,9 +2,9 @@ name: Python Unit Tests
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


hangfei:
we are using two-space. is there special rule for this case?

anencore94:
> we are using two-space. is there special rule for this case?

Thx for your reply :) @hangfei 
I'm not sure what is the intended convention, but as following from the [CONTRIBUTING.md](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md#development-setup), `pyink` automaticallly format that file as four spaces for ValueError wrapping.

image


Maybe it depends on the version of pyink or python? if it doesn't raise an issue in workflow?

However, as I see the other code lines, use four spaces when we raise some Error class
image
image

hangfei:
@Jacksunwei could you help take a look why pyformat or pyink is not working?

hangfei:
closing as this is not valid anymore.

PR #601, opened 2025-05-07
fix(tests): use mock GCS client in artifact service tests to avoid real credentials

thiagoneves:
This PR updates the artifact service tests to properly inject a mock GCS client, avoiding the need to load real Google Cloud credentials when running the tests.


hangfei:
Thanks. This is a good fix!

hangfei:
Thanks. It's merged by bot.

PR #604, opened 2025-05-07
fix: get updated session before appending events

harikrishnang33:
Addresses https://github.com/google/adk-python/issues/537 temporarily

Changed the update time check logic to the below code since we are assigning session.last_update_time in the end of the database session service with storage session update time timestamp
```
if storage_session.update_time.timestamp() != session.last_update_time:
        raise ValueError(
            "Session last_update_time"
            f" {datetime.fromtimestamp(session.last_update_time):%Y-%m-%d %H:%M:%S} is"
            " not equal to update_time in storage"
            f" {storage_session.update_time:%Y-%m-%d %H:%M:%S}"
        )
"""
Already existing code
"""
session.last_update_time = storage_session.update_time.timestamp()
```


harikrishnang33:
@hangfei @DeanChensj Please check this

harikrishnang33:
@hangfei @DeanChensj Could you please review this PR?

benorama:
Minor perf improvement: only refresh the session if required (if the session is not outdated, no need to do an additional DB request) ?

harikrishnang33:
> Minor perf improvement: only refresh the session if required (if the session is not outdated, no need to do an additional DB request) ?

@benorama Only when we check the storage_session(db) and session (in memory) update time, we get to know that the upate time is not in sync. So should we check if the session is not updated, and then get session and then check if the session is not updated again and then throw the error?

hangfei:
@DeanChensj to help take a look?

hangfei:
@harikrishnang33 the tests are failing. could you help take a look?

DeanChensj:
Hi @harikrishnang33 , thanks for the PR. We want to make sure the session object we are appending the new event is not stale, that's why we don't fetch it during the append event.

harikrishnang33:
Thanks for the clarification @DeanChensj . I will close this PR.

PR #605, opened 2025-05-07
fix: improve json serialization by allowing non-ascii characters

nauyiahc:
Updated _safe_json_serialize to use json.dumps(obj, ensure_ascii=False). This allows non-ASCII characters to be preserved in JSON output, improving readability and compatibility for internationalization.

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -129,7 +129,7 @@ def _safe_json_serialize(obj) -> str:
 
   try:
     # Try direct JSON serialization first
-    return json.dumps(obj)
+    return json.dumps(obj, ensure_ascii=False)
   except (TypeError, OverflowError):
     return str(obj)
 


boyangsvl:
I don't know if ensure_ascii is the problem. Ideally even if Chinese character is encoded in dumps(), once you loads() it should restore back to the correct characters.

nauyiahc:
> I don't know if ensure_ascii is the problem. Ideally even if Chinese character is encoded in dumps(), once you loads() it should restore back to the correct characters.

The input to the LLM is a JSON string. I think setting ensure_ascii=False would help the LLM handle non-ASCII responses better.

selcukgun (approved):

selcukgun:
Assuming this is useful for logging purposes (regarding json.loads not needing this).

selcukgun:
merged by copybara: https://github.com/google/adk-python/commit/c4cd1802bf86abb82b0e0fa016a9b2ca9d7917ad

Issue #512, opened 2025-05-02
Support for Immutable Session State Keys


triangle-z:
I have a case where some state keys shouldn't be modified by the agent. Similar keys are in the documentation such as `user:login_count` which should ideally only be modified by the code itself to ensure that it value is correct. Currently, the session state is fully mutable, which creates risk (low but still not 0) for accidental overwrites, inconsistencies in workflows, and unintended side effects.

---

I know that this is against the initial purpose of the state, which is to be the agent's scratchpad, but it might be helpful to add support for immutability in session state keys.

Potential implementation options:

1. **`StateValue` object**  
   A `StateValue` class with an optional `mutable` flag. If a raw value is assigned, it defaults to `mutable=True`.  
   ```python
    state_changes = {
        "task_status": "active",   # Equivalent to StateValue(session.state.get("user:login_count", 0) + 1), default mutable=True
        "user:login_count": StateValue(session.state.get("user:login_count", 0) + 1, mutable=False), # Update user state
    }
   ```

2. **Prefix convention**  
   Use a reserved prefix like `immutable:` for the read-only keys.  
   ```python
   session.state["immutable:booking_id"] = "12345"
   ```

3. **Immutable key list**  
   Support a config-level list (or method) of keys that should not be modified after initial assignment.  
   ```python
   IMMUTABLE_KEYS = ["user:login_count", "booking_id"]
   ```

---

There are of course some alternatives if this is not a good idea :

- Enforcing immutability in the agent logic, but this is error-prone and scattered across the codebase.
- Using `before` callbacks to manually prevent modification of certain keys, but this leads to spaghetti code and lacks a centralized mechanism.
- Fetching data directly from another source in the tools that need it (e.g., database, in memory ...) instead of storing it in state, but this also leads to spaghetti code and decouples business logic from conversational context.

---

If this aligns with the project, I'm happy to contribute a pull request and also update the documentation.

triangle-z:
I've implemented a proposed solution based on the `StateValue` wrapper and committed it here: triangle-z/adk-python@6a4bc27727834c336746047dec585008c60d220e. If you believe this approach is viable, I’d appreciate your input on a few questions:

1. **Silent failure vs. exception**:
   When the LLM (or a user) tries to modify an immutable key, I currently prevent the update and just print a warning, but I don’t raise an exception. I did this because currently, the LLM is not aware of which keys are immutable, so raising would cause unnecessary exceptions.
[...]

PR #615, opened 2025-05-07
Add support for immutable state keys using StateValue wrapper

triangle-z:
This PR addresses the feature request in #512, by introducing a `StateValue` wrapper to support immutability of session state keys.

---

# Changes made

* Introduced a `StateValue` class that allows marking a value as `mutable=False`.
* Enforced that all values stored in session state (`_value` and `_delta`) are wrapped in `StateValue`, even if initially provided as raw values.
* Added logic in the `State` class to prevent updates to immutable values.
* Added serialization in `StateValue` to support pydantic.
* Moved `State` and `StateValue` into a dedicated `state/` folder to avoid circular dependencies. I’m aware this is a relatively large structural change and open to suggestions on how best to modularize it with minimal disruption.

---

# Test results

I ran the tests on both the original `main` branch and my feature branch. A few tests fail in both cases — these issues appear to be unrelated to my changes.

I’ve attached the two screenshots:
* One from the original branch (`main`)
Tests screenshot main branch

* One from my feature branch (with `StateValue` support)
Tests screenshot feature branch

---

# Open questions for maintainers

I have several questions and design decisions that I'd like feedback on before finalizing this PR:

1. Behavior on update attempt of immutable key
   I currently prevent the update and log/print a warning instead of raising an exception, since the LLM may not know which keys are immutable.
   → Should this be a hard error (exception), or is a soft failure preferred?

2. Should the LLM or agent logic be made aware of immutability?
   Right now, immutability is internal to the state system. Should metadata about immutability be exposed to agents?

3. Precedence between `_value` and `_delta`
   In the case of conflicting mutability (e.g., the same key exists in `_value` and `_delta` with different mutability flags), I’ve made `_delta` take precedence.
   → Is this the intended resolution logic?

4. **Serialization/Deserialization of `StateValue`**
   Currently, only the underlying value is serialized (e.g., `StateValue(value="123", mutable=True)` becomes `"123"`). This means immutability is lost on deserialization. I avoided changing the serialization logic too deeply to stay aligned with the current ADK architecture.
   → What’s the preferred approach here: minimal change with this tradeoff, or a more structured serialization format that preserves metadata?

---

I’m tagging the participants on the original issue for visibility:
@boyangsvl @hangfei @DeanChensj

Please let me know your thoughts on the questions above — I’m happy to iterate on this implementation or adapt it to better align with the project’s design principles.


triangle-z:
Hi @boyangsvl @hangfei @DeanChensj,

I made a few updates to the PR:

* Questions 1, 2, and 3 are still open and I'd appreciate your input on those.
* For question 4, I found it too fragile to implement serialization without deserialization, so I updated it. Now the immutability metadata is preserved during both serialization and deserialization.
* That said, calling `str()` or `State.to_dict()` still only returns the underlying values without immutability info — which ties back to question 2 about whether this metadata should be exposed.
* I also added a few tests to cover the updated behavior :

Capture d'écran 2025-05-09 150052

* Here's the screenshot of the main branch’s test results (with 0 of my modifications) :

Capture d'écran 2025-05-07 231136

Let me know what you think, happy to revise based on your feedback!

surajksharma07:
Hi @triangle-z and thanks to everyone who participated in this thread.

Thanks for your contribution and understanding!”

Issue #606, opened 2025-05-07
Only the first part's text is returned from last_event.content.parts, causing loss of multi-part responses


TP9-dev:
agent_tool.py

Description:
The code returns only the text of the first part in last_event.content.parts:
```
tool_result = last_event.content.parts[0].text
```

This implementation causes issues when the response contains multiple parts, as only the first part's text is returned and any additional parts are ignored. As a result, important information in subsequent parts is lost, leading to incomplete or incorrect tool outputs.

Expected behavior:
All parts' texts should be concatenated (e.g., joined with \n) and returned as a single string, so that multi-part responses are handled correctly.

Suggested fix:
Replace the above line with:
```
tool_result = "\n".join(
    [p.text for p in last_event.content.parts if getattr(p, "text", None)]
)
```
This will ensure that all available text parts are included in the tool result.

wuliang229:
@TP9-dev To help us investigate and reproduce, do you have an example code snippet of when a multi-Part response is used?

wuliang229:
Should have been fixed by [this commit](https://github.com/google/adk-python/commit/933f01b89f0212e06079f553db7d7ecdcb21cd9b) Please re-open if you see this issue again.

PR #618, opened 2025-05-08
Fix: concatenate all text parts in last_event.content.parts

TP9-dev:
Previously, only the first part's text was returned, causing incomplete tool outputs when multiple parts were present. Now, all parts' texts are joined with newlines to ensure complete responses.

Related to: https://github.com/google/adk-python/issues/606


wuliang229:
Should have been fixed by [this commit](https://github.com/google/adk-python/commit/933f01b89f0212e06079f553db7d7ecdcb21cd9b)

Issue #398, opened 2025-04-26
Support Enum type for function tool args


paullpp:
** Please make sure you read the contribution guide and file the issues in the rigth place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Is your feature request related to a problem? Please describe.**
Currently function tools can only have primitive / simple types. It'd be useful to support Enums, even other complex types, though this issue will focus on Enums.

**Describe the solution you'd like**
Support for Enum arguments in function tool calls

**Describe alternatives you've considered**
In the case of Enums, we've considered using literals instead, but find that the Agent can make values up rather than using values in the literal. A workaround is to make the function parameter of string type, and put the valid options into the functions docstring. However, that's tedious for large enums.

**Additional context**
N/A

kavinkumar807:
@paullpp can you please give me more context?

thomasglow:
Example error:
```
GenerateContentRequest.tools[0].function_declarations[32].parameters.properties[countryCode].enum[0]: cannot be empty\n', 'status': 'INVALID_ARGUMENT'}}
```

juszzz:
@lucasnobre212 opened a PR to resolve this issue back in May and it seems like its still waiting on a review. Is there anyone with merge permissions who can take a look at it? There are a lot of auto-generated client libraries that make liberal use of enums and it would be nice to drop those library methods in as tools to our agents.

dennisgera:
I'm also running into this issue. Would love to have functions accept an enum as a parameter. Just looked at @juszzz's PR, LGTM. Does anyone know when we can have that merged? Is anyone allowed to approve adk PRs?

jpantsjoha:
I've submitted a documentation PR with Enum workarounds while we wait for native support:

**PR**: https://github.com/google/adk-docs/pull/786

The type annotation guide documents using `typing.Literal` as an interim solution:

```python
from typing import Literal, Dict, Any

def configure_setting(
    mode: Literal["auto", "manual", "disabled"]
) -> Dict[str, Any]:
    """Set configuration mode to one of the allowed values."""
    return {"mode": mode, "updated": True}
```

Benefits:
- Type checkers validate the literal values
- Schema generation includes the enum constraint
- Clear API contract for LLM

This should help developers until native Enum support is implemented (tracked in this issue and #2733).

Jacksunwei:
Thanks @jpantsjoha for documenting the `typing.Literal` workaround in the ADK docs PR.

[...]

PR #625, opened 2025-05-08
feat: Support Enum parameter parsing in function declarations

lucasnobre212:
Added handling for Enum types in `function_parameter_parse_util.py` to enable parsing of Enum parameters. 
Updated schema generation to include enum values and validated default values. 
Included a unit test to ensure correctness of the implementation.

Observations: 
- This is my first PR, so please tell me if I need to provide more context or make more changes 
- I added a unit test for Enum. I couldn't run a integration test though, I tried to run only test_single_function_calls_success from test_tools.py but pytest kept skipping it. 

Related Issues:
#398

### DIFF
--- a/src/google/adk/flows/llm_flows/_code_execution.py
+++ b/src/google/adk/flows/llm_flows/_code_execution.py
@@ -152,7 +152,7 @@ class _CodeExecutionResponseProcessor(BaseLlmResponseProcessor):
     if llm_response.partial:
       return
 
-    for event in _run_post_processor(invocation_context, llm_response):
+    async for event in _run_post_processor(invocation_context, llm_response):
       yield event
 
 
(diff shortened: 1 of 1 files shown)


lucasnobre212:
Should I add default compatibility? The documentations states "Do not set default values for parameters. E.g., def my_func(param1: str = "default"). Default values are not reliably supported or used by the underlying models during function call generation. All necessary information should be derived by the LLM from the context or explicitly requested if missing.", but there are some types that are compatible, as in _is_default_value_compatible

paullpp:
There is one more thing missing. The function needs to actually be called with an enum value. Currently it will try passing a string, which fill fail if the function tries accessing the enum value. Example below:
Screenshot 2025-05-09 at 12 51 48 PM
Screenshot 2025-05-09 at 12 51 52 PM

To actually make it pass an Enum value, you need to modify `run_async` in `function_tool.py`. At least that's how we got it to work, maybe there's a better place to make that modification.

Also, I believe this will not be able to handle 
1) Optional enum parameters
2) enum parameters in a list

you'd need to add extra logic in both `function_tool.py` and `function_parameter_parse_util.py` to support those, but could be out of scope for this PR. just thought I'd mention it

lucasnobre212:
@paullpp thanks for the insights! I agree with everything you said. First, about actually passing the Enum value, is it possible for the LLM to actually pass the Enum to the tool? When working with enums the json schemas are always defined as strings (https://ai.google.dev/gemini-api/docs/function-calling?example=meeting#step_1_define_function_declaration). The idea of passing an enum is to limit the strings the LLM can choose from, but of course, I can be missing something.
I didn't check the other two points, but I can implement them if needed!

lucasnobre212:
I will need to create a new PR to fix the branch I have in my fork. Sorry!

Issue #522, opened 2025-05-03
base_llm_flow.py has running problem


spcBackToLife:
**Describe the bug**
in `base_llm_flow.py#run_async`, 
```python
    async def run_async(
        self, invocation_context: InvocationContext
    ) -> AsyncGenerator[Event, None]:
        """Runs the flow."""
        while True:
            last_event = None
            async for event in self._run_one_step_async(invocation_context):
                last_event = event
                yield event
            if not last_event or last_event.is_final_response():
                break

```
When the model's output exceeds `max_token`, there will be a loop and the logic will be executed again not continue to write the rest content

hangfei:
Thanks for reporting this. 

Could you provide an example to reproduce this error?

spcBackToLife:
ok, here is my example:
- agent
```python
# @title Define the Weather Agent
# Use one of the model constants defined earlier
from app.adk_agents.agents.render_agent.output import PageAnalysis
from app.adk_core.adk.agents import LlmAgent
from app.adk_core.adk.models.lite_llm import LiteLlm

# 用于分析页面内容元素
analysis_page_agent = LlmAgent(
    name="analysis_page_agent",
    model=LiteLlm(
        model="gpt-4.1-mini",
        api_key="xxx",
        # max_tokens=32768,
        api_base="xxxx",
    ),
    description="""
    你是顶级的 UI 分析师，能够对 UI 需求进行详细分析。它会先判断需求中页面或子页面的类型，接着确定页面下模块的情况及重要性排序，
    然后深入分析模块内分组和元素的优先级，帮助清晰梳理 UI 需求结构，明确各部分的重要程度。
# 相关分析概念定义
## 元素 - element
元素是构成 UI 界面的最基础、不可再分的单元，它们具有明确的功能和视觉表现，具体含义示例如下：
- **文本**：用于传达信息，如标题、说明文字、标签等，具有不同的字体、字号、颜色和样式。
- **数字**：用于呈现具体的数据，如统计数字、价格、数量等，可根据设计需求进行格式设置。
- **图标（Icon）**：以简洁的图形形式表达特定的含义，具有识别性高、占用空间小的特点，可增强界面的直观性和易用性。
- **图片**：用于展示视觉内容，如产品图片、宣传海报、用户头像等，可丰富界面的视觉效果。
- **基础组件**：具备特定交互功能的元素，常见的有：
    - **输入框**：供用户输入文本、数字等信息，可设置输入类型、长度限制、默认值等。
[...]

PR #659, opened 2025-05-10
fix(llm): prevent base llm flow truncation (#522)

AlankritVerma01:
Fixes: #522

Previously, `BaseLlmFlow.run_async` would loop indefinitely whenever the LLM response was truncated due to hitting its `max_tokens` limit. The loop’s break condition only checked for “no events” or `last_event.is_final_response()`, neither of which fires for a partial/truncated chunk—so the generator re-invoked `_run_one_step_async` endlessly.

With this change, we:

- Treat any partial (truncated) event as terminal by adding `or last_event.partial` to the break condition in `run_async`.
- Ensure that when an LLM emits a truncated chunk (`partial=True`), `run_async` yields it exactly once and then terminates, avoiding infinite loops.

Test Plan
---------
- **New unit test** `test_base_llm_flow_truncation.py` verifies that:
  - Given a flow that only ever yields one partial event, `run_async` emits exactly one event and then stops.
- Run it with:
  ```bash
  pytest tests/unittests/flows/llm_flows/test_base_llm_flow_truncation.py


AlankritVerma01:
Hey @Jacksunwei @hangfei 
Please review the PR
Thanks team!

Jacksunwei (approved):

Jacksunwei:
Merged in https://github.com/google/adk-python/commit/2a07eefc27b8f4cde9cfb27c1a9ba5bb0e415d31

Issue #311, opened 2025-04-21
Token usage metadata in agent response


Adithya-Krishnan-AK:
Can we get the token usage metadata of llm response at agent level in adk. 
Currently when i look at the **LlmResponse** structure in models and I see it doesn't have usage_metadata(input,output and total token details).
Is it possible to have them in order to get the details in agent response which would help in estimates before deploying and also tracking.

Related to discussion: https://github.com/google/adk-python/discussions/97

yrangana:
You can get `usage_meteada` from response

`"usage_metadata":{
"candidates_token_count":11,
"candidates_tokens_details":[{"modality":"TEXT","token_count":11}],
"prompt_token_count":17,
"prompt_tokens_details":[{"modality":"TEXT","token_count":17}],
"total_token_count":28}`

Adithya-Krishnan-AK:
Thanks for the response @yrangana is this with LiteLLM models or with direct gemini models use also, if possible can you please share an example of getting usage_meteada  from the LLMAgent response.

jordi-ydteam:
@yrangana it would be possible to elaborate a bit more how to extract the usage_metadata when using the agent? I cannot see any option when I am configuring the agent. 

I know that this usage metadata comes in the response of the vertex models, but I think that it is not included in the LlmResponse

yrangana:
@Adithya-Krishnan-AK and @jordi-ydteam,

This is from LiteLLM. There should be a way to directly extract this data. Currently, I’m wrapping my calls through LiteLLM.

Adithya-Krishnan-AK:
@yrangana Thanks for confirming, can you please share an example code to get usage_metadata from response as above. 
I tried using LiteLLM model for an adk LLMAgent but unable to get the usage_metadata in the response.

ajdaling:
I just submitted a PR #694 

I assume there is a reason that they didn't do this already so not sure if it will be merged but I just copied the usage_metadata property from the llm response to the llm_response object so that the usage metadata is accessible via after_model callbacks

Jacksunwei:
Fixed in https://github.com/google/adk-python/commit/1f0fd7bfce622459ef9dbc05ec96ffe6b613e4d1

josippavicic:
The added token usage in the LLM response is helpful in theory, but without the model name being included in the callback, this data is practically useless for cost estimation.

PR #694, opened 2025-05-13
Fixes issue #311: Add usage_metadata to llm_response

ajdaling:
- Copied usage_metadata property from generate_content_response in llm_response

This makes the usage metadata accessible via the model_callbacks so that at least some token tracking is possible.


Jacksunwei (approved):

Jacksunwei:
Merged in https://github.com/google/adk-python/commit/1f0fd7bfce622459ef9dbc05ec96ffe6b613e4d1

yufanyufan:
database_session_service.py should be updated as well

PR #724, opened 2025-05-14
fix: Match argument case when displaying errors for missing CLI arguments

bl3ck:
## Description

This PR contains a final commit from [PR #443](https://github.com/google/adk-python/pull/443) that addresses @hangfei’s feedback. The original PR was closed before this change was merged.

It introduces a static method, `_format_missing_arg_error,` to ensure that missing argument names in error messages preserve the correct case and match the original argument names in the CLI.

## Before
Image

## After
fix-match-arg-case

## How to Test
- Run `adk create` without specifying `APP_NAME`
- Run `adk run` without specifying `AGENT`
- Run `adk eval` without specifying `AGENT_MODULE_FILE_PATH`

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ Agent Development Kit (ADK) is a flexible and modular framework for developing a
 ## 🤖 Agent2Agent (A2A) Protocol and ADK Integration
 
 For remote agent-to-agent communication, ADK integrates with the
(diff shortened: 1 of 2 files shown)


hangfei:
@bl3ck merged in this commit: https://github.com/google/adk-python/commit/b226a06c0bf798f85a53c591ad12ee582703af6d. Thanks!

PR #727, opened 2025-05-14
Add tool: get_city_population to multi_tool_agentAdd population tool

JY123-X:
This pull request adds a new tool function `get_city_population()` to the agent located in the `multi_tool_agent` folder.

The tool returns hardcoded population values for major cities including:
- New York
- Tokyo
- London
- Paris
- Beijing

The agent was tested locally using ADK Dev UI and responded correctly to prompts like "What is the population of London?".

Files modified:
- agent.py (main logic)
- .env (API key)
- __init__.py

This contribution demonstrates how to extend ADK agents with custom tools in a modular and testable way.


multi_tool_agent/.env
teruguparthasaradhi, line 1: is this google API KEY Valid, if yes I guess it should be removed

JY123-X: Thank you for pointing that out! I’ve already revoked the exposed API key.

hangfei:
we don't add simple tools that are specific to certain use cases. 

If you need some samples, you can look into https://github.com/google/adk-python/tree/main/contributing/samples or adk-samples repo.

Issue #761, opened 2025-05-16
Adding tools with LlmRequest.append_tools() does not work for LiteLLM models


efunneko:
**Describe the bug**
Within a 'before model' callback, I am adding some tools to the request by creating a list of these tools and calling `append_tools()` on the llm_request. This puts a new Tool object on the llm_request.config.tools list. 

However, in models/lite_llm.py, it is hardcoded to only look at the first tool (see problematic `[0]`):

```python
tools = None
  if (
      llm_request.config
      and llm_request.config.tools
      and llm_request.config.tools[0].function_declarations  <--- problem here
  ):
    tools = [
        _function_declaration_to_tool_param(tool)
        for tool in llm_request.config.tools[0].function_declarations
    ]
```

**To Reproduce**
1. Create a simple agent that uses the before model callback to add a tool using llm_request.append_tools()
2. Configure the ADK to use a LiteLLM fronted model
3. Send a user input to the model that should call that tool
4. See that the tool is never called (or set a breakpoint before the completion call and observe the tool is not there)
 

**Expected behavior**
Tools added with the llm_request.append_tools() should show up in lite_llm provided model calls

**Screenshots**
These can be created if necessary, but just using hardcoded [0] in the code is clearly wrong

**Desktop (please complete the following information):**
 - OS: Mac
 - Python version(python -V): Python 3.11.10
 - ADK version(pip show google-adk): 0.5.0

**Additional context**

efunneko:
PR available: #762

seanzhou1023:
I feel like it's an expected behavior for Gemini for multiple custom tools in multiple function_declarations in the first tool section, other tools are for builtin tools IIUC. @selcukgun please double confirm and see whether we need to adjust this behavior for different 3rd party models.

efunneko:
@seanzhou1023 , any update on this? Do you need more from me? I would love to clean up my code and use the append_tools() properly

surajksharma07:
Hi @efunneko and thanks to everyone who participated in this thread.

We're currently doing some housekeeping on our issue tracker. Given the rapid pace of development and the number of updates since this was last active, there's a good chance this issue has been resolved in a newer version. To help us keep our backlog focused on current issues, we are closing this as stale.

If you're still facing this problem, could you please update to the very latest version of the library and confirm the issue still exists? If it does, please comment here, and we will be happy to reopen this issue and investigate further.

Thanks for your contribution and understanding!

PR #762, opened 2025-05-16
fix: LiteLLM: Loop over all tools in the llm_request when extracting the tool functions

efunneko:
Fix for #761. This will properly loop over all the llm_request.config.tools when extracting all the tool function declarations for calling models fronted by LiteLLM.


hangfei:
Could you follow the contributing guide to add tests: https://github.com/google/adk-python/blob/main/CONTRIBUTING.md.

Feel free to re-open after the issue is fixed. thanks!

PR #763, opened 2025-05-16
feat!: remove `ToolboxTool` in favor of using `toolbox-core` directly

### LINKED ISSUE #595: ToolboxTool should depend on `toolbox-core` and not `toolbox-langchain`
ADK's `ToolboxTool` current depends on `toolbox-langchain` which requires an additional dependency on `langchain` aswell to use ADK with Toolbox.

https://github.com/google/adk-python/blob/794a70edcd68e838b698716f7ce540756f0c4594/src/google/adk/tools/toolbox_tool.py#L36-L38

MCP Toolbox for databases supports a [`toolbox-core`](https://github.com/googleapis/genai-toolbox?tab=readme-ov-file#integrating-your-application) package that is orchestration framework agnostic. 

```python
from toolbox_core import ToolboxClient

# update the url to point to your server
client = ToolboxClient("http://127.0.0.1:5000")

# these tools can be passed to your application! 
tools = await client.load_toolset("toolset_name")
```

ADK should switch to using `toolbox-core` so that the additional dependency on LangChain is removed for users of ADK.

This would also remove the need for `ToolboxTool` to have to wrap its tools in `LangchainTool` as it currently does. 

This would once again simplify ADK's code base and remove toil for developers to get started with ADK + Toolbox.

jackwotherspoon:
Currently for users to leverage [MCP Toolbox for databases](https://github.com/googleapis/genai-toolbox) they must install `toolbox-langchain` and `langchain` and leverage ADK's `ToolboxTool` which wraps MCP toolbox's ToolboxClient. 

However, ADK's `ToolboxTool` does not expose all of MCP Toolbox's core features and does not allow full control of things like [bound parameters](https://github.com/googleapis/mcp-toolbox-sdk-python/blob/3516be521250f7ed097da1a280bf970f9e48e756/packages/toolbox-core/src/toolbox_core/client.py#L213) etc that MCP Toolbox supports.

Now with the `toolbox-core` package users can leverage toolbox without needing `ToolboxTool`.

Users will need to now `pip install toolbox-core`, a much more lightweight package and not dependent on langchain.

This gives ADK users full control over MCP Toolbox capabilities (such as both sync and async clients/tools).

```python
# NEW IMPORT
from toolbox_core import ToolboxSyncClient

# Full control over MCP Toolbox
toolbox = ToolboxSyncClient("http://127.0.0.1:5000")
toolbox_tools = toolbox.load_toolset("my-toolset")

root_agent = Agent(
    model="gemini-2.0-flash",
    name="root_agent",
    instruction=agent_instruction,
    # Add Toolbox tools to ADK agent
    tools=toolbox_tools,
)
```

Fixes #595


seanzhou1023 (approved):

seanzhou1023 (approved):

Jacksunwei:
Merged in https://github.com/google/adk-python/commit/11ca52809063bde64f5b9fbce577cae3752695b9

thanks!

PR #777, opened 2025-05-17
fix: separate thinking from text parts in streaming mode

almeynman:
Fixes https://github.com/google/adk-python/issues/770

Tested with the same example as in the original issue and got event with 2 parts back correctly


genquan9:
The changes are verified and work well for latest codes.

genquan9 (approved):
The changes are verified and work well.

seanzhou1023:
merged in commit 795605a

PR #853, opened 2025-05-22
fix: Call all tools in parallel calls during partial authentication

calvingiles:
Currently if a model performs parallel tool calls and only the first is authenticated, the others will not be called. When authentication is later provided, the other tool calls will never be executed.

This fixes that by treating all calls equally rather than special casing the first call.


seanzhou1023 (approved):
thanks for the fixing :)

seanzhou1023 (approved):

hangfei:
merged by bot in https://github.com/google/adk-python/commit/0e72efb4398ce6a5d782bcdcb770b2473eb5af2e. Thanks for your contribution!

PR #868, opened 2025-05-22
fix: Enable InMemoryRunner to be used in async tests

calvingiles:
Small tweak to the InMemoryRunner to enable use with async tests.

This is backwards compatible with previous use but enables a run_async method that will not swallow stack traces on errors.

I have deferred changing any tests to use this, but I can provide a follow up MR that makes tests async such that failures will provide a stack trace with the exact error, not fail silently.


boyangsvl (approved):

calvingiles:
Closing as now imported into main with Copybara

PR #869, opened 2025-05-22
Fix: unsafe_local_code_executor global import

qieqieplus:
How to reproduce the error:

```
from google.adk.code_executors import UnsafeLocalCodeExecutor
from google.adk.code_executors.code_execution_utils import CodeExecutionInput

result = UnsafeLocalCodeExecutor().execute_code(
    invocation_context=None,
    code_execution_input=CodeExecutionInput(
        code='''
import math

def pi():
    return math.pi

print(pi())
'''
    )
)

print(result)
```

output:
```
CodeExecutionResult(stdout='', stderr="name 'math' is not defined", output_files=[])
```


src/google/adk/code_executors/unsafe_local_code_executor.py
```diff
@@ -55,10 +55,9 @@ def execute_code(
     error = ''
     try:
       globals_ = {}
```
qieqieplus: maybe
```
globals_ = {"__name__": "__main__"}
```
for better compatibility?

hangfei:
Could you explain what's the issue and how your fix works? 

Please also fix CLA.

qieqieplus:
> Could you explain what's the issue and how your fix works?
> 
> Please also fix CLA.

When exec() is called with only the globals dict, it is shared for both global and local scopes during execution. Modules imported at the top level (e.g., import math) are added to globals_ and are then accessible to functions defined within the executed code, as these functions will use globals_ as their global namespace.

src/google/adk/code_executors/unsafe_local_code_executor.py
seanzhou1023: it's within _prepare_globals

seanzhou1023 (approved):

src/google/adk/code_executors/unsafe_local_code_executor.py
qieqieplus: > it's within _prepare_globals

yes, `_prepare_globals` was added after this PR

seanzhou1023:
merged in https://github.com/google/adk-python/commit/1778490e64830f45ab71a8c3ba9fcebeca8d1d44

PR #882, opened 2025-05-23
Add support for Claude 4 and register anthropic models by default

ptone:
With the release of Claude 4 I wanted to test ADK with Claude.

This PR 

- updates regex for supported claude models
- registers anthropic models by default
- removes a spurious for loop during registration

### DIFF
--- a/src/google/adk/models/__init__.py
+++ b/src/google/adk/models/__init__.py
@@ -16,6 +16,7 @@
 
 from .base_llm import BaseLlm
 from .google_llm import Gemini
+from .anthropic_llm import Claude
 from .llm_request import LlmRequest
 from .llm_response import LlmResponse
 from .registry import LLMRegistry
@@ -27,5 +28,5 @@ __all__ = [
 ]
 
 
-for regex in Gemini.supported_models():
-  LLMRegistry.register(Gemini)
+LLMRegistry.register(Gemini)
+LLMRegistry.register(Claude)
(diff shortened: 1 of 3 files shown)


selcukgun:
LGTM

That said, it looks like supported_models is noop for non-gemini models; so claude 4 already works without this PR.

src/google/adk/models/__init__.py
```diff
@@ -27,5 +28,5 @@
 ]
 
 
-for regex in Gemini.supported_models():
-  LLMRegistry.register(Gemini)
+LLMRegistry.register(Gemini)
+LLMRegistry.register(Claude)
```
selcukgun, line 32: Fixing the loop is fine. But please remove Claude and the import as registry is intended for gemini only.

selcukgun (changes requested):
Good to go after removing claude import in src/google/adk/models/__init__.py

avidelatm:
Is this a duplicate of my PR that [got merged](https://github.com/google/adk-python/commit/7fc09b2c642cc1f84f6c4ee87b63f4f4eff8dd94) earlier today to fix the Claude regex?

ptone:
Yes this is now basically duplicate of 7fc09b2

I'm just going to close this, as I also did not understand the correct usage of the Claude model class and the use of the registry.

I've submitted the very small #1129 to still remove the superfluous loop in `models.__init__`

PR #883, opened 2025-05-23
chore: Fix reqeust -> request typo

calvingiles:
There is only one usage of `_reqeust_credential` in the library, which has been safely resolved, and there are no usages of `_request_credential`. Given it is private, there should not be any public usages of this either.


src/google/adk/tools/openapi_tool/openapi_spec_parser/tool_auth_handler.py
```diff
@@ -260,7 +255,7 @@ def prepare_auth_credentials(
           auth_credential=exchanged_credential,
       )
     else:
-      self._reqeust_credential()
+      self._request_credential()
```
calvingiles, line 258: This is the only usage of `_reqeust_credential` in the library, and there are no usages of `_request_credential`. Given it is private, there should not be any public usages of this either.

seanzhou1023 (approved):

seanzhou1023:
merged in https://github.com/google/adk-python/commit/4075290a1d57569ca778467d2a6b65873f571dde

PR #888, opened 2025-05-23
Allow using hosted Gemma & LearnLM models by whitelisting modelid prefix

### LINKED ISSUE #887: Support calling Google-hosted Gemma models
In `Gemini.supported_models()`, a regular expression checks that model id begins with `gemini``. This prevents calling Gemma models and LearnLM models because their model IDs begin with `gemma-` and `learnlm-` , even when the API access is in place.

The [regex is defined here](https://github.com/google/adk-python/blob/main/src/google/adk/models/google_llm.py#L64).

Image

nileshtrivedi:
This should fix #887.

### DIFF
--- a/src/google/adk/models/google_llm.py
+++ b/src/google/adk/models/google_llm.py
@@ -62,6 +62,8 @@ class Gemini(BaseLlm):
 
     return [
         r'gemini-.*',
+        r'gemma-.*',
+        r'learnlm-.*',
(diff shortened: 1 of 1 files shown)


nileshtrivedi:
Sorry, I will not be signing the CLA.

boyangsvl:
Thanks for the contribution. However, we can not accept the PR without a signed CLA.

PR #913, opened 2025-05-24
Add py.typed for PEP 561 compliance

mohamadghaffari:
## Summary

This PR adds a `py.typed` marker file to the `google.adk` package and updates the `pyproject.toml` to include it in the Flit build.

This change ensures the package is [PEP 561](https://peps.python.org/pep-0561/) compliant, allowing tools like `mypy`, `pyright`, and `pytype` to recognize and use the type hints provided in the source.

## Changes

- Added `src/google/adk/py.typed` (empty file)
- Updated `[tool.flit.module]` to include `py.typed` in the build

## Testing Plan

Built the package locally using:
`uv build`

Installed the .whl locally using:

`uv pip install dist/google_adk-<version>.whl`

Verified the presence of py.typed in the installed package:
```bash
python -c "import google.adk, os; print(os.path.exists(os.path.join(os.path.dirname(google.adk.__file__), 'py.typed')))"
# Output: True
```

Verified that both mypy and pyright detect and use the type hints correctly.

CLA

☑️ I have signed the [Google Contributor License Agreement](https://cla.developers.google.com/)


Jacksunwei (approved):

hangfei:
merged in the commit. thanks!

PR #936, opened 2025-05-26
fix(sample): Correct text artifact saving in artifact_save_text sample

ystory:
### Description of Changes:

This PR addresses an `AttributeError: 'NoneType' object has no attribute 'data'` that occurred in the `artifact_save_text` sample (`contributing/samples/artifact_save_text/agent.py`) when the `log_query` tool attempted to save a text query as an artifact.

The error was caused because `types.Part(text=query)` was used, which leaves `inline_data` as `None`. `GcsArtifactService` (and potentially other artifact services) expect binary data within `inline_data.data`.

The fix updates the `log_query` function within this specific sample to:
1. Encode the query string to UTF-8 bytes.
2. Wrap it in `types.Blob` with `mime_type='text/plain'`.
3. Assign this `Blob` to the `inline_data` field of the `types.Part` object.

### Testing:

Verified that the artifact_save_text sample now runs without the AttributeError and successfully saves the query as a 'text/plain' artifact when an artifact service like GcsArtifactService is configured

Screenshot 2025-05-26 at 14 27 02
Screenshot 2025-05-26 at 14 45 05


DeanChensj (approved):

DeanChensj (approved):

DeanChensj:
Merged in https://github.com/google/adk-python/commit/5c6001d90fe6e1d15a2db6b30ecf9e7b6c26eee4

PR #807, opened 2025-05-20
Update vertex_ai_session_service.py

dpark27:
Fix for [Issue #804 ](https://github.com/google/adk-python/issues/804)

Adding in extra url parsing so that the request for vertex ai sessions returns correctly and is not always empty

Issue #804, opened 2025-05-20
VertexAiSessionService list sessions always returns empty list


dpark27:
**Describe the bug**
When using VertexAiSessionService, the `list_sessions` function always returns an empty result

**To Reproduce**
1. Initialize a VertexAiSessionService
2. Create a session
3. run `session_service.list_sessions(app_name=session.app_name, user_id=your_user_id)`

**Expected behavior**
Return a list of sessions that exist for the user

**Desktop (please complete the following information):**
 - OS: OS X
 - Python version(python -V): 3.11.5
 - ADK version(pip show google-adk): v0.5.0

**Additional context**

DeanChensj:
Hi @dpark27 , what user_id are you using?  Is it a special character?

dpark27:
We are using email addresses as usernames, ie ***@***.***

DeanChensj:
Thanks for fixing the issue. https://github.com/google/adk-python/pull/996 is now merged.

PR #996, opened 2025-05-28
Fix: list vertex ai sessions

dpark27:
Copying from https://github.com/google/adk-python/pull/807 due to CLA issues

Fix for https://github.com/google/adk-python/issues/804

When the user id contains special charactes (i.e. an email), we have added in extra url parsing to address those characters. We have also added an if statement to use the correct url when there is no user_id supplied.

This fixes the request for vertex ai sessions so that the function returns correctly and is not always empty.

### DIFF
--- a/src/google/adk/sessions/vertex_ai_session_service.py
+++ b/src/google/adk/sessions/vertex_ai_session_service.py
@@ -16,6 +16,9 @@ from __future__ import annotations
 import asyncio
 import logging
 import re
+import time
+import urllib.parse
+
 from typing import Any
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


dpark27:
@DeanChensj - here is a carbon copy of https://github.com/google/adk-python/pull/807

I will close the other one.

DeanChensj (approved):

DeanChensj:
Thanks for the PR!

DeanChensj (approved):

DeanChensj:
Merged in https://github.com/google/adk-python/commit/9d4ca4ed44cf10bc87f577873faa49af469acc25

Issue #1010, opened 2025-05-29
non-ASCII characters are escaped in telemetry trace_send_data


genkey6:
**Describe the bug**
In the trace_send_data function of the telemetry module, the JSON-dumped content is set as an attribute of the span.
When non-ASCII characters (such as Japanese Hiragana or Kanji) are passed to the function, they are escaped in the exported span logs.

**Expected behavior**
Non-ASCII characters (such as Japanese Hiragana or Kanji) should be preserved in their original form in the exported span logs, without being escaped.
For example, the string "こんにちは" should appear as "こんにちは" in the span attributes, not as escaped Unicode like "\u3053\u3093\u306b\u3061\u306f".

**Desktop (please complete the following information):**
 - Python version(python -V): 3.12.9
 - ADK version(pip show google-adk): 1.1.1

jwiejulie:
Hi @genkey6!

We're currently doing some housekeeping on our issue tracker. Given the rapid pace of development and the number of updates since this was last active, there's a good chance this issue has been resolved in a newer version. To help us keep our backlog focused on current issues, we are closing this as stale.

If you're still facing this problem, could you please update to the very latest version of the library and confirm the issue still exists? If it does, please comment here, and we will be happy to reopen this issue and investigate further.

Thanks for your contribution and understanding!

PR #1012, opened 2025-05-29
fix: set ensure_ascii=False in telemetry trace_send_data

genkey6:
## Description:
This pull request fixes an issue where non-ASCII characters (e.g., Japanese Hiragana or Kanji) passed to the trace_send_data function were being escaped in the exported span logs. This behavior reduced the readability and usability of the logs in multilingual environments.

## Changes made:
- Modified the json.dumps call in the trace_send_data function to include ensure_ascii=False, allowing non-ASCII characters to be preserved in their original form rather than being Unicode-escaped.

## Related issue:
Fixes: https://github.com/google/adk-python/issues/1010

### DIFF
--- a/src/google/adk/telemetry.py
+++ b/src/google/adk/telemetry.py
@@ -153,12 +153,15 @@ def trace_send_data(
   # information still needs to be recorded by the Agent Development Kit.
   span.set_attribute(
       'gcp.vertex.agent.data',
-      json.dumps([
-          types.Content(role=content.role, parts=content.parts).model_dump(
-              exclude_none=True
-          )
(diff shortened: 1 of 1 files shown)


hangfei:
@selcukgun PTAL.

selcukgun:
Thank you for the PR! But this is already addressed with https://github.com/google/adk-python/commit/d587270327a8de9f33b3268de5811ac756959850 

https://github.com/google/adk-python/blob/65063023a5a7cb6cd5db43db14a411213dc8acf5/src/google/adk/telemetry.py#L223

PR #946, opened 2025-05-26
Fix(DatabaseSessionService): add missing `kwargs` when instantiating database session service required for underlying sqlalchemy's `create_engine`

sanchitrk:
Previously, `DatabaseSessionService` was instantiated without additional `**kwargs`,
which could be added for configuring SQLAlchemy's `create_engine`, such as custom connection arguments, pooling options, or echo settings. This caused the service to ignore optional, yet essential configuration parameters and limited its flexibility.

This fix ensures that `**kwargs` are correctly forwarded to the underlying SQLAlchemy engine setup,
keeping expected behavior and supporting full engine conn customization.

Fixes #944

PR #1042, opened 2025-05-30
chore: follow-up #946 to allow session db kwargs in fast api entrance.

Colstuwjx:


Colstuwjx:
Hi @DeanChensj , this is the follow-up PR to #946 since some end users might directly use `get_fast_api_app` to init the adk application web service, it's also necessary to expose these database settings on the fastapi side.

Thanks.

Colstuwjx:
Here are the customized db kwargs on our internal project:

```
# Call the function to get the FastAPI app instance
# Ensure the agent directory name ('capital_agent') matches your agent folder
app: FastAPI = get_fast_api_app(
    agents_dir=AGENTS_DIR,
    session_db_url=SESSION_DB_URL,
    allow_origins=ALLOWED_ORIGINS,
    # Setting this to True would invalidate any additional routes
    # we did a workaround: https://github.com/google/adk-python/issues/51#issuecomment-2899152791
    web=False,
    session_db_kwargs={
        "pool_pre_ping": True,
        "pool_recycle": 3600,
    },
)
```

rpedela-recurly:
My team needs this change! Thanks for contributing!

src/google/adk/cli/fast_api.py
```diff
@@ -193,6 +194,7 @@ def get_fast_api_app(
     *,
     agents_dir: str,
     session_db_url: str = "",
+    session_db_kwargs: Optional[Mapping[str, Any]] = None,
```
DeanChensj, line 222: The plan is just to expose this to the fast_api_app, not the cli, right?

src/google/adk/cli/fast_api.py
```diff
@@ -202,6 +204,10 @@ def get_fast_api_app(
   trace_dict: dict[str, Any] = {}
   session_trace_dict: dict[str, Any] = {}
 
+  # Database session settings
+  if session_db_kwargs is None:
```
DeanChensj: can be moved to inline line 282?

src/google/adk/cli/fast_api.py
Colstuwjx: I've checked the DatabaseSessionService references, only fast_api.py `get_fast_api_app` has instantiated this class, and also I'm using `get_fast_api_app` to start the web service, it's enough to do so for me. Please let me know if we need to expose more entrances (how?)

src/google/adk/cli/fast_api.py
Colstuwjx: done.

DeanChensj (approved):

DeanChensj:
Merged in https://github.com/google/adk-python/commit/6cc3d9ddd19b221c704b61245f2cb9e6db1241bf

PR #1043, opened 2025-05-30
chore: add missing cli installation guide in CONTRIBUTING.md

Colstuwjx:


hangfei:
@Colstuwjx Thanks for your contribution.

Thanks. let's use this way to install: https://github.com/google/adk-python/blob/main/pyproject.toml#L63. 

So let's add pytest and mention install with that approach.

Colstuwjx:
Hi @hangfei , the previous cli `uv sync --all-extras` didn't install the dependency, and I tried `uv pip install -r pyproject.toml --all-extras` and it works, please take a look again.

Thanks.

wuliang229:
`uv sync --all-extras` will install all the dependencies from project.toml.

PR #1107, opened 2025-06-03
fix: update pyink.yaml to fix the "detect changed files" step

wuliang229:


Jacksunwei (approved):

Jacksunwei (commented):
~test~

.github/workflows/pyink.yml
```diff
@@ -28,6 +28,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
```
Jacksunwei, line 30: Why adding this? The default is checking the "merged version", which would be better to CI/CD.

PR #1206, opened 2025-06-06
fix: Notify agent of failed agent transfer in function response

calvingiles:
# What happens now

If an error is raised when calling a sub agent - for example, because one of the toolsets raises an error on get_tools (expected on an MCP toolset connection error, for example) then the error is silently caught, and the transfer response is returned as is. The means two things:

1. the context window strongly implies that a transfer was successful
2. the last sent message is from the original agent, not the transferee, and so the transfer will _not_ have been picked up, according to [`_find_agent_to_run`](https://github.com/google/adk-python/blob/main/src/google/adk/runners.py#L336-L337)

# What should happen

given resolving the failure in general is not possible, I think the only stable solution is to accept that the transfer failed, and to ensure the `transfer_to_agent` response does not imply otherwise.

# Alternatives
 
## Remove the transfer call

The call and response could be stripped. This seems likely to result in the agent trying again though, without a fix being applied.

## Try again

There are no general solution to a failed run_async - for example, a misconfigured MCPToolset, or an MCP server outage will result in persistent failures until that is resolved. Given this can easily take longer than the request timeout it. seems better to accept that failure may occur sometimes.

# Proposed solution (in this PR)

If the `run_async` or `run_live` methods emit at least one event, consider the subagent to be working and transfer successful. If not, assume a failure and re-write the function response. 

This involves just a little juggling of the yield to delay when it was yielded before until we are sure, which will put a small amount of latency on the result, but before an event is yielded it is not known if the agent was successful.

Happy to implement an alternative solution if this is not the way the project wants to take this.


calvingiles:
Note - if we want to make successful transfer clearer, we can also add `return {'result': f'Transferred to {agent_name}'}` as a default in the `transfer_to_agent` tool (which will be overridden on failure).

rohityan:

PR #1226, opened 2025-06-07
feat: Support adding a name prefix to MCPTool and MCPToolset

calvingiles:
Given the high occurrence of tool name collisions with MCP Servers (for example, creating an agent that support multiple IDEs will often result in multiple `list_directory` tools) it is often desirable to be able to namespace MCP tools under a sensible server name. For mcp servers that do not follow this convention, this feature addition adds a simple `tool_name_prefix` argument that will be added to the names of the `MCPTool` or `MCPToolset`.

### DIFF
--- a/src/google/adk/tools/mcp_tool/mcp_tool.py
+++ b/src/google/adk/tools/mcp_tool/mcp_tool.py
@@ -62,6 +62,7 @@ class MCPTool(BaseTool):
       mcp_session_manager: MCPSessionManager,
       auth_scheme: Optional[AuthScheme] = None,
       auth_credential: Optional[AuthCredential] = None,
+      tool_name_prefix: str = '',
   ):
     """Initializes a MCPTool.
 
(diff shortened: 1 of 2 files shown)


src/google/adk/tools/mcp_tool/mcp_tool.py
seanzhou1023, line 1: I feel the changes in this file will cause the prefix to be added twice, given we already handled prefix  prepending in base toolset, did you test it ? 

I think the only changes needed in this file is 
```
response = await session.call_tool(self._mcp_tool.name, arguments=args)
```

could you try and verify ?

src/google/adk/tools/mcp_tool/mcp_tool.py
```diff
@@ -128,7 +139,7 @@ async def _run_async_impl(
     # Get the session from the session manager
     session = await self._mcp_session_manager.create_session(headers=headers)
 
-    response = await session.call_tool(self.name, arguments=args)
+    response = await session.call_tool(self._mcp_tool.name, arguments=args)
```
seanzhou1023, line 142: besides this line, I feel all the other changes in this file are not necessary and will cause problem as I mentioned in the comment of the whole file, would you please try and test it ?

seanzhou1023:
@calvingiles , just realized you acutally added this prefix feature in mcp tool before we added this feature generally in base_toolset. sorry for the delay, we are really short of hands. would you be able to adapt to the latest code base and address the comments ?

seanzhou1023:
close this PR as same function is implemented in https://github.com/google/adk-python/pull/2823

PR #1245, opened 2025-06-09
Fix: Handle Pydantic v2 ValidationError in AgentLoader

### LINKED ISSUE #1227: pydantic_core._pydantic_core.ValidationError
Describe the bug
When running an ADK agent using google-adk==1.2.1 with pydantic==2.11.5, an internal exception handling block in agent_loader.py attempts to access e.msg, which does not exist in modern Python exceptions (specifically pydantic_core._pydantic_core.ValidationError). This leads to a crash, masking the actual root cause.

To Reproduce

Create a simple ADK agent using the SequentialAgent class.

Pass an invalid input or missing state variable (e.g. forget to pass "user_input").

adk run --input='{"wrong_key": "some topic"}'
You’ll see:

AttributeError: 'ValidationError' object has no attribute 'msg'
Expected behavior
The error should cleanly show the validation problem (e.g., missing input key) without crashing on .msg. ADK should not assume .msg exists and should use str(e) instead.

Desktop (please complete the following information):

OS: Windows 10
Python version: 3.12.3
ADK version: 1.2.1
Pydantic version: 2.11.5

Additional context
This is caused by this line in agent_loader.py:

e.msg = f"Fail to load '{agent_name}' module. " + e.msg
Since e.msg is not a valid attribute of ValidationError in Pydantic v2, this breaks.

Suggested Fix

Replace:

e.msg = f"Fail to load '{agent_name}' module. " + e.msg

With:

raise RuntimeError(f"Fail to load '{agent_name}' module. {str(e)}") from e
This works across both Pydantic v1 and v2, and provides a proper traceback.

hangfei:
The AgentLoader previously attempted to access `e.msg` on exceptions encountered during agent loading. This caused a crash when the exception was a `pydantic_core._pydantic_core.ValidationError` from Pydantic v2+, as this error type does not have a `msg` attribute.

This commit modifies the exception handling in `AgentLoader`'s `_load_from_module_or_package` and `_load_from_submodule` methods to use `str(e)` instead of `e.msg`. This ensures that the error message is correctly retrieved for all exception types, including Pydantic v2 `ValidationError`.

A new test case has been added to specifically verify this scenario by simulating a Pydantic v2 style error during agent loading. Existing tests for other error types (SyntaxError, NameError) within agent code have been updated to reflect the new `RuntimeError` wrapping behavior of the AgentLoader.

Fixes #1227

### DIFF
--- a/src/google/adk/cli/utils/agent_loader.py
+++ b/src/google/adk/cli/utils/agent_loader.py
@@ -76,13 +76,7 @@ class AgentLoader:
         e.msg = f"Fail to load '{agent_name}' module. " + e.msg
         raise e
     except Exception as e:
-      if hasattr(e, "msg"):
-        e.msg = f"Fail to load '{agent_name}' module. " + e.msg
-        raise e
-      e.args = (
-          f"Fail to load '{agent_name}' module. {e.args[0] if e.args else ''}",
-      ) + e.args[1:]
-      raise e
+      raise RuntimeError(f"Fail to load '{agent_name}' module. {str(e)}") from e
 
     return None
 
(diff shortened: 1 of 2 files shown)


MyHackInfo:
Hi @hangfei it look good to me and please make sure run test case

thaq1f:
Hey there! Quick follow up if this is resolved, am getting the same error from my end too. Thanks!

maxbaluev:
Hi! I think i have related to that problem

PR #1269, opened 2025-06-10
feat: Add PubSubListenerAgent sample with README

cbpalumbi:
This pull request introduces a new sample agent, PubSubListenerAgent, which demonstrates how to listen to Google Cloud Pub/Sub messages and emit events within the Google ADK Python framework.

Key features include:
-     Asynchronous listening to a specified Pub/Sub subscription
-     Processing and acknowledging Pub/Sub messages
-     Emitting ADK events to integrate with other agents or components
-     Graceful shutdown support (currently works on macOS and Linux)
-     A sample README explaining setup and usage

This sample aims to help developers integrate real-time Pub/Sub messaging into their multi-agent workflows. It also includes environment variable configuration via .env and dependency requirements.

Looking forward to feedback or suggestions for improvement!


hangfei:
Thanks for the contribution @cbpalumbi .

Please fix conventionalcommits.org check. 

So this agent only does receive and return. 

How does the LLM generation(agentic) come into picture?

cbpalumbi:
> Thanks for the contribution @cbpalumbi .
> 
> Please fix conventionalcommits.org check.
> 
> So this agent only does receive and return.
> 
> How does the LLM generation(agentic) come into picture?

Thanks for the feedback, Hangfei!

You're right - this agent currently focuses on receiving and acknowledging Pub/Sub messages and emitting them as Events into the ADK loop. It doesn’t include LLM-based generation or reasoning of its own.

That said, I tried to design it as a modular utility that could act as a bridge between external infrastructure (like CI systems, background workers, etc.) and ADK-based agent workflows. The idea is to make it easy to plug this into more complex setups.

For example:

- A workflow agent could subscribe to events from this listener and trigger follow-up actions 
- The Pub/Sub messages could encode metadata or task results that are picked up and reasoned over by other agents

I also tried to keep the implementation self-contained and reactive so it would be easy to compose into other workflows if the ADK later supports parallel long-lived agents or background listeners. I imagine this would fit well into an architecture where agents communicate via shared events, and some can persistently listen while others respond.

Please let me know if you have any more feedback or ideas on how listener could be elegantly combined with existing workflow agents or multi agent architectures! I experimented a bit with having a root agent create a parallel agent that acts as a parent for any "listener / background " type agents. But I found it was hard to interleave agents' work with messages without constantly stopping and restarting the listener. 

Also, I edited the PR's name to meet the commit requirements as far as I can tell. Is this what you meant? From my end, the commit itself is passing the conventional commits check.

rohityan:
Hi @cbpalumbi and thanks to everyone who participated in this thread.

Issue #1005, opened 2025-05-29
DatabaseSessionService uses sync DB operations - needs async variant


polaris340:
**Is your feature request related to a problem? Please describe.**
DatabaseSessionService uses async method signatures but internally calls SQLAlchemy’s sync APIs

**Describe the solution you'd like**
Add an AsyncDatabaseSessionService that uses SQLAlchemy’s async APIs (AsyncSession, etc.).

**Describe alternatives you've considered**
Fix existing DatabaseSessionService implementation to use async database operations.

**Additional context**
n/a

mbroton:
I was trying to use `DatabaseSessionService` in an async way (passing `db_url` starting with `postgresql+asyncpg`) and there's more to that issue - I couldn't even initialize the class, because there's creation of sync engine, sync sessionmaker and tables creation (`Base.metadata.create_all(self.db_engine)`) which is causing:
`exc.MissingGreenlet: greenlet_spawn has not been called; can't call await_only() here. Was IO attempted in an unexpected place?`

I was thinking about creating a PR with a quick fix, however I wasn't sure how to handle `__init__` without introducing breaking changes to the API.

Additional notes:
- It would be nice to have more control over tables creation, e.g. I want to use [alembic](https://alembic.sqlalchemy.org/en/latest/) and keep tables creation in migration files.
- There's redundant line `self.inspector = inspect(self.db_engine)` because `self.inspector` is never used

condorcet:
@DeanChensj Hello! do we have news here? In our project we use patched version of database service like this https://github.com/google/adk-python/pull/1316 it works, but wondering, will it be solved by ADK team in next releases?

leslyarun:
@DeanChensj When will this support be added? Please priortise it

hangfei:
We are evaluating if we can merge https://github.com/google/adk-python/pull/2889 or there is a better option.

hangfei:
Also related: https://github.com/google/adk-python/issues/1750

DeanChensj:
This is now closed by https://github.com/google/adk-python/pull/2889

PR #1316, opened 2025-06-11
feat: support asyncio in session database service

condorcet:
Fixes issue https://github.com/google/adk-python/issues/1005

Currently, database session service only has async / await interface, but internally it doesn't use async driver. For production usage it is critical, because database operations block the event loop.

This changes allows to use async drivers in database session service.

This also introduce some breaking changes:
1. Because of async / await we have to initialize tables in separate method (`initialize_tables`) and not in `__init__`. In general such separation of concerns seems appropriate, since table initialization can be a separate process (it could be run before application startup, can be part of CI / CD, can use other DSN etc).
2. [inspection](https://docs.sqlalchemy.org/en/20/core/inspection.html) was removed because it doesn't support async / await in such a simple way. Most likely, if someone needs inspection, it should be utilized as separate method / property of class to be ready to async / await specific. See more [here](https://github.com/sqlalchemy/sqlalchemy/issues/6121)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,6 +106,7 @@ eval = [
 test = [
   # go/keep-sorted start
   "a2a-sdk>=0.3.0,<0.4.0;python_version>='3.10'",
+  "aiosqlite>=0.21.0",
   "anthropic>=0.43.0",                            # For anthropic model tests
   "langchain-community>=0.3.17",
   "langgraph>=0.2.60, <= 0.4.10",                 # For LangGraphAgent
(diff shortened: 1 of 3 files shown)


phuoc-nguyen-ts:
The message ordering seems to be messed up when I use this

Telling the agent to list our convo till now, and it list information in a wrong order. Which did not happen with the sync version

condorcet:
I believe this is because the MR is outdated, the logic of ordering has been changed some time ago.
@phuoc-nguyen-ts MR was actualized, can you check this?

rohityan:
Hi @condorcet and thanks to everyone who participated in this thread.

Issue #1346, opened 2025-06-12
GCS artifact storage should be async


condorcet:
At this moment GCS artifact storage implementation relies on GCS client which does not support asyncio.

This results in the event loop being blocked on every I/O operation, which is critical for production use.

I believe this can be solved in a more generic way with S3 compatible interface (e.g. https://github.com/google/adk-python/issues/555).
But as a first iteration, we can simply run the current methods with a thread pool executor to satisfy the minimal async behavior.

hangfei:
@DeanChensj is this already fixed?

DeanChensj:
Resolved by https://github.com/google/adk-python/pull/1347

PR #1347, opened 2025-06-12
feat: make GCS artifact service async under the hood

condorcet:
Fixes issue https://github.com/google/adk-python/issues/1346

### DIFF
--- a/src/google/adk/artifacts/gcs_artifact_service.py
+++ b/src/google/adk/artifacts/gcs_artifact_service.py
@@ -14,6 +14,8 @@
 
 """An artifact service implementation using Google Cloud Storage (GCS)."""
 
+import asyncio
+from concurrent.futures import ThreadPoolExecutor
 import logging
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


DeanChensj:
Could you resolve the failing checks?

condorcet:
Thanks! I will actualize request soon

condorcet:
@DeanChensj done

condorcet:
> asyncio provides a thread-pool-based mechanism for this. We don't need to manage thread pool executor for us.
>
> Let's use asyncio.to_thread approach instead.

@DeanChensj @Jacksunwei 

In this case we won't be able to configure executor (e.g. number of workers). Under the hood `asyncio.to_thread` use the same class of ThreadPoolExecutor, but with defaults.

Also `asyncio.to_thread` copy context vars and this is really seems reasonable.

So, WDYT will it be acceptable for us don't have ability to configure executor?

Jacksunwei:
I think it's fine.

condorcet:
@Jacksunwei @DeanChensj done. Now we use `asyncio.to_thread` instead thread pool executor explicitly

DeanChensj (approved):

DeanChensj:
This PR is merged in https://github.com/google/adk-python/commit/81e0d4083f7c3ac3e14562ff712e953de7361c67

Issue #1214, opened 2025-06-06
Add Built-in Retry Mechanism for API Errors (e.g., 429 Too Many Requests) in LLM Agent


darkmars27:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

Statement: Before submitting this request, I have reviewed the ADK documentation & python code and existing issues and have not found any readily available or clearly documented built-in retry mechanisms for the LLM Agent, specifically for handling API errors like 429 Too Many Requests. If such an implementation exists, please guide me to the relevant documentation.

**Is your feature request related to a problem? Please describe.**
I am frequently encountering 429 Too Many Requests errors when using the LLM Agent within the Google ADK. The current implementation does not provide any direct or convenient mechanism to handle these errors, specifically for retries. This leads to failed requests and interruptions in our application's workflow, requiring manual intervention or the implementation of complex, external retry logic.

**Describe the solution you'd like**
I would like to request the addition of a built-in, configurable retry mechanism for the LLM Agent within the Google ADK. This mechanism should automatically handle transient API errors, such as 429 Too Many Requests, 500 Internal Server Error, 503 Service Unavailable, etc., by implementing an exponential backoff strategy with configurable parameters.

Ideally, this would involve:

- Automatic Retries: The LLM Agent should automatically retry failed requests based on a predefined set of HTTP status codes.
- Configurable Backoff: Users should be able to configure parameters like:
   Maximum number of retries.
   Initial backoff delay.
   Maximum backoff delay.
- Error Handling: A clear way to differentiate between retriable and non-retriable errors, allowing for appropriate handling of persistent issues.

**Describe alternatives you've considered**
I have considered using the HttpOptions base_url parameter to route requests through a custom proxy or service that would implement the retry logic. However, this approach introduces unnecessary overhead and complexity:

- Additional Hop: It adds an extra network hop between our application and the Google API, potentially increasing latency and introducing a new point of failure.

- Maintenance Burden: It requires us to deploy, manage, and maintain an additional proxy service, increasing operational overhead.

- Security Concerns: Depending on the proxy implementation, it might introduce additional security considerations.

**Additional context**
[...]

PR #1361, opened 2025-06-12
feat: implement retry mechanism for LiteLlm

petitoff:
In case of the issue: https://github.com/google/adk-python/issues/1214


pwwpche:
Hi @petitoff 

Thanks for your contribution! Have you looked at using any libraries for implementing the retry logic? 

For example, ADK already uses tenacity for retry

https://github.com/google/adk-python/blob/main/src/google/adk/sessions/vertex_ai_session_service.py#L121

petitoff:
@pwwpche hi,

I haven't tried using libraries like tenacity before, but I could explore that option. However, when creating this PR, I wanted a simple, fast, and working solution that could be implement immediately (like a hot fix)

pwwpche:
> @pwwpche hi,
> 
> I haven't tried using libraries like tenacity before, but I could explore that option. However, when creating this PR, I wanted a simple, fast, and working solution that could be implement immediately (like a hot fix)

I would highly recommend refactor this change using `tenacity` then. With ~10 lines of code, it could achieve the same outcome as the hundreds of lines of code we have in the lite_llm.py. Since this implementation is in the core of ADK, let's prioritize a bit more on longer term maintainability.

Something like

```
from tenacity import retry
from tenacity import retry_if_result
from tenacity import RetryError
from tenacity import stop_after_attempt
from tenacity import wait_exponential

# ...

    retry_wrapped_acompletion = retry(
          stop=stop_after_attempt(self.stop_attempts),
          wait=wait_exponential(multiplier=1, min=1, max=3),
          retry=retry_if_result(lambda response: not response),
          reraise=True,
      )(acompletion)

  return await retry_wrapped_acompletion(...)
```

Thanks!

rohityan:
Hi @petitoff and thanks to everyone who participated in this thread.

PR #1375, opened 2025-06-13
fix: AgentTool import

kubikb:
Currently, the class `AgentTool` cannot be imported due to it not being listed in the `tools` directory's `__init__.py` file.

The same import pattern that is used for other tools such as `FunctionTool` (imported using `from google.adk.tools import FunctionTool`) cannot be used and `AgentTool` needs to be imported with `from google.adk.tools.agent_tool import AgentTool`.

The change in this PR includes `AgentTool` in the listing in `src/google/adk/tools/__init__.py`, thereby allowing a more standard import with `from google.adk.tools import AgentTool`, adhering to existing import pattern of another tools.


wuliang229:
It's added in this [commit](https://github.com/google/adk-python/commit/5edc493da9525f18dea84cd49342e131db44cb87)

PR #1430, opened 2025-06-16
Create adk-docs dir

J-Dream:
added adk docs directory.  removed redundant docs directory, that is now inside 'adk-docs'


hangfei:
adk-docs is in https://github.com/google/adk-docs. Please revise there. thanks.

PR #1462, opened 2025-06-17
feat: Pass Auth through from AgentTool

calvingiles:
This captures the requested_auth_configs actions from an AgentTool run and forwards them to the caller agent.

The result of this is that an AgentTool with a tool requiring auth will trigger a credentials request at the top level. When the credentials are provided, the function call for the AgentTool call will be replayed. Without this change, the authentication request is simply swallowed.

Note: The code was NOT AI GENERATED and has been tested in our production systems.


rohityan:

PR #1488, opened 2025-06-18
fix: Update google_search_tool.py to support updated Gemini LIVE model name

rayira:
fix: Update google_search_tool.py to support updated Gemini LIVE model name


rayira:
will redo it.

Issue #1512, opened 2025-06-19
New Gemini LIVE model naming got blocked by Google Search tool use


rayira:
**Describe the bug**
New Gemini LIVE model naming got blocked by Google Search tool use. Current check is to avoid model naming 'gemini-1' and allow only 'gemini-2' for Google Search tool use.  The new naming is 'gemini-live-' which stops the LIVE to use Google Search tool.

**To Reproduce**
1. Install 'google-adk==1.2.1'
2. Run agent with 'tools=[google_search],'
3. Start the agent
4. See error: 'Google search tool is not supported for model gemini-live-2.5-flash-preview-native-audio'

**Expected behavior**
Gemini live 2.5 flash should support Google search tool.

**Additional context**
Gemini live 2.5 flash document.
https://cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-flash#live-api-native-audio

hangfei:
This is fixed thanks to @rayira ! Closing this issue.

PR #1515, opened 2025-06-19
fix: Update google_search_tool.py to support updated Gemini LIVE model naming

rayira:
## Description
Fixes [#1512](https://github.com/google/adk-python/issues/1512) by updating google_search_tool.py to support new Gemini LIVE model naming

## Changes
- Update the model name checking in google_search_tool.py


rayira:
will redo to fix cls problem.

PR #1518, opened 2025-06-19
fix: Update google_search_tool.py to support updated Gemini LIVE model naming

rayira:
## Description
Fixes [#1512](https://github.com/google/adk-python/issues/1512) by updating google_search_tool.py to support new Gemini LIVE model naming

## Changes
- Update the model name checking in google_search_tool.py


hangfei:
This is already fixed in : https://github.com/google/adk-python/blob/10197db0d752defc5976d1f276c7b5405a94c75b/src/google/adk/tools/google_search_tool.py#L57

PR #1519, opened 2025-06-19
fix: double json dump evaluation result

sandangel:
Current eval result is double json dumpt. This fix the issue by writing the json string directly.

Signed-off-by: San Nguyen <vinhsannguyen91@gmail.com>


ankursharmas:
Hi @sandangel,

I do want to acknowledge that it the double json conversion in the evalset results manager is not ideal, but making the proposed change will break reading any files that are saved in the old format. And yes we could try to handle the breakage in the code, but that would make the code really ugly. 

So, for now I would keep the existing code as-is.

PR #1529, opened 2025-06-19
Fixes #1514: Handle HttpResponse from google-genai >=1.21.0 in VertexAiSessionService

### LINKED ISSUE #1514: Throw exceptions when using `VertexAiSessionService` with `google-genai` 1.21.0
**Describe the bug**
When using `VertexAiSessionService` with `google-genai` 1.21.0, `create_session` method (maybe also other methods) throw exception with below log.

```
vertex_ai_session_service.py:80 - Create Session response headers={'Content-Type': 'application/json; charset=UTF-8', 'Vary': 'Referer', 'Content-Encoding': 'gzip', 'Date': 'Thu, 19 Jun 2025 06:34:24 GMT', 'Server': 'ESF', 'X-XSS-Protection': '0', 'X-Frame-Options': 'SAMEORIGIN', 'X-Content-Type-Options': 'nosniff', 'Alt-Svc': 'h3=":443"; ma=2592000,h3-29=":443"; ma=2592000', 'Transfer-Encoding': 'chunked'} body='{\n  "name": "projects/813188275284/locations/us-central1/reasoningEngines/1585944367997124608/sessions/6910699411165151232/operations/1616170870357622784",\n  "metadata": {\n    "@type": "type.googleapis.com/google.cloud.aiplatform.v1beta1.CreateSessionOperationMetadata",\n    "genericMetadata": {\n      "createTime": "2025-06-19T06:34:24.124032Z",\n      "updateTime": "2025-06-19T06:34:24.124032Z"\n    }\n  }\n}\n'
File "/usr/local/lib/python3.12/site-packages/google/adk/cli/fast_api.py", line 422, in create_session
File "/usr/local/lib/python3.12/site-packages/google/adk/sessions/vertex_ai_session_service.py", line 92, in create_session
session_id = api_response['name'].split('/')[-3]
TypeError: 'HttpResponse' object is not subscriptable
```

**To Reproduce**
1. Install `google-genai==1.21.0`  and `adk-python>=1.3.0`  and use `VertexAiSessionService` to manage sesions.
2. Run Agent and create a session.
3. See error

**Expected behavior**
No error

 - Python version(python -V): 3.12
 - ADK version(pip show google-adk): 1.3.0
 - Platform: cloud run, agentengine

**Additional context**
From google-genai 1.21.0,  genai api client response is changed with full response.
So it does not return dict object.

https://github.com/googleapis/python-genai/commit/8c9e4ba4c426f55addf49521018db90a360b0374#diff-af6bee559d65a77d542f74d9cdcc4a3fd7fa1b82e9474b5f270873f070d565c9

But `VertexAiSessionService` does not keeping up with the changes.
https://github.com/google/adk-python/blob/main/src/google/adk/sessions/vertex_ai_session_service.py#L92

Current solution is the `google-genai` version must be fixed to 1.20.0 or lower.

swapnita1205:
### What was the issue?
In `google-genai >=1.21.0`, the API methods return an `HttpResponse` object instead of a plain dictionary. This caused `VertexAiSessionService.create_session()` to raise a `TypeError` when trying to subscript the response.

### What is the fix?
Added support for the new response structure by calling `await api_response.json()` before extracting values. This makes the service compatible with both existing and future versions of the library.

### How to reproduce the issue:
1. Use `google-genai==1.21.0` or later
2. Run any agent that uses `VertexAiSessionService.create_session()`
3. Observe the TypeError: `'HttpResponse' object is not subscriptable`

### Impact
Fixes compatibility issue when using latest `google-genai` releases.

Fixes #1514


src/google/adk/sessions/vertex_ai_session_service.py
```diff
@@ -89,7 +89,11 @@ async def create_session(
     )
     logger.info(f'Create Session response {api_response}')
 
-    session_id = api_response['name'].split('/')[-3]
+    if isinstance(api_response, dict):
+        session_id = api_response['name'].split('/')[-3]
+    else:
+        session_json = await api_response.json()
+        session_id = session_json['name'].split('/')[-3]
     operation_id = api_response['name'].split('/')[-1]
```
zack314, line 97: Don't you also need to update this line?

More generally, in the file there are many usages of `api_client.async_request` that probably also need to be updated. For instance
* get_session_api_response.get('sessionState', {})
* list_events_api_response.get('httpHeaders', None)
* list_events_api_response['sessionEvents']
* api_response.get('httpHeaders', None)
* lro_response.get('done', None)

hangfei:
Thanks. This is already fixed by another commit. Thanks though!

PR #1585, opened 2025-06-23
Fix: Corrected handling of parallel function responses to prevent "too many function responses" error..

Th3-C0der:
## What Happens

Currently, when a function call receives multiple responses across different events in parallel with uncorrelated function responses, the event re-writing logic can **incorrectly include too many function responses**. This leads to an error like:

```json
{
  "error": {
    "code": 400,
    "message": "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.",
    "status": "INVALID_ARGUMENT"
  }
}
```

-----

## What Should Happen

The ideal solution for this scenario is somewhat intricate, but several key principles should hold true:

  * When processing events into content, the result should **always be a valid model input**.
  * If multiple **function CALLS** are present with the same ID, only the **most recent** should be taken.
  * If multiple **function RESPONSES** are present, only the **most recent** should be taken.
  * If calls are present in parallel, moving one to ensure the above rule **should not automatically move others**.
  * If responses are present in parallel, moving one to ensure the above rule **should not automatically move others**.
  * If multiple calls should occur in a sequence, they should be collected into a **parallel call**.
  * If multiple responses should occur in a sequence, they should be collected into a **parallel response**.

**Note:** Points 6 and 7 might suggest that models *must* support parallel function calling, which could break compatibility with models that don't. However, this isn't a requirement. The only way these re-writing rules would bring multiple responses together is if a series of function responses were ordered without any interleaved model parts—meaning parallel function responses must already be present. For parallel calls, the logic ensures a matching number of calls and responses in the preceding turn, so parallel function calling also requires parallel responses to already exist. The only other situation for parallel calls is when they are unmatched, which typically only happens at the end of a conversation, and in these cases, they would also need to be present as such.

-----

## Solution

A new method, `_extract_latest_function_parts`, will directly create the `contents` according to these rules. Function calls and responses are specially handled and manipulated, while all other parts remain unchanged. This preserves all existing authentication and invalid event filtering logic, as well as foreign agent part re-writing.

-----

## Code Source

All tests and the proposed implementation were developed **without AI**. Therefore, the likelihood of these code snippets existing in copyrighted source code is negligible.

-----

## Update - Added Validation

An additional validator has been added to the mock model and tests to catch future occurrences of this problem....

### DIFF
--- a/contributing/samples/parallel_functions/README.md
+++ b/contributing/samples/parallel_functions/README.md
@@ -1,0 +1,103 @@
+# Parallel Function Test Agent
+
+This agent demonstrates parallel function calling functionality in ADK. It includes multiple tools with different processing times to showcase how parallel execution improves performance compared to sequential execution.
+
+## Features
+
+- **Multiple async tool types**: All functions use proper async patterns for true parallelism
+- **Thread safety testing**: Tools modify shared state to verify thread-safe operations
+- **Performance demonstration**: Clear time differences between parallel and sequential execution
+- **GIL-aware design**: Uses `await asyncio.sleep()` instead of `time.sleep()` to avoid blocking
+
+## Tools
+
+1. **get_weather(city)** - Async function, 2-second delay
+2. **get_currency_rate(from_currency, to_currency)** - Async function, 1.5-second delay
+3. **calculate_distance(city1, city2)** - Async function, 1-second delay
+4. **get_population(cities)** - Async function, 0.5 seconds per city
+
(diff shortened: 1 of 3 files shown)


mdanowar3 (commented):
mda430325@gmail.com

calvingiles:
Looks like this is very similar to #1582

calvingiles:
I just noticed that the MR description above is a word for word copy of the description in my PR #1582, with. the only differences being some formatting (capitalisation and bold / italics changed). What is particularly odd is that the actual code implementation does not follow the description of the proposed solution - there is no `_extract_latest_function_parts` in this PR (that was the solution I used in #1582).

PR #1619, opened 2025-06-24
Fix: Add optional detailed logging to AgentEvaluator

kubikb:
This PR introduces a `log_detailed_results` parameter to `AgentEvaluator.evaluate()` and `AgentEvaluator.evaluate_eval_set()` in `src/google/adk/evaluation/agent_evaluator.py`.

### Description

This feature enables users to see detailed, per-invocation results during agent evaluations. When `log_detailed_results` is set to `True`, the evaluator will log the actual invocation, expected invocation, score, and evaluation status for each step.

### Motivation

Without this change, users only see a final, aggregated evaluation result (either successful or failed through an `assert`). When an evaluation fails, they don't have a breakdown of the results, which makes debugging difficult. This forces them to "fly blind" or resort to lower-level evaluation methods to get more details, which requires extra work.

While the ADK UI offers detailed result observation, having this capability directly within `pytest` runs is beneficial, especially for CI/CD environments where the UI is not available. This change improves the developer experience by providing a direct and convenient way to debug evaluation tests from the logs.

### DIFF
--- a/src/google/adk/evaluation/agent_evaluator.py
+++ b/src/google/adk/evaluation/agent_evaluator.py
@@ -110,8 +110,7 @@ class AgentEvaluator:
       num_runs: Number of times all entries in the eval dataset should be
         assessed.
       agent_name: The name of the agent.
-      log_detailed_results: Logs detailed results on the console. This is
-        usually helpful during debugging.
+      log_detailed_results: Logs detailed results. All invocation results will be logged if true.
     """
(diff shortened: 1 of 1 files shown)


ankursharmas:
Thank you @kubikb for the PR. 

We actually addressed this in a recent PR: https://github.com/google/adk-python/commit/04de3e197d7a57935488eb7bfa647c7ab62cd9d9

I really appreciate you taking time to propose a fix.

PR #1629, opened 2025-06-24
feat(code_executors): Add GkeCodeExecutor for sandboxed code execution on GKE

syangx39:
### Summary

This PR introduces `GkeCodeExecutor`,  a new code executor that provides a secure and scalable method for running LLM-generated code by leveraging GKE Sandbox. It serves as a robust alternative to local or standard containerized executors by leveraging the **GKE Sandbox** environment, which uses gVisor for workload isolation.

For each code execution request, it dynamically creates an ephemeral Kubernetes Job with a hardened Pod configuration, offering significant security benefits and ensuring that each code execution runs in a clean, isolated environment.

### Key Features of GkeCodeExecutor

* **Dynamic Job Creation**: Uses the Kubernetes `batch/v1` API to create a new Job for each code snippet.
* **Secure Code Mounting**: Injects code into the Pod via a temporary `ConfigMap`, which is mounted to a read-only file.
* **gVisor Sandboxing**: Enforces execution within a `gvisor` runtime for kernel-level isolation.
* **Hardened Security Context**: Pods run as non-root with all Linux capabilities dropped and a read-only root filesystem.
* **Resource Management**: Applies configurable CPU and memory limits to prevent abuse.
* **Automatic Cleanup**: Uses the `ttl_seconds_after_finished` feature on Jobs for robust, automatic garbage collection of completed Pods and Jobs.
* **Node Scheduling**: The executor uses Kubernetes `tolerations` in its Pod specification. This allows the k8s scheduler to place the execution Pod onto a **_pre-configured_** gVisor-enabled node.
* **Module Integration**: The `GkeCodeExecutor` is registered in the `code_executors/__init__.py`, making it available for use by agents. The `ImportError` handling is configured to check for the required `kubernetes` SDK.

### Execution Flow:
54zSPHsQTMaBN3m (1)

1. Agent invokes `GkeCodeExecutor` with the LLM-generated code. 
2. The `GkeCodeExecutor` will `execute_code` – creates a temporary `ConfigMap`, and then create a k8s `Job` to run it.
3. This Job runs a standard `python:3.11-slim` container. The image is pulled once to the node and cached. The Job will mount the ConfigMap as `/app/code.py`
4. The GkeCodeExecutor will monitor the Job to completion, fetch `stdout/stderr` logs from the container, return `CodeExecutionResult` to the LlmAgent, and ensure all temp resources are deleted.
5. The calling agent formats the result and provides a final response to the user. If the result contains error, it will retry up to [error_retry_attempts](https://github.com/google/adk-python/blob/a4d432a9e62a5e759e94168463795cd0e7d7ab72/src/google/adk/code_executors/base_code_executor.py#L36) times.


syangx39:
@hangfei could you assign a reviewer for my PR

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,219 @@
...
+            return f"Error retrieving logs: {e.reason}"
+
+    def _create_code_configmap(self, name: str, code: str) -> None:
+        """Creates a ConfigMap to hold the Python code."""
+        body = client.V1ConfigMap(
+            metadata=client.V1ObjectMeta(name=name), data={"code.py": code}
+        )
+        self._core_v1.create_namespaced_config_map(namespace=self.namespace, body=body)
+
+    def _cleanup_configmap(self, name: str) -> None:
```
eliaslevy: We could patch the `ConfigMap` after the `Job` is created to add an `ownerReference` so that the `ConfigMap` its auto-deleted when the `Job` is deleted. There would still be a race condition that could leave a stale `ConfigMap` is the agent died before the `ownerReference` is added, but it would be a smaller window compared to the agent dying before `_cleanup_configmap` is called.

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,219 @@
...
+            config.load_incluster_config()
+            logger.info("Using in-cluster Kubernetes configuration.")
+        except config.ConfigException:
+            logger.info("In-cluster config not found. Falling back to kubeconfig.")
+            config.load_kube_config()
+
+        self._batch_v1 = client.BatchV1Api()
+        self._core_v1 = client.CoreV1Api()
+
+    def execute_code(
```
eliaslevy: Code executors also support input and output files. Can the GKE executor support them?

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,219 @@
...
+    - Ephemeral, per-execution environments using Kubernetes Jobs.
+    - Secure-by-default Pod configuration (non-root, no privileges).
+    - Automatic garbage collection of completed Jobs and Pods via TTL.
+    - Efficient, event-driven waiting using the Kubernetes watch API.
+    """
+    namespace: str = "default"
+    image: str = "python:3.11-slim"
+    timeout_seconds: int = 300
+    cpu_limit: str = "500m"
+    mem_limit: str = "512Mi"
```
eliaslevy: Should we allow the requested resources and not just the limits to be configured?

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,219 @@
...
+    each execution request. The user's code is mounted via a ConfigMap, and the
+    Pod is hardened with a strict security context and resource limits.
+
+    Key Features:
+    - Sandboxed execution using the gVisor runtime.
+    - Ephemeral, per-execution environments using Kubernetes Jobs.
+    - Secure-by-default Pod configuration (non-root, no privileges).
+    - Automatic garbage collection of completed Jobs and Pods via TTL.
+    - Efficient, event-driven waiting using the Kubernetes watch API.
+    """
```
eliaslevy, line 97: Docs should spell out clearly what K8s RBAC permissions are needed for the GKE executor to work.

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,219 @@
...
+                    return CodeExecutionResult(stderr=f"Job failed. Logs:\n{logs}")
+        finally:
+            watch.stop()
+
+            # If the loop finishes without returning, the watch timed out.
+            raise TimeoutError(
+                f"Job '{job_name}' did not complete within {self.timeout_seconds}s."
+            )
+
+    def _get_pod_logs(self, job_name: str) -> str:
```
eliaslevy: The function is returning errors as strings, which will then be reported as stdout or the code execution result. It should raise exceptions for errors so they can be reported in stderr.

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,219 @@
...
+            logger.error(
+                f"An unexpected error occurred during job '{job_name}': {e}",
+                exc_info=True,
+            )
+            return CodeExecutionResult(stderr=f"An unexpected executor error occurred: {e}")
+        finally:
+            # The Job is cleaned up by the TTL controller, and we ensure the ConfigMap is always deleted.
+            self._cleanup_configmap(configmap_name)
+
+    def _create_job_manifest(self, job_name: str, configmap_name: str) -> client.V1Job:
```
eliaslevy: We should annotate the k8s resources, including the Job, with the invocation id from the `InvocationContext` to have some piece of data to link sandboxes to specific executions. 

There is an execution id (aka execution_session_id) in the execution context, but its only set for "stateful" executors, and its not clear what makes an executor stateful. The `stateful` is not really referenced anywhere from what I can tell.

eliaslevy:
@syangx39 what about blocking network access from the sandbox? Seems like we'd want to add that in. Or are you expecting the developer to create a network policy on his own an apply it to the jobs?

src/google/adk/code_executors/gke_code_executor.py
syangx39: Done. Also added an example deployment_rbac.yaml under samples in this PR.

src/google/adk/code_executors/gke_code_executor.py
syangx39: Sound good. For the readbility, I will add a follow up PR for the PVC support bt adk-agent Pod and the ephemeral code-execution Pods to accomplish this.

High-Level workflow:
1. For each execute_code request that involves files, _create_pvc every time.
2. adk-agent will mount this PVC and write the `input_files` from the CodeExecutionInput into it.
3. GkeCodeExecutor will create the sandboxed Job, also will mount the same PVC.
4. user's code runs inside the sandboxed Pod, it can read the input files from the dir and write its output files to the same dir
5. after Job completes, the adk-agent, can now read the output_files from the shared volume.
6. cleanup

src/google/adk/code_executors/gke_code_executor.py
syangx39: Done.

src/google/adk/code_executors/gke_code_executor.py
syangx39: Done. Added {"adk.agent.google.com/invocation-id": invocation_context.invocation_id}

src/google/adk/code_executors/gke_code_executor.py
syangx39: Done. added raised RuntimeError in _get_pod_log(). Now stdout contains only the actual output from succesful user code; stderr contains output from failed user code OR errors from the executor framework itself

src/google/adk/code_executors/gke_code_executor.py
syangx39: Done. the ConfigMap is now a "child" of the Job.

syangx39 (commented):
**Two Action Items left afterwards**
1. **Follow-up PR on input/output files support with _create_pvc().**
2. regarding this comment "_@syangx39 what about blocking network access from the sandbox? Seems like we'd want to add that in. Or are you expecting the developer to create a network policy on his own an apply it to the jobs?_" \
**gVisor's default network sandboxing** in GKE specifically blocks access to sensitive host-level endpoints like the GKE metadata server. A pod running in the gVisor sandbox can still make network calls to public services on the internet. \
If gke_code_executor blocks network access, the sandboxed code will not be able to `pip install` packages from the internet. So I'm thinking -  \
The best way to handle dependencies is ahead of time. A developer would build an image with the required packages already installed via a requirements.txt file. They would then push this image to Artifact Registry and configure to use their custom image instead of the default `python:3.11`.  \
Anyway, **I will add a block_network_access flag for the flexibility and default to `True`.** But for readbility, will do it in a follow-up PR.

eliaslevy (approved):

hangfei:
What's the startup time for GkeCodeExecutor? Should the user starts it first or it can be triggered on-demand?

contributing/samples/gke_agent_sandbox/deployment_rbac.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: v1
```
hangfei, line 1: should every user write this?

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,248 @@
...
+from google.adk.code_executors.base_code_executor import BaseCodeExecutor
+from google.adk.code_executors.code_execution_utils import CodeExecutionInput, CodeExecutionResult
+
+from kubernetes import client, config
+from kubernetes.client.rest import ApiException
+from kubernetes.watch import Watch
+
+logger = logging.getLogger(__name__)
+
+class GkeCodeExecutor(BaseCodeExecutor):
```
hangfei, line 36: i was wondering if there will be other variant of GkeCode executor other than gVisor-based?

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,248 @@
...
+    - Secure-by-default Pod configuration (non-root, no privileges).
+    - Automatic garbage collection of completed Jobs and Pods via TTL.
+    - Efficient, event-driven waiting using the Kubernetes watch API.
+
+    RBAC Permissions:
+    This executor interacts with the Kubernetes API and requires a ServiceAccount 
+    with specific RBAC permissions to function. The agent's pod needs permissions
+    to create/watch Jobs, create/delete ConfigMaps, and list Pods to read logs.
+    For a complete, working example of the required Role and RoleBinding, see the
+    file at: contributing/samples/gke_agent_sandbox/deployment_rbac.yaml
```
hangfei: source code should not depends on samples which will be changed frequently.

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,248 @@
...
+logger = logging.getLogger(__name__)
+
+class GkeCodeExecutor(BaseCodeExecutor):
+    """Executes Python code in a secure gVisor-sandboxed Pod on GKE.
+
+    This executor securely runs code by dynamically creating a Kubernetes Job for
+    each execution request. The user's code is mounted via a ConfigMap, and the
+    Pod is hardened with a strict security context and resource limits.
+
+    Key Features:
```
hangfei: no need to add this to source code.

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,248 @@
...
+
+    RBAC Permissions:
+    This executor interacts with the Kubernetes API and requires a ServiceAccount 
+    with specific RBAC permissions to function. The agent's pod needs permissions
+    to create/watch Jobs, create/delete ConfigMaps, and list Pods to read logs.
+    For a complete, working example of the required Role and RoleBinding, see the
+    file at: contributing/samples/gke_agent_sandbox/deployment_rbac.yaml
+    """
+    namespace: str = "default"
+    image: str = "python:3.11-slim"
```
hangfei: does it only work with 3.11?

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,248 @@
...
+    timeout_seconds: int = 300
+    cpu_request: str = "200m"
```
hangfei: cpu_requested?

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,248 @@
...
+    mem_request: str = "256Mi"
+    cpu_limit: str = "500m"
```
hangfei: please add comment on what does the unit "m" mean

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,248 @@
...
+    def execute_code(
+        self,
+        invocation_context: InvocationContext,
+        code_execution_input: CodeExecutionInput,
+    ) -> CodeExecutionResult:
+        """Orchestrates the secure execution of a code snippet on GKE."""
+        job_name = f"adk-exec-{uuid.uuid4().hex[:10]}"
+        configmap_name = f"code-src-{job_name}"
+
+        try:
```
hangfei, line 121: add doc to explain the logic

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -0,0 +1,248 @@
...
+    timeout_seconds: int = 300
+    cpu_request: str = "200m"
+    mem_request: str = "256Mi"
+    cpu_limit: str = "500m"
+    mem_limit: str = "512Mi"
+
+    _batch_v1: client.BatchV1Api
+    _core_v1: client.CoreV1Api
+
+    def __init__(self, **data):
```
hangfei: does this work for AI Studio api key or EasyGCP?

If not, is it possible to throw an exception when it inits?

hangfei:
How long does it take to spin up the code executor and finish it?

hangfei:
Please add unit tests.

Please add a sample under contributing/samples

hangfei:
Please make sure you add a corresponding docs in adk-docs before submission. Thanks.

contributing/samples/gke_agent_sandbox/deployment_rbac.yaml
syangx39: No, the default namespace is ["default"](https://github.com/syangx39/adk-python/blob/5533fbf31085a98184b91bda48b36a515524ea84/src/google/adk/code_executors/gke_code_executor.py#L35). The file deployment_rbac.yaml located in the contributing/samples/ directory is intended to be a sample. It demonstrates all the necessary RBAC components (Namespace, ServiceAccount, Role, RoleBinding) and how they connect. The name agent-sandbox was chosen to be descriptive for the sample, but it's not a required value.

src/google/adk/code_executors/gke_code_executor.py
syangx39: No. We've designed the GkeCodeExecutor to be exclusively gVisor-based because its core mission is to safely run untrusted, LLM-generated code. Offering a non-sandboxed variant would create a significant security liability, as it would expose the node's kernel to the code being executed.

src/google/adk/code_executors/gke_code_executor.py
syangx39: I think this 'Key Features' serve as a high-level summary so a future developer can immediately grasp the component's security posture and design without having to read the whole implementation. 

Do you think we could keep it for that reason? I feel it adds a lot of value for future maintainability.

src/google/adk/code_executors/gke_code_executor.py
syangx39: Sounds good, if I replace the "see the
   file at: contributing/samples/gke_agent_sandbox/deployment_rbac.yaml" with """"...
   RBAC Permissions:
   This executor requires a ServiceAccount with specific RBAC permissions. The
   Role granted to the ServiceAccount must include rules to manage Jobs,
   ConfigMaps, and Pod logs. Below is a minimal set of required permissions:

   rules:
     # For creating/deleting code ConfigMaps and patching ownerReferences
     - apiGroups: [""]  # Core API Group
       resources: ["configmaps"]
       verbs: ["create", "delete", "get", "patch"]
     # For watching Job completion status
     - apiGroups: ["batch"]
       resources: ["jobs"]
       verbs: ["get", "list", "watch", "create", "delete"]
     # For retrieving logs from the completed Job's Pod
     - apiGroups: [""] # Core API Group
       resources: ["pods", "pods/log"]
       verbs: ["get", "list"]
"""

Are you okay with this comment then?

src/google/adk/code_executors/gke_code_executor.py
syangx39: Not necessary 3.11, but it has to be Python https://github.com/syangx39/adk-python/blob/5533fbf31085a98184b91bda48b36a515524ea84/src/google/adk/code_executors/gke_code_executor.py#L110

I can patch a follow-up PR to make the command configurable, turning it into a more generic script executor.

src/google/adk/code_executors/gke_code_executor.py
syangx39: It's the guaranteed minimum amount of CPU that Kubernetes reserves for a container.

src/google/adk/code_executors/gke_code_executor.py
syangx39: Done.

src/google/adk/code_executors/gke_code_executor.py
syangx39: AI Studio API key is irrelevant for this component. The GkeCodeExecutor's only job is to communicate with the Kubernetes API, either via ServiceAccount token or kubeconfig fail. 

If load_kube_config() fails, it raises a kubernetes.config.ConfigException. This exception will halt the initialization and cause the GkeCodeExecutor constructor to fail, which I think is the correct behavior when no valid Kubernetes credentials can be found.

src/google/adk/code_executors/gke_code_executor.py
syangx39: Done.

syangx39:
@hangfei 
1. "Please add a sample under contributing/samples"
added, `contributing/samples/gke_agent_sandbox/deployment_rbac.yaml` is the permission file and `contributing/samples/code_execution/gke_sandbox.agent.py` is the agent app. I feel like adding them at different places might confuse reader -- **maybe I should move both under `contributing/samples/code_execution/`??**  

2. "Please add unit tests."
added 
```
tests/unittests/code_executors/test_gke_code_executor.py::TestGkeCodeExecutor::test_create_job_manifest_structure[VERTEX] PASSED                                                            [100%]
```

3. "What's the startup time for GkeCodeExecutor?"
We haven't really tested this but I can give you a ballpark figure
Here's the breakdown of startup time:
- API latency: agent making API call to k8s control plane to create ConfigMap and Job. This is negligible (miliseconds).
- Pod Scheduling: k8s scheduler assign new Pod to a gVisor-enabled node. If healthy cluster w available nodes, this is miliseconds to seconds.
- Pod Initialization: 
	- image pull is the biggest factor. The node must have the container image. For the first time (cold start), it download from registry, since python:3.11-slim is a lightweight image, it takes seconds. For the following times (cache), it's neglible.
	- gVisor Sandbox creation: slight overhead (miliseconds?)
- From my previous experience, it usually takes a few seconds (for the 1st time, the following times are instant due to cache) on a healthy cluster. We're working on cold start optimization.

4. "Should the user starts it first or it can be triggered on-demand?"
triggered on-demand, [code_executor.execute_code(...)](https://github.com/google/adk-python/blob/a858d79b3af1f42e4675b352626dfeef507a48aa/src/google/adk/flows/llm_flows/_code_execution.py#L230
) automatically run analysis code on any new data files the user provides.

5. "please create a issue and doc to update docs ad adk-docs repo. Please make sure you add a corresponding docs in adk-docs before submission."
Sure. Do you have a specific page you want me to put? 
maybe https://google.github.io/adk-docs/tools/built-in-tools/#code-execution?

hangfei:
> 5\. maybe https://google.github.io/adk-docs/tools/built-in-tools/#code-execution?

this works: maybe https://google.github.io/adk-docs/tools/built-in-tools/#code-execution?

src/google/adk/code_executors/gke_code_executor.py
hangfei: fair point.

src/google/adk/code_executors/gke_code_executor.py
hangfei: that's better. thanks.

src/google/adk/code_executors/gke_code_executor.py
hangfei: please. thanks.

src/google/adk/code_executors/gke_code_executor.py
hangfei: i mean the naming is confusing. usually i see "cpu_requested" or "cpu"

hangfei:
Overall LGTM. Thanks!

Let's do the following before merge: 
1. have a adk-docs PR for this change.
2. fix the tests.

syangx39:
> Overall LGTM. Thanks!
> 
> Let's do the following before merge:
> 
> 1. have a adk-docs PR for this change.
> 2. fix the tests.

@hangfei 
1. I added the "adk-docs" PR here: https://github.com/google/adk-docs/pull/621/files 
2. What do you mean by "fix the tests"? The unit tests are already added in `tests/unittests/code_executors/test_gke_code_executor.py` and they are all passed. Here's the log
```
$ pip install -e .[test,gke]
$ pytest tests/unittests/code_executors/test_gke_code_executor.py
========================================================== test session starts ==========================================================
platform linux -- Python 3.12.3, pytest-8.4.1, pluggy-1.6.0
rootdir: /home/user/adk-python
configfile: pyproject.toml
plugins: asyncio-1.1.0, mock-3.14.1, anyio-4.10.0, langsmith-0.4.14, xdist-3.8.0
asyncio: mode=Mode.AUTO, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 14 items                                                                                                                      

tests/unittests/code_executors/test_gke_code_executor.py ..............                                                           [100%]

========================================================== 14 passed in 7.94s ===========================================================
```

syangx39:
@hangfei 
The doc PR got merged. Please merge this PR as well for consistency.
https://google.github.io/adk-docs/tools/built-in-tools/#gke-code-executor

hangfei:
thanks. 

1. there are some formatting and tests failures. plz fix. 

2. could you share the steps to test the code execuctor so we can test it?

contributing/samples/gke_agent_sandbox/deployment_rbac.yaml
seanzhou1023: great, could you add a README.md to this sample agent and putting information above in it , so that users understand the usage of this file ?

pyproject.toml
```diff
@@ -135,6 +135,10 @@ extensions = [
   "toolbox-core>=0.1.0",                  # For tools.toolbox_toolset.ToolboxToolset
 ]
 
+# For GKE-specific features, primarily the GkeCodeExecutor.
+gke = [
```
seanzhou1023: let's put it into extensions.

src/google/adk/code_executors/__init__.py
```diff
@@ -49,3 +49,14 @@
       ' Executor with agents, please install it. If not, you can ignore this'
       ' warning.'
   )
+
+try:
```
seanzhou1023: the original codes in this file is problematic, we've just fixed it, let's follow the latest style.

src/google/adk/code_executors/gke_code_executor.py
seanzhou1023: I think  hangfei meant, for sandboxed solution, will there be other option besides gVisor (e.g. Kata Containers) , thus shall we add some thing related to gVisor in the name ?

syangx39:
> 1. there are some formatting and tests failures. plz fix.
> 2. could you share the steps to test the code execuctor so we can test it?

1. sounds good. I forgot to add "kubernetes" in [test] in pyproject.toml, fixed it
2. Sounds good. This is the plan:

### GkeCodeExecutor Validation Plan

#### Step 1: Set Up the GKE Autopilot Cluster
```
gcloud container clusters create-auto adk-sandbox-cluster \
  --project="<YOUR_PROJECT_ID>" \
  --region="<YOUR_GCP_REGION>"
```

#### Step 2: Deploy the Agent and Send Predictions
1. Apply RBAC Permissions `kubectl apply -f deployment_rbac.yaml`.
2. Build and Push the adk Agent Image 
3. Update `deployment_agent.yaml` with your image path and run `kubectl apply -f ...`
4. `kubectl port-forward svc/adk-agent-service 8080:80 -n adk-sandbox`
5. gVisor Sandbox Execution
```
curl -X POST http://localhost:8080/invoke \
-H "Content-Type: application/json" \
-d '{
    "prompt": "Write a Python script to query the GKE metadata server for the node'\''s instance ID and print the result. If it fails, print the error."
}'
```

GWeale:
LGTM 
all scenarios tested worked!

GWeale (approved):

GWeale:
Merged!

PR #1644, opened 2025-06-25
fix: issue 1641 database session service leaks events

sindrigunnars:
fix: issue 1641 database session service leaks events

### DIFF
--- a/src/google/adk/sessions/database_session_service.py
+++ b/src/google/adk/sessions/database_session_service.py
@@ -459,7 +459,9 @@ class DatabaseSessionService(BaseSessionService):
 
       storage_events = (
           session_factory.query(StorageEvent)
+          .filter(StorageEvent.app_name == app_name)
           .filter(StorageEvent.session_id == storage_session.id)
+          .filter(StorageEvent.user_id == user_id)
           .filter(timestamp_filter)
           .order_by(StorageEvent.timestamp.desc())
           .limit(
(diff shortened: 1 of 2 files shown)


DeanChensj (approved):

DeanChensj (approved):
Could we run auto_format.sh to resolve the error?

DeanChensj (approved):

DeanChensj:
Could you fix:
```
This pull request has 7 commits.
Please squash them into a single commit before merging.
You can use git rebase -i HEAD~N
...where N is the number of commits you want to squash together. The PR check conveniently tells you this number! For example, if the check says you have 3 commits, you would run: git rebase -i HEAD~3.
Because you have rewritten the commit history, you must use the --force flag to update the pull request: git push --force
```

sindrigunnars:
Should be good now :)

src/google/adk/sessions/database_session_service.py
```diff
@@ -460,6 +460,7 @@ async def get_session(
       storage_events = (
           session_factory.query(StorageEvent)
           .filter(StorageEvent.session_id == storage_session.id)
+          .filter(StorageEvent.user_id == user_id)
           .filter(timestamp_filter)
           .order_by(StorageEvent.timestamp.desc())
           .limit(
```
rpedela-recurly: Is `.filter(StorageEvent.app_name == app_name)` also needed since the `events` table PK is defined as `primary key (id, app_name, user_id, session_id)`?

sindrigunnars: You bring up a good point that I hadn't considered. Yes it should probably also filter by app_name. I will get to fixing it and cleaning up the pull request later today.

rpedela-recurly: Awesome, thanks!

sindrigunnars:
@DeanChensj Main has been merged in to the branch and the commits squashed. Should be ready for merge into main.

Issue #1637, opened 2025-06-25
TypeError: transfer_to_agent() got an unexpected keyword argument 'parameters'


woosung1223:
```bash
 File "/.venv/lib/python3.12/site-packages/google/adk/flows/llm_flows/base_llm_flow.py", line 417, in _postprocess_handle_function_calls_async
    if function_response_event := await functions.handle_function_calls_async(
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.12/site-packages/google/adk/flows/llm_flows/functions.py", line 165, in handle_function_calls_async
    function_response = await __call_tool_async(
                        ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.12/site-packages/google/adk/flows/llm_flows/functions.py", line 430, in __call_tool_async
    return await tool.run_async(args=args, tool_context=tool_context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.12/site-packages/google/adk/tools/function_tool.py", line 82, in run_async
    return self.func(**args_to_call) or {}
           ^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: transfer_to_agent() got an unexpected keyword argument 'parameters'
```

I'm getting an error like this in a rare case when calling a subagent (about 1 out of 20 times).

I've only modified the tool and sub-agent configuration, so I don't think there's anything I need to change. How do I fix this? thanks.

Apollovishwas:
I Have the exact same problem. 

When transferring, in your case, it's passing  "parameters".
In my case it as passing "query" and sometimes "questions"

I've solved this by explicitly mentioning not to add any parameters when transferring, in the prompt
(I know, I know, it's not the fix, but it works)

cschmidterNBTC:
I haven't encountered this since they closed this issue #458. But, for a workaround, you may want to add `**kwargs` to `transfer_to_agent` or change the docstring there yourself.

hangfei:
@woosung1223 thanks for reporting this. 

Could you share the error log(what function returned by model)? 

Could you share a minimal example to reproduce it?

woosung1223:
@hangfei 

```
Content(parts=[
Part(video_metadata=None, thought=None, code_execution_result=None, executable_code=None, file_data=None, 
[...]

PR #1656, opened 2025-06-26
Fix: Handle unexpected 'parameters' argument in FunctionTool.run_async

hangfei:
The LLM occasionally includes an unexpected 'parameters' argument when calling tools, specifically observed with 'transfer_to_agent'. This change makes FunctionTool.run_async more robust by filtering arguments against the function signature before invocation.

This resolves issue #1637.

This is generated by Jules Agent.

### DIFF
--- a/src/google/adk/tools/function_tool.py
+++ b/src/google/adk/tools/function_tool.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


hangfei:
@boyangsvl does this change make sense or we should still let model handle it all?

boyangsvl (approved):

boyangsvl (approved):

PR #1679, opened 2025-06-26
chore: updating CONTRIBUTING.md with missing extra

jackwotherspoon:
Contributing doc says to do the following:

```sh
uv sync --extra test --extra eval
pytest ./tests/unittests
```

If you follow this the tests will fail:

```sh
tests/unittests/a2a/executor/test_task_result_aggregator.py:27: in <module>
    from a2a.types import Message
E   ModuleNotFoundError: No module named 'a2a'
```

This makes sense since the `a2a` package is not part of ADK's core dep, it is an extra:

https://github.com/google/adk-python/blob/e79651cd86ba3f0c998109f2140f1db2cab78708/pyproject.toml#L79-L83

Thus for a2a tests to pass we must include the extra in the sync command:

```sh
uv sync --extra test --extra eval --extra a2a
pytest ./tests/unittests
```


jackwotherspoon:
@seanzhou1023 mind giving this a stamp? 😄

wuliang229 (approved):

PR #1709, opened 2025-06-28
chore: Add util to build our llms.txt and llms-full.txt files

hangfei:
Add a script to build the llms and llms-full txt files. 

1. both files will be in markdown by convention
2. keep only python code samples (can add java support later)
3. will keep these two files in this repo so it's easier to find.
4. later, we can add this as part of our workflow CI/CD

Token count: 
llms.txt tokens: 1523
llms-full.txt tokens: 240465


llms.txt
```diff
@@ -0,0 +1,309 @@
...
+
+
+
+
+Happy Agent Building!
+
+**Source:** [adk-python repository](https://github.com/google/adk-python)
+
+## Documentation
+- [Custom agents](https://google.github.io/adk-docs/agents/custom-agents.md)
```
boyangsvl: does the LLM follow the link?

contributing/dev/utils/build_llms_txt.py
boyangsvl, line 1: Are there existing tools that can convert a doc site to llms.txt?

hangfei: There is few, for example: https://wordlift.io/generate-llms-txt/. 

The unique need for us is that we need to split python and java so it doesn't waste the context window. So i didn't find a tool that can do this effectively.

llms.txt
hangfei: Yes. it follows when enabled with search/url. 

It's the recommended format as well: https://llmstxt.org/#format

PR #1790, opened 2025-07-04
Add support for including history windowin current turn contents

sourabh-patil:
Added one parameters history_window_size for LlmAgent class that is dependent on include_contents parameters. Now, include_contents can have 3 values "none","default" and "windowed". All works like before but if user sets "windowed", user also needs to set "history_window_size" as well (which has to be positive). It is used in adk-python/src/google/adk/flows/llm_flows/contents.py to get the windowed history.


surajksharma07:
Hi @sourabh-patil and thanks to everyone who participated in this thread.

Issue #1803, opened 2025-07-06
Agent() doesn't accept tools in `list[FunctionTool]`


jokester:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**

```py
from google.adk import Agent
from google.adk.tools import FunctionTool

tools: list[FunctionTool] = ...
Agent(..., tools=tools)
```

Will fail to type check (like in Pylance) like:

```
Argument of type "list[FunctionTool]" cannot be assigned to parameter "tools" of type "list[ToolUnion]" in function "__init__"
  "list[FunctionTool]" is not assignable to "list[ToolUnion]"
    Type parameter "_T@list" is invariant, but "FunctionTool" is not the same as "ToolUnion"
    Consider switching from "list" to "Sequence" which is covariantPylance[reportArgumentType](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportArgumentType.md)
```

Because ADK had class Agent like:

```
class Agent(BaseModel)
  ...
  tools: list[ToolUnion] = Field(default_factory=list)
```

and `list[T]` seems to be invariant on T in Python.

**Expected behavior**

Passing a `list[FunctionTool]` should type check.

**Screenshots**

**Additional context**

jokester:
Changing `Agent.tools` to `Sequence[ToolUnion]` in my venv fixed the type mismatch.

IMO there is no much reason to use invariant `list[T]`? All uses of `agent.tools` I could find are like `for t in tools`.

(I'm not a type guru, opinions welcome)

jokester:
If this is worth fixing I have a tiny PR at https://github.com/google/adk-python/pull/1804

hangfei:
I tested the following and it works. Could you try with my example? 

If you still see this issue and feel free to re-open. 
```

print('list of tools with types.....')
tools: list[FunctionTool] = [FunctionTool(roll_die), FunctionTool(check_prime)]


root_agent = Agent(
    model='gemini-2.0-flash',
    name='hello_world_agent',
    description=(
        'hello world agent that can roll a dice of 8 sides and check prime'
        ' numbers.'
    ),
    instruction="""
      You roll dice and answer questions about the outcome of the dice rolls.
      You can roll dice of different sizes.
      You can use multiple tools in parallel by calling functions in parallel(in one request and in one round).
      It is ok to discuss previous dice roles, and comment on the dice rolls.
[...]

PR #1804, opened 2025-07-06
change Agent.tools to covariant Sequence[T]

jokester:
fixes https://github.com/google/adk-python/issues/1803

### DIFF
--- a/src/google/adk/agents/llm_agent.py
+++ b/src/google/adk/agents/llm_agent.py
@@ -25,6 +25,7 @@ from typing import ClassVar
 from typing import Dict
 from typing import Literal
 from typing import Optional
+from typing import Sequence
 from typing import Type
 from typing import Union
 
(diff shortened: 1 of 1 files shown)


hangfei:
Could you fix the failed checks?

jokester:
@hangfei I think I fixed it. Can you run the check again?

surajksharma07:
Hi @jokester and thanks to everyone who participated in this thread.

Jacksunwei:
@jokester 

Thank you for the PR! However, we're closing this as we cannot accept the change.

While `Sequence[ToolUnion]` would fix the type checking issue, we need to keep `list[ToolUnion]` because:
1. The API is already released and stable (changing it would be a breaking change)
2. The field needs to be mutable - ADK itself modifies `agent.tools` at runtime (e.g., [SequentialAgent appends tools](https://github.com/google/adk-python/blob/main/src/google/adk/agents/sequential_agent.py#L149))

See the explanation and workaround in https://github.com/google/adk-python/issues/1803#issuecomment-3487621904

Closing the associated issue as won't fix.

Issue #1816, opened 2025-07-07
How do I pass authorization header to LlmAgent tool?


matino:
I have an instance that has a `Runner` object as a dependency. I want to pass the request "authorization" header value to the runner agent tool. As I understand, the correct way is to do it via session service (per https://github.com/google/adk-python/issues/267). However, this doesn't work when I prefix the key with `temp:` (since I don't want to persist tokens). 

```
        state_changes = {
            "temp:token": "<my token>",
        }
        actions_with_update = EventActions(state_delta=state_changes)
        system_event = Event(
            invocation_id="inv_login_update",
            author="system",  # Or 'agent', 'tool' etc.
            actions=actions_with_update,
        )

        await self.runner.session_service.append_event(session, system_event)
```

If I change the key from `temp:token` to `user:token`, the state is visible in the LlmAgent tool inside the `ToolContext` state.

Can somebody suggest a better approach?

sindrigunnars:
I agree that the temp session key is fairly misleading. The way I've done this is just to make it user: specific. You could do some redis acrobatics with TTL to have it truly temporary or even write an empty value to it in the agent when you have retrieved it.

Let me know if you have any other suggestions/fixed I may have missed.

matino:
I think another option would be to use the credential service. The `Runner` instance could call `save_credentials`, then the LLM tool could call `load_credentials` with `tool_context._invocation_context.credential_service.load_credential(...)`. That's not super clean, but I guess it's more secure than persisting the token in the session service (even if only for a short while).

sindrigunnars:
But does this solve the problem in the context of MCP tools?

matino:
> But does this solve the problem in the context of MCP tools?

I haven't used MCP tools, so can't really tell :/

seanzhou1023:
the key prefixed with `temp:` won't be added to the session state. One thing you can do is that each time your tool read it your tool can remove it from the session state.

seanzhou1023:
feel free to reopen if the above suggestion still doesn't work for you.

matino:
@seanzhou1023 Could you please help with reviewing/merging https://github.com/google/adk-python/pull/1829? It would really help to use credential service for secrets instead of session service 🙏

PR #1829, opened 2025-07-08
feat: expose credential service in readonly context

matino:
The problem is described in https://github.com/google/adk-python/issues/1816.
TLDR; currently, there's no clean way to pass secrets to the LLM.

This PR makes it possible to access the credential service in a read-only context. It can be use to pass the `authorization` header to the `LlmAgent` tool:

- In `Runner` instance, call `save_credential()`
- In LLM, use `tool_context.credential_service.load_credential()` to load the credential (tool_context is of `ToolContext` type)


src/google/adk/tools/agent_tool.py
```diff
@@ -116,6 +116,7 @@ async def run_async(
         artifact_service=ForwardingArtifactService(tool_context),
         session_service=InMemorySessionService(),
         memory_service=InMemoryMemoryService(),
+        credential_service=tool_context.credential_service,
```
matino: Not sure if we should do the same (i.e. use `tool_context` services) for session and memory service.

matino:
@hangfei Do you think we can merge this? It's a real blocker for my use case (especially passing credential service to the `Runner` in agent_tool)

src/google/adk/agents/readonly_context.py
```diff
@@ -52,3 +53,8 @@ def agent_name(self) -> str:
   def state(self) -> MappingProxyType[str, Any]:
     """The state of the current session. READONLY field."""
     return MappingProxyType(self._invocation_context.session.state)
+
+  @property
```
seanzhou1023: I think you don't need to expose credential service here as bridge , your purpose is to pass it on to agent tool , right ?

seanzhou1023 (approved):

src/google/adk/tools/agent_tool.py
seanzhou1023: no need for session and probably needed for memory service. but we can address in another PR.

src/google/adk/agents/readonly_context.py
matino: @seanzhou1023 It was my intention. This way I can call credential service from the agent tool via the `ToolContext`. Currently, I need to rely on the "protected" `_invocation` field: `tool_context._invocation_context.credential_service` - not so great TBH.

Issue #929, opened 2025-05-25
LiteLLM giving error with OpenAI models and Grafana's MCP server


umangk-2410:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
I was integrating OpenAI's `gpt-4o` model ADK using LiteLLM, but it gave me error regarding the schema for one of tools for the Grafana's MCP server.

**To Reproduce**
1. Defining the Agent
```python
root_agent = LlmAgent(
            model=LiteLlm(model="openai/gpt-4o"),
            name='grafana_retrieval_agent',
            description="Helps user retrieve metrics and logs from Grafana Datasources.",
            instruction=REVISED_RETRIEVAL_PROMPT,
            tools=tools,
            output_key="retrieval_agent_output"
        )
```
2. Defining the MCP server
```python
async def grafana_mcp():
    try:
        print(f"Attempting to connect to Grafana MCP Server")
        tools, exit_stack = await MCPToolset.from_server(
            connection_params = StdioServerParameters(
                command="mcp-grafana",
                args=[],
                env={
                    "GRAFANA_URL": "http://localhost:3000",
                    "GRAFANA_API_KEY": GRAFANA_API_KEY
                }
            )
        )
        return tools, exit_stack
    except Exception as e:
        print(f"Failed to connect to Grafana MCP Server: {e}")
        return [], None
```
3. Defining the `run` logic
```python
    async def run(self, query: str):
        session_service = InMemorySessionService()
        session = session_service.create_session(
            state={}, app_name='mcp_grafana_app', user_id='user_fs'
        )

        print(f"\nUser Query for Retrieval Agent:\n '{query}'")
        content = types.Content(role='user', parts=[types.Part(text=query)])

        root_agent, exit_stack = await self.init_agent()
        runner = Runner(
            app_name='mcp_grafana_app',
            agent=root_agent,
            session_service=session_service,
        )

        print("Running Retrieval Agent...")
        events_async = runner.run_async(
            session_id=session.id, user_id=session.user_id, new_message=content
        )

        output = None
        async for event in events_async:
            # print(event.content.parts[0])
            text = event.content.parts[0].text
            function_response = event.content.parts[0].function_response
            # if text is not None:
                # print(text)
            if function_response is not None:
                print(function_response.response["result"])
            output = text

        print("Closing MCP server connection...")
        await exit_stack.aclose()
[...]

PR #2020, opened 2025-07-17
fix: litellm tool serialization

thiagosalvatore:
This PR fixes the issue reported on #929 

When serializing tools with a deep level of nesting, it was leaving behind some attributes, causing issues like `Invalid schema for function 'search_shop_catalog': 'STRING' is not valid under any of the given schemas`.

This PR does a small refactor on the `_schema_to_dict` function adding support for that.
The core issue was that after calling `model_dump`, there's no longer an entry of type `types.Schema`, so the logic inside `properties` was not being able to serialize the dict there.

I've also added one extra test for this specific scenario so we have coverage there too.

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -311,46 +311,45 @@ TYPE_LABELS = {
 
 
 def _schema_to_dict(schema: types.Schema) -> dict:
-    """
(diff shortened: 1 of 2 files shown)


seanzhou1023 (approved):

seanzhou1023:
thanks for the fix :)

seanzhou1023 (approved):

Issue #2053, opened 2025-07-19
OpenAPI tool ignores required parameters for rest interfaces


jeffbryner:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
When attempting to call a rest api via the openapi tool/spec parser, required api parameters not passed by the model are not used, even if defaults are present in the spec. This is especially true for 'path' parameters like are used in many google apis. 

**To Reproduce**
Attempt to use an open api rest interface like gmail. Most api calls for the gmail api expect a userId parameter. The default is listed as 'me', but if it is not sent the call to the function/tool fails noting a required parameter. 

**Expected behavior**
Expected that the api parser/rest api tool would fill in required parameters, especially if they are given defaults. 

 **Model Information:**
gemini pro/flash 2.5

**Additional context**
I'm using the gmail api as represented by the open api spec. 
A sample function declaration:

```
    "/gmail/v1/users/{userId}/messages": {
      "get": {
        "description": "Lists the messages in the user's mailbox.",
        "operationId": "gmail.users.messages.list",
        "responses": {
          "200": {
            "description": "Successful response",
            "content": {
              "*/*": {
                "schema": {
                  "$ref": "#/components/schemas/ListMessagesResponse"
                }
              }
            }
          }
        },
        "parameters": [
          {
            "name": "userId",
            "in": "path",
            "description": "The user's email address. The special value `me` can be used to indicate the authenticated user.",
            "required": true,
            "schema": {
              "type": "string",
              "default": "me"
            }
          },
          {
            "name": "maxResults",
            "in": "query",
            "description": "Maximum number of messages to return. This field defaults to 100. The maximum allowed value for this field is 500.",
            "schema": {
              "type": "integer",
              "default": 100
            }
          },
          {
            "name": "pageToken",
            "in": "query",
            "description": "Page token to retrieve a specific page of results in the list.",
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "q",
            "in": "query",
            "description": "Only return messages matching the specified query. Supports the same query format as the Gmail search box. For example, `\"from:someuser@example.com rfc822msgid: is:unread\"`. Parameter cannot be used when accessing the api using the gmail.metadata scope.",
            "schema": {
              "type": "string"
            }
          },
```
 
[...]

PR #2054, opened 2025-07-19
fix: Add required parameters with defaults if missing

jeffbryner:
fixes https://github.com/google/adk-python/issues/2053


seanzhou1023:
please update the commit message to follow https://www.conventionalcommits.org/en/v1.0.0/ and update the PR title to match it.

codeninja:
We have noticed a serious issue with tool calling that may be related to this which is preventing us from updating from 1.6 -> 1.8.

When calling our tools with 1.8, ADK is sending undefined values for optional fields, rather than omitting them entirely from the request. This gets interpreted downstream by our GO API as an invalid value. Since Null != undefined and it's expecting the request to omit the values.

I suspect it may be related to this code.

Would it be possible to drop values entirely from the request which have undefined values?

jeffbryner:
> We have noticed a serious issue with tool calling that may be related to this which is preventing us from updating from 1.6 -> 1.8.
@codeninja I don't believe this code has been merged into 1.8 as the PR is still open.

codeninja:
@jeffbryner I understand.

I believe our issue is introduced by line 379 in this code, above your diff, as it stands in 1.8. I found my way here while researching our issue. 

I believe that the the API params are being polluted with undefined values which are making their way into the request.

jeffbryner:
> I believe our issue is introduced by line 379 in this code, above your diff, as it stands in 1.8. I found my way here while researching our issue.
> 
> I believe that the the API params are being polluted with undefined values which are making their way into the request.


Ah, I see. Probably best as a separate bug? This PR handles the case where the model doesn't send default parameters (probably assuming they are defaulted in the API) and explicitly sets defined, default values.

jeffbryner:
Thanks!

Issue #2101, opened 2025-07-22
actions parameter in ApplicationIntegrationToolset has wrong type


manifoldtimblin:
**Describe the bug**
The actions parameter of the ApplicationIntegrationToolset class has the wrong type. The type [currently in the codebase](https://github.com/google/adk-python/blob/13ff009d34836a80f107cb43a632df15f7c215e4/src/google/adk/tools/application_integration_tool/application_integration_toolset.py#L90) is `Optional[str]`, however the [example in the comments](https://github.com/google/adk-python/blob/13ff009d34836a80f107cb43a632df15f7c215e4/src/google/adk/tools/application_integration_tool/application_integration_toolset.py#L71) and usage as an [argument in the instantiation of IntegrationClient](https://github.com/google/adk-python/blob/13ff009d34836a80f107cb43a632df15f7c215e4/src/google/adk/tools/application_integration_tool/application_integration_toolset.py#L150) indicate the type should actually be `Optional[list[str]]`

jwiejulie:
Commit merged! Closing, thank you!

PR #2102, opened 2025-07-22
Use correct type for actions parameter in ApplicationIntegrationToolset

manifoldtimblin:
Addresses https://github.com/google/adk-python/issues/2101

I've ran `pytest ./tests/unittests`, and all tests passed.


boyangsvl (approved):

hangfei:
merged. thanks for the fix!

PR #2148, opened 2025-07-24
fix: eval module not found exception string

### LINKED ISSUE #2071: no matches found: google-adk[eval]
**Describe the bug**
when I try to use **eval** in adk, there is an exception```Eval module is not installed, please install via "pip install google-adk[eval]"```. I followed the exception and tried to install eval dependencies using **pip install google-adk[eval]** . `I'm getting no matches found: google-adk[eval]`

**To Reproduce**
1. Install 'pip install google-adk[eval]'
2. See error `I'm getting no matches found: google-adk[eval]`

**Expected behavior**
The eval dependencies to be installed correctly.

 - OS: [macOs Sequoia]
 - Python version: 3.13.2
 - ADK version: 1.7.0

**Additional context**
I guess the latest package is not published with the extras

kavinkumar807:
This PR fixes #2071 exception string from `pip install google-adk[eval]` to `pip install "google-adk[eval]"` which makes it compatible for all the bash, zsh and other terminals


hangfei:
Thanks for the fix.

hangfei:
Merged. thanks for the fix!

PR #2151, opened 2025-07-24
Let runner create session if no session exists with that id

uneetkumarsingh:
The [Documentation](https://google.github.io/adk-docs/sessions/session/#the-session-lifecycle) mentions that: 
`Start or Resume: Your application's Runner uses the SessionService to either create_session (for a new chat) or get_session (to retrieve an existing one).`

However as of now the session has to be created manually before running the runner. This PR makes sure that runner creates a session of not already existing.


ankursharmas:
@Jacksunwei Please help review this PR

Jacksunwei:
It's still the application's responsibility to create a new session before run.  I think the document meant `adk web`'s behavior, it's still recommend the callsite the create a new session before running the agent. 

---

I'll circle back to update the doc.

PR #2155, opened 2025-07-24
fix: construct cloud run url for agent card when using: adk deploy cloud_run --a2a

andrewlarimer:
Currently, using **adk deploy cloud_run --a2a** invokes the deployment with **adk api_server**, however the URL on the agent card (even if the correct URL is provided) is overwritten by **https://0.0.0.0:port/a2a/agent_name**.

In order for another agent, such as an ADK RemoteA2aAgent to use the Agent Card and access the agent, the URL needs to be a correct Cloud Run service URL.

This fix sets a new IS_DEPLOY_TO_CLOUD_RUN env variable in the Dockerfile, and updates the api_server such that if that env variable is present, it constructs the Cloud Run service URL appropriately to update the url on the Agent Card.

This has been tested by modifying the Dockerfile to deploy to cloud run with this env variable and copy the modified fast_api.py script into the Cloud Run container. The hosted agent card then displays the correct URL, and an ADK RemoteA2aAgent can access the remote agent.

Feel free to ping me for test steps in a lab environment, if that is helpful.

Screenshot 2025-07-24 at 1 28 17 PM
Screenshot 2025-07-24 at 1 27 44 PM

### DIFF
--- a/src/google/adk/cli/cli_deploy.py
+++ b/src/google/adk/cli/cli_deploy.py
@@ -239,7 +239,7 @@ def to_cloud_run(
             log_level.lower() if log_level else verbosity,
             '--labels',
             'created-by=adk',
-            '-set-env-vars',
+            '--set-env-vars',
             'GOOGLE_CLOUD_RUN_K_SERVICE=1'
         ],
         check=True,
--- a/src/google/adk/cli/cli_deploy.py
+++ b/src/google/adk/cli/cli_deploy.py
@@ -39,6 +39,7 @@ ENV PATH="/home/myuser/.local/bin:$PATH"
 ENV GOOGLE_GENAI_USE_VERTEXAI=1
 ENV GOOGLE_CLOUD_PROJECT={gcp_project_id}
 ENV GOOGLE_CLOUD_LOCATION={gcp_region}
+ENV IS_DEPLOY_TO_CLOUD_RUN=1
 
 # Set up environment variables - End
 
--- a/src/google/adk/cli/cli_deploy.py
+++ b/src/google/adk/cli/cli_deploy.py
@@ -39,7 +39,6 @@ ENV PATH="/home/myuser/.local/bin:$PATH"
 ENV GOOGLE_GENAI_USE_VERTEXAI=1
 ENV GOOGLE_CLOUD_PROJECT={gcp_project_id}
 ENV GOOGLE_CLOUD_LOCATION={gcp_region}
-ENV IS_DEPLOY_TO_CLOUD_RUN=1
 
 # Set up environment variables - End
 
@@ -240,6 +239,8 @@ def to_cloud_run(
             log_level.lower() if log_level else verbosity,
             '--labels',
             'created-by=adk',
+            '-set-env-vars',
+            'GOOGLE_CLOUD_RUN_K_SERVICE=1'
         ],
         check=True,
     )
--- a/src/google/adk/cli/cli_deploy.py
+++ b/src/google/adk/cli/cli_deploy.py
@@ -223,26 +223,26 @@ def to_cloud_run(
     region_options = ['--region', region] if region else []
     project = _resolve_project(project)
     subprocess.run(
-        [
-            'gcloud',
-            'run',
-            'deploy',
-            service_name,
-            '--source',
-            temp_folder,
-            '--project',
-            project,
-            *region_options,
-            '--port',
-            str(port),
-            '--verbosity',
-            log_level.lower() if log_level else verbosity,
-            '--labels',
-            'created-by=adk',
-            '--set-env-vars',
-            'GOOGLE_CLOUD_RUN_K_SERVICE=1'
-        ],
-        check=True,
+      [
+        'gcloud',
+        'run',
+        'deploy',
+        service_name,
+        '--source',
+        temp_folder,
+        '--project',
+        project,
+        *region_options,
+        '--port',
+        str(port),
+        '--verbosity',
+        log_level.lower() if log_level else verbosity,
+        '--labels',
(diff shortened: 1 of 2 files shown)


zeroasterisk:
I like this solution (thanks Andrew!) but I don't really like the `IS_DEPLOY_TO_CLOUD_RUN` configuration being added to `cli_deploy.py`.

We talked off-thread about `K_SERVICE` being set by [cloud run](https://cloud.google.com/run/docs/container-contract#services-env-vars) but also by [other kubernetes deployments](https://github.com/knative/specs/blob/main/specs/serving/runtime-contract.md), and we need to assemble the URL from Cloud Run but not make too many assumptions about other environments.

Perhaps better isolating the effect of this change to just the `to_cloud_run()` function and something like `GOOGLE_CLOUD_RUN_K_SERVICE`.

ankursharmas:
@DeanChensj Please help review this PR

src/google/adk/cli/cli_deploy.py
```diff
@@ -239,6 +239,8 @@ def to_cloud_run(
             log_level.lower() if log_level else verbosity,
             '--labels',
             'created-by=adk',
+            '--set-env-vars',
+            'GOOGLE_CLOUD_RUN_K_SERVICE=1'
```
zeroasterisk: ADK core maintainers, is this the right config you want to key off?

src/google/adk/cli/fast_api.py
```diff
@@ -1102,7 +1103,15 @@ async def _get_a2a_runner_async() -> Runner:
         logger.info("Setting up A2A agent: %s", app_name)
 
         try:
-          a2a_rpc_path = f"http://{host}:{port}/a2a/{app_name}"
+          if "GOOGLE_CLOUD_RUN_K_SERVICE" in os.environ:
```
zeroasterisk, line 321: if <running on cloud run> derive the host URL
else use the default host:port

zeroasterisk (approved):
LGTM and we need this to land soon... but I'd like an ADK core maintainer to opine on the new env variable.

seanzhou1023:
actually we don't need to do something specific to cloud run. 

we need to support 0.0.0.0 as host name when we start api server anyway.
e.g. 

```
adk api_server --a2a --host 0.0.0.0 --port 8001 agent_folders
```
this is a valid command that we should support.

 a2a sdk is working on some feature so that agent card can be dynamically built from request. in the case of 0.0.0.0 the host name in rpc url should be whatever from the request. 

I think this PR is kind of workaround or hacky.  Will let @Jacksunwei who developed cloud run deployment feature to review.

src/google/adk/cli/fast_api.py
```diff
@@ -317,7 +318,15 @@ async def _get_a2a_runner_async() -> Runner:
...
```
seanzhou1023, line 321: we should not put cloud run specific logic into fast api server

src/google/adk/cli/fast_api.py
seanzhou1023: also such logic should not be a2a specific, features like allowed origin also needs to allowlist the actual host name. we should not create discrepancy here.

andrewlarimer:
> actually we don't need to do something specific to cloud run.

Right now when using `adk deploy cloud_run --a2a`, the api_server overwrites the URL in the agent card with a host of 0.0.0.0. So while it may not be specific to Cloud Run, we do need to prevent that overwrite.

src/google/adk/cli/fast_api.py
andrewlarimer: We could construct the URL in the to_cloud_run function and pass that as the env variable for fast api server to look for and use if it's present.

Jacksunwei:
I think this is a needed feature, but the approach is too specific to cloud run. When deploying an agent with a2a, I'd expect user needs to publish with a different url other than the cloud_run k service in real case. 

---

I think we can decouple the flag with cloud run. 

Something like: `adk api_server --a2a --a2a_url=https://my-fancy-agent.acme.com/`. 

---

As for deploying to cloud run, we should allow user to specify the url, `adk deploy cloud_run ... --a2a --a2a_url=xxxx`. If `--a2a_url=xxxx` is given, we use it. If `--a2a_url=xxxx` is not given, we could set the k service url.

andrewlarimer:
> As for deploying to cloud run, we should allow user to specify the url, `adk deploy cloud_run ... --a2a --a2a_url=xxxx`. If `--a2a_url=xxxx` is given, we use it. If `--a2a_url=xxxx` is not given, we could set the k service url.

That makes sense to me. Would you like me to implement that @Jacksunwei, or is that something you'd prefer to implement?

Jacksunwei:
Yes, please. We love your contribution 😀

seanzhou1023:
> I think this is a needed feature, but the approach is too specific to cloud run. When deploying an agent with a2a, I'd expect user needs to publish with a different url other than the cloud_run k service in real case.
> 
> I think we can decouple the flag with cloud run.
> 
> Something like: `adk api_server --a2a --a2a_url=https://my-fancy-agent.acme.com/`.
> 
> As for deploying to cloud run, we should allow user to specify the url, `adk deploy cloud_run ... --a2a --a2a_url=xxxx`. If `--a2a_url=xxxx` is given, we use it. If `--a2a_url=xxxx` is not given, we could set the k service url.

How does `--a2a_url` work with 
```
adk api_server --a2a --host 0.0.0.0 ? 
```

andrewlarimer:
Discussed further with Sean and Wei. The request became to dynamically generate the URL using a function instead of a pre-supplied agent card, which sounds like it needs to be implemented by the A2A team.

PR #2167, opened 2025-07-24
feat: Make MAX_DOWNLOADED_QUERY_RESULT_ROWS configurable

lupuletic:
- Add max_downloaded_rows field to BigQueryToolConfig with default value 50
- Add optional max_rows parameter to execute_sql function
- Parameter takes precedence over config when both provided
- Maintain backward compatibility with existing code
- Add comprehensive tests for config and parameter functionality
- Update truncation logic to use resolved max_downloaded_rows value

🤖 Generated with [Claude Code](https://claude.ai/code)


src/google/adk/tools/bigquery/query_tool.py
```diff
@@ -175,9 +175,10 @@ def execute_sql(
       rows.append(row_values)
 
     result = {"status": "SUCCESS", "rows": rows}
+    max_downloaded_rows = max_rows or (config.max_downloaded_rows if config else 50)
```
ankursharmas: can you please create a private constant for "50", something like:

_DEFAULT_MAX_DOWNLOADED_QUERY_RESULT_ROWS = 50

That way, we ensure that we have to update the default value (if need be) only at one place and it has a name.

Also, instead of duplicating the logic, please evaluate the value into a single var and reuse the var.

src/google/adk/tools/bigquery/query_tool.py
```diff
@@ -175,9 +175,10 @@ def execute_sql(
...
     if (
-        MAX_DOWNLOADED_QUERY_RESULT_ROWS is not None
-        and len(rows) == MAX_DOWNLOADED_QUERY_RESULT_ROWS
+        max_downloaded_rows is not None
```
ankursharmas: This appears to be a redundant, as we ensure that max_downloaded_rows cannot be None.

ankursharmas:
The changes proposed in this PR are already made. Closing this PR as it is no longer needed.

Issue #2175, opened 2025-07-25
Live Mode: User text messages aren't saved to session history


ac-machache:
### Issue: Text input not saved in session history during `run_live`

When I use an agent in `run_live` mode, I've found a critical bug: if the user sends a plain text message, it's sent to the model, but it never gets saved to the session history. The model gets the message for the current turn, but the conversation context is immediately lost for all future turns.

This only happens with text. Audio input works fine because the Gemini API sends back a transcription event, which the ADK correctly saves. The API doesn't "echo" back plain text, so the ADK was never creating a user event for it.

---

### The Workaround

The solution is to make sure the ADK creates a user event for text messages before they are sent to the model. After a lot of debugging, we found the right way to do this without breaking the asyncio loop or causing other bugs.

The workaround involves two changes to `src/google/adk/flows/llm_flows/base_llm_flow.py`:

---

#### 1. Update `_send_to_model` to be an async generator that yields a user event for text messages

This is the core of the workaround. It creates the event at the source:

```python
if (
    live_request.content
    and live_request.content.parts
    and live_request.content.parts[0].text
):
    user_event = Event(
        invocation_id=invocation_context.invocation_id,
        author='user',
        content=live_request.content,
    )
    yield user_event  # <-- This creates the event we were missing
    await llm_connection.send_content(live_request.content)
```

#### 2. Update the `run_live` method to handle the event generator properly

To avoid a `TypeError`, use a background handler that consumes the generator and puts the events onto an `asyncio.Queue`. The main `run_live` loop then pulls events from that queue.

```python
# In run_live:
async def send_handler():
    """Consumes the _send_to_model generator and puts events on the queue."""
    async for event in self._send_to_model(llm_connection, invocation_context):
        await event_queue.put(event)

# ... (similar handler for _receive_from_model)

send_task = asyncio.create_task(send_handler())
receive_task = asyncio.create_task(receive_handler())

# The main loop is now simple and just pulls from the queue
while True:
    event = await event_queue.get()
    if event is None:
        break
    yield event
```
### Full Updated Code

```python
async def run_live(
    self,
    invocation_context: InvocationContext,
) -> AsyncGenerator[Event, None]:
    """Runs the flow using live api."""
    llm_request = LlmRequest()
    event_id = Event.new_id()

    async for event in self._preprocess_async(invocation_context, llm_request):
        yield event
    if invocation_context.end_invocation:
        return

    llm = self.__get_llm(invocation_context)
[...]

PR #2201, opened 2025-07-27
feat: add support for context window compression config to RunConfig  ---

ac-machache:
### Description

This PR adds support for the **Context Window Compression** configuration to `RunConfig`.  
It enables users to configure and activate context window compression during live streaming, allowing the model to manage its context length by sliding or trimming once a token threshold is reached.

Key changes:
- Added `ContextWindowCompressionConfig` support with `trigger_tokens` and `SlidingWindow` (`target_tokens`).
- Integrated the config into the live run path.
- Added unit test to verify correct integration and propagation.

---

### Testing Plan

- Added unit test: `test_streaming_with_context_window_compression_config` in `test_live_streaming_configs.py`.

### DIFF
--- a/src/google/adk/agents/run_config.py
+++ b/src/google/adk/agents/run_config.py
@@ -82,6 +82,11 @@ class RunConfig(BaseModel):
   session_resumption: Optional[types.SessionResumptionConfig] = None
   """Configures session resumption mechanism. Only support transparent session resumption mode now."""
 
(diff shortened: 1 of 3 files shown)


ac-machache:
@hangfei, hello i added the support for the context window compression, and can you pls check #2175 and tell me if i should add PR for that issue fix too or wait.

PR #2248, opened 2025-07-30
docs: fix import statement in sample code

jnuvola:
Use correct [tool name](https://github.com/google/adk-python/blob/main/src/google/adk/tools/load_artifacts_tool.py#L113) for the import statement in `generate_image` sample agent.


wuliang229:
Thanks for the contribution. This is no longer needed.

PR #2253, opened 2025-07-30
feat: Add fallback_to_parent functionality to LlmAgent

takashikik:
## Summary

This PR implements the `fallback_to_parent` feature for `LlmAgent`, which allows agents to automatically transfer control back to their parent agent after completing their execution. This provides a declarative way to ensure that child agents reliably return to the parent's context after completing their tasks.

## Motivation

Currently, the Agent has a `disallow_transfer_to_parent` setting to prevent transferring control to the parent agent. However, there are scenarios where the opposite behavior is desired: forcing the agent to always transfer control back to its parent. Without this feature, it's difficult to ensure that a child agent reliably returns to the parent's context after completing its task.

### Core Implementation
- **LlmAgentConfig**: Added `fallback_to_parent: Optional[bool] = None` property to support YAML configuration
- **LlmAgent**: Added `fallback_to_parent: bool = False` field with comprehensive documentation explaining the fallback conditions
- **BaseLlmFlow**: Implemented `_handle_fallback_to_parent_async` method and integrated fallback logic in both `_postprocess_async` and `_postprocess_live` methods

### Fallback Behavior Logic
The fallback behavior is only activated when no model transfer occurs. Fallback to parent only happens when ALL of the following conditions are met:
- The agent is an `LlmAgent` instance
- `fallback_to_parent=True`
- `parent_agent` exists
- The model response contains no function calls (indicating no explicit `transfer_to_agent`)

### Test Coverage
Added comprehensive test suite in `tests/unittests/flows/llm_flows/test_enforce_transfer.py` with 12 test cases covering:
- Basic fallback functionality
- No fallback when model already has transfer calls
- No fallback when `fallback_to_parent=False`
- No fallback when no parent agent exists
- No fallback for non-LlmAgent instances
- Multi-level hierarchy fallback cascading
- Live mode functionality (implementation verified)

## Implementation Details

When `fallback_to_parent=True`, the agent will:
1. Execute normally and generate its response
2. Check if no function calls (including `transfer_to_agent`) are present in the response
3. Automatically create a `transfer_to_agent` function call event to the parent
4. Create a function response event with the transfer action
5. Execute the parent agent with the same invocation context

### Use Cases
This feature is particularly useful for creating hierarchical agent structures where:
- A parent agent delegates specific, self-contained tasks to child agents
- The conversation flow must be guaranteed to return to the parent to continue the main process
- More robust and predictable control over agent routing is required

### Example
```python
task_executor = LlmAgent(
    name="task_executor",
    model="gemini-2.0-flash",
    fallback_to_parent=True,  # Always return to parent after execution
    instruction="Execute the specific task and report results"
)

coordinator = LlmAgent(
    name="coordinator", 
    model="gemini-2.0-flash",
    sub_agents=[task_executor],
    instruction="Coordinate tasks and manage workflow"
)
```

## Breaking Changes
None. This is a purely additive feature with default value `False`, maintaining backward compatibility.

## Testing
- All 3,810 existing unit tests pass without regression
- New test suite with 12 comprehensive test cases
- Live mode functionality tested and verified
- Transfer-related tests (24 tests) all pass
- Built and confirmed that it works as expected, with child agents forwarding data to the parent agent.

screenshot_adk_web


m-breitbach:
I was unsure if I understand the multi-agent behavior of ADK correctly, but the description of that PR seems to confirm my observations. I think a feature like this is absolutely necessary to control the agent flow reliably. Sure, one could just use `AgentTool`, but to me this feels like a hack and also breaks nice traceability via the ADK frontend.

lazharichir:
This feature should indeed be built-in as it's a very common flow to have a parent require work from a sub agent and have the control handed back to the parent once the sub agent's work is done.

_Also noting that using `AgentTool` feels more like a workaround than a solution and the developer/debugging experience isn't as ideal as with sub agents._

hangfei:
Thanks for the PR and discussion.

We are going to support this feature in a different way so it's more flexible. I will close this PR and will share more updates once the new support is ready.

marcopellegrino-devoteam:
@hangfei is there any update?
This is a very critical production-level feature when having a `PlanReActPlanner()` coordinator. Sub-agents must always return to the coordinator. How can this be achieved otherwise?

Issue #2257, opened 2025-07-30
Cannot import HttpAuth and HttpCredentials


vv-varun:
**Describe the bug**
When defining OpenAPIToolset, we cannot import HttpAuth and HttpCredentials because they are not imported in `google/adk/auth/__int.py__`

**To Reproduce**
Steps to reproduce the behavior, define AuthCredential like this:
```
http_auth_credential = AuthCredential(
        auth_type=AuthCredentialTypes.HTTP,
        http=HttpAuth(
            scheme='bearer',
            credentials=HttpCredentials(
                token=""
            )
        )
    )
```

At build time, complier gives error that `HttpAuth` and `HttpCredentials` are of unknown type.

**Expected behavior**
It should be possible to import these modules like : `from google.adk.auth import OAuth2Auth, HttpAuth, HttpCredentials
`

**Desktop:**
 - OS: iOS
 - Python version(python -V): 3.12.6
 - ADK version(pip show google-adk): 1.8.0

seanzhou1023:
thanks for reporting the issue. you can do `from google.adk.auth.auth_credential import OAuth2Auth, HttpAuth, HttpCredentials`

PR #2258, opened 2025-07-30
Fix - Import HttpAuth and HttpCredentials modules

vv-varun:
Fixes: https://github.com/google/adk-python/issues/2257

Just added 2 import statements :)


hangfei:
Close as it's needed based on the original ticket.

Issue #2059, opened 2025-07-20
Persist Mcp-Session-Id Across Streamable HTTP Sessions in MCPToolset


abprime:
** Please make sure you read the contribution guide and file the issues in the right place. **  
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**  
When using `MCPToolset` with `StreamableHTTPConnectionParams` (Streamable HTTP transport), the ADK does not persist the `Mcp-Session-Id` header returned by the MCP server on the first Initialize handshake. As a result, every new tool invocation re-triggers the authentication flow and returns a fresh login URL instead of reusing the existing session.

**To Reproduce**  
1. Install the ADK package:  
  ```bash
  pip install google-adk
  ```
2. Use the following agent definition:
  ```python
  from google.adk.agent import Agent
  from google.adk.tools.mcp_tool.mcp_toolset import MCPToolset, StreamableHTTPConnectionParams
  from your_visualization_module import visualization_agent
  
  root_agent = Agent(
      model="gemini-2.5-flash",
      name="root_agent",
      description="A helpful assistant for user questions.",
      instruction="You are a helpful assistant. Please help the user to the best of your knowledge",
      tools=[
          MCPToolset(
              connection_params=StreamableHTTPConnectionParams(
                  url="http://mcp.fi.money:8080/mcp/stream",
              ),
          ),
      ],
  )
  ```
  # Run a simple query that triggers the first tool call
  response1 = await root_agent.run("fetch bank transactions")
  # Run a second query in the same process
  response2 = await root_agent.run("fetch bank transactions again")
  

3. Observe that on each run, the MCP server returns a login URL instead of reusing the session ID.

Expected behavior
After the MCP server returns a Mcp-Session-Id header on the initial handshake, the ADK should automatically store that session ID and include it on all subsequent Initialize or tool calls over the same transport, avoiding repeated auth prompts.

Screenshots
N/A

Desktop (please complete the following information):
	•	OS: [e.g. Ubuntu 22.04 LTS] - MacOS 15.5 (24F74)
	•	Python version (python -V): Python 3.11.8
	•	ADK version (pip show google-adk): 1.6.1

Model Information:
	•	Model invoked by the agent - "gemini-2.5-flash".

Additional context
[...]

PR #2259, opened 2025-07-30
feat(mcp): add auto_initialize parameter to StreamableHTTPConnectionParams

abprime:
# Fix MCP Session Persistence Across StreamableHTTP Sessions

## Summary

This PR fixes [issue #2059](https://github.com/google/adk-python/issues/2059) by adding an `auto_initialize` parameter to `StreamableHTTPConnectionParams` that controls MCP session initialization and enables session persistence across tool invocations.

## Problem

When using `MCPToolset` with `StreamableHTTPConnectionParams`, the ADK was not persisting the `Mcp-Session-Id` header returned by the MCP server on the first Initialize handshake. This caused every new tool invocation to re-trigger the authentication flow and return a fresh login URL instead of reusing the existing session.

## Solution

The fix introduces an `auto_initialize` parameter to `StreamableHTTPConnectionParams` that:

- **Default value**: `True` (maintains backward compatibility)
- **When `True`**: Automatically initializes the session and re-initializes across tool invocations, if there is a disconnection (`adk web` guaranteed to happen because of agent reloading)
- **When `False`**: Skips automatic initialization, allowing manual control over session lifecycle. Pass custom mcp-session-id (initialised outside the agent context)

This approach is similar to the [LangGraph MCP implementation](https://langchain-ai.github.io/langgraph/reference/mcp/?h=auto_in#langchain_mcp_adapters.client.MultiServerMCPClient.session) and follows the [MCP Specification for Session Management](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management).

### 1. Enhanced StreamableHTTPConnectionParams

- Added `auto_initialize: bool = True` parameter to control session initialization
- Updated docstring to document the new parameter

### 2. Updated Session Initialization Logic

- Modified `MCPSessionManager.create_session()` to check the `auto_initialize` parameter
- Only calls `session.initialize()` when `auto_initialize=True` for StreamableHTTP connections
- Maintains existing behavior for other connection types

## Usage Example

```python
from google.adk.tools.mcp_tool.mcp_toolset import MCPToolset, StreamableHTTPConnectionParams

# Enable session persistence (default behavior)
toolset = MCPToolset(
    connection_params=StreamableHTTPConnectionParams(
        url="http://mcp.fi.money:8080/mcp/stream",
        auto_initialize=True,  # Session will be auto-initialized every tool invocation
    ),
)

# Disable auto-initialization for manual control
toolset = MCPToolset(
    connection_params=StreamableHTTPConnectionParams(
        url="http://mcp.fi.money:8080/mcp/stream",
        headers={
          "mcp-session-id": "<custom-session-id>",
          },
        auto_initialize=False,  # Manual initialization required but session is persistent
    ),
)
```

## Testing

- ✅ All existing tests pass
- ✅ Integration tests for session creation logic

## Backward Compatibility

This change is fully backward compatible:

- `auto_initialize=True` is the default value
- Existing code continues to work without modification
- No breaking changes to the public API

## Related Links

- [Issue #2059](https://github.com/google/adk-python/issues/2059) - Original bug report
- [MCP Specification - Session Management](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management)
- [LangGraph MCP Adapters](https://langchain-ai.github.io/langgraph/reference/mcp/?h=auto_in#langchain_mcp_adapters.client.MultiServerMCPClient.session) - Similar implementation

## Checklist

- [x] Code follows the project's style guidelines
- [x] Self-review of code completed
- [x] Code is documented with appropriate docstrings
- [x] Tests are added and pass
- [x] Changelog is updated
- [x] Demo sample is provided
- [x] Backward compatibility is maintained

### DIFF
--- a/src/google/adk/tools/mcp_tool/mcp_session_manager.py
+++ b/src/google/adk/tools/mcp_tool/mcp_session_manager.py
@@ -15,25 +15,20 @@
 from __future__ import annotations
 
 import asyncio
-from contextlib import AsyncExitStack
-from datetime import timedelta
 import functools
 import hashlib
 import json
 import logging
 import sys
-from typing import Any
-from typing import Dict
-from typing import Optional
-from typing import TextIO
-from typing import Union
+from contextlib import AsyncExitStack
+from datetime import timedelta
+from typing import Any, Dict, Optional, TextIO, Union
 
 import anyio
 from pydantic import BaseModel
 
 try:
-  from mcp import ClientSession
-  from mcp import StdioServerParameters
+  from mcp import ClientSession, StdioServerParameters
   from mcp.client.sse import sse_client
   from mcp.client.stdio import stdio_client
   from mcp.client.streamable_http import streamablehttp_client
@@ -99,6 +94,7 @@ class StreamableHTTPConnectionParams(BaseModel):
         Streamable HTTP server.
       terminate_on_close: Whether to terminate the MCP Streamable HTTP server
         when the connection is closed.
+      auto_initialize: Whether to automatically initialize the session
   """
 
   url: str
@@ -106,6 +102,7 @@ class StreamableHTTPConnectionParams(BaseModel):
   timeout: float = 5.0
   sse_read_timeout: float = 60 * 5.0
   terminate_on_close: bool = True
+  auto_initialize: bool = True
 
 
 def retry_on_closed_resource(func):
@@ -359,7 +356,13 @@ class MCPSessionManager:
           session = await exit_stack.enter_async_context(
               ClientSession(*transports[:2])
           )
-        await session.initialize()
+
+        initialize = True
+        if isinstance(self._connection_params, StreamableHTTPConnectionParams):
+          # Streamable HTTP connections can skip auto-initialization
+          initialize = self._connection_params.auto_initialize
+        if initialize:
+          await session.initialize()
 
         # Store session and exit stack in the pool
         self._sessions[session_key] = (session, exit_stack)


abprime:
@hangfei @seanzhou1023 
please review and suggest changes if any.

src/google/adk/tools/mcp_tool/mcp_session_manager.py
```diff
@@ -359,7 +356,13 @@ async def create_session(
           session = await exit_stack.enter_async_context(
               ClientSession(*transports[:2])
           )
-        await session.initialize()
+
+        initialize = True
```
seanzhou1023, line 360: how does skipping initialize preserve the header ? 
when codes reach here, user should already been authenticated, you won't be able to skip the auth by adding logic here if there were additional auth required.

 client = self._create_client(merged_headers)
 
 above should contain the auth token in merged_headers

seanzhou1023:
@abprime thank you so much for the fix :) Could you explain or add comments how your fix preserve the Mcp-Session-Id  header ?  you mentioned that 
```
This caused every new tool invocation to re-trigger the authentication flow and return a fresh login URL instead of reusing the existing session.
```

However the codes you modified is already after the authentication flow got triggered, how does that prevent triggering the authentication flow ? 

BTW the authentication flow is triggered in BaseAuthenticatedTool ( which MCPTool inherits from ) , by the time below codes are invoked :

```
    # Get the session from the session manager
    session = await self._mcp_session_manager.create_session(headers=headers)
```

the credentials are already there and access tokens are put into the headers , which means the auth flow is already done.

azoom-y-ishii:
Hi @seanzhou1023 — I’m trying to confirm whether the team is aware of (and planning to address) a session-continuity issue I’m observing with ADK + Streamable HTTP.

**Summary**

* With ADK’s `McpSessionManager` and `streamablehttp_client`, the **transport/stream is rebuilt** between user turns.
* When that happens, **`Mcp-Session-Id` is not carried over** to the new transport, so the MCP server treats it as a **new logical session**.
* Within a single user turn (multiple tool calls) things look fine; **across turns**, server-side state tied to the session is lost.

**Observed behavior**

* On EOF in `ClientSession`’s receive loop, ADK cleans up and calls `create_session()` again.
* The new client is created without reusing a previously issued `Mcp-Session-Id` (there doesn’t appear to be a place where ADK stores and re-injects it on subsequent requests).
* Result: the server starts a brand-new session on the next user message.

**Expected (per Streamable HTTP semantics)**

* The client should **keep the logical session stable** across transport restarts by **including the same `Mcp-Session-Id` header on every subsequent request**, and only re-initialize when the server returns “unknown/expired session” (e.g., 404).
* In other words: *recreate the transport, not the session*.

**Why this matters**

* Many MCP servers/tools cache per-session state (dataset IDs, progress tokens, etc.). Losing the session between user messages is functionally breaking in those scenarios.

**What would help**

* Clarification on ADK’s intended behavior here.
* If this is acknowledged as a gap, guidance on the preferred approach:

  * Maintain a session map keyed by logical user/auth headers (excluding `Mcp-Session-Id`),
  * Save the `Mcp-Session-Id` returned on initialize,
  * Re-attach it on every subsequent transport/RPC,
  * Fall back to re-initialize **only** when the server rejects the ID.

If I’ve misunderstood the intended model or if there’s an existing mechanism to access and persist `Mcp-Session-Id` that I should use, please point me to it. Otherwise I’m happy to help test a fix. Thanks!

yoshinoriishii:
Hi @seanzhou1023 — I’m trying to confirm whether the team is aware of (and planning to address) a session-continuity issue I’m observing with ADK + Streamable HTTP.

**Summary**

* With ADK’s `McpSessionManager` and `streamablehttp_client`, the **transport/stream is rebuilt** between user turns.
* When that happens, **`Mcp-Session-Id` is not carried over** to the new transport, so the MCP server treats it as a **new logical session**.
* Within a single user turn (multiple tool calls) things look fine; **across turns**, server-side state tied to the session is lost.

**Observed behavior**

* On EOF in `ClientSession`’s receive loop, ADK cleans up and calls `create_session()` again.
* The new client is created without reusing a previously issued `Mcp-Session-Id` (there doesn’t appear to be a place where ADK stores and re-injects it on subsequent requests).
* Result: the server starts a brand-new session on the next user message.

**Expected (per Streamable HTTP semantics)**

* The client should **keep the logical session stable** across transport restarts by **including the same `Mcp-Session-Id` header on every subsequent request**, and only re-initialize when the server returns “unknown/expired session” (e.g., 404).
* In other words: *recreate the transport, not the session*.

**Why this matters**

* Many MCP servers/tools cache per-session state (dataset IDs, progress tokens, etc.). Losing the session between user messages is functionally breaking in those scenarios.

**What would help**

* Clarification on ADK’s intended behavior here.
* If this is acknowledged as a gap, guidance on the preferred approach:

  * Maintain a session map keyed by logical user/auth headers (excluding `Mcp-Session-Id`),
  * Save the `Mcp-Session-Id` returned on initialize,
  * Re-attach it on every subsequent transport/RPC,
  * Fall back to re-initialize **only** when the server rejects the ID.

If I’ve misunderstood the intended model or if there’s an existing mechanism to access and persist `Mcp-Session-Id` that I should use, please point me to it. Otherwise I’m happy to help test a fix. Thanks!

wukath:
closing since @abprime hasn't responded to sean's comment since august -- feel free to reopen if you have a fix!

@flybae thank you for reporting this issue as well, pls track https://github.com/google/adk-python/issues/2979 for updates

Issue #2273, opened 2025-07-31
Streaming Conversation History Fragmentation Issue


gianluca-henze-parloa:
# Streaming Conversation History Fragmentation Issue

## Problem Description

When using the Gemini Live API with ADK for streaming conversations, the conversation history becomes severely fragmented due to how streaming chunks are stored. Each individual streaming token/chunk is saved as a separate `Content` entry in the conversation history instead of being consolidated into meaningful conversation turns.

### Example of Fragmented History

```python
Content(parts=[Part(text=' Hey')], role='user'),
Content(parts=[Part(text=',')], role='user'), 
Content(parts=[Part(text=' how')], role='user'),
Content(parts=[Part(text=' you')], role='user'),
Content(parts=[Part(text=' do')], role='user'),
Content(parts=[Part(text='ing')], role='user'),
Content(parts=[Part(text='?')], role='user'),
Content(parts=[Part(text="I'm doing well, thank you for asking! How can I help you today?")], role='model'),
Content(parts=[Part(text=' I')], role='user'),
Content(parts=[Part(text=' would')], role='user'),
Content(parts=[Part(text=' like')], role='user'),
# ... and so on
```

### Impact

1. **Token Waste**: Massive token consumption due to fragmented history
2. **Context Degradation**: Important conversation context gets lost as history grows
3. **Performance Issues**: Slower response times due to excessive tokens
4. **Cost Escalation**: Higher API costs due to unnecessary token usage
5. **Model Confusion**: Fragmented context may confuse the model's understanding

hangfei:
Thanks for reporting. I am working on a fix.

hangfei:
Fixed in: https://github.com/google/adk-python/commit/1819ecb4b8c009d02581c2d060fae49cd7fdf653. 

Please retry this in next release or build a dev wheel yourself. Feel free to re-open or create a new issue if it's not fixed.

Issue #2261, opened 2025-07-30
Sequential LlmAgents not returning audio with bidi streaming


ZachTB123:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
I have two separate `LlmAgent`'s who are instructed to only say a specific. After the first agent completes, I want the next `LlmAgent` to say its message. When using either a `SequentialAgent` or custom agent to orchestrate this, I'm not getting an audio response.

**To Reproduce**
1. Clone: https://github.com/google/adk-docs/tree/main/examples/python/snippets/streaming/adk-streaming-ws/app
2. Update `google_search_agent/agent.py` with the following code:
```python
from google.adk.agents import Agent, SequentialAgent

msg1_agent = Agent(
    name="message_agent_1",
    model="gemini-live-2.5-flash",
    description="Agent to say a message to the user.",
    instruction=(
        """
        Your goal is to say a specific message to the user. The message is:

        This is message one.

        Do not say anything else other than the message. You should only
        say the message exactly as it is.
        """
    ),
    include_contents="none"
)

msg2_agent = Agent(
    name="message_agent_2",
    model="gemini-live-2.5-flash",
    description="Agent to say a message to the user.",
    instruction=(
        """
        Your goal is to say a specific message to the user. The message is:

        This is message two.

        Do not say anything else other than the message. You should only
        say the message exactly as it is.
        """
    ),
    include_contents="none"
)

root_agent = SequentialAgent(
    name="MessagesAgent",
    sub_agents=[msg1_agent, msg2_agent],
    description="An agent that plays messages in sequence.",
)
```
4. Use the directions here: https://google.github.io/adk-docs/streaming/custom-streaming-ws/#3.-interact-with-your-streaming-app to run the app.
5. Say "hello".
6. You should see `This is message one.This is message two.` displayed in the page.

**Expected behavior**
I expect to hear those two messages said back to me as audio. Instead, I only see the text representations.

This seems like it is possibly due to the tool call that gets added [here](https://github.com/google/adk-python/blob/v1.5.0/src/google/adk/agents/sequential_agent.py#L65-L72). If I instead using a custom agent like follows:

```python
class CustomAgent(BaseAgent):

    def __init__(
        self,
        name: str,
    ):
        super().__init__(
            name=name,
            sub_agents=[msg1_agent, msg2_agent],
        )

    @override
    async def _run_live_impl(
[...]

Issue #2162, opened 2025-07-24
When using VertexAiSessionService, streaming audio applications become unusable due to excessive HTTP traffic.


frankfarzan:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
When using `VertexAiSessionService`, streaming audio applications become unusable due to excessive HTTP traffic. During streaming scenarios, every event generated results in an immediate HTTP call to the Vertex AI session service (`reasoningEngines/{id}/sessions/{session_id}:appendEvent`), creating a very chatty communication pattern that introduces significant latency and network overhead.

**To Reproduce**
2. Configure `VertexAiSessionService`:
   ```python
   session_service = VertexAiSessionService("PROJECT_ID", "LOCATION")
   runner = adk.Runner(
       agent=root_agent,
       app_name=app_name,
       session_service=session_service
   )
   ```
3. Start any streaming session with audio modality with `run_live`
4. Observe network traffic showing frequent HTTP POST requests to the session service
5. Experience degraded performance

**Expected behavior**
Streaming applications should have optimized session management that doesn't create excessive network chatter. The session service should either batch events, buffer them, or use other optimization strategies to minimize HTTP calls during high-frequency streaming scenarios.

In addition, for our use case specifically, persisting audio blobs to session is not useful. We only care about persisting transcribed input/output text. There should be a way to filter audio out as well. But note that even if only persisting text, the performance degradation is a problem.

**Desktop (please complete the following information):**
 - OS: macOS 24.5.0 (darwin) 
 - Python version(python -V): Python 3.13.4
 - ADK version(pip show google-adk): 1.6.1

 **Model Information:**
gemini-live-2.5-flash-preview-native-audio

rdaros:
I just want to corroborate the OP report. My streaming app works perfectly with InMemoryService, if I just switch to VertexAiSessionService without any other modification, the latency of the audio chunks is 1.5s per chunk, unusable. With memory service it is basically realtime.

EDIT: 

1) I tested with both live models and the behavior is the same.
gemini-2.0-flash-live-preview-04-09
gemini-live-2.5-flash-preview-native-audio

2) I tested with ADK versions 1.6.1, 1.7.0 and 1.8.0

3) Python version 3.12.4

DeanChensj:
Hi @hangfei , I think this is something that we should improve in the run_live level as @frankfarzan suggested

[...]

Issue #2045, opened 2025-07-18
Text Messages Don't Create Session Events in Live Streaming Mode


frankfarzan:
## Summary
In live streaming mode, text messages sent via `live_request_queue.send_content()` do not create Session events with `role="user"`, while audio messages sent via `live_request_queue.send_realtime()` do create these events.

## Current Behavior
- **`live_request_queue.send_content()`**: Bypasses Session event creation and goes directly to LLM via `send_content()`
- **`live_request_queue.send_realtime()`**: Creates Session events with `role="user"` through model transcription feature

## Expected Behavior
Both `send_content()` and `send_realtime()` should consistently create Session events with `role="user"` for proper conversation history tracking.

## Root Cause
Text content sent via `send_content()` in live mode bypasses the normal `_append_new_message_to_session()` pathway used in non-live mode. Audio blobs work because the LLM model's transcription feature generates `LlmResponse` objects with `role="user"` content that flow back through the system.

## Impact
- Inconsistent Session history between text and audio inputs
- Missing user messages in Session events when using `send_content()`
- Potential issues with conversation tracking and debugging

## Code Location
- `LiveRequestQueue.send_content()` and `send_realtime()` in `src/google/adk/agents/live_request_queue.py`
- `BaseLlmFlow._send_to_model()` in `src/google/adk/flows/llm_flows/base_llm_flow.py`
- `Runner._append_new_message_to_session()` in `src/google/adk/runners.py`
- `GeminiLlmConnection.receive()` transcription handling in `src/google/adk/models/gemini_llm_connection.py`

NikitaZemtsov:
I assumed that's the way it was intended. After all, it makes no sense to add the audio user to the event, it just clutters the event and takes up extra space. Text messages can be added separately, controlling what you add.

It would be cool to have a flag to delete agent audio data. Since they clog up the database a lot and don't carry any meaningful load. I made a wrapper for myself over DatabaseSessionService and delete agent audio data. It saves a lot of space.

frankfarzan:
@NikitaZemtsov This issue is about sending text content, e.g.:

```python
content = Content(
    role="user", parts=[Part.from_text(text="Hello")]
)
live_request_queue.send_content(content=content)
```

This is not added to the Session. Audio blobs work fine as they are transcribed and added to Session as expected.

ac-machache:
Hi @frankfarzan ,

[...]

PR #2295, opened 2025-08-01
feat: Refactor live SequentialAgent worklow and live event handling for reliability and cleaner session history

ac-machache:
# Architectural Improvements for Live-Running (BIDI) in ADK

## 1. SequentialAgent Transition Refactor #2261

- Removed reliance on the `task_completed` tool injection in `sequential_agent.py`.
- Modified `gemini_llm_connection.py` to listen for the Gemini API’s `generation_complete` signal.
- Yield a dedicated `LlmResponse` with `generation_complete=True` and `partial=True`.
- Updated `_run_live_impl` to break the current sub-agent loop on `generation_complete` events, enabling reliable sequential agent transitions based on deterministic signals.

## 2. Capturing User Text Messages as Events #2175,  #2045

- Re-architected `run_live` in `base_llm_flow.py` to use an `asyncio.Queue` as a unified event bus for both user and model events.
- `_send_to_model` detects user text messages, creates proper Event objects (`author='user'`), and enqueues them.
- Concurrent `send_handler` and `receive_handler` tasks feed events into the queue.
- Main `run_live` loop consumes the queue, yielding user text messages alongside other events for proper session recording.

## 3. Preventing Session Clutter from Partial Events, #2162

- In `gemini_llm_connection.py`, set `partial=True` for `LlmResponse` instances with audio `inline_data` to avoid saving transient audio chunks.
- Implemented text accumulation for user speech #2273:
  - Partial transcriptions yield real-time feedback events.
  - Model turn start signals end of user speech; then a consolidated full transcription event (`partial=False`) is yielded.
- Result: cleaner session history, saving only complete, meaningful events.


ac-machache:
@hangfei,

Can you check here as well? I’ve added a number of fixes related to events in live streaming and improvements to the sequential agents in live streaming, making them more deterministic.

If everything looks good to you, I can proceed with implementing the `goaway` message and the `setup_complete` event as requested in #2103, and also address #2161

polong-lin:
@hangfei Could you please take a look? This can potentially solve multiple issues with the Live API usage raised across users.

gianluca-henze-parloa:
Hi,
Just checking in, has a decision been made on when this might be merged? My conversation history is getting quite cluttered, which I believe could be affecting agent performance. A timeline would be much appreciated. Thanks!

hangfei:
Thanks for the contribution. need to get this in first: https://github.com/google/adk-python/pull/1867 and then we merge this change.

gianluca-henze-parloa:
> Thanks for the contribution. need to get this in first: https://github.com/google/adk-python/pull/1867 and then we merge this change.

Any updates here?

hangfei:
@ac-machache Could you resolve the conflicts?

ac-machache:
@hangfei  
Hello Hangfei,

I’ve updated the code and resolved the previous conflicts, but there are still a few issues to address:

1. **Thinking capabilities**
   - The live model `models/gemini-2.5-flash-native-audio-preview-09-2025` has thinking capabilities enabled.
   - I wasn’t able to disable them — setting `thinking_budget` to `0` or changing its value doesn’t have any effect.
   - The `include_thoughts` flag can’t be set to `false`, and the `thought_signature` is always `None`.
   - For now, when the model produces thoughts, we only see `thought=true` in the session service.

2. **Transcription behavior**
   - All Gemini models return both **audio** and **text**, even when the input is plain text.
   - Their text outputs are treated as **transcriptions**, not as text completions.
   - According to the Google documentation, there should be a `"finished"` signal for transcription to indicate the end, but it’s always `None`.
   - For now, we rely on **control events** to determine turns in conversations.

3. **Session service**
   - Input/output transcriptions are accumulated before being flushed and saved to the session service.
   - No more audio blobs are being saved to the session service.
   - However, the database session service doesn’t yet have `input_transcription` and `output_transcription` fields, so you’ll see `null` events.

4. **Sequential agent behavior**
   - Sequential agents now rely on the `generate_complet` signal, and that part works fine.
   - If the **first agent** makes a tool call, the **second one** doesn’t fire.
   - If the **second agent** is the one performing the tool call, everything works as expected.

5. **Remaining issues**
   - Missing `thought_signature`.
   - Sequential agent chaining (as described above).

Let me know if you want me to focus on fixing a specific part next.

hangfei:
2 and 3 should be fixed by other PRs. Please take another look and resolve merge conflicts. 

For 1, the purpose of task_completed is to ensure that the first agent knows when the task are completed. In bi-di streaming, as it's continuous streaming, it's not clear when it's completed. So that's why we need this function. the user needs to trigger this function by saying something like the task is completed. Without this, we will have another model where when the turn is complete, we transfer to next task. 

For both of the two design options, i'd like to learn more about your use cases to see which make more sense. We can also consider support both if both cases are reasonable.

hangfei:
> @hangfei Hello Hangfei,
> 
> I’ve updated the code and resolved the previous conflicts, but there are still a few issues to address:
> 
> 1. **Thinking capabilities**
>    
>    * The live model `models/gemini-2.5-flash-native-audio-preview-09-2025` has thinking capabilities enabled.
>    * I wasn’t able to disable them — setting `thinking_budget` to `0` or changing its value doesn’t have any effect.
>    * The `include_thoughts` flag can’t be set to `false`, and the `thought_signature` is always `None`.
>    * For now, when the model produces thoughts, we only see `thought=true` in the session service.
> 2. **Transcription behavior**
>    
>    * All Gemini models return both **audio** and **text**, even when the input is plain text.
>    * Their text outputs are treated as **transcriptions**, not as text completions.
>    * According to the Google documentation, there should be a `"finished"` signal for transcription to indicate the end, but it’s always `None`.
>    * For now, we rely on **control events** to determine turns in conversations.
> 3. **Session service**
>    
>    * Input/output transcriptions are accumulated before being flushed and saved to the session service.
>    * No more audio blobs are being saved to the session service.
>    * However, the database session service doesn’t yet have `input_transcription` and `output_transcription` fields, so you’ll see `null` events.
> 4. **Sequential agent behavior**
>    
>    * Sequential agents now rely on the `generate_complet` signal, and that part works fine.
>    * If the **first agent** makes a tool call, the **second one** doesn’t fire.
>    * If the **second agent** is the one performing the tool call, everything works as expected.
> 5. **Remaining issues**
>    
>    * Missing `thought_signature`.
>    * Sequential agent chaining (as described above).
> 
> Let me know if you want me to focus on fixing a specific part next.

1. Does all other model has thinking enabled by default? If not, let's test other models first. 

2. transcription should be fixed as i tested last time, i do see finished signal. Which model are you using? Maybe it's different from one model to the other. 

3. yes. it's not supported yet. i am working on it.

Let's focus on fixing the sequential agent issues first.

hangfei:
Let's focus on solving one problem in one PR so it's easier and faster to merge. Thanks!

hangfei:
Closing due to inactivity. Let's propose a new PR with a single focus.

Issue #2299, opened 2025-08-01
ModuleNotFoundError: No module named 'llama_index'


MIK-RC:
**ModuleNotFoundError: No module named 'llama_index'**
Reopening a ticket for a bug where optional dependencies are flagged as unavailable due to an import issue.
Code:
```
from google.adk.agents import Agent
from google.adk.tools.retrieval.vertex_ai_rag_retrieval import VertexAiRagRetrieval

ask_vertex_retrieval  = VertexAiRagRetrieval(
    name='retrieve_rag_documentation',
    description=(
        'Use this tool to retrieve documentation and reference materials for the question from the RAG corpus,'
    ),
    rag_resources=[
        rag.RagResource(
            rag_corpus=corpus_name
        )
    ],
    similarity_top_k=10,
    vector_distance_threshold=0.6,
)

grounded_agent = Agent(
    name="Grounded_Agent",
    model=GEMINI_MODEL,
    description="Agent to answer questions using rag_retrieval_tool.",
    instruction="I can answer questions using the data provided, always use the relevant tool to answer any question. Don't assume answers and tell where you found the answer.",
    tools=[ask_vertex_retrieval]
)
GROUNDED_APP_NAME="grounded_agent"
session_service = InMemorySessionService()  # Can be replaced easily {https://google.github.io/adk-docs/sessions/session/#sessionservice-implementations}

# Session and Runner
async def setup_session_and_runner(user_id):
    session = await session_service.create_session(app_name=GROUNDED_APP_NAME, user_id=user_id)
    runner = Runner(agent=grounded_agent, app_name=GROUNDED_APP_NAME, session_service=session_service)
    return session, runner

# Agent Interaction
async def call_agent_async(query, user_id):
    content = types.Content(role='user', parts=[types.Part(text=query)])
    session, runner = await setup_session_and_runner(user_id=user_id)
    events = runner.run_async(user_id=user_id, session_id=session.id, new_message=content)

    async for event in events:
        if event.is_final_response():
            final_response = event.content.parts[0].text
            print("Agent Response: ", final_response)
```

Error:
```
ModuleNotFoundError                       Traceback (most recent call last)
[/tmp/ipython-input-161051703.py](https://localhost:8080/#) in <cell line: 0>()
      1 from google.adk.agents import Agent
----> 2 from google.adk.tools.retrieval.vertex_ai_rag_retrieval import VertexAiRagRetrieval
      3 
[...]

PR #2301, opened 2025-08-01
fix: import issue when using base adk retrieval tools

Muhammad-Ibrahim-Khan:
[Ticket] (https://github.com/google/adk-python/issues/2299)
There is a bug where an optional dependency is causing issues when importing base ADK vertex tools. This is meant to fix that import issue.


Jacksunwei:
I think this will not fix that issue. Instead, we need to have a check of llamaindex before importing it in file_retrieval.py.

Muhammad-Ibrahim-Khan:
Here's what I thought.

There are currently 3 files, 2 of them use Llama-Index.
If someone wishes to use the tool supported by Llama-index, they'll have to specifically import the relevant class right? In which case, the custom error will pop up at the init level.

In the other case, if someone wishes to use the base tool, they wouldn't have to worry about the other tools being imported.

Adding a llama-index package search for *every* file would be counter intuitive as opposed to doing it on the base level.

Muhammad-Ibrahim-Khan:
1. For squashing commits, simply do it before the merge option (squash and merge PR).
2. For logs/screenshots of evidence that this fix works. @Jacksunwei, I apologise for tagging you in this comment, but let me know if there is anything else I should add or if you believe it would help here.

Uploading the *Before and After* pictures

### Before
Screenshot 2025-08-06 195650
Screenshot 2025-08-06 195701


### After
Screenshot 2025-08-06 200518

Jacksunwei:
Thanks for the screenshot! Your agent samples does help a lot to illustrate. 

 I see your point. Your agent is importing from one python file, but you don't want to be blocked by other modules in the package. 

---

@seanzhou1023 I think your latest commit should also address this need.

Muhammad-Ibrahim-Khan:
@Jacksunwei , is there anything else I need to do before the PR will be reviewed (and merged)?

I checked the latest change in the main branch, and it only changes how imports are done (migrated from always loaded import to a lazy import using __getattr__). The problem was that this was only done for VertexAIRagRetrieval, whereas this PR aims to do the same for the optional imports (especially the ones that use LLamaIndex).

Muhammad-Ibrahim-Khan:
@Jacksunwei 
@seanzhougoogle , you just merged a PR 10 hrs ago which is the same thing (same issue and code) I raised a PR for 1 week ago?

May I ask what is the purpose of contribution guidelines if this kind of behavior is to be expected?

You may as well add a note that "just DM us your fixes, we don't want you to do anything".

seanzhou1023:
sorry, we didn't mean to ignore your fix. Your contribution is definitely welcome. Actually before we saw your PR, we already discussed about the final solution (the one in my second PR) when I raised the first PR. But somehow the first PR got merged by some other reviewer who wasn't involved in the discussion before I applied the final solution. As you can see our solution is a little bit different from yours. Since this change is quite small and we are quite short of hands for following up all the PRs and issues on the this repo , we felt that it may not worth back and forth comments on your PR. However I just realized that we didn't consider from your perspective too much. If that hurts you,  we do apologize. We do appreciate your passion for contribution :)

Let me know if you have further suggestions :)

Jacksunwei:
Hi @Muhammad-Ibrahim-Khan , sorry for the communication gaps. This issue was also raised from a Google internal channel and you two worked on this independently. When I posted comment (e528749a1c9653e2b457e21e4ae6b0b2da834b2c), Sean already had a partial fix checked in. That's why. I tagged Sean here. 

Please don't get us wrong. We do welcome contribution, but this is just like the race condition we see in all softwares :)

Muhammad-Ibrahim-Khan:
Considering how quick you were to reply, we all know what this is.

I shared from the beginning what this was but this review was stalled, and no my recent commit isn't "different" from what you pushed except for a minor improvement (a string that is returned for when an import fails).

You say that you guys fix all of this internally so just close the contrib guide, it is a waste of time.

It isn't that you're short on hands, it's morals and ethics that no one is willing to go through this crap.

You guys aren't the first to do this nor the last.
Just close the PR.

PR #2320, opened 2025-08-04
fix: Remove thoughts from contents in llm requests

### LINKED ISSUE #843: Exclude model's "thought" responses from the contents of subsequent model calls
Currently they are not filtered out, and they can add substantial amount of content without providing actual value. Ideally, make their exclusion configurable in case it does add value for some agents.

CAROLZXYZXY:
Fix #843


Jacksunwei (approved):

Jacksunwei:
Merged by copabara.

PR #2369, opened 2025-08-05
Fixing issues with adk_app.py

### LINKED ISSUE #2298: Agent Engine Deployment with GCSArtifact via custom adk_app.py not working as expected
When trying to deploy ADK Agents with GCS Artifact via ADK CLI. The given custom adk_app.py is not transferred to the temporary folder and the default app is always created. Causing not able to deploy with custom services. 

On inspection of ADK CLI deploy.py file, it doesn't seem to check whether the custom adk app file exists or not it just creates a default app irrespective with the default settings(No additional services). 

Please fix this bug by adding a condition to check whether custom app file exists.

vinodseshadri:
## Pull Request Description

This pull request addresses a critical issue in the `cli_deploy.py` module -where in the `adk_app` was ignored

### Issues fixed

#2298 

### Changes done

* Added code to check if `adk_app` is set and if it is then copy the py file from agent source folder instead of creating
* Moved the hardcoded `add_app` to a constant so that it can be revisited later easily
* Added test cases to ensure custom file for add_app works
* library built and command line tested

### DIFF
--- a/src/google/adk/cli/cli_deploy.py
+++ b/src/google/adk/cli/cli_deploy.py
@@ -307,6 +307,7 @@ def to_agent_engine(
       values of `GOOGLE_CLOUD_PROJECT` and `GOOGLE_CLOUD_LOCATION` will be
       overridden by `project` and `region` if they are specified.
   """
+  ADK_APP_OBJECT_NAME = 'adk_app'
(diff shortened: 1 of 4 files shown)


vinodseshadri:
Closing to raise the same from another branch

Issue #2407, opened 2025-08-07
Bug: AgentTool uses a hardcoded user_id, preventing user context propagation to sub-agents


akshaypachpute-1998:
### Bug: AgentTool uses a hardcoded user_id, preventing user context propagation to sub-agents
"user_id="tmp_user" hardcoded into the agent tool"

### Description:
The AgentTool in /google/adk/tools/agent_tool.py uses a hardcoded user_id='tmp_user' when creating a new session for the agent it wraps. This happens within the run_async method.

code snippet 
`
...
  @override
  async def run_async(
      self,
      *,
      args: dict[str, Any],
      tool_context: ToolContext,
  ) -> Any:
...
    session = await runner.session_service.create_session(
        app_name=self.agent.name,
        user_id='tmp_user',  # <-- This is hardcoded
        state=tool_context.state.to_dict(),
    )
...
`

**Why is this a problem?**
This hardcoding breaks the chain of user identity. When a parent agent calls a sub-agent via the AgentTool, the original user_id is lost. Any tool or logic inside the sub-agent that needs to perform user-specific actions (e.g., accessing user data from a database, retrieving user-specific memory, checking permissions) will fail or operate on the wrong context because it receives 'tmp_user' instead of the actual user's ID.

**Impact:**
This prevents the creation of robust, multi-agent applications where user context must be maintained across different agents and tools. It limits the utility of AgentTool to only stateless sub-agents that do not require user-specific information.

**Suggested Fix:**
The user_id should be retrieved from the parent context, which is available via the tool_context parameter passed into run_async. The create_session call should be updated to use the dynamic user_id from the parent session.For example, the fix might involve accessing the user ID from the tool_context.

code-snippet
`session = await runner.session_service.create_session(
        app_name=self.agent.name,
        user_id=tool_context._invocation_context.user_id,
        state=tool_context.state.to_dict(),
    )`

**To Reproduce**
To reproduce this bug, we need to set up a two-agent system: a ParentAgent that calls a ChildAgent using the AgentTool. The ChildAgent will have a tool designed to simply return the user_id it receives from its context.

**Expected behavior**
It should return the user_id of the user calling the agent, 

but, in current situation we are getting `tmp_user `

akshaypachpute-1998:
create PR --> https://github.com/google/adk-python/pull/2409
@Jacksunwei

polong-lin:
Thank you for spotting this, @akshaypachpute-1998 and for your PR and explanation of the solution.

[...]

PR #2409, opened 2025-08-07
fix-issue 2407: replaced hard coded value for user_id to the value from the tool context from parent agent. Fixes google/adk-python#2407

akshaypachpute-1998:
Description:

This PR Fixes: #2407 

The AgentTool in /google/adk/tools/agent_tool.py uses a hardcoded user_id='tmp_user' when creating a new session for the agent it wraps. This happens within the run_async method.

code snippet
... @override async def run_async( self, *, args: dict[str, Any], tool_context: ToolContext, ) -> Any: ... session = await runner.session_service.create_session( app_name=self.agent.name, user_id='tmp_user',  # <-- This is hardcoded state=tool_context.state.to_dict(), ) ...

Why is this a problem?
This hardcoding breaks the chain of user identity. When a parent agent calls a sub-agent via the AgentTool, the original user_id is lost. Any tool or logic inside the sub-agent that needs to perform user-specific actions (e.g., accessing user data from a database, retrieving user-specific memory, checking permissions) will fail or operate on the wrong context because it receives 'tmp_user' instead of the actual user's ID.

Impact:
This prevents the creation of robust, multi-agent applications where user context must be maintained across different agents and tools. It limits the utility of AgentTool to only stateless sub-agents that do not require user-specific information.

Suggested Fix:
The user_id should be retrieved from the parent context, which is available via the tool_context parameter passed into run_async. The create_session call should be updated to use the dynamic user_id from the parent session.For example, the fix might involve accessing the user ID from the tool_context.

code-snippet
session = await runner.session_service.create_session( app_name=self.agent.name, user_id=tool_context._invocation_context.user_id, state=tool_context.state.to_dict(), )

To Reproduce
To reproduce this bug, we need to set up a two-agent system: a ParentAgent that calls a ChildAgent using the AgentTool. The ChildAgent will have a tool designed to simply return the user_id it receives from its context.

Expected behavior
It should return the user_id of the user calling the agent,

but, in current situation we are getting tmp_user


akshaypachpute-1998:
#2407 https://github.com/google/adk-python/issues/2407

akshaypachpute-1998:
Conversation:-
image

logs-after the change :- 
image

results :- 

[
    {
        "title": "Tomatina Festival", #record created before change
        "description": "in Bu\u00f1ol, Valencia, Spain",
        "time": "2025-08-20T12:00:00",
        "created_at": "2025-08-07T13:58:58.036026",
        "created_by": "tmp_user"
    },
    {
        "title": "Tomatina Festival", #record created after change with web ui
        "description": "The festival will be held on August 27, 2025, in the town of Bu\u00f1ol, Valencia, Spain.",
        "time": "2025-08-20T12:00:00",
        "created_at": "2025-08-07T14:04:09.024231",
        "created_by": "user"
    },
    {
        "title": "Tomatina festival", #record created after change with api
        "description": "The festival will be held on August 27, 2025, in Bu\u00f1ol, Valencia, Spain.",
        "time": "2025-08-20T12:00:00",
        "created_at": "2025-08-07T14:12:18.411700",
        "created_by": "akshay.pachpute"
    }
]

akshaypachpute-1998:
> Description:
> 
> The AgentTool in /google/adk/tools/agent_tool.py uses a hardcoded user_id='tmp_user' when creating a new session for the agent it wraps. This happens within the run_async method.
> 
> code snippet ... @OverRide async def run_async( self, *, args: dict[str, Any], tool_context: ToolContext, ) -> Any: ... session = await runner.session_service.create_session( app_name=self.agent.name, user_id='tmp_user', # <-- This is hardcoded state=tool_context.state.to_dict(), ) ...
> 
> Why is this a problem? This hardcoding breaks the chain of user identity. When a parent agent calls a sub-agent via the AgentTool, the original user_id is lost. Any tool or logic inside the sub-agent that needs to perform user-specific actions (e.g., accessing user data from a database, retrieving user-specific memory, checking permissions) will fail or operate on the wrong context because it receives 'tmp_user' instead of the actual user's ID.
> 
> Impact: This prevents the creation of robust, multi-agent applications where user context must be maintained across different agents and tools. It limits the utility of AgentTool to only stateless sub-agents that do not require user-specific information.
> 
> Suggested Fix: The user_id should be retrieved from the parent context, which is available via the tool_context parameter passed into run_async. The create_session call should be updated to use the dynamic user_id from the parent session.For example, the fix might involve accessing the user ID from the tool_context.
> 
> code-snippet session = await runner.session_service.create_session( app_name=self.agent.name, user_id=tool_context._invocation_context.user_id, state=tool_context.state.to_dict(), )
> 
> To Reproduce Steps to reproduce the behavior: To reproduce this bug, we need to set up a two-agent system: a ParentAgent that calls a ChildAgent using the AgentTool. The ChildAgent will have a tool designed to simply return the user_id it receives from its context.
> 
> Expected behavior It should return the user_id of the user calling the agent,
> 
> but, in current situation we are getting tmp_user

Closes #2407

Jacksunwei:
Merged in https://github.com/google/adk-python/commit/0b89f1882dccc1acd0ee109832053edecec04850

PR #2427, opened 2025-08-07
feat(plugins): add thread-safe LogCollectorPlugin for event logging

ivanmkc:
This commit introduces the `LogCollectorPlugin`, a plugin that collects execution logs from ADK callbacks in asynchronous environments. Logs are organized by session ID for retrieval.

The plugin uses `asyncio.Lock` to ensure thread-safe logging. Logs are stored in a `defaultdict(list)` indexed by session ID and can be retrieved with the `get_logs_by_session` method.

The plugin implements the following callback hooks to log contextual data:
- `on_user_message`
- `before_run`, `after_run`
- `before_agent`, `after_agent`
- `before_model`, `after_model`, `on_model_error`
- `before_tool`, `after_tool`, `on_tool_error`
- `on_event`

Each log entry includes the callback type and relevant data, such as invocation ID, agent name, tool name, function call ID, arguments, and results or errors.


hangfei:
It's supported now through: https://github.com/google/adk-python/blob/main/src/google/adk/plugins/logging_plugin.py. 

Closing this. If this doesn't work for you, feel free to raise a issue or PR.

PR #2453, opened 2025-08-09
add CLAUDE.md to all directories

ckl-axel:


ckl-axel:
Opened by accident, please ignore.

PR #2471, opened 2025-08-11
chore: Add middleware support agent 

SudharsaneSivamany:
Middleware support agent perform server maintenance activity based on user prompt.


hangfei:
We don't accept contributions for agents in samples if there is no intended feature testing.

Issue #2351, opened 2025-08-05
Support for --add-cloudsql-instances flag in adk deploy cloud_run command


addodelgrossi:
**Is your feature request related to a problem? Please describe.**
I’m unable to specify the Cloud SQL instance when deploying an agent with the `adk deploy cloud_run` command. Currently, there is no way to pass the --add-cloudsql-instances parameter like we do with the gcloud run services update command. This causes issues when trying to deploy an agent that uses `session_service_uri` with a Cloud SQL database.

**Describe the solution you'd like**
I would like the adk deploy cloud_run command to support an option for specifying the Cloud SQL instance (e.g., --add-cloudsql-instances), similar to the existing parameter in gcloud run services update. This would allow proper configuration of database connectivity during deployment.

**Describe alternatives you've considered**
I would like the adk deploy cloud_run command to support an option for specifying the Cloud SQL instance (e.g., --add-cloudsql-instances), similar to the existing parameter in gcloud run services update. This would allow proper configuration of database connectivity during deployment.

**Additional context**
Currently, the only workaround is to manually run gcloud run services update after deploying with ADK, to add the Cloud SQL instance. However, this adds extra manual steps and breaks the fully automated deployment flow.

jackwotherspoon:
Instead of adding individual support for flags, there should be a mechanism for `adk deploy cloud_run` to assume all unknown flags are Cloud Run flags and add them to the deployment.

This would future proof ADK to support all future Cloud Run flags out of the box and not have to add one off support for each flag. 

A much more robust solution.

jackwotherspoon:
Working PR is up to allow passthrough of all flags after `AGENT_PATH` as Cloud Run flags 😄

PR #2544, opened 2025-08-14
feat: passthrough extra args for `adk deploy cloud_run` as Cloud Run args

jackwotherspoon:
The command `adk deploy cloud_run` supports limited `gcloud run deploy` args 😢.

Which makes the command fine for simple deployments... 

It should support all current and future Cloud Run deployment args for the command to be widely adopted.

This can easily be done by passing through all extra args passed to `adk deploy cloud_run` to gcloud...

This PR assumes any extra args/flags passed after `AGENT_PATH` are gcloud flags.

## Example

```sh
# ADK flags
adk deploy cloud_run \
--project=$GOOGLE_CLOUD_PROJECT \
--region=$GOOGLE_CLOUD_LOCATION \
$AGENT_PATH \
# Use the -- separator for gcloud args
-- \
--min-instances=2 \
--no-allow-unauthenticated
```

This gives full Cloud Run feature support to ADK users 🤖 🚀 

## Test Plan

To test you can just build locally or pip install feature branch directly:

```
uv venv
uv pip install git+https://github.com/jackwotherspoon/adk-python.git
```

Deploy to Cloud Run using additional arguments following `AGENT_PATH`, such as `--min-instance=2` or `--description="Cloud Run test"`:

```sh
uv run adk deploy cloud_run \
--project=$GOOGLE_CLOUD_PROJECT \
--region=$GOOGLE_CLOUD_LOCATION \
--with_ui \
$AGENT_PATH \
-- \
--labels=test-label=adk \
--min-instances=2
```

You can click on the Cloud Run service after deployment and check the service yaml, you should see the additional label etc.

image

Fixes https://github.com/google/adk-python/issues/2351


tests/unittests/cli/utils/test_cli_tools_click.py
```diff
@@ -196,6 +196,49 @@ def _boom(*_a: Any, **_k: Any) -> None:  # noqa: D401
...
+    print(f"Exception: {result.exception}")
+    
+  assert result.exit_code == 0
+  assert rec.calls, "cli_deploy.to_cloud_run must be invoked"
+  
+  # Check that extra_gcloud_args were passed correctly
+  called_kwargs = rec.calls[0][1]
+  extra_args = called_kwargs.get("extra_gcloud_args")
+  assert extra_args is not None
+  assert "--labels=test-label=test" in extra_args
```
hangfei, line 238: where is adk-labels? could you assert that?

src/google/adk/cli/cli_deploy.py
```diff
@@ -222,26 +223,48 @@ def to_cloud_run(
...
+    extra_args_without_labels = []
+    
+    if extra_gcloud_args:
+        for arg in extra_gcloud_args:
+            if arg.startswith('--labels='):
+                # Extract user-provided labels
+                user_labels_value = arg[9:]  # Remove '--labels=' prefix
+                user_labels.append(user_labels_value)
+            else:
+                extra_args_without_labels.append(arg)
```
hangfei: what if extra arg conflicts with existing args? raise error(other than the labels)?

hangfei:
Please add testing plan(log or screenshot)

Jacksunwei:
@jackwotherspoon I see there is a merge conflict, could you help resolve?

src/google/adk/cli/cli_tools_click.py
```diff
@@ -858,7 +858,13 @@ def cli_api_server(
   server.run()
 
 
-@deploy.command("cloud_run")
+@deploy.command(
+    "cloud_run",
+    context_settings={
```
Jacksunwei, line 863: This would make it harder to maintain long-term. 

I'd recommend adding a `--gcloud_extra_arg` option (with `--multiple=True`) to host. 

Usage pattern could be `--gcloud_extra_arg="--args1=val1" --gcloud_extra_arg="--args2=val2" `

jackwotherspoon: > This would make it harder to maintain long-term.

@Jacksunwei can you explain why this makes it harder to maintain...?

Seems way easier to maintain than the current approach of adding supported flags one by one.

jackwotherspoon: >Usage pattern could be --gcloud_extra_arg="--args1=val1" --gcloud_extra_arg="--args2=val2" 

From a DevX perspective I find this much worse than my current approach (in my opinion), you are asking folks to learn something new and have to do extra work to convert...

Versus telling folks to "take any of your existing Cloud Run deployments and copy your flags and just append them to your command".

The second option feels a bit more magical and it "just works"

Jacksunwei: I think it makes things confusing. From user's perspective, when they do `adk deploy cloud_run --arg1 val1 --arg2 val2`. 

There is no easy way to tell whether `arg1` is for `adk deploy cloud_run` or for the underlying `gcloud` command. 

User cannot have help text from `adk deploy cloud_run --help`, either. 

It appears to be doing less things, but it will just confuse people and shut our door to provide any additional options in the future, because if gcloud also has an option of the same name, we may accidentally break users' existing setup.

jackwotherspoon: @Jacksunwei I think you may be mis-understanding...

Anything after `AGENT_PATH` is considered a gcloud arg, everything before is ADK flag...
```sh
# ADK flags
adk deploy cloud_run \
--project=$GOOGLE_CLOUD_PROJECT \
--region=$GOOGLE_CLOUD_LOCATION \
$AGENT_PATH \
# NEW gcloud run deploy flags
--min-instances=2 \
--no-allow-unauthenticated
```

> There is no easy way to tell whether `arg1` is for `adk deploy cloud_run` or for the underlying gcloud command.

Yes there is, anything before `AGENT_PATH` is an ADK arg, if I try and add a gcloud flag before `AGENT_PATH` it will error:

```
Try 'adk deploy cloud_run --help' for help.

Error: No such option: --allow-unauthenticated
```

> User cannot have help text from `adk deploy cloud_run --help`, either.

Yes they can, `adk deploy cloud_run --help`, works fine. Only time it would potentially not work is `--help` after the `AGENT_PATH`.

> because if gcloud also has an option of the same name, we may accidentally break users' existing setup.

This is a fair point, but do we really see this being the case...? Is there a plan to add a bunch more flags to `adk deploy cloud_run`? The benefits here seem to be huge.

src/google/adk/cli/cli_deploy.py
jackwotherspoon: Updated with a validation function 👍

tests/unittests/cli/utils/test_cli_tools_click.py
jackwotherspoon: added a separate test to test label merging logic 👍

src/google/adk/cli/cli_tools_click.py
Jacksunwei: Let's add a `--` in between passthrough args and adk deploy args.

The ending goal would be something like below, though we don't yet support multiple agents yet: 

```
adk deploy cloud_run agent1 agent2 agent3 -- --region=us-west1 --min-instances=2
```

Issue #1785, opened 2025-07-04
Windows log symlink creation fails without admin rights; suggest configurable log dir and graceful fallback


flegare:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**

On Windows, running ADK commands that trigger logging (such as adk run ...) fails with OSError: [WinError 1314] due to an attempt to create a symlink for the latest log file in a temp directory. This requires admin privileges on Windows, causing the CLI to crash for non-admin users.

**To Reproduce**
1. Install google-adk on Windows.
2. Run adk run <agent> as a non-admin user. (followed quick start guide)
3. Observe the error:
`OSError: [WinError 1314] A required privilege is not held by the client: ...`

**Expected behavior**

ADK should log to a user-writable directory and not crash if symlink creation is not permitted. Ideally, the log directory should be configurable, and symlink creation should be skipped or handled gracefully on platforms where it is not supported.

**Screenshots**

![Image](https://github.com/user-attachments/assets/1b600440-d3c6-4d86-82ec-efda7308d250)

**Desktop (please complete the following information):**
 - OS: Windows 10/11
 - Python version(python -V): Python 3.13.0
 - ADK version(pip show google-adk): 

Name: google-adk
Version: 1.5.0
Summary: Agent Development Kit
Home-page: https://google.github.io/adk-docs/
Author: 
Author-email: Google LLC <googleapis-packages@google.com>
License: 
Location: -Redacted-
Requires: anyio, authlib, click, fastapi, google-api-python-client, google-cloud-aiplatform, google-cloud-secret-manager, google-cloud-speech, google-cloud-storage, google-genai, graphviz, mcp, opentelemetry-api, opentelemetry-exporter-gcp-trace, opentelemetry-sdk, pydantic, python-dateutil, python-dotenv, PyYAML, requests, sqlalchemy, starlette, typing-extensions, tzlocal, uvicorn, websockets
Required-by: 

 **Model Information:**
N/A (applies to CLI/logging, not model-specific)

**Additional context**

Related to closed issue #6 ([https://github.com/google/adk-python/issues/6](vscode-file://vscode-app/c:/Users/flega/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)) , but the problem persists.

The root cause is in logs.py, which always attempts to create a symlink for the latest log file.

[...]

PR #2582, opened 2025-08-17
fix: graceful fallback for symlink creation failures

lorenzofavaro:
## Description
Fixes issue #1785 where ADK commands crash on Windows due to symlink creation requiring admin privileges.

## Problem
On Windows, running ADK commands (like `adk run`) fails with `OSError: [WinError 1314] A required privilege is not held by the client: ...` because the logging system attempts to create a symlink for the latest log file. Windows requires administrator privileges for symlink creation by default (unless Developer Mode is enabled), causing crashes for non-admin users.

## Root Cause
The issue was in [`logs.py`](src/google/adk/cli/utils/logs.py) where `os.symlink()` was called unconditionally without error handling, causing the entire CLI to crash when symlink creation failed.

## Solution
This PR implements graceful symlink handling. Changes made:
- Extracted symlink creation into separate function with error handling
- Added graceful fallback when symlink creation fails
- Replaced crashes with warnings to keep CLI functional
- Improved user messaging about log file locations

### Before (broken)
```bash
> adk run my_agent
Log setup complete: C:\Users\username\AppData\Local\Temp\agents_log\agent.20250817_215119.log
  File "google\adk\cli\utils\logs.py", line 72, in log_to_tmp_folder
    os.symlink(log_filepath, latest_log_link)
OSError: [WinError 1314] A required privilege is not held by the client
```

### After (fixed)
```bash
> adk run my_agent
Log setup complete: C:\Users\username\AppData\Local\Temp\agents_log\agent.20250817_215319.log
UserWarning: Cannot create symlink for latest log file: [WinError 1314] A required privilege is not held by the client
To access latest log: tail -F C:\Users\username\AppData\Local\Temp\agents_log\agent.20250817_215319.log
Running agent my_agent, type exit to exit.
```

### DIFF
--- a/src/google/adk/cli/utils/logs.py
+++ b/src/google/adk/cli/utils/logs.py
@@ -18,6 +18,7 @@ import logging
 import os
 import tempfile
 import time
+import warnings
 
 LOGGING_FORMAT = (
     '%(asctime)s - %(levelname)s - %(filename)s:%(lineno)d - %(message)s'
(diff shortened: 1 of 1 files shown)


aamanlamba:
Please merge this fix - adk is broken in most common windows environments

charliedoyne:
> Please merge this fix - adk is broken in most common windows environments

Agreed - please merge this!! This issue is a big inconvenience for those using Windows + ADK. Thanks Lorenzo for providing the fix!

PR #2640, opened 2025-08-20
fix: return invocation_id when using tools + output_schema

### LINKED ISSUE #2631: When using output schema with tools, invocation_id is not being set
When calling an agent which has tools and an output schema, the `invocation_id` is not being set on the final model response event. This prevents the agent from running on Agent Engine where `invocation_id` is required on all events.

**To Reproduce**
- Create an Agent with tools and an output schema defined
- Run the agent
- Review the events in ADK web, the event titled `functionResponse:set_model_response` has an `invocation_id`
- The event after this does not have an `invocation_id` set.

**Expected behaviour**
All the events should have an `invocation_id`

**ADK Version**
1.11.0

**Model**
gemini-2.5-pro

**Additional context**
I believe the bug lies [here](https://github.com/google/adk-python/blob/5b999ed6fd23a0fc1da56ccff4c09621f433846b/src/google/adk/flows/llm_flows/_output_schema_processor.py#L81) as only the author is being set on the event and not the invocation_id.

Running on Agent Engine results in this error:
```
"google.genai.errors.ClientError: 400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': 'List of found errors:\t1.Field: event.invocation_id; Message: Required field is not set.\t', 'status': 'INVALID_ARGUMENT', 'details': [{'@type': 'type.googleapis.com/google.rpc.BadRequest', 'fieldViolations': [{'field': 'event.invocation_id', 'description': 'Required field is not set.'}]}]}}
```

dylan-apex:
Fixes #2631

### DIFF
--- a/src/google/adk/flows/llm_flows/_output_schema_processor.py
+++ b/src/google/adk/flows/llm_flows/_output_schema_processor.py
@@ -81,10 +81,12 @@ def create_final_model_response_event(
   final_event = Event(
       author=invocation_context.agent.name,
       invocation_id=invocation_context.invocation_id,
+      branch=invocation_context.branch,
+      content=types.Content(
+          role='model', parts=[types.Part(text=json_response)]
+      ),
   )
-  final_event.content = types.Content(
-      role='model', parts=[types.Part(text=json_response)]
-  )
+
   return final_event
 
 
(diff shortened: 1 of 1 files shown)


Jacksunwei (approved):

dylan-apex:
@Jacksunwei Noting that I added more after you approved the PR

dylan-apex:
Overshadowed by #2648

PR #2735, opened 2025-08-26
returned session id and used it in evals.

ShaharKatz:


DeanChensj:
Hi @ShaharKatz , thanks for the fix, @ankursharmas  could you help review the PR from eval's perspectvie?

DeanChensj (approved):

src/google/adk/evaluation/evaluation_generator.py
```diff
@@ -28,6 +28,7 @@
 from ..memory.in_memory_memory_service import InMemoryMemoryService
 from ..runners import Runner
 from ..sessions.base_session_service import BaseSessionService
+from ..sessions.vertex_ai_session_service import VertexAiSessionService  # TODO: some users won't have this dependency.
```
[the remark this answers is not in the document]
ShaharKatz: I've added a try/catch block on the Vertex service import so that users that don't use the `google-cloud-aiplatform` package would not get the `ImportError`.
Note - since `response_evaluator.py` already has a dependency on this package in the following line: 
```
from vertexai import types as vertexai_types
```
This meas that now (after the latest fix) users running the evaluation would result in the correct message: 
```
Eval module is not installed, please install via `pip install "google-adk[eval]"
```

src/google/adk/cli/cli_eval.py
```diff
@@ -230,7 +230,8 @@ async def run_evals(
         print(f"Running Eval: {eval_set_id}:{eval_name}")
         session_id = f"{EVAL_SESSION_ID_PREFIX}{str(uuid.uuid4())}"
 
-        inference_result = (
+        # TODO: this overrides the seesion_id that was generated locally in the prefix format. 
+        inference_result, session_id = (
```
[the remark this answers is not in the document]
ShaharKatz: The logic is required for the `session_id` to be set from the `cli_eval.py` according to the `EVAL_SESSION_ID_PREFIX` pattern of 
```
f"{EVAL_SESSION_ID_PREFIX}{str(uuid.uuid4())}"
```
The logic defines the session_id and it is used within `EvaluationGenerator._generate_inferences_from_root_agent` method unless the Vertex Session Service is used - in which we must use the `session_id` generated there.

src/google/adk/evaluation/evaluation_generator.py
```diff
@@ -158,14 +160,25 @@ async def _generate_inferences_from_root_agent(
...
+    if isinstance(session_service, VertexAiSessionService):
+      vertex_session = await session_service.create_session(
+          app_name=app_name,
+          user_id=user_id,
+          state=initial_session.state if initial_session else {}
+      )
+      if hasattr(vertex_session, "id"):
+          session_id = vertex_session.id
+      else:
+        raise ValueError("Session ID was not returned by the Vertex AI session service during session creation.")
```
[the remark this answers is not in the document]
ShaharKatz: Fixed.

ShaharKatz:
@DeanChensj thanks for the review!
@ankursharmas I fixed all comments and issues, ready for review

src/google/adk/evaluation/evaluation_generator.py
```diff
@@ -158,14 +163,22 @@ async def _generate_inferences_from_root_agent(
         initial_session.app_name if initial_session else "EvaluationGenerator"
     )
     user_id = initial_session.user_id if initial_session else "test_user_id"
-    session_id = session_id if session_id else str(uuid.uuid4())
 
-    _ = await session_service.create_session(
+    if VertexAiSessionService and isinstance(session_service, VertexAiSessionService):
```
ankursharmas: Doing this breaks the Abstraction and make Eval Service aware about the several implementations that Session Service has.

Lets not do this.

ankursharmas: What we should do here instead is:
1) If the session_id is not passed in then we create a new session id. This keeps this module backward compatible.

2) If session id is passed in, then we should first check if the session exsits or not? We use session_service.get_session(...) method to do that.
    If the session exists: We do nothing
    if the session doesn't exist: We create the session using the passed in session id. Again this keeps the method backwards compatible. of course this doesn't work for VertexAiSessionService, but we handle that problem in LocalEvalService. (Refer to my comments there)

src/google/adk/evaluation/local_eval_service.py
```diff
@@ -359,7 +359,7 @@ async def _perform_inference_sigle_eval_item(
     )
 
     try:
-      inferences = (
+      inferences, session_id = (
```
ankursharmas: Revert this, and please refer to other comments.

ankursharmas: We take advantage of the session_id_supplier.

Right now, session_id_supplier has a very simple implementation that returns a random UUID.

We update signature to a formal interface:

class SessionIdSupplier(ABC):
    async def get_session_id(app_name: str, user_id: str, [initial_state: dict](state: Optional[dict[str, Any]] = None):


class PreCreatedSessionIdSupplier:
   def __init__(session_service: BaseSessionService):
      self.session_service = session_service
  
   @override
   async def get_session_id(app_name: str, user_id: str, [initial_state: dict](state: Optional[dict[str, Any]] = None):
      created_session = await self.session_service.create_session(...)
      return created_session.id


We then make the PreCreatedSessionIdSupplier as the default session id supplier.

ShaharKatz:
@ankursharmas thanks for the review and comments, using PreCreatedSessionIdSupplier would be extensible and keep the abstraction. 
I'll update according to the comments and re-request a review.

ShaharKatz:
@ankursharmas i've implemented your design of the `PreCreatedSessionIdSupplier` and `BaseSessionIdSupplier` and removed any dependency of the evaluation service to specific implementation of the session service. 

A couple of notes: 
1. i've added the implementation of the session supplier service under the `sessions` folder of the package, as it relates to the implementation of the sessions. 
2. To pass all unittests, I have had to implement a small fix in the cli tests for `MockSessionService` under the `test_fast_api.py` here -  `[test_fast_api.py](https://github.com/google/adk-python/pull/2735/files#diff-ced5cec191008b8c3fb6b73b0383e6cad466d10363581bf3cb48ae7c32a93524)`. 
It appears that the `MockSessionService` did not conform to the `BaseSessionService` interface by returning a session as a dictionary in the `create_session` method, and not a `Session` as defined by the interface for all session services. The specific errors happened when the test service tried to access the session_id by expecting a dictionary `session[id]` and not a session object - `session.id`.

ShaharKatz:
Notice that in [this](https://github.com/google/adk-python/pull/2735/commits/f98a5c8a273cb0cd551ece2bdf1a60200ac6654b) commit i had to change old unavailable gemini model 1.5 to up-to-date 2.5 to get the tests to pass. 
@ankursharmas should I add this as a separate PR, as this is outside the scope of the session ID fix?

ankursharmas:
Regarding [this](https://github.com/google/adk-python/pull/2735/commits/f98a5c8a273cb0cd551ece2bdf1a60200ac6654b) commit, please have a separate PR.

ShaharKatz:
@ankursharmas removed the [commit](https://github.com/google/adk-python/pull/2735/commits/f98a5c8a273cb0cd551ece2bdf1a60200ac6654b) you asked to remove. rebased and updated.

rohityan:
Hi @ShaharKatz , Thank you for your contribution through this pull request!  This PR has merge conflicts that require changes from your end. Could you please rebase your branch with the latest main branch to address these? Once this is complete, please let us know so we can proceed with the review.

ShaharKatz:
This was approved and pushed to main in commit https://github.com/google/adk-python/commit/7d4326c3606a7ff2ba3c0fdef08d4f6af52ee71e, closing.

PR #2762, opened 2025-08-27
fix indentation for generation_params assignment

### LINKED ISSUE #2763: Indentation Error Causes generation_params to be None in LiteLLM
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
A critical indentation error in the _get_completion_inputs function causes generation_params to be set to None during the iteration process. The if not generation_params: check is incorrectly placed inside the for loop, which means if any iteration doesn't find a matching key in config_dict, the entire generation_params dictionary gets set to None. This causes subsequent iterations to fail with TypeError: 'NoneType' object does not support item assignment when trying to assign values.

**To Reproduce**

1.     Install Google ADK: pip install google-adk
2.     Initialize a LiteLLM model with a configuration that has some but not all generation parameters
3.     Run a query that triggers LLM generation
4.     See error in the _get_completion_inputs function

Example code that may trigger the issue:
```python
from google.adk.models.lite_llm import LiteLLM

# This might cause the issue if generation_config is not properly handled
model = LiteLLM(model_name="gpt-3.5-turbo")
```

**Expected behavior**
The generation_params should remain a dictionary throughout the iteration process, and only be set to None after all parameters have been processed if no valid parameters were found.

**Screenshots**

 - OS: Ubuntu 24.04
 - Python version(python -V): Python 3.13
 - ADK version(pip show google-adk): 1.12.0

 **Model Information:**

-     The issue is not model-specific but occurs in the LiteLLM configuration parsing logic
-     Affects all models used through the LiteLLM interface

**Additional context**
The problematic code in /usr/local/lib/python3.13/site-packages/google/adk/models/lite_llm.py

```python
generation_params = {}
param_mapping = {
    "max_output_tokens": "max_completion_tokens",
    "stop_sequences": "stop",
}
for key in (
    "temperature",
    "max_output_tokens",
    "top_p",
    "top_k",
    "stop_sequences",
    "presence_penalty",
    "frequency_penalty",
):
    if key in config_dict:
        mapped_key = param_mapping.get(key, key)
        generation_params[mapped_key] = config_dict[key]

    if not generation_params:  # ← WRONG INDENTATION - INSIDE LOOP!
        generation_params = None  # This executes after every iteration

return messages, tools, response_format, generation_params
```

File "/usr/local/lib/python3.13/site-packages/google/adk/models/lite_llm.py", line 590, in _get_completion_inputs
    generation_params[mapped_key] = config_dict[key]
TypeError: 'NoneType' object does not support item assignment

Sabreclaw:
Fixes #2763
generation_params is giving NoneType exception for this error

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -575,12 +575,11 @@ def _get_completion_inputs(
     config_dict = llm_request.config.model_dump(exclude_none=True)
     # Generate LiteLlm parameters here,
     # Following https://docs.litellm.ai/docs/completion/input.
-    generation_params = {}
     param_mapping = {
         "max_output_tokens": "max_completion_tokens",
         "stop_sequences": "stop",
     }
-    for key in (
+    keys_to_extract = (
         "temperature",
         "max_output_tokens",
         "top_p",
@@ -588,13 +587,13 @@ def _get_completion_inputs(
         "stop_sequences",
         "presence_penalty",
         "frequency_penalty",
-    ):
-      if key in config_dict:
-        mapped_key = param_mapping.get(key, key)
-        generation_params[mapped_key] = config_dict[key]
-
-    if not generation_params:
-      generation_params = None
+    )
+    
+    generation_params = {
+        param_mapping.get(key, key): config_dict[key]
+        for key in keys_to_extract
+        if key in config_dict
+    } or None
 
   return messages, tools, response_format, generation_params
 


ankursharmas (changes requested):
Can you please add a unit test case to cover this case.

https://github.com/google/adk-python/blob/main/tests/unittests/models/test_litellm.py

GWeale:
Hi thank you for your contribution! This has been fixed, thank you for creating an issue for this :)

Issue #2776, opened 2025-08-28
Tools without parameters cause AttributeError in LiteLLM integration


ar0it:
Tools without parameters cause an AttributeError in the LiteLLM integration's _function_declaration_to_tool_param() function because it tries to access function_declaration.parameters.required when parameters is None.

The fix is to add a null check: if function_declaration.parameters and function_declaration.parameters.required:

ar0it:
Same bug as issue #2763

ankursharmas:
@GWeale I think the check needs to happen here:
https://github.com/google/adk-python/blob/a503a0c807e50ec9dde7d5095f8e020861d1375d/src/google/adk/models/lite_llm.py#L395

GWeale:
Thanks for finding this issue. The fix is in for next release! If it does not work for you please reopen.

PR #2777, opened 2025-08-28
Fix issue #2776 function_declaration.parameters.required fails if parameters is none in models/lite_llm.py

ar0it:
Fixes Issue  #2776 function_declaration.parameters.required fails if parameters is none.

Add check for function_declaration.parameters.
Add test parameters to test_litellm.py for case where parameters is none.

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -392,7 +392,7 @@ def _function_declaration_to_tool_param(
       },
   }
 
-  if function_declaration.parameters.required:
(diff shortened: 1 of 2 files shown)


ar0it:
Similair to PR #2762, however this also adds a test and checks function_declaration.parameters individually (not via indent into previous scope).

njbrake:
Hi! I'd love for this to get prioritized since it's a quick fix (and it's blocking us on https://github.com/mozilla-ai/any-agent/issues/780 and likely others too)

Thanks 🙏

ar0it:
Fixed by @GWeale in commit 1e23652

PR #2779, opened 2025-08-28
feat(agent-tool): support propagating runner plugins

davidkl97:


src/google/adk/tools/agent_tool.py
```diff
@@ -45,11 +45,13 @@ class AgentTool(BaseTool):
   Attributes:
     agent: The agent to wrap.
     skip_summarization: Whether to skip summarization of the agent output.
+    include_plugins: Whether to include plugins from parent runner context
   """
 
-  def __init__(self, agent: BaseAgent, skip_summarization: bool = False):
+  def __init__(self, agent: BaseAgent, skip_summarization: bool = False, include_plugins: bool = False):
```
Jacksunwei: Let's add *, to make all future parameters keyword-only.

davidkl97: done

PR #2782, opened 2025-08-28
feat(cli): Enhance GKE deployment with editable install and service account options

syangx39:


hangfei:
close due to inactivity.

PR #2805, opened 2025-08-31
fix(oauth): add client id to token exchange

davidkl97:
add client id to token request, adhering to [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3)


davidkl97:
Hey @Jacksunwei , can you review this please?

xuanyang15:
@davidkl97 Thanks for creating this PR! Have you verified it works as expected end to end?

davidkl97:
Hey @xuanyang15 , yes it works as expected once client id is passed

PR #2823, opened 2025-09-02
fix(mcp): Initialize tool_name_prefix in MCPToolse

### LINKED ISSUE #2814: MCPToolset doesn't support inherited tool_name_prefix parameter from BaseToolset
Describe the bug
  The MCPToolset class inherits from BaseToolset but doesn't properly accept or handle the
  tool_name_prefix parameter that is defined in the parent class constructor. This causes a TypeError when
   trying to use the documented functionality.

  To Reproduce
  1. Try to create an MCPToolset with tool_name_prefix parameter:
  from google.adk.tools.mcp_tool.mcp_toolset import MCPToolset
  from google.adk.tools.mcp_tool.mcp_session_manager import SseConnectionParams

  toolset = MCPToolset(
      connection_params=SseConnectionParams(url="http://localhost:8000/sse"),
      tool_name_prefix="my_prefix_"  # This should work based on inheritance
  )
  2. Observe the TypeError

  Expected behavior
  Based on the inheritance from BaseToolset, MCPToolset should:
  - Accept tool_name_prefix parameter in its constructor
  - Properly initialize the parent class with this parameter
  - Apply the prefix to tool names as implemented in BaseToolset.get_tools_with_prefix()

  Actual behavior
  - TypeError: MCPToolset.__init__() got an unexpected keyword argument 'tool_name_prefix'
  - The super().__init__() call in line 124 only passes tool_filter but omits tool_name_prefix

  Root cause analysis
  In MCPToolset.__init__() (around line 124):
  super().__init__(tool_filter=tool_filter)  # Missing tool_name_prefix parameter

  Should be:
  super().__init__(tool_filter=tool_filter, tool_name_prefix=tool_name_prefix)

  Environment
  - google-adk version: 1.12.0
  - Python version: 3.13.3
  - OS: macOS Darwin 24.6.0

  Current workaround
  toolset = MCPToolset(connection_params=...)
  toolset.tool_name_prefix = "my_prefix_"  # Manual assignment after creation

  Impact
  This prevents proper tool naming in multi-agent scenarios where different agents need prefixed tools to
  avoid naming conflicts, which is a common use case in agent orchestration systems.

shsha4:
Description
  This change introduces a tool_name_prefix attribute to McpToolset and McpToolsetConfig. This allows for adding a prefix to the
  names of all tools within the toolset, which can help avoid naming collisions and provide better organization.

  The implementation involves updating the McpToolset's __init__ and from_config methods to handle the new tool_name_prefix and
  adding the corresponding field to McpToolsetConfig.

  Testing Plan
  A new unit test file has been added to ensure the functionality works as expected.

   - `tests/unittests/tools/test_mcp_toolset.py`:
     - The test_mcp_toolset_with_prefix test case verifies that the tool_name_prefix is correctly applied to the tool names
       retrieved from the toolset.
     - All tests were run via pytest and passed.

  Related Issue
   - Closes #2814


alberto-mate:
When applying this pull request, a "Tool not found" error will occur if the MCP does not have the prefix. For improved agent behavior, the prefix should be included to resolve naming conflicts in the declaration. However, when making a request to the MCP, the agent should use the tool's name without the prefix. 

Check this PR as example

https://github.com/google/adk-python/pull/1226

alberto-mate:
Hi, the only change that this PR needs to be merged is the following:
https://github.com/google/adk-python/blob/0dce65c97f2109ca586084cffbca45a0aa0733f1/src/google/adk/tools/mcp_tool/mcp_tool.py#L134

`response = await session.call_tool(self._mcp_tool.name, arguments=args)`

alberto-mate:
Hi @shsha4, can you apply the changes that I suggested above and push them into this PR? I don't have the permissions to do so :)

shsha4:
@alberto-mate 
Apologies for the delayed response - I was caught up with work and just had a chance to review this. Please take a look when you have time.

alberto-mate:
Thanks for the contribution!

src/google/adk/tools/mcp_tool/mcp_toolset.py
```diff
@@ -100,6 +100,7 @@ def __init__(
           StreamableHTTPConnectionParams,
       ],
       tool_filter: Optional[Union[ToolPredicate, List[str]]] = None,
+      tool_name_prefix: str = "",
```
seanzhou1023: should default to None and make it optional

src/google/adk/tools/mcp_tool/mcp_toolset.py
```diff
@@ -239,6 +245,8 @@ class McpToolsetConfig(BaseToolConfig):
 
   tool_filter: Optional[List[str]] = None
 
+  tool_name_prefix: str = ""
```
seanzhou1023: should default to None and make it optional

shsha4:
@seanzhou1023 
Hi , I have made the requested changes

seanzhou1023:
please fix the test failure. follow other mcp test to skip mcp tests when python version is less than 3.10

shsha4:
@seanzhou1023 
Hello, I have made the requested changes. Please review them.

seanzhou1023 (approved):

seanzhou1023:
please run ./autoformat.sh to resolve the format error

alberto-mate:
ping @shsha4

shsha4:
@seanzhou1023 
Addressed the requested formatting changes. Please review.

cc. @alberto-mate 
pong🤣

PR #2826, opened 2025-09-02
GitHub agent example

ThomasCantuti:
Example of Google ADK agent using MCP protocol to connect with GitHub server


hangfei:
Those samples are for feature testing. There are some existing agents like this. If you can't find simliar one, let me know. 

Closing this for now.

ThomasCantuti:
Thank you to review my proposal.
I tried to make a quick search online and I figured out that there aren't any examples like this one.
What I developed is a very easy, but powerful example where people can find in the same code:
- how to use open-weight models (already seen in other examples);
- how to connect the agent to GitHub server through MCP thanks to the PAT;
- how the agent interacts with a tool developed by developers uniquely when it is called. This solved a big real life problem in one my project.

Anyway, it is not a problem if this solution is not going to be merged into the google adk samples, I've just wanted to let people know about this possibility.

PR #2849, opened 2025-09-04
feat: Add persist_memory flag to AgentTool

### LINKED ISSUE #2850: AgentTool do not persist memory
AgentTool do not persist memory. This means that every time a AgentTool is called, it lost the context from the previous call. It becomes very transactional but less helpful when follow up question arise.

ThibaultScalbert:
This PR adds a `persist_memory` flag to the `AgentTool` class, which allows the conversation history between a main agent and a sub-agent to be persisted across multiple calls to the `AgentTool` within the same session.

Fixes #2850

### Testing Plan

I have added a new test case `test_lru_cache` to verify that the LRU cache evicts runners when the cache is full.

I have also updated the existing tests `test_persist_memory` and `test_no_persist_memory` to use a fixture to reduce code duplication.


hangfei:
You could use sub-agent which will persist by default. Closing this.

PR #2870, opened 2025-09-07
feat: add token_endpoint_auth_method support to OAuth2 credentials

sully90:
## Summary

  Add `token_endpoint_auth_method` field to OAuth2Auth class to allow configuring OAuth2 token endpoint authentication methods. This enables users to specify how the client should authenticate with the authorization server's token
  endpoint.

  • Add `token_endpoint_auth_method` field to `OAuth2Auth` with default value `"client_secret_basic"`
  • Update `create_oauth2_session()` to pass the authentication method to `OAuth2Session`
  • Maintain backward compatibility with existing OAuth2 configurations

  ## Unit Tests
  Added unit test coverage with 3 new test methods:

  1. `test_create_oauth2_session_with_token_endpoint_auth_method()` - Tests explicit auth method setting (`client_secret_post`)
  2. `test_create_oauth2_session_with_default_token_endpoint_auth_method()` - Tests default behavior (`client_secret_basic`)
  3. `test_create_oauth2_session_oauth2_scheme_with_token_endpoint_auth_method()` - Tests with OAuth2 scheme using `client_secret_jwt`

  **Test Results:**
  ✅ 16/16 OAuth2 credential utility tests passed
  ✅ 240/240 auth module tests passed (no regressions)
  ✅ Tests cover both GOOGLE_AI and VERTEX variants
  ✅ Pylint score: 9.41/10

  ## Changes Made

  **src/google/adk/auth/auth_credential.py**
  - Added `token_endpoint_auth_method: Optional[str] = "client_secret_basic"` to `OAuth2Auth` class

  **src/google/adk/auth/oauth2_credential_util.py** 
  - Updated `create_oauth2_session()` to pass `token_endpoint_auth_method` parameter to `OAuth2Session`

  **tests/unittests/auth/test_oauth2_credential_util.py**
  - Added 3 comprehensive test methods covering different authentication scenarios

  ## Backward Compatibility

  ✅ **Non-breaking change** - All existing OAuth2 configurations continue to work unchanged with the default `client_secret_basic` authentication method.

  ## Supported Authentication Methods

  - `client_secret_basic` (default) - Client credentials in Authorization header
  - `client_secret_post` - Client credentials in request body
  - `client_secret_jwt` - JWT with client secret
  - `private_key_jwt` - JWT with private key

### DIFF
--- a/src/google/adk/auth/auth_credential.py
+++ b/src/google/adk/auth/auth_credential.py
@@ -80,6 +80,7 @@ class OAuth2Auth(BaseModelWithConfig):
   expires_at: Optional[int] = None
   expires_in: Optional[int] = None
   audience: Optional[str] = None
+  token_endpoint_auth_method: Optional[str] = "client_secret_basic"
 
 
 class ServiceAccountCredential(BaseModelWithConfig):
--- a/src/google/adk/auth/oauth2_credential_util.py
+++ b/src/google/adk/auth/oauth2_credential_util.py
@@ -82,6 +82,7 @@ def create_oauth2_session(
           scope=" ".join(scopes),
           redirect_uri=auth_credential.oauth2.redirect_uri,
           state=auth_credential.oauth2.state,
+          token_endpoint_auth_method=auth_credential.oauth2.token_endpoint_auth_method,
       ),
       token_endpoint,
   )
--- a/tests/unittests/auth/test_oauth2_credential_util.py
+++ b/tests/unittests/auth/test_oauth2_credential_util.py
@@ -122,6 +122,93 @@ class TestOAuth2CredentialUtil:
     assert client is None
     assert token_endpoint is None
 
+  def test_create_oauth2_session_with_token_endpoint_auth_method(self):
+    """Test create_oauth2_session with token_endpoint_auth_method specified."""
+    scheme = OpenIdConnectWithConfig(
+        type_="openIdConnect",
+        openId_connect_url=(
+            "https://example.com/.well-known/openid_configuration"
+        ),
+        authorization_endpoint="https://example.com/auth",
+        token_endpoint="https://example.com/token",
+        scopes=["openid", "profile"],
+    )
+    credential = AuthCredential(
+        auth_type=AuthCredentialTypes.OPEN_ID_CONNECT,
+        oauth2=OAuth2Auth(
+            client_id="test_client_id",
+            client_secret="test_client_secret",
+            redirect_uri="https://example.com/callback",
+            state="test_state",
+            token_endpoint_auth_method="client_secret_post",
+        ),
+    )
+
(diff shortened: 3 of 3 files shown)


sully90:
https://github.com/google/adk-python/issues/2871

sully90:
@seanzhou1023 any chance of a review on this? It's been open almost 2 months

DinoChiesa:
I know this has been open a long time, just a quick comment. 

The suggested methods are: 
- client_secret_basic (default) - Client credentials in Authorization header
- client_secret_post - Client credentials in request body
- client_secret_jwt - JWT with client secret
- private_key_jwt - JWT with private key

Which systems require `client_secret_post`?  That seems like a diversion from RFC 6749; can we not just omit that from formal support in ADK, and provide some sort of sample or guidance for letting people do this "manually" if they really need it? 

re: JWT options, I think these are both JWT bearer grants as per RFC 7523, see [sec 2.1] (https://datatracker.ietf.org/doc/html/rfc7523#section-2.1) and [sec 2.2] (https://datatracker.ietf.org/doc/html/rfc7523#section-2.2).  I think the names you are suggesting attempt to distinguish between the signing type? Why not just call it `jwt_bearer` ?  And accept a `signingKey` parameter and an optional `algorithm`, which determines whether private key or secret key is used?

But, there is a different axis of variance: it is possible to use a JWT:  
- as the client credential in a 2-legged grant (supplanting the client_credentials grant defined in RFC 6749), OR
- as the client credential in a 3-legged (authorization_code) grant.  

So ideally whatever solution is adopted would need to distinguish these cases .
- auth_code with client ID+secret basicauth encoded (RFC 6749), and/or PKCE
- auth_code with JWT for client credentials (RFC 7523 Sec 2.2)  
- client_credentials with client ID+secret basicauth encoded (RFC 6749)
- jwt_bearer with JWT for client credentials (RFC 7523 Sec 2.1)

sully90:
> Which systems require `client_secret_post`? That seems like a diversion from RFC 6749

https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1

Alternatively, the authorization server MAY support including the
   client credentials in the request-body using the following
   parameters:

   client_id
         REQUIRED.  The client identifier issued to the client during
         the registration process described by [Section 2.2](https://datatracker.ietf.org/doc/html/rfc6749#section-2.2).

   client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.


This is the mechanism our OAuth server requires (`client_secret_post`). It's also directly supported by `authlib` (https://docs.authlib.org/en/latest/client/oauth2.html#client-authentication). I added `client_secret_jwt` and `private_key_jwt` as they are also supported by `authlib` so made sense to add support for them all in one PR.

DinoChiesa:
ok.  Maybe there are 5 cases then. It's complicated.

Issue #2839, opened 2025-09-03
AttributeError: 'NoneType' object has no attribute 'required' when using LoopAgent as sub-agent with LiteLLM


supradeep-thinkbeyond:
**Describe the bug**
When a root Agent delegates to a LoopAgent configured as a sub-agent, the application crashes with an AttributeError. The error occurs in `/usr/local/lib/python3.11/site-packages/google/adk/models/lite_llm.py` at line 395 when the `_function_declaration_to_tool_param` function tries to access `parameters.required` on a `None` object. This happens specifically during agent-to-agent delegation when ADK attempts to convert the LoopAgent into a function declaration for LiteLLM.

**To Reproduce**
1. Install `pip install google-adk==1.13.0` (or any recent version after 1.5.0)
2. Create a file with the following code structure:
```python
from google.adk.agents import Agent, LoopAgent
from google.adk.models.lite_llm import LiteLlm

# Create a model
model = LiteLlm(
    model="openai/gpt-4.1-mini",
    api_key="your-api-key"
)

# Create a sub-agent
bidding_agent = Agent(
    name="bidding_agent ",
    model=model,
    instruction="Execute given tasks",
    tools=[]
)

# Create a LoopAgent
auction_loop = LoopAgent(
    name="auction_loop",
    sub_agents=[bidding_agent],
    max_iterations=100
)

# Create root agent with LoopAgent as sub-agent
root_agent = Agent(
    name="root_agent",
    model=model,
    instruction="Coordinate auction bidding",
    sub_agents=[auction_loop]  # This causes the error
)
```
3. Run the agent with any input that triggers delegation to the LoopAgent
4. See error when the root agent attempts to delegate to the auction_loop

**Expected behavior**
The root agent should successfully delegate tasks to the LoopAgent without any errors. The LoopAgent should be properly converted to a tool representation that LiteLLM can use for function calling.

**Full error traceback**:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/google/adk/flows/llm_flows/base_llm_flow.py", line 937, in _run_and_handle_error
    raise model_error
  File "/usr/local/lib/python3.11/site-packages/google/adk/flows/llm_flows/base_llm_flow.py", line 921, in _run_and_handle_error
    async for response in agen:
  File "/usr/local/lib/python3.11/site-packages/google/adk/models/lite_llm.py", line 740, in generate_content_async
    _get_completion_inputs(llm_request)
[...]

PR #2877, opened 2025-09-08
fix: #2839 argument reference error

Lin-Nikaido:
target issue
closes: #2839

LiteLLM has a bug cannot perse None arguments.

# Testing plan
added test code into `tests/unittests/models/test_litellm.py` and it was passed


Lin-Nikaido:
Oops...
This fix already fixed in [this](https://github.com/google/adk-python/commit/1e23652968164c5fdfa5564e966e78799237d94b)

Issue #2883, opened 2025-09-09
`_pickle.UnpicklingError: pickle data was truncated` raised in database session using MySql


Lin-Nikaido:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
When put leage data into event and load it. the `_pickle.UnpicklingError` was occurred.
The root caurse is `DynamicPickleType` mapping `BLOB` as default in case of `MySql`, not `LONGBLOB`. And learge data will be able to cut off tail of data. And raise pickle error.

**To Reproduce**
1. Install '...'
2. Run '....'
3. Open '....'

**Expected behavior**

**Desktop (please complete the following information):**
 - OS: AL2023
 - Python version(python -V): 3.10
 - ADK version(pip show google-adk): 1.13.0

**Additional context**

PR #2884, opened 2025-09-09
fix: #2883 pickle data was truncated error in database session using MySql

Lin-Nikaido:
closes: #2883 
# Fix
When put leage data into event and load it. the _pickle.UnpicklingError was occurred.
The root caurse is `DynamicPickleType` mapping `BLOB` as default in case of MySql, not `LONGBLOB`. And learge data will be able to cut off tail of data. And raise pickle error.

# What todo
Defined `LONFBLOB` as default explicitly.

# Question
Where should we code the test code like this case? I cannot found the test code the DB and table was created expectedly.

### DIFF
--- a/src/google/adk/sessions/database_session_service.py
+++ b/src/google/adk/sessions/database_session_service.py
@@ -112,6 +112,8 @@ class DynamicPickleType(TypeDecorator):
   impl = PickleType
 
   def load_dialect_impl(self, dialect):
+    if dialect.name == "mysql":
+      return dialect.type_descriptor(mysql.LONGBLOB)
     if dialect.name == "spanner+spanner":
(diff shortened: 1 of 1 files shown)


DeanChensj:
Merged in https://github.com/google/adk-python/commit/36c96ec5b356109b7c874c85d8bb24f0bf6c050d

khoirulanam160125-sys:
from google.adk.tools import google_search

root_agent = Agent(
    name="search_assistant",
    model="gemini-2.5-flash", # Or your preferred Gemini model
    instruction="You are a helpful assistant. Answer user questions using Google Search when needed.",
    description="An assistant that can search the web.",
    tools=[google_search]
)

Issue #2416, opened 2025-08-07
Automatically saving memory to memory bank at the end of a session


zyn-google:
**Is your feature request related to a problem? Please describe.**
When using memory bank vertex ai service. The memory is lost when a session ends. We have to explicitly call add_session_to_memory to save it.

**Describe the solution you'd like**
Can we have the ADK to automatically save the memory to memory bank when a session ends?

DeanChensj:
I think that's something we should consider, how do we define end of a session?  If the session is end, can the user continue that session?

ac-machache:
@DeanChensj

I think another potential solution would be to add support for creating memory that is directly controlled by the agent — not just through the `generate_memories` method, which ingests entire sessions.

According to the documentation, the memory bank does support this kind of fine-grained control. However, after reviewing the ADK memory module, I couldn’t find any implementation that allows agents to create or manage individual memories using the `create_memory` endpoint.

It would be helpful to expose or document how agents can use `create_memory` directly, rather than relying solely on `generate_memories`.

Would it make sense to extend or clarify the current memory API to support this?

guillaumeblaquiere:
The end of session is hard (impossible) to define by the framework, it's known from the external environment. 

The solution is to invoke the add_session_to_memory on the memory service. However, when we are using the api_server runner, it's impossible to access to this feature.

I recommend to add an endpoint to trigger externally to add the session to the memory, something like this:
```
 @app.post("/apps/{app_name}/users/{user_id}/sessions/{session_id}/add_to_memory")
    async def add_session_to_memory( app_name: str, user_id: str, session_id: str):
        session = await self.session_service.get_session(
            app_name=app_name,
            user_id=user_id,
            session_id=session_id
        )
        await self.memory_service.add_session_to_memory(session)
```

I will propose a pull request with this if you are OK

hangfei:
After discussion, i think it make sense to add an endpoint to update the memory from session. The endpoint is added in above commit. And then you can call that endpoint from your agentic application. Closing this.

PR #2900, opened 2025-09-10
feat: add endpoint to generate memory from session

guillaumeblaquiere:
In relation with #2416


src/google/adk/cli/adk_web_server.py
```diff
@@ -1017,6 +1017,16 @@ async def delete_artifact(
...
 
+    # Add to memory the current session
+    @app.post("/apps/{app_name}/users/{user_id}/sessions/{session_id}/add_to_memory")
+    async def custom_dummy_endpoint( app_name: str, user_id: str, session_id: str):
+        session = await self.session_service.get_session(
+            app_name=app_name,
+            user_id=user_id,
+            session_id=session_id
+        )
+        await self.memory_service.add_session_to_memory(session)
```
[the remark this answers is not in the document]
guillaumeblaquiere: All items fixed

src/google/adk/cli/adk_web_server.py
```diff
@@ -1017,6 +1017,32 @@ async def delete_artifact(
           filename=artifact_name,
       )
 
+    @app.post(
+        "/apps/{app_name}/users/{user_id}/sessions/{session_id}/add_to_memory",
```
DeanChensj: use `add-to-memory` which is more RESTful.

src/google/adk/cli/adk_web_server.py
```diff
@@ -1017,6 +1017,32 @@ async def delete_artifact(
...
+    ) -> None:
+        """Adds all events from a given session to the memory service.
+
+        Args:
+            app_name: The name of the application.
+            user_id: The ID of the user.
+            session_id: The ID of the session to add to memory.
+
+        Raises:
+            HTTPException: If the session with the given IDs is not found.
```
DeanChensj: nit.  "given ID is"

src/google/adk/cli/adk_web_server.py
guillaumeblaquiere: Done

src/google/adk/cli/adk_web_server.py
DeanChensj: nit. "ID", not "IDs"

DeanChensj (approved):

DeanChensj:
Could you fix the failing tests?

guillaumeblaquiere:
> Could you fix the failing tests?

Should be OK. I also added a test for my new endpoint. I had to change the mock type for the memory service.
I let you review these changes (both recommended by Gemini Code Assist)

And sorry for the delay, I'm CET timezone based (Europe/Paris)

DeanChensj:
Could you fix the style check using autoformat.sh?

guillaumeblaquiere:
> ou fix the style check using autoformat.

Done. Thanks for you patience

src/google/adk/cli/adk_web_server.py
```diff
@@ -1017,6 +1018,33 @@ async def delete_artifact(
           filename=artifact_name,
       )
 
+    @app.post(
+        "/apps/{app_name}/users/{user_id}/sessions/{session_id}/add-to-memory",
```
DeanChensj: After syncing with the team, we decide to move this endpoint to be memory centric, shall we change this to
```
PATCH /apps/{app_name}/users/{user_id}/memory 
``` 
?  
In the Request Body, let's define a `UpdateMemoryRequest` which has session id in it.  It would be more flexible.
Also should we just return None instead?

guillaumeblaquiere: Right. I updated the endpoint, created the request, and updated the test.

The 204 return code has been recommended by Gemini at the beginning of this PR. 

* When I specify a `status_code=204` into the decorator, I have to use a Response as return type. If not, ALL the tests are broken.
* If I remove it (or set 200), the tests are OK with a return None.

Let me know what you prefer.

src/google/adk/cli/adk_web_server.py
```diff
@@ -1017,6 +1025,54 @@ async def delete_artifact(
...
+    )
+    async def patch_memory(
+        app_name: str, user_id: str, memory_request: UpdateMemoryRequest
+    ) -> Response:
+      """Adds all events from a given session to the memory service.
+
+      Args:
+          app_name: The name of the application.
+          user_id: The ID of the user.
+          memory_request: The request containing the session ID to add to memory.
```
DeanChensj: memory_request is a bit confusing, maybe update_memory_request? 

Also let's not talk about what exact fields it has (subject to change), just talk about why we need it.

src/google/adk/cli/adk_web_server.py
```diff
@@ -1017,6 +1025,54 @@ async def delete_artifact(
...
+
+      Raises:
```
DeanChensj, line 1166: Can we merge these in the python doc?

src/google/adk/cli/adk_web_server.py
```diff
@@ -1017,6 +1025,54 @@ async def delete_artifact(
...
+          )
+          HTTPException(
+              status_code=400, detail="Session ID cannot be empty"
+          )
+      """
+      if not self.memory_service:
+        raise HTTPException(
+            status_code=400, detail="Memory service is not configured."
+        )
+      if memory_request is None:
```
DeanChensj: can be merged now

src/google/adk/cli/adk_web_server.py
DeanChensj: We don't expect a response here so I think return None should be fine, then we don't need to set statu_code=204

src/google/adk/cli/adk_web_server.py
guillaumeblaquiere: I didn't catch what you mean.

src/google/adk/cli/adk_web_server.py
guillaumeblaquiere: It is not clear for me. What do you expect?

DeanChensj:
Merged in https://github.com/google/adk-python/commit/25958242db890b4d2aac8612f7f7cfbb561727fa

PR #2904, opened 2025-09-10
feat: Allow for use of default anthropic client via `anthropic_llm`

EItanya:
Currently the `ClaudeLLM`class will fail if vertex creds aren't available, but the code should work with upstream anthropic clients as well with a couple minor changes.


rohityan:
Hi @EItanya , Thank you for this contribution!
Closing this PR here as it belongs to [adk-python-community](https://github.com/google/adk-python-community) repo.
We highly recommend releasing as a standalone package that we will then share through: https://google.github.io/adk-docs/integrations/

Issue #3082, opened 2025-10-03
Missing field when using some MCP servers


nickvander:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
When using the ADK to call an MCP server, I am seeing these errors:

GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[segmentation_classes].items: missing field.
* GenerateContentRequest.tools[0].function_declarations[1].parameters.properties[segmentation_classes].items: missing field.

**To Reproduce**
Please share a minimal code and data to reproduce your problem.

Build a simple agent that calls an MCP server like:

https://github.com/GoogleCloudPlatform/vertex-ai-creative-studio/blob/main/experiments/mcp-genmedia/mcp-genmedia-go/mcp-imagen-go/README.md

The issue is in this part of the code:

https://github.com/google/adk-python/blob/main/src/google/adk/tools/_gemini_schema_util.py#L140-L184

The error "parameters.properties[segmentation_classes].items: missing field" indicates _to_gemini_schema isn't correctly handling array types. 

**Expected behavior**
Tools should work fine

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: [e.g. macOS, Linux, Windows]
 - Python version(python -V):
 - ADK version(pip show google-adk):

 **Model Information:**
 - Are you using LiteLLM: No
 - Which model is being used: gemini-2.5-flash, but happens to any Gemini model

**Additional context**

jimpeebles:
Would love this to get fixed!

nickvander:
> Would love this to get fixed!

Try it out! The fix should now have landed

danistrebel:
Thanks for the great issue description @nickvander. I just ran into the same problem in 1.21.0. 

From what I can see the sanitization and regression tests were removed in https://github.com/google/adk-python/commit/ab00c41fc91c06e6faec56f14daefc1a97d877c6.

I suggest re-opening this.

jimpeebles:
FWIW, the patch seemed to be working well until it was reverted.

nickvander:
I actually don't have the powers to reopen this issue, but I will work on a fix now and get that out for review. Thanks for flagging this! I didn't realize that my fix got promptly reverted

seanzhou1023:
@nickvander found his fixes got somehow reverted , thus reopen it for re-fixing.

PR #3086, opened 2025-10-03
fix: Fix issue with MCP tools throwing an error.

nickvander:
**Problem**

When an MCP tool is resolved using _gemini_schema_util.py, if it has an array it wasn't correctly resolving them and showing an error:

GenerateContentRequest.tools[0].function_declarations[1].parameters.properties[segmentation_classes].items: missing field

**Root Cause**

The error "parameters.properties[segmentation_classes].items: missing field" indicates _to_gemini_schema isn't correctly handling array types.

**Implementation Details**

Correctly handle array types by adding a default items field of {"type": "string"} to schemas with type: array but no items field.

**Associated Issue (fixes)**

https://github.com/google/adk-python/issues/3082

**Testing Plan**

Reran the MCP server that wasn't working, and this now works. This is the MCP server that I used:

https://github.com/GoogleCloudPlatform/vertex-ai-creative-studio/blob/main/experiments/mcp-genmedia/mcp-genmedia-go/mcp-imagen-go/README.md

**Verification**

With this update, the MCP servers now work without issues!


jimpeebles (approved):
LGTM! Lets get this fixed.

PR #3102, opened 2025-10-06
fix: change LlmAgent.tools to covariant Sequence

jokester:
recreated from https://github.com/google/adk-python/pull/1804 which got closed as "stale"

Fixes https://github.com/google/adk-python/issues/1803

(the type mismatch still happens in google-adk v1.15.1)

### DIFF
--- a/src/google/adk/agents/llm_agent.py
+++ b/src/google/adk/agents/llm_agent.py
@@ -25,6 +25,7 @@ from typing import ClassVar
 from typing import Dict
 from typing import Literal
 from typing import Optional
+from typing import Sequence
 from typing import Type
 from typing import Union
 import warnings
@@ -218,7 +219,7 @@ class LlmAgent(BaseAgent):
   ```
   """
 
-  tools: list[ToolUnion] = Field(default_factory=list)
+  tools: Sequence[ToolUnion] = Field(default_factory=list)
   """Tools available to this agent."""
 
   generate_content_config: Optional[types.GenerateContentConfig] = None


rohityan:
Hi @jokester, Thanks for submitting this pull request! 
To move this forward, could you please add a few unit tests to this PR? Having solid test coverage will significantly expedite the review and merging process.

jokester:
@ryanaiagent hi, thanks for checking this. However I'm not sure if adding "unit test" is simple or possible.

The issue in question is about type annotation, not a runtime bug. It would require a type checker to "test" the change, but I don't see this repo has one (pyright or mypy) in CI.

I can confirm the fix works with [the snippet pyright complained about](https://github.com/google/adk-python/issues/1803#issuecomment-3186253227). Do you think this will work too?

Jacksunwei:
@jokester 

Thank you for the PR! However, we're closing this as we cannot accept the change.

While `Sequence[ToolUnion]` would fix the type checking issue, we need to keep `list[ToolUnion]` because:
1. The API is already released and stable (changing it would be a breaking change)
2. The field needs to be mutable - ADK itself modifies `agent.tools` at runtime (e.g., [SequentialAgent appends tools](https://github.com/google/adk-python/blob/main/src/google/adk/agents/sequential_agent.py#L149))

See the explanation and workaround in https://github.com/google/adk-python/issues/1803#issuecomment-3487621904

Closing the associated issue #1803 as won't fix.

jokester:
oh I didn't notice the consequences. Thanks for pointing that out. totally understood.

PR #3127, opened 2025-10-09
fix: Fix error when query job destination is None

na0fu3y:
# What this PR does
This PR fixes a bug in the `execute_sql function` within `bigquery/query_tool.py`. The bug caused an error when executing queries that do not return a result set, such as a `CREATE TEMP FUNCTION` statement.

# The problem
Within `execute_sql,` when a user executed a query that doesn't produce a result set, like the following:

```sql
CREATE TEMP FUNCTION f() AS (0); SELECT f();
```

The corresponding BigQuery job object would have its `destination` property set to `None`. This caused an error because the code was written with the expectation of a valid destination table.

# The solution
The fix modifies `execute_sql` to check if the `query_job.destination` is None. If it is, the function correctly identifies the query as a statement that doesn't require a destination and allows it to complete successfully.

This ensures that `execute_sql` can robustly handle these types of queries.


rohityan:
Hi @na0fu3y Thanks for submitting this PR! 

Before we can proceed with this, we need two things to be added:

Unit Test: Please add a new unit test that fails with the old code and passes with your fix. A test case using a CREATE TEMP FUNCTION or another statement that doesn't return results would be perfect to validate the solution and prevent future regressions.

Linked Issue: Please create a new issue describing the bug, then link it to this PR by adding Fixes #<issue-number> in the PR description. If an issue already exists, you can just link that one.

Let us know if you need any help with these steps. Thanks again!

Issue #3116, opened 2025-10-08
Granular Per-Agent Speech Configuration


qyuo:
**Is your feature request related to a problem?**
In multi-agent conversational applications, distinguishing between different agents is crucial for user experience. Assigning a unique voice to each agent provides clear audio cues, helping users identify which persona is speaking and making the interaction more intuitive.

Currently, the speech configuration is set globally for the entire session. This prevents developers from defining per-agent voices, which is a significant limitation, especially in bidirectional streaming contexts.

**Describe the solution you'd like**
I propose adding an optional `speech_config` parameter to the individual agent's configuration object. This would allow developers to specify unique voice and other speech settings on a per-agent basis.

**Example Implementation:**
```python
llm_agent.LlmAgent(
    model=models.Gemini(
        model="gemini-live-2.5-flash-preview",
        retry_options=types.HttpRetryOptions(initial_delay=1, attempts=2),
    ),
    name="recommendation_agent",
    description="Recommends Fantasy Football Picks",
    instruction="Provide data-driven draft recommendations",
    speech_config=types.SpeechConfig(
        voice_config=types.VoiceConfig(
            prebuilt_voice_config=types.PrebuiltVoiceConfig(
                voice_name="Kore",
            )
        )
    ),
)

```

hangfei:
Supported now.

PR #3170, opened 2025-10-14
Granular Per-Agent Speech Configuration

qyuo:
Addresses Feature Request: #3116 

This PR adds a `speech_config` to the **LLM Agent configuration** for the **live use case**. When an **asynchronous LLM** call is made to the **Gemini Live API**, it prioritizes the most specific agent configuration's speech_config. If that is null, it then uses the run configuration's speech_config. Unit tests have been added to verify this behavior.

### DIFF
--- a/src/google/adk/artifacts/base_artifact_service.py
+++ b/src/google/adk/artifacts/base_artifact_service.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 from abc import ABC
 from abc import abstractmethod
+from datetime import datetime
 from typing import Any
 from typing import Optional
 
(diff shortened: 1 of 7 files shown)


hangfei:
If there are multiple agents with different speech config, will that work?

hangfei:
Could you update this sample to have different configs for each agent: https://github.com/google/adk-python/tree/main/contributing/samples/live_bidi_streaming_multi_agent?

illuisionary:
> If there are multiple agents with different speech config, will that work?

Yes. They will reply with a different voice based on which one provides the final output.

hangfei (approved):

hangfei (approved):

hangfei:
this is merged. thanks for your contribution. cc: @joefernandez to help add a documentation for this change.

kazunori279:
As per a request from @hangfei I've added this feature to the draft of Bidi-streaming dev guide (which I plan to publish next week on the adk-docs streaming section). Please check: https://github.com/kazunori279/adk-streaming-guide/blob/main/docs/part5_audio_and_video.md#voice-configuration-speech-config cc: @joefernandez for vis.

PR #3180, opened 2025-10-15
Fix: Bug when callback_context._invocation_context is missing in Glob…

### LINKED ISSUE #3163: AttributeError when using the GlobalInstructionPlugin with a Runner
### Describe the bug

When using the `GlobalInstructionPlugin` with a `Runner`, an `AttributeError` is raised during the `before_model_callback`. The error message, `'CallbackContext' object has no attribute 'invocation_context'`, indicates that the plugin is trying to access an attribute that does not exist on the context object provided by the runner's execution flow.

***

### To Reproduce

1.  Initialize a `Runner` and include `GlobalInstructionPlugin` in the `plugins` list.
2.  Call the `runner.run_async()` method to start an agent execution.
3.  The following error occurs.

**Code:**

```python
# Simplified reproduction code based on provided snippets
from google.adk.plugins import GlobalInstructionPlugin
# ... other necessary imports for agent, services, etc.

async def get_runner_async():
    """Returns a runner configured with the GlobalInstructionPlugin."""
    # Assume root_agent and services are defined elsewhere
    root_agent = ... 
    artifact_service = ...
    session_service = ...
    memory_service = ...
    
    runner = Runner(
        app_name="my_app",
        agent=root_agent,
        artifact_service=artifact_service,
        memory_service=memory_service,
        plugins=[
            GlobalInstructionPlugin(
                global_instruction="This is a global instruction."
            )
        ]
    )
    return runner

async def main():
    """Main execution function to demonstrate the bug."""
    runner = await get_runner_async()
    
    # The error occurs when run_async is called and the plugin's callback is triggered
        user_id="test_user",
        session_id="test_session",
        new_message="Hello, agent!",
    ):
        print(event)

# To run:
# import asyncio
# asyncio.run(main())

UlookEE:
Fix #3163 

Hello,
Since global_instruction has been deprecated, I’m migrating to GlobalInstructionPlugin.
During the migration, I encountered an error and am submitting this PR to fix it.

In [df05ed6](https://github.com/google/adk-python/commit/df05ed6b3b7b218d85fddc1acd6617802cdf6f2a) ,
GlobalInstructionPlugin references invocation_context, but CallbackContext actually contains _invocation_context.
This mismatch always causes an error during execution.


UlookEE:
The previous test failure was caused by other commits. Please test it again.

GWeale:
Hi thank you for this PR! [df05ed6](https://github.com/google/adk-python/commit/df05ed6b3b7b218d85fddc1acd6617802cdf6f2a) this change has been rolled back, when this change gets added in I will merge this in as I think it is a great change!

UlookEE:
@GWeale 
Hello. Was this change rolled back only internally within the Google team? I'm trying to use it, but the inaccessible variable still remains, so I'm asking.

GWeale (approved):

GWeale:
This has been merged! Thank you for the contribution

PR #3186, opened 2025-10-15
fix: Support list[BaseModel] annotation about FunctionTool args

t-miyak:
## Summary
Add support for `list[pydantic.BaseModel]` type arguments in FunctionTool, enabling automatic conversion of JSON arrays from LLM tool call responses to lists of Pydantic model instances.

## Problem
Currently, when a function tool is defined with `list[PydanticModel]` type annotation, the LLM tool call response provides `list[dict]`, requiring manual conversion in each tool function. This leads to repetitive boilerplate code and makes tool definitions less clean.

### DIFF
--- a/src/google/adk/tools/function_tool.py
+++ b/src/google/adk/tools/function_tool.py
@@ -32,7 +32,7 @@ from ._automatic_function_calling_util import build_function_declaration
 from .base_tool import BaseTool
 from .tool_context import ToolContext
 
-logger = logging.getLogger("google_adk." + __name__)
+logger = logging.getLogger('google_adk.' + __name__)
 
 
 class FunctionTool(BaseTool):
@@ -57,22 +57,22 @@ class FunctionTool(BaseTool):
         the callable returns True, the tool will require confirmation from the
         user.
     """
-    name = ""
-    doc = ""
+    name = ''
+    doc = ''
     # Handle different types of callables
-    if hasattr(func, "__name__"):
+    if hasattr(func, '__name__'):
       # Regular functions, unbound methods, etc.
       name = func.__name__
-    elif hasattr(func, "__class__"):
+    elif hasattr(func, '__class__'):
       # Callable objects, bound methods, etc.
       name = func.__class__.__name__
 
     # Get documentation (prioritize direct __doc__ if available)
-    if hasattr(func, "__doc__") and func.__doc__:
+    if hasattr(func, '__doc__') and func.__doc__:
       doc = inspect.cleandoc(func.__doc__)
     elif (
-        hasattr(func, "__call__")
-        and hasattr(func.__call__, "__doc__")
+        hasattr(func, '__call__')
+        and hasattr(func.__call__, '__doc__')
         and func.__call__.__doc__
     ):
       # For callable objects, try to get docstring from __call__ method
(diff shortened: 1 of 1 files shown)


t-miyak:
I've tested locally and confirmed it works.

```
➜ uv run pytest tests/unittests/tools/test_function_tool_pydantic.py
================================================================ test session starts ================================================================
platform darwin -- Python 3.11.4, pytest-8.4.2, pluggy-1.6.0
configfile: pyproject.toml
plugins: mock-3.15.1, asyncio-1.2.0, anyio-4.11.0, xdist-3.8.0, langsmith-0.4.35
collected 14 items                                                                                                                                  

tests/unittests/tools/test_function_tool_pydantic.py ..............                                                                           [100%]

================================================================ 14 passed in 6.55s =================================================================
```

t-miyak:
@GWeale 

I fixed the CI errors.

Thank you!

GWeale:
Thanks for adding list support! One concern though if any item fails`model_validate`, we skip it and still overwrite the argument with the (now shorter) converted list. That is a behavior change from the singlevalue path, it kept the original value on failure. Can you fix so we would it will not silently drop data?

t-miyak:
@GWeale 

https://github.com/google/adk-python/compare/b139406...fe575f7

Thanks for the review!

I've updated the logic. Now, if an item in the list fails validation, it retains the original data for that specific item rather than skipping it.

I debated whether to revert the entire list to the original or just the failed items. I chose this approach (keeping successful conversions while reverting failed ones) because it provides clearer feedback to the end-user about which specific items failed the conversion, rather than reverting the whole batch.

That said, if you feel reverting the entire list is better, I'm happy to change it.

~~P.S. I ran ./autoformat.sh, and it seems to have introduced some unexpected formatting changes.~~
→ It was due to my local environment settings. I fixed it.

src/google/adk/tools/function_tool.py
```diff
@@ -129,8 +130,39 @@ def _preprocess_args(self, args: dict[str, Any]) -> dict[str, Any]:
...
+                  try:
+                    converted_list.append(element_type.model_validate(item))
+                  except Exception as e:
+                    logger.warning(
+                        f"Failed to convert item in '{param_name}' to Pydantic"
+                        f' model {element_type.__name__}: {e}'
+                    )
+
+                    # Keep the original value if conversion fails
+                    converted_list.append(item)
```
GWeale: When a list element fails Pydantic validation you now append the original dict back into converted_list. The downstream tool function still receives that list, so any comprehensions or attribute access (user.name) will now raise (dict has no attribute name). 

I think  we skipped the bad element before, so the tool could still run on the valid ones. If we want to preserve the original data for feedback, we either need to keep the entire argument unconverted so tool decides or continue skipping/raising

GWeale:
Hi thank you for fixing, I also noticed something on line 160. Could you fix so i can pull in :)

t-miyak:
@GWeale 

Thank you for the review and feedback.

After considering the trade-offs, I've decided to revert the implementation back to skipping invalid items.

My reasoning is that if we pass a list containing the original dict, the downstream tool will almost certainly raise an unexpected error anyway. The tool's logic is written expecting a list of Pydantic objects that match the type annotation (e.g., attribute access like user.name), but it would receive a dict instead.

Furthermore, an item that fails validation is not the data the downstream tool is designed to process, so I believe skipping it is the correct and most robust behavior in this case. This allows the tool to proceed with the remaining valid items without crashing.

I've pushed the change（revert commits）. Please check them.

GWeale:
Hi thank you for this PR, but we are looking to solve this problem another way.

PR #3216, opened 2025-10-18
Public Invocation Context for Global Instructions Plugin

juanfcastropiccolo:
## Summary
This PR addresses the TODO comment in `CallbackContext.__init__` by exposing the `invocation_context` as a public property for Agent Development Kit plugin use.

## Problem
The `GlobalInstructionPlugin` and potentially other ADK plugins need access to the `invocation_context` to create `ReadonlyContext` instances. Currently, only the private `_invocation_context` attribute is available, causing `AttributeError` when plugins attempt to access `callback_context.invocation_context`.

## Solution
Added a public `@property` method `invocation_context()` to the `CallbackContext` class that returns `self._invocation_context`. This follows the same pattern used for other context properties like `state` and `session`.

## Changes
- Added `invocation_context` property to `CallbackContext` class
- Added comprehensive docstring explaining this if for ADK plugin development
- Mantains backward compatibility (no breaking changes)

## Testing
- Verified that `GlobalInstructionPlugin.before_model_callback` now works correctly
- No linting errors
- Follows existing code style and patterns in the codebase

## Related
Resolves the TODO: "make this public for Agent Development Kit, but private for users."

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -206,7 +206,7 @@ asyncio_mode = "auto"
 python_version = "3.9"
 exclude = "tests/"
 plugins = ["pydantic.mypy"]
-# Start with non-strict mode, and swtich to strict mode later.
+# Start with non-strict mode, and switch to strict mode later.
 # strict = true
 disable_error_code = ["import-not-found", "import-untyped", "unused-ignore"]
 follow_imports = "skip"


juanfcastropiccolo:
hi @hangfei @polong-lin , could you please review this? :) thanks in advance!

rohityan:
Hi @juanfcastropiccolo , Thank you for your contribution! 
Thank you for your contribution! We need additional information to start the review. Can you please provide unit tests.

juanfcastropiccolo:
Hello @hangfei , I fixed the CI issue. Could you please try the merge again? Thanks!

Jacksunwei:
Hi @juanfcastropiccolo, thank you for the contribution, but I need to decline this PR.

The TODO comment was about `_event_actions` (line 50), not `_invocation_context`. `_event_actions` is an internal data structure used by the ADK runtime to track event metadata (state changes, transfers, etc.). It shouldn't be exposed in the public `CallbackContext` API because users manipulating it directly would break event handling.

The TODO was asking for something Python can't provide—there's no way to make an attribute "public for framework code but private for users." The underscore prefix already signals internal use, and ADK plugins can access `_event_actions` when needed.

I'd suggest just ignoring that TODO—the current design is working as intended.

Thanks again for looking into this!

Issue #3075, opened 2025-10-02
Local Safe Code Executor with docker and output file capturing support


AlexFierro9:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Is your feature request related to a problem? Please describe.**
As someone using ADK in corporate environments, I am surprised to see the UnSafeLocalCodeExecutor is only a exec tool, I think it is possible to make it more robust for a lot of usecases.

**Describe the solution you'd like**
I'd like to the current implementation more robust so that gives us a secure and isolated space to run our code. Here are my 2 ideas for how making it work:

Better Process Handling:

Running in a Sub-process: It could run code in a separate child process. That way, it's isolated from the main ADK tool process, and we can cleanly capture all the output (stdout, stderr) and exit codes without worrying about it interfering with anything else.

Docker Integration: For total isolation, it would be amazing if it could run code inside a Docker container. We could point it to a Dockerfile or a pre-built image, ensuring the environment is always clean and consistent, just like in our CI/CD pipelines.

Capturing Generated Files:

It would also be super helpful if the executor could grab files created during a run, like logs or reports. Maybe we could tell it which files or folders to look for, and it would pull them out of the execution environment for us to inspect.

**Describe alternatives you've considered**
Adding security checks to current execs just seems more like a band-aid, this could be something that over time may be developed or repurposed for other use cases. I also considered a VM but that is definitely an overkill

AlexFierro9:
Hey Guys anything? can I be assigned to this if the other maintainers are busy? @seanzhou1023 @boyangsvl

boyangsvl:
I've assigned the issue to you @AlexFierro9 , thank you for the contribution!

AlexFierro9:
#3225

klateefa:
@AlexFierro9 Thank you for reporting this. I understand the request is currently under review. Please keep me posted if anything else is needed. Thanks for your patience.

GWeale:
closing as a duplicate of #5302, which asks for the part still missing: ContainerCodeExecutor copying input files in and generated files back out. a pull request for it is open in #5321. the other two asks exist today. ContainerCodeExecutor runs code in docker from a prebuilt image or a Dockerfile path, and UnsafeLocalCodeExecutor runs each program in its own child python process and reports stdout, stderr and the exit code, 2.9.2 included. open a new issue if you want us to look at it again.

PR #3225, opened 2025-10-20
added isolated code executor

AlexFierro9:
Have added a new Isolated Code Executor that uses subprocesses module to run the code on a new interpreter process, there is a bonus of isolated code execution for local setups as well. Working on support for Docker Integration as well but that is going to be a while.

Implements feature plan partly described in #3075

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ dependencies = [
   "google-cloud-storage>=3.0.0, <4.0.0",                     # For GCS Artifact service
   "google-genai>=1.45.0, <2.0.0",                            # Google GenAI SDK
   "graphviz>=0.20.2, <1.0.0",                                # Graphviz for graph rendering
+  "jsonschema>=4.23.0, <5.0.0",                              # Agent Builder config validation
   "mcp>=1.8.0, <2.0.0;python_version>='3.10'",               # For MCP Toolset
   "opentelemetry-api>=1.37.0, <=1.37.0",                     # OpenTelemetry - limit upper version for sdk and api to not risk breaking changes from unstable _logs package.
   "opentelemetry-exporter-gcp-logging>=1.9.0a0, <2.0.0",
--- a/src/google/adk/tools/openapi_tool/openapi_spec_parser/rest_api_tool.py
+++ b/src/google/adk/tools/openapi_tool/openapi_spec_parser/rest_api_tool.py
@@ -428,7 +428,7 @@ class RestApiTool(BaseTool):
               f"Tool {self.name} execution failed. Analyze this execution error"
               " and your inputs. Retry with adjustments if applicable. But"
               " make sure don't retry more than 3 times. Execution Error:"
-              f" {error_details}"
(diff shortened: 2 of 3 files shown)


AlexFierro9:
#3075

AlexFierro9:
Implemented unit-tests in line with those from unsafe local code executor

AlexFierro9:
@boyangsvl please assign a reviewer as time permits

hangfei:
Thanks for your contribution!

#1 Why is it running in a separate process helpful? We have UnsafeCodeExecutor that does very similar things but in the same process. (https://github.com/google/adk-python/blob/main/src/google/adk/code_executors/unsafe_local_code_executor.py
#2 If it make sense, it should be a config on current UnsafeCodeExecutor. 

@AlexFierro9

AlexFierro9:
#1 Unsafe LocalCodeExecutor crashed the main adk process for me in a few scenarios, eg. Recursion or brute force codes and it allowed access to internal data from the main process as well. Running it as a subprocess allows for better stability and much better data protection at a slightly higher overhead. I know that it's literally called UnsafeCodeExecutor but this solution seemed pretty obvious to me, and it may appeal to certain developers.

#2 That makes sense to me but the current implementation of UnsafeCodeExecutor seems pretty tight to me, I didn't want to touch that, though if that's better for long term maintenance and upkeep, I can do that as well

@hangfei

boyangsvl:
@AlexFierro9 I think merging with UnsafeCodeExecutor makes sense as we want people to know very clearly that this is an unsafe executor from a security point of view. Could you work on it? Thanks!

AlexFierro9:
On it! @boyangsvl

AlexFierro9:
@boyangsvl I've added config option in UnsafeLocalCodeExecutor to use isolated process for code execution as discussed.

src/google/adk/code_executors/unsafe_local_code_executor.py
```diff
@@ -47,39 +51,57 @@ class UnsafeLocalCodeExecutor(BaseCodeExecutor):
   # optimize_data_file.
   optimize_data_file: bool = Field(default=False, frozen=True, exclude=True)
 
-  def __init__(self, **data):
+  use_isolated_process: bool = False
```
boyangsvl: what does 'isolated' mean? I am worried that some people may interpret this as a sandbox. Could you rename it to use_separate_process?

AlexFierro9: On it

AlexFierro9:
@boyangsvl, I've completed the changes as requested

hangfei:
Could you describe the manual testing you have done? 

Please fix the pyink issue with ./autoformat.sh we provided.

hangfei:
Tests seems are still failing. PTAL. Thanks.

hangfei:
the tests, formatting are still failing. PTAL.

AlexFierro9:
Closing this PR to start a new one, will push it out shortly

PR #3297, opened 2025-10-27
Agent card skills usage

luis5tb:
**Problem:**
The subagents agent cards information is not being used by the root agent to decide on which subagent should the request be transfered to

**Solution:**
This PR ensures the subagents agent cards information (description, skills) are being used as context for the root agent so that it can better decide on the subagent to transfer the request.

It also includes a bit or rewording/formatting so that information about each subagent can be better processed

### Testing Plan

Unit test (test_agent_transfer_system_instructions.py) has been modified to cover the new expected context

**Unit Tests:**

```
$ pytest ./tests/unittests/flows/llm_flows/test_agent_transfer_system_instructions.py                                                                                                                                                     agent_card_skills_usage 
==================================================================================================================================================== test session starts =====================================================================================================================================================
platform linux -- Python 3.11.14, pytest-8.4.2, pluggy-1.6.0
configfile: pyproject.toml
plugins: langsmith-0.4.38, xdist-3.8.0, asyncio-1.2.0, mock-3.15.1, anyio-4.11.0
collected 4 items                                                                                                                                                                                                                                                                                                            

tests/unittests/flows/llm_flows/test_agent_transfer_system_instructions.py ....                                                                                                                                                                                                                                        [100%]

====================================================================================================================================================== warnings summary ======================================================================================================================================================
.venv/lib64/python3.11/site-packages/google/genai/types.py:9562
  /home/ltomasbo/AI/Claude/adk-python/.venv/lib64/python3.11/site-packages/google/genai/types.py:9562: PydanticDeprecatedSince212: Using `@model_validator` with mode='after' on a classmethod is deprecated. Instead, use an instance method. See the documentation at https://docs.pydantic.dev/2.12/concepts/validators/#model-after-validator. Deprecated in Pydantic V2.12 to be removed in V3.0.
    @model_validator(mode='after')  # type: ignore[arg-type]

================================================================================================================================================ 4 passed, 1 warning in 3.26s ================================================================================================================================================
```

_Please provide instructions on how to manually test your changes, including any
necessary setup or configuration. Please provide logs or screenshots to help
reviewers better understand the fix._

### Checklist

### Additional context

### DIFF
--- a/src/google/adk/flows/llm_flows/agent_transfer.py
+++ b/src/google/adk/flows/llm_flows/agent_transfer.py
@@ -16,6 +16,8 @@
 
 from __future__ import annotations
 
+import asyncio
+import logging
 import typing
 from typing import AsyncGenerator
 
@@ -35,6 +37,8 @@ if typing.TYPE_CHECKING:
   from ...agents import BaseAgent
   from ...agents import LlmAgent
   from ...agents.remote_a2a_agent import RemoteA2aAgent
+
+logger = logging.getLogger('google_adk.' + __name__)
 
 
 class _AgentTransferLlmRequestProcessor(BaseLlmRequestProcessor):
@@ -138,10 +142,14 @@ async def _build_target_agents_info_async(target_agent: BaseAgent) -> str:
         return _build_target_agent_info_from_card(
             target_agent, target_agent._agent_card
         )
-    except Exception:
+    except Exception as e:
       # If resolution fails, fall through to default behavior
+      logger.warning(
+          'Failed to resolve A2A agent card for agent "%s", falling back to' ' basic info. Error: %s',
+          target_agent.name,
+          e,
+      )
       pass
-
   # Fallback to original behavior for non-A2A agents or if card unavailable
   return _build_target_agents_info(target_agent)
 
@@ -192,11 +200,12 @@ async def _build_target_agents_instructions(
       f'`{name}`' for name in available_agent_names
   )
 
-  # Build agent info asynchronously to support A2A agent card resolution
-  agent_info_list = []
-  for target_agent in target_agents:
-    agent_info = await _build_target_agents_info_async(target_agent)
-    agent_info_list.append(agent_info)
+  # Build agent info asynchronously and concurrently to support A2A agent card resolution
+  tasks = [
+      _build_target_agents_info_async(target_agent)
+      for target_agent in target_agents
+  ]
+  agent_info_list = await asyncio.gather(*tasks)
 
   # Create a separator for visual clarity
   agents_section = '\n\n'.join(agent_info_list)
--- a/src/google/adk/flows/llm_flows/agent_transfer.py
+++ b/src/google/adk/flows/llm_flows/agent_transfer.py
@@ -30,8 +30,11 @@ from ...tools.transfer_to_agent_tool import transfer_to_agent
 from ._base_llm_processor import BaseLlmRequestProcessor
 
 if typing.TYPE_CHECKING:
+  from a2a.types import AgentCard
+
   from ...agents import BaseAgent
   from ...agents import LlmAgent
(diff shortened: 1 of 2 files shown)


bparees:
@luis5tb thinking about our particular needs/use cases a few things come to mind:

1) when would we want to reload the agent card in case it has changed?  (currently i assume this just does it when the ADK process starts up/initializes the remote agent).  Tbh this is probably a bigger can of worms that is more closely tied to the A2A protocol/agent-card format itself (agent cards having some sort of guidance about reloading/expiration)

2) ability to choose *not* to use the agent card details as part of the supervisor routing instructions?  (maybe this just means not providing an agent card url/string to the remote agent in the first place?)

luis5tb:
> @luis5tb thinking about our particular needs/use cases a few things come to mind:
> 
> 1. when would we want to reload the agent card in case it has changed?  (currently i assume this just does it when the ADK process starts up/initializes the remote agent).  Tbh this is probably a bigger can of worms that is more closely tied to the A2A protocol/agent-card format itself (agent cards having some sort of guidance about reloading/expiration)

This is a  good point. I agree it must be handled but not as part of this PR, right? I think it is loaded the first time the root agent needs to invoke the subagents (at least it is when I see the request on the subagent)

> 2. ability to choose _not_ to use the agent card details as part of the supervisor routing instructions?  (maybe this just means not providing an agent card url/string to the remote agent in the first place?)

Another great point! Thanks for raising! Perhaps one way to configure using or not the description/skills on the agent card is by defining or not a description when creating the RemoteA2aAgent. What do you think?

Note the agent card must be read as the endpoint is being retrieved from there (as well as other details such as the security needed for instance)

bparees:
> This is a good point. I agree it must be handled but not as part of this PR, right? I think it is loaded the first time the root agent needs to invoke the subagents (at least it is when I see the request on the subagent)

no, not as part of this PR per se, but it's the next logical question once the ADK has this behavior (to be fair this concern also applies to the existing ADK behavior because for example a new version of the agent card could update how security is managed), so maybe "when should agent cards be refreshed/how is it controlled" is a distinct RFE that just has implications for this behavior.

> Another great point! Thanks for raising! Perhaps one way to configure using or not the description/skills on the agent card is by defining or not a description when creating the RemoteA2aAgent. What do you think?

yeah, ultimately these feel like config/parameter choices when constructing the remote agent:

1) to use the agent card or not for routing/supervisor instructions
2) to provide a description directly or not
3) interplay of the description and the agent card (can you provide both?  if so, do they get merged together?)

luis5tb:
> > This is a good point. I agree it must be handled but not as part of this PR, right? I think it is loaded the first time the root agent needs to invoke the subagents (at least it is when I see the request on the subagent)
> 
> no, not as part of this PR per se, but it's the next logical question once the ADK has this behavior (to be fair this concern also applies to the existing ADK behavior because for example a new version of the agent card could update how security is managed), so maybe "when should agent cards be refreshed/how is it controlled" is a distinct RFE that just has implications for this behavior.
> 
> > Another great point! Thanks for raising! Perhaps one way to configure using or not the description/skills on the agent card is by defining or not a description when creating the RemoteA2aAgent. What do you think?
> 
> yeah, ultimately these feel like config/parameter choices when constructing the remote agent:
> 
> 1. to use the agent card or not for routing/supervisor instructions
> 2. to provide a description directly or not
> 3. interplay of the description and the agent card (can you provide both?  if so, do they get merged together?)

Yeah, currently (with this PR I mean) if both are provided, both are included:
```
  if target_agent.description:
    descriptions.append(f'Description: {target_agent.description}')
  if agent_card.description and agent_card.description != target_agent.description:
    descriptions.append(f'Agent card description: {agent_card.description}')
```

rohityan:
Hi @luis5tb closing this since it already merged into codebase in this commit https://github.com/google/adk-python/commit/e1e0d6361675e7b9a2c9b2523e3a72e2e5e7ce05

PR #3303, opened 2025-10-27
chore: Skip events without content when finding the last transfer agent.

hanmingalbertyang:
Events without content, such as those from callbacks with state_deltas, do not represent a transfer to an agent that should continue the conversation.


hanmingalbertyang:
this was not the correct fix for the issue

Issue #3429, opened 2025-11-06
Detailed List Apps


dylan-apex:
The current implementation of the `/list-apps` endpoint only returns the name of the folder that the agents are in. This is unhelpful for a production deployment because we want to include the description and a human presentable name from the API, rather than having to write it in the frontend code.

PR #3430, opened 2025-11-06
feat(web): add list-apps-detailed endpoint

dylan-apex:

- Closes: #3429

_If applicable, please follow the issue templates to provide as much detail as
possible._

**Problem:**
The existing `/list-apps` endpoint only returns the name of the folder that each agent is in

**Solution:**
This adds a new endpoint `/list-apps-detailed` which will load each agent using the existing `AgentLoader.load_agent` method, and then return the folder name, display name (with underscores replaced with spaces for a more readable version), description, and the agent type.

This does introduce overhead if you had multiple agents since they all need to be loaded, but by maintaining the existing `/list-apps` endpoint, users can choose which one to hit if they don't want to load all agents. Since the existing `load_agents` method will cache results, there's only a penalty on the first hit.

### Testing Plan

Created a unit test for this, similar to the `/list-apps`. Also tested this with my own ADK instance to verify it loaded correctly.
```
curl --location "localhost:8000/list-apps-detailed"
```
```json
{
    "apps": [
        {
            "name": "agent_1",
            "displayName": "Agent 1",
            "description": "A test description for a test agent",
            "agentType": "package"
        },
        {
            "name": "agent_2",
            "displayName": "Agent 2",
            "description": "A test description for a test agent ",
            "agentType": "package"
        },
        {
            "name": "agent_3",
            "displayName": "Agent 3",
            "description": "A test description for a test agent",
            "agentType": "package"
        }
    ]
}


```

**Unit Tests:**

- [X] I have added or updated unit tests for my change.
- [X] All unit tests pass locally.

 3054 passed, 2383 warnings in 46.96s 

### Checklist

### DIFF
--- a/src/google/adk/cli/utils/base_agent_loader.py
+++ b/src/google/adk/cli/utils/base_agent_loader.py
@@ -36,6 +36,14 @@ class BaseAgentLoader(ABC):
   def list_agents(self) -> list[str]:
     """Lists all agents available in the agent loader in alphabetical order."""
 
-  @abstractmethod
   def list_agents_detailed(self) -> list[dict[str, Any]]:
-    """Lists all agents with detailed metadata (name, display name, description, type)."""
+    agent_names = self.list_agents()
+    return [
+        {
+            'name': name,
+            'display_name': None,
+            'description': None,
+            'type': None,
+        }
+        for name in agent_names
+    ]


src/google/adk/cli/utils/agent_loader.py
```diff
@@ -321,6 +323,57 @@ def list_agents(self) -> list[str]:
...
+          agent = loaded
+
+        agent_type = self._determine_agent_type(agent_name)
+
+        # Convert underscores to spaces for display_name
+        display_name = agent.name.replace("_", " ")
+
+        app_info = {
+            "name": agent_name,
+            "display_name": display_name,
```
boyangsvl: could you remove display_name? We do not have a concept of a display name.

src/google/adk/cli/utils/agent_loader.py
```diff
@@ -321,6 +323,57 @@ def list_agents(self) -> list[str]:
...
+    if (base_path / "root_agent.yaml").exists():
+      return "yaml"
+    elif (base_path / "agent.py").exists():
+      return "package"
+    elif (base_path / "__init__.py").exists():
+      return "package"
+    else:
+      module_path = Path.cwd() / self.agents_dir / f"{agent_name}.py"
+      if module_path.exists():
+        return "python"
```
boyangsvl: what's the difference between package and python? I think we support only the folder with __init__.py.

src/google/adk/cli/adk_web_server.py
```diff
@@ -702,6 +713,11 @@ async def internal_lifespan(app: FastAPI):
     async def list_apps() -> list[str]:
       return self.agent_loader.list_agents()
 
+    @app.get("/list-apps-detailed")
```
boyangsvl: could you do this under the existing list-apps? maybe adding a query parameter?

src/google/adk/cli/utils/agent_loader.py
dylan-apex: __init__.py is not technically required if theres a root_agent defined

src/google/adk/cli/utils/agent_loader.py
dylan-apex: Its not explicitly a display name in the library, but theres the folder name (agent_name) and the name the agent has when created `Agent(name="")`, which is closer to a display name.

src/google/adk/cli/utils/agent_loader.py
dylan-apex: actually nvm you're right

dylan-apex:
@boyangsvl 
Used existing list-apps endpoint with `detailed=true` query param
used `simple_name` for the file/folder name
use `agent_name` for the value that comes from `agent.name` - the end user can convert underscores to spaces if they wish
removed `python` type - confused it with `yaml`

src/google/adk/cli/adk_web_server.py
```diff
@@ -280,6 +280,17 @@ class ListMetricsInfoResponse(common.BaseModel):
   metrics_info: list[MetricInfo]
 
 
+class AppInfo(common.BaseModel):
+  simple_name: str
```
boyangsvl: is this the name of the app? if so could you just call it 'name'?

src/google/adk/cli/adk_web_server.py
```diff
@@ -280,6 +280,17 @@ class ListMetricsInfoResponse(common.BaseModel):
...
+  agent_name: str
```
boyangsvl: this is root_agent_name? If so could you rename?

src/google/adk/cli/adk_web_server.py
```diff
@@ -280,6 +280,17 @@ class ListMetricsInfoResponse(common.BaseModel):
...
+  description: str
+  agent_type: Literal["yaml", "package"]
```
boyangsvl: how about naming this 'language', so it can be 'yaml', or 'python'. it can be other languages too.

dylan-apex:
@boyangsvl recommended changes made, autoformat ran, and all unit tests pass

hangfei (approved):

hangfei:
@dylan-apex You also need to implement it for InMemoryAgentLoader. 

Right now, it's failing some tests. Please help fix.

dylan-apex:
@hangfei I don't have access to see which tests are failing. I also assume you mean `src/google/adk/cli/utils/agent_loader.py` when you say `InMemoryAgentLoader`, but want to confirm.

I found [this commit](https://github.com/google/adk-python/commit/bda3df24802d0456711a5cd05544aea54a13398d) which hints at an InMemory implementation, but not sure if this is what you mean.

src/google/adk/cli/utils/base_agent_loader.py
```diff
@@ -34,3 +35,7 @@ def load_agent(self, agent_name: str) -> Union[BaseAgent, App]:
   @abstractmethod
   def list_agents(self) -> list[str]:
     """Lists all agents available in the agent loader in alphabetical order."""
+
+  @abstractmethod
```
hangfei: this is a backward incompatible change. let's make this optional or add a default implementation(if possible)

hangfei:
> @hangfei I don't have access to see which tests are failing. I also assume you mean `src/google/adk/cli/utils/agent_loader.py` when you say `InMemoryAgentLoader`, but want to confirm.
> 
> I found [this commit](https://github.com/google/adk-python/commit/bda3df24802d0456711a5cd05544aea54a13398d) which hints at an InMemory implementation, but not sure if this is what you mean.

i meant this one: https://github.com/google/adk-python/pull/3430/files#r2547636742. Thanks.

src/google/adk/cli/utils/base_agent_loader.py
dylan-apex: Added

dylan-apex:
@hangfei added a default implementation that just pulls the name from the regular `list_agents` method

PR #3660, opened 2025-11-22
fix(migration): constrain alembic targets and reset version state.

MasakiMu319:

  - Closes: N/A
  - Related: N/A

### Problem:
In production, leaving Alembic unconstrained is extremely dangerous—autogenerated migrations could drop or alter unrelated business tables. The migration script also left stale alembic_version state and did not scope tables to ADK session metadata.

### Solution:
Constrain Alembic to ADK session tables via targeted metadata and an include_object filter, using a dedicated alembic_version_adk as alembic version table. Rewrite the migration script to emit this safe env.py, clean backups, and reset any old version table before autogeneration—here we only need the DB tables to align with the latest Base metadata—keeping migrations isolated and idempotent.

### Testing Plan

- Unit Tests: Not run (no code-level changes).
- Manual Verification: Ran the migration flow locally and confirmed the Alembic version table matched expectations.

### Checklist
- [x] I have performed a self-review of my changes.

### DIFF
--- a/contributing/samples/migrate_session_db/sample-output/alembic/env.py
+++ b/contributing/samples/migrate_session_db/sample-output/alembic/env.py
@@ -53,6 +53,7 @@ def include_object(obj, name, type_, reflected, compare_to):
       return False
   return True
 
+
 # other values from the config, defined by the needs of env.py,
 # can be acquired:
 # my_important_option = config.get_main_option("my_important_option")
--- a/contributing/samples/migrate_session_db/sample-output/alembic/env.py
+++ b/contributing/samples/migrate_session_db/sample-output/alembic/env.py
@@ -49,7 +49,7 @@ def include_object(obj, name, type_, reflected, compare_to):
   if type_ == "index":
     try:
       return obj.table.name in _ALLOWED_TABLE_NAMES
-    except Exception:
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @MasakiMu319 ,Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Before we can proceed with the review can you fix the lint errors.

MasakiMu319:
> Hi @MasakiMu319 ,Thank you for your contribution! We appreciate you taking the time to submit this pull request. Before we can proceed with the review can you fix the lint errors.

Already done.

rohityan:
Hi @ankursharmas , can you please review this

GWeale:
the env.py this patches doesn't exist any more. the sample moved to contributing/samples/context_management/migrate_session_db and was rewritten around the built-in adk migrate session command, so there's nothing left to constrain there.

the other half of your diff, scripts/db_migration.sh, is still in the tree. nothing points at it now that the sample stopped using it, so it's dead code, and removing it seems like a better answer than hardening it.

you were right about the underlying risk. an unconstrained autogenerate will happily drop tables it doesn't own. if you think the script is worth keeping and fixing rather than deleting, open a new issue and make that case.

PR #3667, opened 2025-11-22
Fix: Add fallback logic to retrieve user message when resuming async …

ycsmiley:
Problem: When using run_async to resume an invocation (e.g., after a ToolConfirmation pause), the strict lookup for the user message using invocation_id often fails, leading to a ValueError: No user message available. This happens because the session persistence layer sometimes disconnects the original user message ID from the resumed invocation ID.

Solution: Added a fallback mechanism in _setup_context_for_resumed_invocation. If the strict lookup returns None, the runner now searches the session history (in reverse order) to find the most recent message authored by 'user'. This ensures the agent allows execution to resume with the correct context.

Testing: Verified locally with a Human-in-the-Loop workflow involving ToolConfirmation and run_async. The fix successfully recovered the user message and allowed the agent to proceed without crashing.


rohityan:
Hi @ycsmiley,Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Before we can proceed with the review can you fix the failing lint errors using autoformat.sh.

ycsmiley:
Hi @ryanaiagent 
Thanks for the guidance. I've run autoformat.sh and pushed the fixes. The lint errors should be resolved now.

rohityan:

ycsmiley:
Hi @haranrk, the merge conflicts and unit test failure have been resolved. The fork CI is green, and the PR is currently waiting for maintainer workflow approval and a write-access review. Could you please help approve the workflows and review the PR when available? Thank you!

DeanChensj:
Hi @ycsmiley,

Thank you for your contribution and for looking into the invocation resumption flow in ADK! We truly appreciate you taking the time to submit this PR.
After a thorough architectural review, we have decided **not to merge** this fallback mechanism. Here is the technical context and rationale:

---

### 1. Risk of Cross-Turn Context Contamination
Searching backwards across the entire session history (`for event in reversed(session.events): if event.author == 'user': return event.content`) introduces a critical bug in multi-turn or branching conversations:
- Suppose **Turn 1 (`inv_1`)** executes a long-running tool and pauses for user confirmation.
- Meanwhile, the user asks an unrelated question in **Turn 2 (`inv_2`)** (e.g., *"What is the weather today?"*).
- When **Turn 1 (`inv_1`)** resumes with the tool result, falling back to the latest user message would grab Turn 2's prompt (*"What is the weather today?"*) and inject it as Turn 1's `user_content`.
- This causes prompt corruption, cross-turn context leakage, and severe model hallucinations.

### 2. Session Persistence Invariants

The session persistence layer (`BaseSessionService` and its database implementations) does not alter or disconnect `Event.invocation_id`. Each event's `invocation_id` is preserved as an immutable field throughout its lifecycle.

---

### 3. Root-Cause Fixes Already Landed in `main`
The historical scenarios that previously triggered `ValueError: No user message available` stemmed from distinct upstream routing issues, all of which have recently been resolved structurally:
1. **Multimodal User Turns**: Prompts structured as `[image_blob, text_query]` were previously rejected because the lookup strictly expected text in `parts[0]`. This has been widened to check text across all parts.
2. **Caller Invocation ID Mismatches**: When resuming via tool responses, caller-supplied IDs on the node path were previously trusted blindly instead of being reconciled against the originating `FunctionCall.id`. This is now automatically reconciled.
3. **Workflow Node Inputs**: Single-turn workflow node inputs are now formally written through session persistence ahead of execution.
---

Because the underlying failure modes have been addressed at their source without compromising session isolation, we will be closing this PR. 
If you have a concrete, reproducible end-to-end scenario that still triggers this error on the latest `main` branch, please feel free to open an issue with a minimal reproduction script and we will gladly investigate!
Thanks again for helping improve ADK!

Issue #4035, opened 2025-12-29
ADK is unsupported RemoteA2aAgent


jacksunmico:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)
https://google.github.io/adk-docs/a2a/quickstart-exposing/

**Describe the bug**
adk-python/contributing/samples/a2a_basic/agent.py:94: UserWarning: [EXPERIMENTAL] RemoteA2aAgent: ADK Implementation for A2A support (A2aAgentExecutor, RemoteA2aAgent and corresponding supporting components etc.) is in experimental mode and is subjected to breaking changes. A2A protocol and SDK arethemselves not experimental. Once it's stable enough the experimental mode will be removed. Your feedback is welcome.
  prime_agent = RemoteA2aAgent(

jacksunmico:
Failed to initialize remote A2A agent: Failed to initialize remote A2A agent prime_agent: Failed to resolve AgentCard from URL http://localhost:8001/.well-known/agent-card.json: HTTP Error 502: Failed to fetch agent card from http://localhost:8001/.well-known/agent-card.json: Server error '502 Bad Gateway' for url 'http://localhost:8001/.well-known/agent-card.json' For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502

jacksunmico:
<img width="1437" height="763" alt="Image" src="https://github.com/user-attachments/assets/a55de2dc-d934-46df-950c-b931a892b407" />

jacksunmico:
<img width="1435" height="769" alt="Image" src="https://github.com/user-attachments/assets/d07c07eb-458a-48cc-80eb-13e087825aed" />

GWeale:
@jacksunmico What was the error for this issue. Where you able to find a fix

luckyxue:
I'm facing the same problem as you, and it's still unresolved. The solutions mentioned above haven't worked.

luckyxue:
It's possible there's a VPN network issue; do not set it to global.

llalitkumarrr:
Hello @luckyxue ,

Kindly try latest google-adk version (`google-adk==1.23.0`)to check if this issue persists. If you still observe this issue, kindly share the error you observe and minimal reproducible code, it would be help us analyze this issue and assist you better.

PR #3674, opened 2025-11-23
Add log_level option for adk run CLI

vadakattu:
Added log_level option for adk run CLI, inline with other CLI capabilities

**Problem:**
The CLI option was overlooked for this entry point & the logs utility lot_to_tmp_folder was not receiving a log level argument.

**Solution:**
Cloned the click option as defined for other entry points, added function argument, and passed to logging utility function

### Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md) document.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.
- [x] I have manually tested my changes end-to-end.
- [ ] Any dependent changes have been merged and published in downstream modules.

Issue #2014, opened 2025-07-17
Master issue: [Streaming Tools] support streaming intermediate results for tools for non-streaming case


hangfei:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

AdamPaternostro:
I have the following use case:

1. I have a single tool that runs for 2 minutes which does a series of steps (each step can take 2 to 30 seconds)
2. The user interface waits on the tool for 2 minutes and the user does not know if the tool is frozen
3. I would like to provide progress updates during the long running execution

Example output from tool: 
- Creating image prompt
- Creating the image
- Scanning the image for policy violations...  (maybe this takes 30 seconds)
- Done scanning image: No policy violations found
- Adding watermark to image
- Checking image into source control
- Updating readme 
- { returns }

The user would see the outputted status and the final LLM output.

Pseudo code for tool.  "raise_status_message" would send back statuses, to the UI via the REST SSE, during the tool execution.

```
def image_curation_process(prompt: str)-> dict:
    """
    Creates an image, scans it and checks it into source control.

    Returns:
        dict: A dictionary containing the status and a boolean result.
        {
            "status": "success" or "failed",
            "tool_name": "image_curation_process",
            "generated_image_url" : "generated_image_url"
            }
        }
    """ 
    generate_enhanced_prompt_result = generate_enhanced_prompt(prompt)
    raise_status_message("Enhanced prompt generated")
    
    generate_image_result = generate_image(generate_enhanced_prompt_result["enhanced_prompt"]"])
    raise_status_message("Image generated")
    
    raise_status_message("Starting image scan...")
    scan_image_result = scan_image(generate_image_result["generated_image_url"])
    raise_status_message("Image scanned")
    
    if scan_image_result == False:
        return { "status": "failed", "tool_name" : "image_curation_process", "messages": ["Image voliated policy 123."] }

    add_watermark_result = add_watermark(generate_image_result["generated_image_url"])
    raise_status_message("Watermark added")
    
    check_into_source_control_result = check_into_source_control(add_watermark_result["generated_image_url"])
    raise_status_message("Image checked into source control")
    
    update_readme_result = update_readme(check_into_source_control_result["generated_image_url"])
    raise_status_message("Readme updated")  

[...]

PR #4060, opened 2026-01-01
Update A2A Agent Card path to expected name

vadakattu:
When using the CLI fastapi server with the --a2a flag to expose an agent for remote access, the agent card address it's looking for is out of date with the latest documentation.

- Related: #4035

**Problem:**
the old path that was expected `agent.json` vs `agent-card.json`
See also `a2a/utils/constants.py`

**Solution:**
updating to look for the new path `agent-card.json`

### DIFF
--- a/src/google/adk/cli/fast_api.py
+++ b/src/google/adk/cli/fast_api.py
@@ -344,12 +344,12 @@ def get_fast_api_app(
         return _get_a2a_runner_async
 
       for p in base_path.iterdir():
-        # only folders with an agent.json file representing agent card are valid
+        # only folders with an agent-card.json file representing agent card are valid
         # a2a agents
         if (
             p.is_file()
             or p.name.startswith((".", "__pycache__"))
-            or not (p / "agent.json").is_file()
+            or not (p / "agent-card.json").is_file()
         ):
           continue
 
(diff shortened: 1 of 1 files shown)


vadakattu:
It seems there are quite a few doc pages, and samples linked which reference the old filename, despite some sporadic updates to the new filename in places

An example of a documentation page which mixes both old and new is here: https://google.github.io/adk-docs/a2a/quickstart-consuming/#overview (search for `agent.json` and `agent-card.json` references)

updates to the ADK examples in here https://github.com/google/adk-python/tree/main/contributing/samples would also be necessary.

Other option would be to clarify that for the CLI a2a utilities, agent.json is expected rather than agent-card.json as a few places state, and just call out the discrepancy with the A2A spec in how the file is ultimately exposed as agent-card.json the docs.

rohityan:
Hi @vadakattu, Thank you for your contribution! We appreciate you taking the time to submit this pull request. 
Like you pointed out clarification of the documentation is required. Would you be willing to revise the PR to incorporate those changes?

vadakattu:
Hey @ryanaiagent, so option A revising and updating all relevant references to the latest protocol, rather than the second option just to signpost users right?

It seems like this would be a more involved PR across a at least 2 repos -- I currently have had a much simpler PR https://github.com/google/adk-python/pull/3674  languishing for little while, so happy to take a look once that's resolved and there's a sense that a PR here could be reviewed & merged relatively expeditiously, as the current protracted review cycle requires more conflict resolution & other updates to maintain alignment with the latest protocol standards and ADK release.

GWeale:
these are two different things rather than a mismatch. agent.json is the file adk api_server --a2a looks for inside an agent folder, and the url it then serves is already /a2a/<app>/.well-known/agent-card.json. the docs that mixed the two were the real problem and got fixed in adk-docs #2014. renaming the on-disk file breaks the four a2a samples and two tests in test_fast_api.py. thanks for spotting the inconsistency.

PR #4072, opened 2026-01-05
feature/auto-create-new-session

lwangverizon:

**Problem:**

When building frontend applications with ADK, there's a limitation where frontends cannot always guarantee that `create_session` is called before initiating a conversation. This creates friction in the user experience because:

- Users may refresh the page or navigate directly to a conversation URL with a specific session_id
- Frontend state management may lose track of whether a session was already created
- Mobile apps or single-page applications have complex lifecycle management where ensuring `create_session` is called first adds unnecessary complexity
- This forces developers to implement additional logic to check session existence before every conversation

Currently, if `get_session` is called with a non-existent session_id, it returns `None`, requiring the frontend to explicitly handle this case and call `create_session` separately.

**Solution:**

Modified the `get_session` method in `DatabaseSessionService` to automatically create a session if it doesn't exist in the database. This "get or create" pattern is common in many frameworks and provides a more developer-friendly API.

The implementation:
1. Attempts to fetch the session from the database
2. If the session doesn't exist (returns `None`), automatically calls `create_session` with the provided parameters
3. Retrieves and returns the newly created session
4. Maintains backward compatibility - existing code continues to work without changes

This allows frontends to simply call `get_session` with a session_id and be confident that the session will be available, regardless of whether it was previously created.

**Benefits:**
- Simplifies frontend integration by removing the need to track session creation state
- Reduces API calls (no need to check existence before calling get_session)
- Follows the principle of least surprise - getting a session with an ID should work reliably
- No breaking changes to existing code that checks for `None` return values

### Testing Plan

**Unit Tests:**

**pytest results:**

### DIFF
--- a/contributing/samples/bigquery/agent.py
+++ b/contributing/samples/bigquery/agent.py
@@ -71,7 +71,7 @@ bigquery_toolset = BigQueryToolset(
 # The variable name `root_agent` determines what your root agent is for the
 # debug CLI
 root_agent = LlmAgent(
-    model="gemini-2.0-flash",
+    model="gemini-2.5-flash",
     name=BIGQUERY_AGENT_NAME,
     description=(
         "Agent to answer questions about BigQuery data and models and execute"
--- a/src/google/adk/agents/live_request_queue.py
+++ b/src/google/adk/agents/live_request_queue.py
@@ -30,13 +30,29 @@ class LiveRequest(BaseModel):
   """The pydantic model config."""
 
   content: Optional[types.Content] = None
-  """If set, send the content to the model in turn-by-turn mode."""
+  """If set, send the content to the model in turn-by-turn mode.
+
+  When multiple fields are set, they are processed by priority (highest first):
+  activity_start > activity_end > blob > content.
+  """
   blob: Optional[types.Blob] = None
-  """If set, send the blob to the model in realtime mode."""
+  """If set, send the blob to the model in realtime mode.
+
+  When multiple fields are set, they are processed by priority (highest first):
+  activity_start > activity_end > blob > content.
+  """
   activity_start: Optional[types.ActivityStart] = None
-  """If set, signal the start of user activity to the model."""
+  """If set, signal the start of user activity to the model.
+
+  When multiple fields are set, they are processed by priority (highest first):
+  activity_start > activity_end > blob > content.
+  """
   activity_end: Optional[types.ActivityEnd] = None
-  """If set, signal the end of user activity to the model."""
+  """If set, signal the end of user activity to the model.
+
+  When multiple fields are set, they are processed by priority (highest first):
+  activity_start > activity_end > blob > content.
+  """
   close: bool = False
   """If set, close the queue. queue.shutdown() is only supported in Python 3.13+."""
 
--- a/src/google/adk/cli/cli_tools_click.py
+++ b/src/google/adk/cli/cli_tools_click.py
@@ -36,7 +36,6 @@ from . import cli_create
 from . import cli_deploy
 from .. import version
 from ..evaluation.constants import MISSING_EVAL_DEPENDENCIES_MESSAGE
-from ..sessions.migration import migration_runner
 from .cli import run_cli
 from .fast_api import get_fast_api_app
 from .utils import envs
@@ -1589,6 +1588,8 @@ def cli_migrate_session(
   """Migrates a session database to the latest schema version."""
   logs.setup_adk_logger(getattr(logging, log_level.upper()))
   try:
+    from ..sessions.migration import migration_runner
+
     migration_runner.upgrade(source_db_url, dest_db_url)
     click.secho("Migration check and upgrade process finished.", fg="green")
   except Exception as e:
--- a/src/google/adk/evaluation/eval_config.py
+++ b/src/google/adk/evaluation/eval_config.py
@@ -23,7 +23,9 @@ from pydantic import alias_generators
 from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Field
+from pydantic import model_validator
 
+from ..agents.common_configs import CodeConfig
 from ..evaluation.eval_metrics import EvalMetric
 from .eval_metrics import BaseCriterion
 from .eval_metrics import Threshold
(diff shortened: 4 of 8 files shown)


src/google/adk/sessions/database_session_service.py
```diff
@@ -305,7 +305,14 @@ async def get_session(
...
       if storage_session is None:
-        return None
+        await self.create_session(
+          app_name=app_name,
+          user_id=user_id,
+          session_id=session_id,
+        )
+        storage_session = await sql_session.get(
+            schema.StorageSession, (app_name, user_id, session_id)
+        )
```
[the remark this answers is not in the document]
lwangverizon: i have updated the code per your suggestion

rohityan:
Hi @lwangverizon, Thank you for your contribution! We appreciate you taking the time to submit this pull request. 
Can you please fix the failing unit tests and lint errors before we proceed with the review.

lwangverizon:
> Hi @lwangverizon, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing unit tests and lint errors before we proceed with the review.

Hi, I have resolved the unit tests and linting errors. I also updated the implementation approach: instead of modifying the session service directly, I updated the logic within the runner. This should minimize the impact of the changes and make the review process easier.

sasha-gitg:
@lwangverizon Thank you for the contribution. Can you make this behavior configurable in the Runner constructor(ie: `auto_create_session: bool = False`) and default to the current behavior to make auto session creation opt-in?

lwangverizon:
Thx i have made the update please take a look at the change summary below:

### Change Summary

* **New Flag:** Added `auto_create_session: bool = False` to `Runner.__init__` to control missing-session behavior. (See `runners.py`)
* **Session Gating:** Updated session retrieval paths to conditionally create or error:
  * `run_async()`: If missing and `auto_create_session=True`, create; else raise `ValueError(self._format_session_not_found_message(session_id))`. (See `runners.py`)
  * `rewind_async()`: Same conditional create-or-error. (See `runners.py`)
  * `run_live()`: Same conditional create-or-error. (See `runners.py`)
* **Error Messaging:** Uses `_format_session_not_found_message(session_id)` so errors include helpful app-name alignment hints when applicable. (See `runners.py`)

### Tests

* **Extended coverage in `test_runners.py`:**
  * `test_session_not_found_message_includes_alignment_hint`: Verifies missing-session error content with alignment hints.
  * `test_session_auto_creation`: Verifies successful execution with `auto_create_session=True` when a session is missing.

src/google/adk/runners.py
```diff
@@ -459,8 +466,13 @@ async def _run_with_trace(
             app_name=self.app_name, user_id=user_id, session_id=session_id
         )
         if not session:
-          message = self._format_session_not_found_message(session_id)
-          raise ValueError(message)
+          if self.auto_create_session:
```
sasha-gitg: Since this is done in 3 places, preference to refactor into a single helper method `async def _get_or_create_session(...)`

tests/unittests/test_runners.py
```diff
@@ -237,6 +237,40 @@ def _infer_agent_origin(
...
+  session_service = InMemorySessionService()
+  runner = RunnerWithMismatch(
+      app_name="expected_app",
+      agent=MockLlmAgent("test_agent"),
+      session_service=session_service,
+      artifact_service=InMemoryArtifactService(),
+      auto_create_session=True,
+  )
+
+  agen = runner.run_async(
```
sasha-gitg, line 278: Preference to include tests for `rewind_async` and `run_live` as well. Thanks!

lwangverizon: The requested changes have been addressed. Please note that I decided to keep the auto_create_session flag outside of the helper function to ensure better code readability. New test cases have also been included. I appreciate your review.

src/google/adk/runners.py
```diff
@@ -966,15 +1020,12 @@ async def run_live(
           DeprecationWarning,
           stacklevel=2,
       )
-    if not session:
```
sasha-gitg, line 969: Even though this param is deprecated, we don't want to break it as part of this change, so let's still follow the same precedence of using the session param if provided.

src/google/adk/runners.py
```diff
@@ -343,9 +348,56 @@ def _format_session_not_found_message(self, session_id: str) -> str:
       return message
     return (
         f'{message}. {self._app_name_alignment_hint} '
-        'The mismatch prevents the runner from locating the session.'
+        'The mismatch prevents the runner from locating the session. '
+        'To automatically create a session when missing, set '
+        'auto_create_session=True when constructing the runner.'
     )
 
+  async def _get_or_create_session(
```
sasha-gitg, line 356: This and _get_session can be refactored into one method to handle both cases and make the calling site simpler. Something like:

```python
async def _get_or_create_session(self, *, user_id: str, session_id: str) -> Session:
  session = await self.session_service.get_session(...)
  if not session:
      if self.auto_create_session:
          session = await self.session_service.create_session(...)
      else:
          raise ValueError(...)
   return session
```

sasha-gitg (changes requested):

src/google/adk/runners.py
lwangverizon: of course just fixed please check

src/google/adk/runners.py
lwangverizon: yes of course just combined them as you suggested, please take another look

sasha-gitg (approved):

src/google/adk/sessions/database_session_service.py
```diff
@@ -436,8 +436,8 @@ async def append_event(self, session: Session, event: Event) -> Event:
       if storage_session.update_timestamp_tz > session.last_update_time:
         raise ValueError(
             "The last_update_time provided in the session object"
-            f" {datetime.fromtimestamp(session.last_update_time):'%Y-%m-%d %H:%M:%S'} is"
-            " earlier than the update_time in the storage_session"
+            f" {datetime.fromtimestamp(session.last_update_time):'%Y-%m-%d %H:%M:%S'}"
```
sasha-gitg: Please revert these lines. They are causing linting failures.

lwangverizon: done

sasha-gitg (approved):

lwangverizon:
Hi, just checking in on this PR. Is there anything else required on my end to get this merged?

sasha-gitg:
> Hi, just checking in on this PR. Is there anything else required on my end to get this merged?

Not on your side. It's going through internal review. I'll update if there are any additional requests.

hymanme:
This is an excellent feature optimization that allows us to automatically create sessions. However, I noticed that when using get_fast_api_app to create a FastAPI object and start the service, the initialization is done through an AdkWebServer object, which currently doesn't support the auto_create_session parameter. Is it possible to further extend support to the AdkWebServer version?

Issue #3263, opened 2025-10-23
Update GkeCodeExecutor with option to use Agent Sandbox


brandonroyal:
For users deploying ADK on Kubernetes and using GKE for Code Execution (https://google.github.io/adk-docs/tools/built-in-tools/#gke-code-executor), it would be great to leverage the faster startup latency and more granular control in Agent Sandbox (https://github.com/kubernetes-sigs/agent-sandbox)

I would like to see an additional `type` option in the GkeCodeExecutor which expands for the `Sandbox` type (https://github.com/kubernetes-sigs/agent-sandbox/blob/main/k8s/crds/agents.x-k8s.io_sandboxes.yaml).  This would augment the current `type: job` which exists today.  For backwards compatibility and stability, I'd expect the `type = 'job'` to be the default - at least until the Sandbox API graduates and stabilizes.

```
gke_executor = GkeCodeExecutor(
    ...
    type="sandbox"
    ...
)
```

When ADK instantiates a code execution request via the `GkeCodeExecutor` of type `sandbox`, ADK will create a `SandboxClaim` object.  Once fulfilled with a `Sandbox`, ADK will execute the code and return the result.  K8s administrators will have the option to configure things like WarmPools for low-latency execution and SandboxTemplates for ease of use.  I will expect these administrator configuration will work as expected when `SandboxClaims` are instantiated via `GkeCodeExecutor`.

aditya-shantanu:
@brandonroyal - I slightly alternate suggestion (in terms of implementation).
We're creating a Python SDK in AgentSandbox.
I'd like to see ADK directly use the Python SDK instead of us recoding very similar logic in two places.

brandonroyal:
Using the general Python SDK as an intermediary makes sense.

omarcevi:
> [@brandonroyal](https://github.com/brandonroyal) - I slightly alternate suggestion (in terms of implementation). We're creating a Python SDK in AgentSandbox. I'd like to see ADK directly use the Python SDK instead of us recoding very similar logic in two places.

Hey any updates regarding the python sdk. if possible i would like to work on this issue would this be ok? @seanzhou1023 @hangfei

klateefa:
@brandonroyal Thanks for reporting!
This feature request is under review ,Thanks for your patience.

DeanChensj:
[...]

PR #4111, opened 2026-01-10
execute-type param addition in GkeCodeExecutor

SHRUTI6991:
***Description:***

**Link to an existing issue (if applicable):**

https://github.com/google/adk-python/issues/3263

This PR updates the `GkeCodeExecutor` to add a new param called `executor_type` which takes either `job` / `sandbox`. Currently `GkeCodeExecutor` can only execute python code, so the changes made also focusses on executing ONLY python code via sandbox client.
### Testing Plan

**Unit Tests:**

Addition of new Unit tests to verify forking of code between `sandbox` vs `job`.

- [yes] I have added or updated unit tests for my change.
- [yes] All unit tests pass locally.

_Please include a summary of passed `pytest` results._

======================== 9 passed, 2 warnings in 2.72s =========================
Finished running tests!

Integration test by running a simple AI Agent: https://google.github.io/adk-docs/tools/google-cloud/gke-code-executor/ to write and execute Python code. The sandbox was connected via local tunnel for testing.

```
2026-01-10 00:46:28,105 - INFO - sandbox_client.py:159 - Creating SandboxClaim 'sandbox-claim-4d0f82e0' in namespace 'agent-sandbox-system' using template 'python-sandbox-template'...
2026-01-10 00:46:28,340 - INFO - sandbox_client.py:180 - Watching for Sandbox to become ready...
2026-01-10 00:46:39,284 - INFO - sandbox_client.py:208 - Sandbox sandbox-claim-4d0f82e0 is ready.
2026-01-10 00:46:39,285 - INFO - sandbox_client.py:243 - Starting Dev Mode tunnel: localhost:49605 -> svc/sandbox-router-svc:8080...
2026-01-10 00:46:39,288 - INFO - sandbox_client.py:259 - Waiting for port-forwarding to be ready...
2026-01-10 00:46:40,289 - INFO - sandbox_client.py:271 - Dev Mode ready. Tunneled to Router at http://127.0.0.1:49605
2026-01-10 00:46:40,790 - INFO - gke_code_executor.py:156 - Executing code in sandbox:

def fibonacci(n):
  """
  This function calculates the nth Fibonacci number.

  Args:
    n: The index of the Fibonacci number to calculate (starting from 0).

  Returns:
    The nth Fibonacci number.
  """
  if n <= 0:
    return 0
  elif n == 1:
    return 1
  else:
    a = 0
    b = 1
    for _ in range(2, n + 1):
      c = a + b
      a = b
      b = c
    return b

# Get the first fibonacci number (index 0)
first_fibonacci = fibonacci(0) 

# Print the first fibonacci number
print(first_fibonacci)

2026-01-10 00:46:40,958 - INFO - sandbox_client.py:464 - File 'script.py' uploaded successfully.
2026-01-10 00:46:41,031 - INFO - sandbox_client.py:356 - Stopping port-forwarding...
2026-01-10 00:46:41,034 - INFO - sandbox_client.py:368 - Deleting SandboxClaim: sandbox-claim-4d0f82e0
```

Output: 

Screenshot 2026-01-09 at 4 48 10 PM

### Checklist

- [ yes] I have read the [CONTRIBUTING.md](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md) document.
- [yes] I have performed a self-review of my own code.
- [yes] I have commented my code, particularly in hard-to-understand areas.
- [yes] I have added tests that prove my fix is effective or that my feature works.
- [yes] New and existing unit tests pass locally with my changes.
- [yes] I have manually tested my changes end-to-end.
- [yes] Any dependent changes have been merged and published in downstream modules.

### Additional context

_Add any other context or screenshots about the feature request here._


src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -78,12 +82,17 @@ class GkeCodeExecutor(BaseCodeExecutor):
...
+  
+  # Sandbox constants
+  python_sandbox_template: str = "python-sandbox-template"
 
   _batch_v1: k8s.client.BatchV1Api
   _core_v1: k8s.client.CoreV1Api
 
   def __init__(
       self,
+      executor_type: str = "job",
```
vicentefb, line 103: We should add validation here to ensure `executor_type` is either 'job' or 'sandbox'. Currently, if a user makes a typo (e.g., 'Sandbox'), the code in `execute_code` will silently fall back to the else block and try to run it as a Job, which might be confusing.

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -136,11 +147,26 @@ def __init__(
...
+        ) as sandbox:
+            # Execute the code as a python script
+            logger.debug("Executing code in sandbox:\n```\n%s\n```", code)
+            sandbox.write("script.py", code)
+            result = sandbox.run("python3 script.py")
+
+            return CodeExecutionResult(stdout=result.stdout)
+    except Exception as e:
+        return CodeExecutionResult(
+            stderr=f"Sandbox execution failed: {str(e)}",
```
vicentefb: We should add a log statement here with exc_info=True. If the sandbox execution fails (e.g., connection timeout or internal sandbox error), we'll lose the stack trace, which makes debugging difficult.

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -78,12 +82,17 @@ class GkeCodeExecutor(BaseCodeExecutor):
 
   kubeconfig_path: str | None = None
   kubeconfig_context: str | None = None
+  
+  # Sandbox constants
+  python_sandbox_template: str = "python-sandbox-template"
```
vicentefb: Should we allow python_sandbox_template to be passed into __init__? If we ever want to support different template (e.g., a data-science specific image), we'll need to make this configurable.

Maybe i'm missing some context or forgetting something but yeah, why are we hardcoding the template name ?

tests/unittests/code_executors/test_gke_code_executor.py
```diff
@@ -225,3 +228,77 @@ def test_create_job_manifest_structure(self, mock_invocation_context):
     assert sec_context.allow_privilege_escalation is False
     assert sec_context.read_only_root_filesystem is True
     assert sec_context.capabilities.drop == ["ALL"]
+
```
vicentefb, line 237: We are missing a test for when SandboxClient throws an exception (e.g., connection failure) as well as input validation mentioned in the other file.

src/google/adk/code_executors/gke_code_executor.py
SHRUTI6991: I hardcoded it because, currently we only support python runtime. If we provide any runtime which doesn't support python. This code might fail.

igooch: Agreed on allowing the user to override the value. Whatever template they use must be available in their cluster, so even with this value hard-coded it can fail if not properly set up.

That being said I'd recommend being able to override the value as part of the `execute_code` rather than the init method. That way the executor can be used with multiple different templates (assuming that each template is valid and available in the cluster).

src/google/adk/code_executors/gke_code_executor.py
igooch: Could you modify this comment block to indicate the difference between Job vs Sandbox? In particular that Sandbox will need other resources available in the cluster (agent-sandbox controller, sandbox-template, client router and gateway, etc.) in order for this code to properly work.

The comment here doesn't need to be extensive. A more detailed update should be added to the docs https://github.com/google/adk-docs/blob/main/docs/tools/google-cloud/gke-code-executor.md after this PR is merged.

rohityan:
Hi @SHRUTI6991, Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Can you please fix the formatting errors.

src/google/adk/code_executors/gke_code_executor.py
SHRUTI6991: Sounds good. I prefer passing the value via init instead of `execute_code`. 

The reason being `execute_code` is an overridden method. Any changes to this method needs to be reflected in all the other code executors where sandbox_template doesn't make sense. Also, it keeps the calling code (the Agent) decoupled from the implementation details of the executor.

src/google/adk/code_executors/gke_code_executor.py
```diff
@@ -136,10 +163,28 @@ def __init__(
...
+          template_name=self.sandbox_template,
+          gateway_name=self.sandbox_gateway_name,
+          namespace=self.namespace,
+      ) as sandbox:
+        # Execute the code as a python script
+        logger.debug("Executing code in sandbox:\n```\n%s\n```", code)
+        sandbox.write("script.py", code)
+        result = sandbox.run("python3 script.py")
+
+        return CodeExecutionResult(stdout=result.stdout)
```
tomergee: Also pass stderr=result.stderr to raise any errors

rohityan:
Hi @SHRUTI6991 , can you please fix the failing unit tests.

pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
   # go/keep-sorted start
   "PyYAML>=6.0.2, <7.0.0",                                   # For APIHubToolset.
   "aiosqlite>=0.21.0",                                       # For SQLite database
+  "agentic_sandbox @ git+https://github.com/kubernetes-sigs/agent-sandbox.git@dbac66ecba5497ac493ca5e4ab5e0fcb1c945134#subdirectory=clients/python/agentic-sandbox-client", # For Agent Sandboxed Code Execution
```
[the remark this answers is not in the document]
rohityan: Is there a plan to transition this to a versioned release instead of a direct Git dependency to a specific commit?

rohityan:
Hi @SHRUTI6991 , looks like the unit test failures are not related to your code.

pyproject.toml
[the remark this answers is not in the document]
SHRUTI6991: Yes, we are going to publish the python client to PyPI. This is an interim solution.

SHRUTI6991:
@ryanaiagent I have addressed the comment. Ptal.

rohityan:
Hi @SHRUTI6991 , Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

sasha-gitg:
Review is continuing in a separate channel.

PR #4137, opened 2026-01-13
fix:Bug #4133: AutomaticFunctionCallingConfig Settings Ignored

### LINKED ISSUE #4133: Bug: `AutomaticFunctionCallingConfig(disable=True)` is ignored - Planner hook bypassed
# Bug: `AutomaticFunctionCallingConfig(disable=True)` is ignored - Planner hook bypassed

## Describe the bug
When setting `automatic_function_calling=AutomaticFunctionCallingConfig(disable=True)` in the `generate_content_config` of an `Agent`, the ADK ignores this configuration. The custom `Planner.process_planning_response()` hook is never called, even when using `maximum_remote_calls=0` as a workaround.

### 1. Install
```bash
pip install google-adk==1.22.1
```

### 2. Create a custom planner with logging (`planner.py`):
```python
import logging
from typing import List, Optional
from google.adk.planners import BuiltInPlanner
from google.adk.agents.callback_context import CallbackContext

# File logger for guaranteed visibility
logger = logging.getLogger("planner_debug")
file_handler = logging.FileHandler("planner_debug.log")
file_handler.setFormatter(logging.Formatter('%(asctime)s - %(message)s'))
logger.addHandler(file_handler)
logger.setLevel(logging.DEBUG)

class StrictLimitPlanner(BuiltInPlanner):
    def __init__(self, max_iterations: int, key_name: str, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.max_iterations = max_iterations
        self._state_key = f"_counter_{key_name}"

    def process_planning_response(
            self,
            callback_context: CallbackContext,
            response_parts: List[types.Part],
    ) -> Optional[List[types.Part]]:
        # THIS LOG NEVER APPEARS - proving the hook is bypassed
        logger.warning(f"[PLANNER CALLED] {self._state_key}")
        
        processed_parts = super().process_planning_response(callback_context, response_parts)
        return processed_parts if processed_parts else response_parts
```

### 3. Create an agent with AFC disabled (`agent.py`):
```python
from google.adk.tools.google_search_tool import GoogleSearchTool
from google.genai.types import GenerateContentConfig, AutomaticFunctionCallingConfig, ThinkingConfig
from planner import StrictLimitPlanner

strict_config = GenerateContentConfig(
    automatic_function_calling=AutomaticFunctionCallingConfig(
        disable=True,
        maximum_remote_calls=0  # Workaround that stops loop but still bypasses planner
    ),
)

agent = Agent(
    name="test_agent",
    model="gemini-2.0-flash-exp",
    tools=[GoogleSearchTool()],
    planner=StrictLimitPlanner(max_iterations=3, key_name="test", thinking_config=ThinkingConfig(include_thoughts=False)),
    generate_content_config=strict_config,
)

app = App(name="test_app", root_agent=agent)
```

### 4. Run and send a message that triggers tool use:
```bash
adk api_server
# Then send: "query ~~~~"
```

### 5. Check logs:
```bash
# Server log shows:
# WARNING - max_remote_calls in automatic_function_calling_config 0 is less than or equal to 0. Disabling automatic function calling.

# BUT planner_debug.log is EMPTY (0 bytes):
ls -la planner_debug.log
# -rw-r--r--  1 user  staff  0 Jan 13 11:39 planner_debug.log
```

## Expected behavior
1. `disable=True` should disable AFC without requiring `maximum_remote_calls=0`
2. Custom `Planner.process_planning_response()` should be called after each model response, regardless of AFC settings
3. The planner should be able to intercept, count, and modify tool calls

## Actual behavior
1. `disable=True` alone does nothing - logs still show "AFC is enabled with max remote calls: 10"
2. `maximum_remote_calls=0` stops the loop but **completely bypasses** the planner hook
3. `process_planning_response()` is never called (proven by empty log file)

## Desktop
- **OS**: macOS (Apple Silicon)
- **Python version**: 3.14.2
- **ADK version**: 1.22.1 (`pip show google-adk`)
- **google-genai version**: 1.56.0

## Model Information
- **Are you using LiteLLM**: No
- **Which model is being used**: gemini-2.0-flash-exp, gemini-3-flash-preview (Vertex AI backend)

### Workaround Status
- Setting `maximum_remote_calls=0` successfully stops the infinite loop (acting as a circuit breaker)
- **BUT** it still bypasses `Planner.process_planning_response()` entirely
- **Result**: Planners cannot count iterations, modify tool inputs, or inject thoughts

### Impact
- Custom planners are completely non-functional when tools are present
- Cost control via iteration limits is impossible at the planner level
- The documented Planner API is effectively broken for agents with tools

### Suggested Fix
The `Agent` class should:
1. Respect `disable=True` in `AutomaticFunctionCallingConfig`
2. Always call `Planner.process_planning_response()` after each model response, regardless of AFC settings

sajanlamsal:

- Closes: #4133

**2. Description of change:**

**Problem:**

The Automatic Function Calling (AFC) loop ignored `AutomaticFunctionCallingConfig`:
- `disable=True` had no effect and AFC continued running
- `maximum_remote_calls` was not enforced
- LLMs could persistently return function calls after the limit, causing infinite loops and no final user response

**Solution:**

This PR implements two fixes:

1. **AFC Configuration Enforcement**
   - `disable=True` now prevents AFC execution and stops further AFC iterations
   - `maximum_remote_calls` is enforced in both `run_async()` and `run_live()` modes
   - Pre-execution checks prevent function execution once the limit is reached
   - Event counting semantics: one LLM response with multiple parallel function calls counts as **one** toward `maximum_remote_calls`

2. **Guardrail Mechanism (Async Mode)**
   - Prevents infinite loops when LLMs keep returning function calls after the limit
   - Triggers after 3 consecutive refused function-call events
   - Performs one final LLM call with:
     - Tools removed
     - AFC disabled
     - A system instruction forcing a text-only response
   - Guarantees loop termination and a final user-visible response

**Real-world example:**  
An LLM searches for "unicorn startups in Antarctica". The tool returns empty results, but the LLM keeps retrying with variations. Without the guardrail: infinite loop. With the guardrail: after 3 refused calls, a final response such as "I couldn't find any unicorn startups in Antarctica" is returned and the loop ends.

### Testing Plan

**Unit Tests:**

**Coverage:**
- 32 tests across 2 new files
  - `test_afc_config.py` (20 tests): AFC disable flag and `maximum_remote_calls` enforcement in async and live modes
  - `test_guardrail.py` (12 tests): guardrail triggering, tool removal, AFC disabling, and final-response enforcement

**Results:**
```bash
pytest tests/unittests/flows/llm_flows/test_afc_config.py -v
# 20 passed

pytest tests/unittests/flows/llm_flows/test_guardrail.py -v
# 12 passed

pytest tests/unittests/flows/llm_flows/ -q
# 329 passed

pytest tests/unittests/ -v
# 3,832 passed (no regressions)
```

**Manual End-to-End Tests:**

**Scenario 1: `disable=True`**
```
WARNING - AFC is disabled. Stopping AFC loop.
INFO - Generated 2 events (no function calls executed)
```
✅ PASS

**Scenario 2: `maximum_remote_calls=3`**
```
INFO - Iteration 1: Executing function call
INFO - Iteration 2: Executing function call
INFO - Iteration 3: Executing function call
INFO - Generated 7 events (exactly 3 FC events, then stopped)
```
(Previously: 20+ events due to infinite loop)
✅ PASS

**Scenario 3: Guardrail triggered (persistent FCs)**
```
DEBUG - LLM returned FCs after limit (1/3)
DEBUG - LLM returned FCs after limit (2/3)
DEBUG - LLM returned FCs after limit (3/3)
INFO  - Guardrail triggered: forcing final text response
INFO  - Tools removed, AFC disabled
INFO  - LLM returned final text response
INFO  - Loop terminated
```
✅ PASS

### Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md).
- [x] I have commented complex logic.
- [x] I have added tests that prove the fix is effective.
- [x] New and existing unit tests pass locally.
- [x] I have manually tested the changes end-to-end.
- [x] Any dependent changes have been merged and published.

### Additional context

- `maximum_remote_calls` counts LLM response events containing function calls, not individual tool invocations.
- Guardrail logic is applied only in async mode; live mode enforces pre-execution limits only.
- No public API changes; this is a backward-compatible bug fix.
- Performance impact is negligible and reduces wasted LLM calls by preventing runaway loops.
- **Reviewer focus:** `_should_stop_afc_loop()` semantics and guardrail cleanup in `run_async()`.


rohityan:
Hi @sajanlamsal, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:
Hi @ankursharmas , can you please review this.

akdong91:
@sajanlamsal I have been poking this PR and am testing in local environment. It seems that some of unittests are failing. For example, `test_afc_disabled_stops_loop` is failing at the first `assert` statement: assert call_count == 0. Do you have any thought of why it is failing?

In 'test_afc_disabled_stops_loop'  response setting, two responses are fed into the mock model. The first one was supposed to be processed but the second one shouldn't. If I understand the test case correctly, isn't 'call_count' supposed to be assserted to 1 instead of 0? Because the first function is called and 'call_count' is updated to 1. You can correct me if I am missing anything.

sajanlamsal:
@akdong91 

Thank you for the question
The assertion assert call_count == 0 is correct. Here's why:
When disable=True is set, no function calls should execute at all - not even the first one.

What happens:
1. First LLM call returns a response with function calls
2. Before executing those calls, _postprocess_async() checks _should_stop_afc_loop() with count_current_event=True
3. Since disable=True, it returns early without executing the function calls
4. call_count stays at 0
5. The loop then checks and breaks due to disable=True


**Why two mock responses?**
The test provides two responses as defensive programming—if the implementation incorrectly continues the loop despite disable=True, the second response would be used and expose the bug. With correct implementation, only the first response is used, and the loop exits immediately.

If you're seeing call_count == 1:

The pre-execution check in _postprocess_async() might not be present or working correctly. Can you verify:
The check exists around line 1064 in base_llm_flow.py:

```
should_stop, _ = _should_stop_afc_loop(
    llm_request,
    invocation_context, 
    model_response_event,
    count_current_event=True,  # Pre-execution check
)
if should_stop:
    return  # Prevents FC execution
```

Run with verbose output: pytest tests/unittests/flows/llm_flows/test_afc_config.py::test_afc_disabled_stops_loop -vv -s
currently test passes in my environment (20/20 AFC tests, 12/12 guardrail tests all passing), so there may be a code sync issue??

akdong91:
@sajanlamsal Thanks for explaining in detail.
Somehow, my local repo was pointing to the old copy of adk installed in the local environment. Yes, you were right. It was a code sync issue. All test passed in my environment too!!! 

I can see how AFC(disable=True) comes into play by stepping into the execution: runner -> base_agent ->llm_agent ->base_llm_flow where `pre_execution` checks in '_postprocess_async()`.

PR #4188, opened 2026-01-17
Fix infinite loop in tool calling with FunctionCallingConfig mode='AN…

### LINKED ISSUE #4179: FunctionCallingConfig(mode="ANY") causes infinite tool-calling loop when sub-agent is used as a tool (google-adk 1.21.0)
Context

We are using a sub-agent as a tool inside a larger agent orchestration flow.

This sub-agent must always rely on an external structured source as its single source of truth. However, in some cases, the sub-agent produces responses without calling its tool, resulting in hallucinated outputs being returned to the root agent.

To prevent this, we attempted to force tool usage via GenerateContentConfig and FunctionCallingConfig.

tool_config = types.ToolConfig(
    function_calling_config=types.FunctionCallingConfig(
        mode="ANY",
        allowed_function_names=["external_search_tool"]
    )
)

SubAgent = LlmAgent(
    model=MODEL_NAME,
    name="SubAgent",
    description="Sub-agent used as a tool that must always rely on an external source.",
    instruction=return_agent_instruction(),
    tools=[external_search_tool],
    generate_content_config=types.GenerateContentConfig(
        tool_config=tool_config,
    ),
    planner=BuiltInPlanner(
        thinking_config=types.ThinkingConfig(
            include_thoughts=True,
            thinking_budget=1024,
        )
    )
)

Expected Behavior
	•	The sub-agent should call the tool at least once before responding.
	•	After receiving the tool result, it should produce a final response and exit.
	•	The sub-agent should not be able to answer without using the tool.

Observed Behavior
	•	When mode="ANY" is enabled, the sub-agent enters an infinite loop of tool calls.
	•	The same tool is repeatedly invoked, even when the previous result is sufficient.
	•	No final response is produced.
	•	Removing FunctionCallingConfig avoids the loop, but the original hallucination issue returns.

Environment
	•	google-adk == 1.21.0

Questions
	•	Is this infinite loop expected behavior when using mode="ANY" with sub-agents used as tools?
	•	Is there a recommended way in ADK to:
	•	enforce mandatory tool usage
	•	limit the number of tool calls
	•	guarantee a single tool call followed by a final response?

jayy-77:
…Y' (Issue #4179)

This commit fixes the infinite loop bug where agents with FunctionCallingConfig mode='ANY' could continuously call tools without ever providing a final response, causing the agent to hang indefinitely.

Fixes #4179

- Closes: #4179

### DIFF
--- a/src/google/adk/agents/invocation_context.py
+++ b/src/google/adk/agents/invocation_context.py
@@ -46,6 +46,10 @@ class LlmCallsLimitExceededError(Exception):
   """Error thrown when the number of LLM calls exceed the limit."""
 
 
+class ToolIterationsLimitExceededError(Exception):
+  """Error thrown when the number of tool iterations exceed the limit."""
+
+
 class RealtimeCacheEntry(BaseModel):
   """Store audio data chunks for caching before flushing."""
 
@@ -76,6 +80,9 @@ class _InvocationCostManager(BaseModel):
   _number_of_llm_calls: int = 0
   """A counter that keeps track of number of llm calls made."""
 
+  _number_of_tool_iterations: int = 0
+  """A counter that keeps track of consecutive tool calling iterations in current agent call."""
+
   def increment_and_enforce_llm_calls_limit(
       self, run_config: Optional[RunConfig]
   ):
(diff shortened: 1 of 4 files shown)


rohityan:
Hi @jayy-77 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you fix the formatting errors before we can proceed with the review. You can use autoformat.sh.

sasha-gitg:
Closing due to lack of response from the author.

PR #4200, opened 2026-01-19
feat: durable runtime support

marcusmotill:
**Problem:**
The current implementation of `SessionService`, `Event`, and other core components relies directly on `time.time()` and `uuid.uuid4()`. These standard library functions are non-deterministic, which prevents the ADK from being used in deterministic execution environments (like Temporal workflows). In such environments, logic that depends on time or random IDs must be replayable and consistent across executions.

**Solution:**
Introduced `google.adk.platform.time` and `google.adk.platform.uuid` modules to abstract time and UUID generation.
- Created `google.adk.platform.time` with `get_time()` and `set_time_provider()`.
- Created `google.adk.platform.uuid` with `new_uuid()` and `set_id_provider()`.
- Updated `Event`, `SessionService`, `InvocationContext`, and `SqliteSessionService` to use these new platform abstractions instead of `time` and `uuid` directly.

This allows runtimes to inject deterministic providers (e.g., Temporal's `workflow.now()` and side-effect-safe UUIDs) when running in a deterministic context, while defaulting to standard `time` and `uuid` for standard execution.

### Testing Plan

**Unit Tests:**

Added new unit tests for the platform modules:
- [tests/unittests/platform/test_time.py](cci:7://file:///usr/local/google/home/marcusmotill/Documents/code/temporal/adk-python-temporal/tests/unittests/platform/test_time.py:0:0-0:0): Verifies `get_time`, provider overriding, and resetting.
- [tests/unittests/platform/test_uuid.py](cci:7://file:///usr/local/google/home/marcusmotill/Documents/code/temporal/adk-python-temporal/tests/unittests/platform/test_uuid.py:0:0-0:0): Verifies `new_uuid`, provider overriding, and resetting.
- Updated [tests/unittests/artifacts/test_artifact_service.py](cci:7://file:///usr/local/google/home/marcusmotill/Documents/code/temporal/adk-python-temporal/tests/unittests/artifacts/test_artifact_service.py:0:0-0:0) to match the new patterns.

Verified that the ADK continues to function correctly in standard (non-deterministic) environments, ensuring the default providers for time and UUIDs work as expected.
- Ran existing agent workflows to confirm session creation and event logging still produce valid timestamps and IDs.

### Checklist

### Additional context

These changes are a prerequisite for the [Temporal integration](https://github.com/temporalio/sdk-python/pull/1282), allowing the ADK to run safely inside Temporal workflows without breaking determinism guarantees.

### DIFF
--- a/tests/unittests/platform/__init__.py
+++ b/tests/unittests/platform/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.


boyangsvl:
Thank you for the PR! Could you use a different name than 'runtime'? runtime is too general and may be confused with something else. Or you don't need a wrapper for time and id provider and make them top level.

src/google/adk/platform/time.py
```diff
@@ -0,0 +1,43 @@
...
+"""Platform module for abstracting system time generation."""
+
+import time
+from typing import Callable
+
+_default_time_provider: Callable[[], float] = time.time
+_time_provider: Callable[[], float] = _default_time_provider
+
+
+def set_time_provider(provider: Callable[[], float]) -> None:
```
sasha-gitg, line 29: It seems like these should be behind a threading.Lock() since these are altering global mutable state.

src/google/adk/agents/invocation_context.py
```diff
@@ -409,4 +410,4 @@ def _find_matching_function_call(
 
 
 def new_invocation_context_id() -> str:
-  return "e-" + str(uuid.uuid4())
+  return "e-" + platform_uuid.new_uuid()
```
sasha-gitg: There are multiple uuid.uuid4() and time.time() call sites throughout the SDK that are not changed in this PR. Is that intentional?

src/google/adk/events/event.py
```diff
@@ -18,6 +18,8 @@
 from typing import Optional
 import uuid
```
sasha-gitg: Do we still need to import this?

tests/unittests/platform/test_time.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2026 Google LLC
```
sasha-gitg, line 1: test/unittests/platform should have an `__init__.py` file

src/google/adk/platform/time.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2026 Google LLC
```
sasha-gitg, line 1: Slight preference to pick module names that don't shadow stdlib names in `.../platform`

src/google/adk/platform/time.py
```diff
@@ -0,0 +1,43 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Platform module for abstracting system time generation."""
+
+import time
+from typing import Callable
+
+_default_time_provider: Callable[[], float] = time.time
```
sasha-gitg, line 23: A single global value can't serve multiple execution contexts in the same process, also the case for async contexts. Preference to use ContextVar. It also remove the need for a lock.

Also, Temporal's Python SDK uses contextvars.

src/google/adk/agents/invocation_context.py
marcusmotill: yes, the intention is to only change the things we need to limit the PR. While we _could_ swap out all instances, we don't need to. Now the real question is, did I get all the ones we need? I am not positive but this batch is a good start for the most common paths I see.

open to taking more of a blanket approach and change more than we need, open to thoughts...

marcusmotill:
@sasha-gitg contextvar and hygiene changes made, if you have preference on naming (platform) feel free to pick a one and I can reflect.. this is our second name iteration from runtime and have low energy on that decision 😄

src/google/adk/agents/invocation_context.py
sasha-gitg: Having two patterns side-by-side is a maintenance trap. A new contributor won't know which to use, and nobody will want to audit whether their code is "on the workflow path" before picking one.

A single rule is much easier to follow than "use them only if your code might run in a Temporal workflow." The cost of converting the remaining sites is low and it eliminates the ambiguity entirely.

Honestly it would be best to
  1. Convert all remaining call sites in this PR
  2. Add to create a linting rule to prevent drift (this can be a follow up)

marcusmotill: I don't disagree, will do a further audit of other locations and look into a linting rule 👍

tests/unittests/sessions/test_session_service.py
```diff
@@ -544,6 +544,7 @@ async def test_append_event_complete(session_service):
       ),
       citation_metadata=types.CitationMetadata(),
       custom_metadata={'custom_key': 'custom_value'},
+      timestamp=1700000000.123,
```
marcusmotill, line 572: easiest solution to avoid nasty percision issues in other tests when saving a timestamp to the DB... basically our new global time (time.time) gave 1 more decimal of precision than `datetime.now().timestamp()` and this issue surfaced

marcusmotill: open to other ideas

src/google/adk/agents/invocation_context.py
marcusmotill: This should be resolved, suggestion on the mechanism for lint rule?

src/google/adk/agents/invocation_context.py
```diff
@@ -413,4 +414,4 @@ def _find_matching_function_call(
 
 
 def new_invocation_context_id() -> str:
-  return "e-" + str(uuid.uuid4())
+  return "e-" + str(cast(str, platform_uuid.new_uuid()))
```
sasha-gitg: Why do we need to cast to str twice?

sasha-gitg:
Please resolve the failing checks.

PR #4214, opened 2026-01-20
Update bug_report.md

ymodak:
Updating github bug template as per new intake processs


rohityan:
Hi @sasha-gitg , can you please review this. LGTM.

PR #4356, opened 2026-02-02
fix: add function_response check to _part_has_payload

filipecaixeta:
## Summary

- Add `function_response` check to `_part_has_payload` function in `lite_llm.py`
- Fixes issue where `_append_fallback_user_content_if_missing` was adding the fallback message "Handle the requests as specified in the System Instruction." even when a valid function response was present in user content
- Added comprehensive unit tests for both `_part_has_payload` and `_append_fallback_user_content_if_missing` with function_response scenarios

## Problem

The `_part_has_payload` function checks whether a `Part` contains usable payload for the model, but it was missing a check for `function_response`. This caused the fallback user content to be injected unnecessarily when a user message contained only a function response, leading to:
- Extra tokens being added to requests
- Potential confusion in model responses
- Incorrect behavior in function call/response flows

## Changes

**`src/google/adk/models/lite_llm.py`:**
```python
if part.function_response and part.function_response.response:
    return True
```

**`tests/unittests/models/test_litellm.py`:**
- Added `TestPartHasPayload` class with 6 unit tests
- Added `TestAppendFallbackUserContentIfMissing` class with 5 unit tests

## Test plan

- [x] All new unit tests pass (11 tests)
- [x] Existing related tests pass (`test_generate_content_async_adds_fallback_user_message`, `test_maybe_append_user_content`)
- [x] Tests cover edge cases: empty response, valid response, text content, empty parts

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -470,6 +470,8 @@ def _part_has_payload(part: types.Part) -> bool:
     return True
   if part.file_data and (part.file_data.file_uri or part.file_data.data):
     return True
+  if part.function_response and part.function_response.response:
+    return True
   return False
 
 
--- a/tests/unittests/models/test_litellm.py
+++ b/tests/unittests/models/test_litellm.py
@@ -24,6 +24,7 @@ from unittest.mock import AsyncMock
 from unittest.mock import Mock
 import warnings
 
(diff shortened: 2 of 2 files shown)


filipecaixeta:
I'm closing this PR since someone was faster and solved it already

Issue #1733, opened 2025-07-01
The last_update_time provided in the session object '2025-07-01 18:59:42' is earlier than the update_time in the storage_session '2025-07-01 18:59:56'. Please check if it is a stale session


code2tan:
**Describe the bug**
```json
{"error": "The last_update_time provided in the session object '2025-07-01 18:59:42' is earlier than the update_time in the storage_session '2025-07-01 18:59:56'. Please check if it is a stale session."}
```

main.py
```python
session_mysql_url = os.getenv('SESSION_MYSQL_URL')

app = get_fast_api_app(
    agents_dir="./sql_agent",
    session_service_uri=session_mysql_url,
    allow_origins=["*"],
    web=False,
    trace_to_cloud=False,
)

if __name__ == "__main__":
    import uvicorn

    config = uvicorn.Config(
        "main:app",
        log_level="info",
        port=8000,
        reload=True,
        # Disable Uvicorn's default logging config
        log_config=None
    )
    server = uvicorn.Server(config)
    server.run()
```

**Desktop (please complete the following information):**
 - OS:  
 - Python version(python -V): 3.12
 - ADK version(pip show google-adk): 1.4.1
 - uvicorn: 0.34.3

 **Model Information:**
 - qwen-max model

veevortexiq:
Same problem..Happens when I use MySQL DB for session service uri

xuanyang15:
Thanks for reporting this bug! Could you please provide the following information? which should be helpful for me to understand the problem
1. Does your mysql db has a different timezone from your environment where you run the agent?
2. Does the same issue happens when using different databases, e.g. sqlite?

code2tan:
> 1. Does your mysql db has a different timezone from your environment where you run the agent?
```mysql
SELECT @@global.time_zone;
output：+08:00

SELECT @@session.time_zone;
outpu: +08:00

SELECT @@system_time_zone;
outpu: CST
```

I used Alibaba Cloud's RDS database service, which is based on MySQL.

```mysql
SELECT VERSION();
outpu: 5.7.44-log
```

xuanyang15:
Thanks for the additional information! Is it possible for you to share a minimal runnable example to reproduce it?

xuanyang15:
I tried with the following simple example and I cannot reproduce the error, so I suspect the error you saw was caused by some race conditions, for example, your application might be processing multiple events for the same session concurrently, leading to a race condition where an older session object is used to update the state. 

[...]

Issue #1848, opened 2025-07-09
Session lastUpdateTime doesn't handle timezone correctly


xuanyang15:
**Describe the bug**
Similar to #1180 (which is for SQLite), for other databases like mysql or postgresql, if the database server is using a different timezone than the adk app server's timezone. The lastUpdateTime we get in adk for a session will have an offset compared to the real time.

For example, if the adk app server's local timezone is PDT (UTC-7) and mysql server's timezone is UTC:
1. We create a session at 2025-07-09 9:30 PDT
2. Mysql server store it as 2025-07-09 16:30 UTC
3. ADK get session will return it as 2025-07-09 17:30 **PDT** (which is actually 2025-07-08 23:30 UTC)

**To Reproduce**
1. Start the mysql server which has a UTC timezone (or any other timezone different from your local timezone)
2. Call DatabaseSessionService to create a new session at time T1
3. Call DatabaseSessionService to get the created session and check its lastUpdateTime which is T2
4. You can see T1 and T2 do not equal (i.e. there is an offset equals to timezone difference)

Example code:
```python
import asyncio
from datetime import datetime, timezone

from google.adk.sessions import DatabaseSessionService

APP_NAME = "test_app"
USER_ID = "test_user"

db_url = "mysql+pymysql://root:0@127.0.0.1:3306/test_session"
session_service = DatabaseSessionService(db_url=db_url)


async def main_async():
  now = datetime.now()
  print(
      f"Current local time is {now.timestamp()} ({now}), UTC time is:"
      f" {now.astimezone(timezone.utc)}"
  )

  session = await session_service.create_session(
      app_name=APP_NAME,
      user_id=USER_ID,
  )
  session_id = session.id
  last_update_time = session.last_update_time
  print(
      f"Created new session: {session_id}, with last update time"
      f" {last_update_time} ({datetime.fromtimestamp(last_update_time)})"
  )

  got_session = await session_service.get_session(
      app_name=APP_NAME, user_id=USER_ID, session_id=session.id
  )
  last_update_time = got_session.last_update_time
  print(
      "Got last update time"
      f" {last_update_time} ({datetime.fromtimestamp(last_update_time)}), UTC"
      f" time is: {datetime.fromtimestamp(last_update_time, tz=timezone.utc)}"
  )


if __name__ == "__main__":
  asyncio.run(main_async())

```

**Expected behavior**
[...]

PR #4365, opened 2026-02-03
fix: Strip timezone for PostgreSQL timestamps in DatabaseSessionService

filipecaixeta:
## Summary
- Fixes `DataError` when using PostgreSQL with `asyncpg` for session storage
- PostgreSQL's default `TIMESTAMP` type is `WITHOUT TIME ZONE`, which cannot accept timezone-aware datetime objects
- The existing code handled this for SQLite but not PostgreSQL - this fix applies the same timezone stripping

## Error
When creating a session with PostgreSQL + asyncpg, the following error occurs:

```
sqlalchemy.dialects.postgresql.asyncpg.Error: <class 'asyncpg.exceptions.DataError'>: 
invalid input for query argument $5: datetime.datetime(2026, 2, 3, 21, 32, 50, 353909, 
tzinfo=datetime.timezone.utc) (can't subtract offset-naive and offset-aware datetimes)
```

During the INSERT:
```sql
INSERT INTO sessions (app_name, user_id, id, state, create_time, update_time) 
VALUES ($1, $2, $3, $4, $5, $6)
```

Where `$5` and `$6` are timezone-aware datetimes being inserted into `TIMESTAMP WITHOUT TIME ZONE` columns.

## Root Cause
Commit 1063fa53 changed from database-generated timestamps (`func.now()`) to explicit Python datetimes (`datetime.now(timezone.utc)`). The SQLite case was handled by stripping the timezone, but PostgreSQL was overlooked.

## Test plan
- [x] Verified fix resolves the error when creating sessions with PostgreSQL + asyncpg
- [ ] Existing unit tests pass

Fixes regression from #1733

### DIFF
--- a/src/google/adk/sessions/database_session_service.py
+++ b/src/google/adk/sessions/database_session_service.py
@@ -373,7 +373,9 @@ class DatabaseSessionService(BaseSessionService):
 
       # Store the session
       now = datetime.now(timezone.utc)
-      if self.db_engine.dialect.name in (_SQLITE_DIALECT, _POSTGRESQL_DIALECT):
+      is_sqlite = self.db_engine.dialect.name == _SQLITE_DIALECT
+      is_postgresql = self.db_engine.dialect.name == _POSTGRESQL_DIALECT
+      if is_sqlite or is_postgresql:
         now = now.replace(tzinfo=None)
 
       storage_session = schema.StorageSession(


rohityan:
Hi @filipecaixeta , Thank you for your contribution! We appreciate you taking the time to submit this pull request.

rusherman:
Hi, thanks for this fix — we hit the exact same issue with `asyncpg` + PostgreSQL.

One thing worth noting: while this PR fixes the `create_session` path, there's a related **timezone semantic inconsistency** in `append_event` (around line 516-521):

```python
if is_sqlite:
    update_time = datetime.fromtimestamp(
        event.timestamp, timezone.utc
    ).replace(tzinfo=None)
else:
    update_time = datetime.fromtimestamp(event.timestamp)
```

The `else` branch (which covers PostgreSQL) uses `datetime.fromtimestamp(event.timestamp)` **without a timezone argument**, so it produces a naive datetime in the **local timezone**. The SQLite branch produces a naive datetime in **UTC**.

This doesn't cause an `asyncpg` error (since the result is already naive), but it means the same `event.timestamp` gets stored as different wall-clock values depending on the database backend — and on the server's local timezone. On a non-UTC server:

- SQLite: stores `2026-02-06 06:48:16` (UTC)
- PostgreSQL: stores `2026-02-06 14:48:16` (local, e.g. UTC+8)

The `create_session` path (which this PR fixes) had the same kind of issue — `datetime.now(timezone.utc)` produces UTC, so after `.replace(tzinfo=None)` the stored value is UTC-based. It might be worth making `append_event` consistent too:

```python
# Consistent UTC for all backends
update_time = datetime.fromtimestamp(event.timestamp, timezone.utc)
if self.db_engine.dialect.name in ("sqlite", "postgresql"):
    update_time = update_time.replace(tzinfo=None)
```

vietnamesekid:
Hi @filipecaixeta.

I have a related PR #4388 that takes a different approach fixing this at the schema level by using `postgresql.TIMESTAMP(timezone=True)` in `PreciseTimestamp.load_dialect_impl()`

I'd like to share some analysis on why I think stripping timezone (`replace(tzinfo=None)`) may not be sufficient for PostgreSQL, and why a schema-level fix might be more appropriate

### The `append_event` inconsistency

As @rusherman also pointed out, there's a semantic inconsistency in `append_event`

```python
if is_sqlite:
    update_time = datetime.fromtimestamp(
        event.timestamp, timezone.utc
    ).replace(tzinfo=None)
else:
    update_time = datetime.fromtimestamp(event.timestamp)  # uses LOCAL timezone
```

For the else branch (PostgreSQL/MySQL):
- `datetime.fromtimestamp(event.timestamp)` return a naive datetime in the app server's local timezone
- This gets stored as-is into a `TIMESTAMP WITHOUT TIME ZONE` column
- If the app server is in PDT (UTC-7) and the DB server is in UTC, the stored value is off by 7 hours

This is exactly the bug described in #1848 

### Why `replace(tzinfo=None)` doesn't fully solve this?
Even if we add replace(tzinfo=None) for PostgreSQL in create_session, the append_event path still uses datetime.fromtimestamp(event.timestamp) which produces local-timezone naive datetimes. We'd need to fix that too, and at that point we're stripping timezone info from UTC datetimes just to store them in a timezone-unaware column fighting against the database rather than working with it.

### Schema-level approach (PR #4388)

My PR instead makes PreciseTimestamp return postgresql.TIMESTAMP(timezone=True) for PostgreSQL, so the column type is TIMESTAMPTZ. This means:

- create_session: datetime.now(timezone.utc) is stored correctly with timezone info preserved
- append_event: Even naive datetimes are interpreted as the DB server's timezone (typically UTC), which is more predictable than depending on the app server's local timezone
- Reading back: PostgreSQL always returns timezone-aware datetimes, so update_time.timestamp() computes correctly

The tradeoff is that existing PostgreSQL users need a one-time migration

```sql
ALTER TABLE sessions ALTER COLUMN create_time TYPE TIMESTAMP WITH TIME ZONE;
-- (and similar for other timestamp columns)
```

### Ideally
The best fix would combine both approaches
1. Schema-level: TIMESTAMP WITH TIME ZONE for PostgreSQL (my PR #4388)
2. Code-level: Ensure append_event also uses UTC-aware datetimes consistently:
```python
update_time = datetime.fromtimestamp(event.timestamp, timezone.utc)
```

This way all timestamps are explicitly UTC throughout the entire pipeline, regardless of the app server's timezone.
What do you think? Happy to coordinate on this

filipecaixeta:
@vietnamesekid @rusherman I'll close this PR and we can wait for yours to be merged, though I’ve lost a bit of faith in how this repo is being maintained. It’s incredibly frustrating when breaking changes are merged without basic testing. Honestly, why was this even allowed to break?
I've been waiting two weeks for another PR to be merged already, so I have zero expectation that this will be handled by the end of the month.

mynewestgitaccount:
@filipecaixeta your PR is the better one. You should reopen it. re: https://github.com/google/adk-python/pull/4388#issuecomment-3859066054

vietnamesekid:
> @vietnamesekid @rusherman I'll close this PR and we can wait for yours to be merged, though I’ve lost a bit of faith in how this repo is being maintained. It’s incredibly frustrating when breaking changes are merged without basic testing. Honestly, why was this even allowed to break? I've been waiting two weeks for another PR to be merged already, so I have zero expectation that this will be handled by the end of the month.

@filipecaixeta Thanks a ton for #4365, really appreciate the quick fix with zero migration hassle. It's honestly super helpful when production is breaking like mine right now.
Im still getting crashes every day because of this, so it'd be great if the maintainers could take a look at both PRs and figure out what makes the most sense (merge one first, go hybrid, whatever fits best).
You nailed the quick and reliable part seriously, thank you. Happy to jump in and help test more if needed

aidanandrews22:
+1

Here is a minimal monkey patch that fixes it. Just add this at the top of an early imported file (before you use any adk stuff):
# Patch: use a datetime subclass so we don't touch immutable built-in
from datetime import datetime as _RealDateTime, timezone
import google.adk.sessions.database_session_service as _db_session_mod

class _NaiveUTCDateTime(_RealDateTime):
    @classmethod
    def now(cls, tz=None):
        r = _RealDateTime.now(tz or timezone.utc)
        return r.replace(tzinfo=None) if r.tzinfo else r

_db_session_mod.datetime = _NaiveUTCDateTime

DeanChensj:
Thanks for this PR!  Reopened and will merge

Issue #3343, opened 2025-10-30
Update database schema automatically


flansen:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Is your feature request related to a problem? Please describe.**

We have setup our google-adk application with a database using DatabaseSessionService in a cloud deployment with a PostgreSQL database.

There were two adk version updates (1.14.0 and 1.17.0) that enforced schema changes.
We as the users are obliged to manually update the schema and have to catch updates to existing tables ourselves.

This is also stated in 1.14.0 Release Notes:

```markdown
NOTE: This requires DB migration, run ALTER TABLE events ADD COLUMN custom_metadata JSON; to migrate existing database tables.
```

With 1.17.0, the required changes were not even mentioned and I had to manually find out what changes to the DB schema are necessary.

The existing migration mechanisms - to my understanding - does only create new tables.
It does not update existing tables by either editind or deleting columns.

**Describe the solution you'd like**

The db schema should be automatically updated by the google-adk.
That must be done on startup, but only if a DB is configured for sessions.

**Describe alternatives you've considered**

%

**Additional context**

Our solution looks like this:

```python
if session_db_config is None:
    logger.info("using no database")
    session_db_url = None
else:
    try:
        ensure_schema_created(session_db_config)
    except Exception as e:
        logger.error(f"Error creating schema: {e}, ignoring and continuing...", exc_info=e)
    try:
        migrate_if_needed(session_db_config)
    except Exception as e:
        logger.error(f"Error migrating tables: {e}, ignoring and continuing...", exc_info=e)
    session_db_url = session_db_config.get_url(schema=SCHEMA_NAME)

app = get_fast_api_app(
    agents_dir=str(agents_dir),
    web=True,
    session_service_uri=session_db_url,
)

def migrate_if_needed(db_config: DBConfig):
    engine = create_engine(db_config.get_url(schema=SCHEMA_NAME))
    with engine.begin() as connection:
        # required for 1.14.0 migration
        # https://github.com/google/adk-python/releases/tag/v1.14.0
        connection.execute(text("ALTER TABLE sessions.events ADD COLUMN IF NOT EXISTS custom_metadata JSONB"))
        # required for 1.17.0 migration
        # https://github.com/google/adk-python/commit/6ab1498aa0bad783e17e97a2ceede97b4bd3016f
[...]

Issue #917, opened 2025-05-24
getting an error when using DatabaseSessionService


rajib76:
Below is my code 
```
import asyncio
from typing import Any

from dotenv import load_dotenv
from google.adk.agents import Agent
from google.adk.models.lite_llm import LiteLlm
from google.adk.runners import Runner
from google.adk.tools import LongRunningFunctionTool
from google.adk.sessions import InMemorySessionService, DatabaseSessionService
from google.adk.events import Event
from google.genai import types

load_dotenv()
APP_NAME = "demo_app"
USER_ID = "user1"
SESSION_ID = "session1"
# session_service = InMemorySessionService()
db_url = "sqlite:///./my_agent_data.sqlite"
session_service = DatabaseSessionService(db_url=db_url)
session = session_service.create_session(app_name=APP_NAME,
                                         user_id=USER_ID,
                                         session_id=SESSION_ID)
MODEL_GPT_4O = "openai/gpt-4o"
MODEL_GPT_4O_LITE_LLM = LiteLlm(model=MODEL_GPT_4O)


# Step 1: Define a long-running function
def ask_for_manager(task: str) -> dict[str, Any]:
    return {
        'status': 'pending',
        'task': task,
        'manager': 'Taylor',
        'ticket_id': 'task-123'
    }


# Step 2: Wrap the function
approval_tool = LongRunningFunctionTool(func=ask_for_manager)

# Step 3: Create agent
agent = Agent(
    model=MODEL_GPT_4O_LITE_LLM,
    name="approval_agent",
    instruction="Ask manager approval for any important task using the tool.",
    tools=[approval_tool]
)

# Step 4: Setup runner and session

runner = Runner(agent=agent, app_name=APP_NAME, session_service=session_service)


# Step 5: Define helper functions
def get_long_running_call(event: Event):
    if not event.long_running_tool_ids or not event.content:
        return
    for part in event.content.parts:
        if part and part.function_call and part.function_call.id in event.long_running_tool_ids:
            return part.function_call


def get_function_response(event: Event, call_id: str):
    if not event.content:
        return
    for part in event.content.parts:
        if part and part.function_response and part.function_response.id == call_id:
            print("function response ", part.function_response)
            return part.function_response


# Step 6: Interaction logic
async def interact(query: str):
    content = types.Content(role='user', parts=[types.Part(text=query)])
    events = runner.run_async(session_id=SESSION_ID, user_id=USER_ID, new_message=content)
    print(events)
    long_call = None
[...]

PR #4388, opened 2026-02-06
fix: use TIMESTAMP WITH TIME ZONE for PostgreSQL in PreciseTimestamp

vietnamesekid:
## Summary

The `PreciseTimestamp` TypeDecorator uses `DateTime` (without timezone) as its default implementation. When used with **PostgreSQL + asyncpg**, this causes the SQL query to cast parameters as `TIMESTAMP WITHOUT TIME ZONE`, which conflicts with timezone-aware datetime objects (those with `tzinfo=UTC`) that ADK now creates internally (since v1.24.0).

This results in the asyncpg error:
```
asyncpg.exceptions.DataError: invalid input for query argument $5: datetime.datetime(2026, 2, 6, 4, 16, 14,... 
(can't subtract offset-naive and offset-aware datetimes)

[SQL: INSERT INTO sessions (app_name, user_id, id, state, create_time, update_time) 
VALUES ($1::VARCHAR, $2::VARCHAR, $3::VARCHAR, $4::JSONB, $5::TIMESTAMP WITHOUT TIME ZONE, $6::TIMESTAMP WITHOUT TIME ZONE)]
```

## Associated Issue

Related to #1848

## Root Cause

- `PreciseTimestamp.load_dialect_impl()` only has special handling for MySQL (`mysql.DATETIME(fsp=6)`), falling back to plain `DateTime` for all other dialects including PostgreSQL.
- Plain `DateTime` maps to `TIMESTAMP WITHOUT TIME ZONE` in PostgreSQL.
- Since ADK v1.24.0, `DatabaseSessionService` creates datetime objects with `tzinfo=datetime.timezone.utc` (offset-aware).
- asyncpg strictly validates parameter types and rejects offset-aware datetimes for `TIMESTAMP WITHOUT TIME ZONE` columns.

## Fix

Add PostgreSQL-specific dialect implementation using `postgresql.TIMESTAMP(timezone=True)`, consistent with how MySQL already has its own dialect-specific handling.

## Testing Plan

- Ran the full session test suite (`tests/unittests/sessions/`) — **109 tests passed, 0 failed**
- SQLite tests remain unaffected since `PreciseTimestamp` only returns `TIMESTAMP(timezone=True)` when dialect is `postgresql`, SQLite falls back to default `DateTime`
- Manually tested against a live PostgreSQL database (DigitalOcean managed DB) with `postgresql+asyncpg://` connection string — session creation, event appending, and session retrieval all work correctly after the fix

## Verification

**Before fix** — session creation crashes:

```
sqlalchemy.exc.DBAPIError: (sqlalchemy.dialects.postgresql.asyncpg.Error)
<class 'asyncpg.exceptions.DataError'>: invalid input for query argument $5:
datetime.datetime(2026, 2, 6, 4, 16, 14,...
(can't subtract offset-naive and offset-aware datetimes)
[SQL: INSERT INTO sessions (app_name, user_id, id, state, create_time, update_time)
VALUES ($1::VARCHAR, $2::VARCHAR, $3::VARCHAR, $4::JSONB,
$5::TIMESTAMP WITHOUT TIME ZONE, $6::TIMESTAMP WITHOUT TIME ZONE)]
```

**After fix** — SQL now correctly uses `TIMESTAMP WITH TIME ZONE`:

```
[SQL: INSERT INTO sessions (app_name, user_id, id, state, create_time, update_time)
VALUES ($1::VARCHAR, $2::VARCHAR, $3::VARCHAR, $4::JSONB,
$5::TIMESTAMP WITH TIME ZONE, $6::TIMESTAMP WITH TIME ZONE)]
```

**Unit tests:**

$ python3 -m pytest tests/unittests/sessions/ -v
======================== 109 passed, 1 warning in 2.67s ========================

## Database Migration Required

Users with existing PostgreSQL databases need to run:
```sql
ALTER TABLE sessions ALTER COLUMN create_time TYPE TIMESTAMP WITH TIME ZONE;
ALTER TABLE sessions ALTER COLUMN update_time TYPE TIMESTAMP WITH TIME ZONE;
ALTER TABLE events ALTER COLUMN timestamp TYPE TIMESTAMP WITH TIME ZONE;
ALTER TABLE app_states ALTER COLUMN update_time TYPE TIMESTAMP WITH TIME ZONE;
ALTER TABLE user_states ALTER COLUMN update_time TYPE TIMESTAMP WITH TIME ZONE;
```

A Python migration script is also included: src/google/adk/sessions/migration/migrate_postgresql_timestamptz.py

### DIFF
--- a/src/google/adk/sessions/database_session_service.py
+++ b/src/google/adk/sessions/database_session_service.py
@@ -344,7 +344,12 @@ class DatabaseSessionService(BaseSessionService):
       )
 
       if config and config.after_timestamp:
-        after_dt = datetime.fromtimestamp(config.after_timestamp)
+        if self.db_engine.dialect.name == "sqlite":
+          after_dt = datetime.fromtimestamp(config.after_timestamp)
+        else:
+          after_dt = datetime.fromtimestamp(
+              config.after_timestamp, timezone.utc
+          )
         stmt = stmt.filter(schema.StorageEvent.timestamp >= after_dt)
 
       stmt = stmt.order_by(schema.StorageEvent.timestamp.desc())
@@ -513,7 +518,7 @@ class DatabaseSessionService(BaseSessionService):
             event.timestamp, timezone.utc
         ).replace(tzinfo=None)
       else:
-        update_time = datetime.fromtimestamp(event.timestamp)
+        update_time = datetime.fromtimestamp(event.timestamp, timezone.utc)
       storage_session.update_time = update_time
       sql_session.add(schema.StorageEvent.from_event(session, event))
 
--- a/src/google/adk/sessions/migration/migrate_postgresql_timestamptz.py
+++ b/src/google/adk/sessions/migration/migrate_postgresql_timestamptz.py
@@ -1,0 +1,150 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Migration script to convert TIMESTAMP to TIMESTAMPTZ for PostgreSQL.
+
+Starting from ADK v1.24.0, DatabaseSessionService creates timezone-aware
+datetime objects (with tzinfo=UTC). When using PostgreSQL with asyncpg,
+this causes a conflict if existing timestamp columns are defined as
+TIMESTAMP WITHOUT TIME ZONE, resulting in:
+
+  asyncpg.exceptions.DataError: can't subtract offset-naive and
+  offset-aware datetimes
+
+This migration alters all timestamp columns in ADK tables to use
+TIMESTAMP WITH TIME ZONE. It is safe to run on existing data as
+PostgreSQL will interpret existing naive timestamps as being in the
+server's timezone (typically UTC).
+
+Usage:
+  python -m google.adk.sessions.migration.migrate_postgresql_timestamptz \
+    --db_url postgresql+asyncpg://user:pass@host:port/dbname
+"""
+
+from __future__ import annotations
+
+import argparse
+import logging
+import sys
+
+from sqlalchemy import create_engine
+from sqlalchemy import text
+
+from . import _schema_check_utils
+
+logger = logging.getLogger("google_adk." + __name__)
+
+# Columns to migrate: (table_name, column_name)
+_TIMESTAMP_COLUMNS = [
+    ("sessions", "create_time"),
+    ("sessions", "update_time"),
+    ("events", "timestamp"),
+    ("app_states", "update_time"),
+    ("user_states", "update_time"),
+]
+
+
+def migrate(db_url: str) -> None:
+  """Migrates TIMESTAMP columns to TIMESTAMP WITH TIME ZONE for PostgreSQL.
+
+  Args:
+    db_url: The database URL (sync or async format).
+  """
+  sync_url = _schema_check_utils.to_sync_url(db_url)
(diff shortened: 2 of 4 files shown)


mynewestgitaccount:
I hope #4365 will be merged rather than this. Solutions like this create major issues for migrating production workloads, and they they will continue to do so until #3343 is resolved.

vietnamesekid:
> I hope #4365 will be merged rather than this. Solutions like this create major issues for migrating production workloads, and they they will continue to do so until #3343 is resolved.

Thanks for sharing your perspective. I agree that without better migration support (#3343), schema related changes are tough to roll out in production. What i was trying to address here is a correctness issue where timestamps depend on the server’s local timezone and can silently drift on non UTC setups. #4365 avoids the immediate error, but this change is mainly about making that behavior more explicit and predictable, scoped to the service layer where the dialect context is available. Totally open to feedback on whether this is the right trade-off at this stage

vietnamesekid:
> > I hope #4365 will be merged rather than this. Solutions like this create major issues for migrating production workloads, and they they will continue to do so until #3343 is resolved.
> 
> Thanks for sharing your perspective. I agree that without better migration support (#3343), schema related changes are tough to roll out in production. What i was trying to address here is a correctness issue where timestamps depend on the server’s local timezone and can silently drift on non UTC setups. #4365 avoids the immediate error, but this change is mainly about making that behavior more explicit and predictable, scoped to the service layer where the dialect context is available. Totally open to feedback on whether this is the right trade-off at this stage

am also running into this issue in production, which is why i put these PRs together. I upgraded to 1.24.0 earlier today, but it didn’t resolve the problem on our end 😂

SemihAkar:
+1 here for 1.24 version
I need urgent fix for the production environment. please hurry up :)

aidanandrews22:
+1

Here is a minimal monkey patch that fixes it. Just add this at the top of an early imported file (before you use any adk stuff):
# Patch: use a datetime subclass so we don't touch immutable built-in
from datetime import datetime as _RealDateTime, timezone
import google.adk.sessions.database_session_service as _db_session_mod

class _NaiveUTCDateTime(_RealDateTime):
    @classmethod
    def now(cls, tz=None):
        r = _RealDateTime.now(tz or timezone.utc)
        return r.replace(tzinfo=None) if r.tzinfo else r

_db_session_mod.datetime = _NaiveUTCDateTime

mynewestgitaccount:
Now that #4365 was merged, I really hope this one won't be merged. The benefits of this change would be so small now. Absolutely not worth it without #3343

wuliang229:
Since #4365 is merged, let's put this on hold until the auto schema migration support is added. Meanwhile, we will look into the semantic inconsistency here https://github.com/google/adk-python/pull/4365#issuecomment-3858380100

vietnamesekid:
Hey @wuliang229 

Still running into this stale session error constantly, even after bumping to 1.26.0:

ValueError: The last_update_time provided in the session object is earlier than the update_time in storage. Please check if it is a stale session.

I'm on PostgreSQL + Cloud Run with multiple instances hitting the same session concurrently. Checked the related issues (#1733, #4751, #4352, #917, etc.) and it really seems the check gets triggered super easily during fast successive append_event calls.

What I've been doing:
- Always reload a fresh session with session_service.get_session() right before any append/modify
- Never reuse the same session object for long — reload fresh every single time
- Still, under real production load with quick concurrent writes, it flakes out occasionally
- Suspect it's something with timestamp precision drifting a tiny bit (float → datetime → DB round-trip, maybe timezone or microsecond issues?)

Couple quick questions if you have a moment:
1. Any plans to improve/tweak the stale session detection in upcoming releases? (e.g. less strict < check, built-in retry logic, better microsecond handling)
2. Beyond the usual "reload + retry manually", is there a more reliable pattern or approach for truly concurrent environments? (maybe some internal coordination or different way to handle appends)

This one's blocking proper scaling for our agent deployment right now, so any update or workaround tips would be hugely helpful.

Thanks a ton

wukath:
Closing PR for now, feel free to reopen if you face problems: at head PostgreSQL is already in _NAIVE_DATETIME_DIALECTS, so asyncpg only sees naive datetimes and #1848 doesn't reproduce. What's left is a breaking change that makes every existing PostgreSQL deployment fail in prepare_tables until someone migrates it by hand.

Issue #3881, opened 2025-12-09
The agent doesn't work as expected when "skip_summarization" is set to True in AgentTool


slilichenko:
In [the code](https://github.com/google/adk-python/blob/82e6623fa97fb9cbc6893b44e228f4da098498da/src/google/adk/tools/agent_tool.py#L124), the skip_summarization of Agent Tool causes the parent context to emit an event with "is_final_response()" returning True. This causes [base flow interruption](https://github.com/google/adk-python/blob/82e6623fa97fb9cbc6893b44e228f4da098498da/src/google/adk/flows/llm_flows/base_llm_flow.py#L373), which means that the loop ends with the last function response. But most of the UIs, including ADK web, don't display the output of the tool response and it appears that the agent is stuck. Furthermore, if the agent is supposed to continue working on the response, perhaps by calling other tools with the output of the first AgentTool call as input - it does't happen because the response is marked as final.

[The documentation](https://google.github.io/adk-docs/tools-custom/function-tools/#key-difference-from-sub-agents) is a bit confusing because a) it doesn't cover these nuances and b) if you open the code snippet the "how it works session" covers what's not in the code (it describes what the way things works with skip_summarization = False). 

IMHO, it's a code bug or a design flaw, or a documentation bug.  A reproducible case: if you update [this simple agent](https://github.com/slilichenko/adk-playground/blob/5fa7b315dde451366179dc976df6ad354bdecd0a/patterns/image-processing-using-agent-tool/agent.py#L99) to use skip_summarization - the answer is never displayed to the user via ADK web interface. The skip summarization path generates 4 events vs 5 with summarization; the last event being the function response (and the function response text is never displayed in the ADK web UI). Happy to provide detailed screenshots if needed.

BnnaFish:
Also, for me, tests Eval works fine, but Agent is exposed as A2A and the response is empty. I have a very simple agent that calls only one tool. Plus tracing shows good response, but in the caller agent its empty.

kkrusher:
## Issue Confirmed - Root Cause Analysis

I can confirm this is a bug/design flaw. Here's what's happening:

### Root Cause

The issue is in `src/google/adk/events/event.py:90`:

```python
def is_final_response(self) -> bool:
    if self.actions.skip_summarization or self.long_running_tool_ids:
      return True  # ← This causes premature termination
```

[...]

PR #4389, opened 2026-02-06
fix(agent_tool): ensure text output when skip_summarization is True

ItsMacto:
When skip_summarization=True, AgentTool previously terminated the flow with a FunctionResponse event that contained no text, leaving UIs stuck.

This change appends a Text part to the response event when summarization is skipped, so the tool's output is displayed. The implementation avoids duplication, safely serializes non-string results, and preserves existing behavior for normal flows.

Added regression tests to verify text output is present for both string and JSON-like tool results.

- Closes: #3881

**Problem:**
When `skip_summarization=True`, the final response event emitted by AgentTool contains only a FunctionResponse and no displayable text. Most UIs do not render function responses, making the agent appear to hang or return an empty result.

**Solution:**
When skip summarization is enabled, the function result is additionally included as a Text part in the response event. If the result is a string, it is used directly; otherwise, it is safely serialized to JSON for display. This ensures the final event remains displayable while preserving existing behavior for non-skip paths.

### Testing Plan

**Unit Tests:**

_Passed tests:_
- `pytest tests/unittests/tools/test_agent_tool.py`

Not required. The change affects event construction and is fully covered by unit tests validating the final response content.

### Checklist

### DIFF
--- a/src/google/adk/flows/llm_flows/functions.py
+++ b/src/google/adk/flows/llm_flows/functions.py
@@ -19,8 +19,8 @@ from __future__ import annotations
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
 import copy
-import functools
 import inspect
+import json
 import logging
 import threading
 from typing import Any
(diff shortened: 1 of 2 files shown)


ItsMacto:
Used the wrong linter. Going back

PR #4390, opened 2026-02-06
fix(agent_tool): ensure text output when skip_summarization is True

### LINKED ISSUE #3881: The agent doesn't work as expected when "skip_summarization" is set to True in AgentTool

ItsMacto:
When skip_summarization=True, AgentTool previously terminated the flow with a FunctionResponse event that contained no text, leaving UIs stuck.

This change appends a Text part to the response event when summarization is skipped, so the tool's output is displayed. The implementation avoids duplication, safely serializes non-string results, and preserves existing behavior for normal flows.

Added regression tests to verify text output is present for both string and JSON-like tool results.

- Closes: #3881

**Problem:**
When `skip_summarization=True`, the final response event emitted by AgentTool contains only a FunctionResponse and no displayable text. Most UIs do not render function responses, making the agent appear to hang or return an empty result.

**Solution:**
When skip summarization is enabled, the function result is additionally included as a Text part in the response event. If the result is a string, it is used directly; otherwise, it is safely serialized to JSON for display. This ensures the final event remains displayable while preserving existing behavior for non-skip paths.

### Testing Plan

**Unit Tests:**

_Passed tests:_
- `pytest tests/unittests/tools/test_agent_tool.py`

Not required. The change affects event construction and is fully covered by unit tests validating the final response content.

### Checklist


rohityan:
Hi @ItsMacto , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the formatting errors. You can run autoformat.sh. Also can you also fix the failing unit tests.

ItsMacto:
Hi @ryanaiagent , I believe I made the changes correctly. One thing when running `autoformat.sh.` it affects like 50 files that are not mine. I just didn't commit these changes, but I'm not sure what the best course of action is.

ItsMacto:
@ryanaiagent If you get a chance would you be able to review this

rohityan:
Hi @ItsMacto , Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:
Hi @Jacksunwei , can you please review this. LGTM.

ItsMacto:
Hi @rohityan @Jacksunwei Was wondering on the next steps!

GWeale:
same fix landed in main via b983fcf (closes #3881). closing.

Issue #4302, opened 2026-01-28
VertexAiSessionService does not paginate when invoking list_sessions


arryon:
**Describe the Bug:**
The VertexAiSessionService does not return any Session objects beyond the initial page default of 100 when using its `list_sessions()` interface.

**Steps to Reproduce:**
See code snippet example below.

1. Create a VertexAiSessionService with more than 100 sessions for a given `user_id` (`app_name` does not matter, the session service does not filter on `app_name` currently)
2. Create an instance of VertexAiSessionService with optional properties and fetch the user sessions using `session_service.list_sessions(app_name="[app_name]", user_id="[user_id]")
3. The resulting list will be capped to a maximum of 100 items

**Expected Behavior:**
We expect pagination to automatically be applied, since the `sessions_iterator` object is of type `google.genai.pagers.AsyncPager` (or perhaps more specifically `google.cloud.spanner_v1.services.spanner.pagers.ListSessionsAsyncPager`?). 

**Observed Behavior:**
The list is capped to 100 items. This is because AsyncPager does not implement automated pagination on non-async iterator dunder methods (__iter__ or __next__), but instead only implements `__aiter__` and `__anext__`. The docs specifically state:
```
If there are more pages, the ``__aiter__`` method will make additional
    ``ListSessions`` requests and continue to iterate
    through the ``sessions`` field on the
    corresponding responses.
```

But, in VertexAiSessionService, we loop the iterator using `for` instead of `async for`, therefore not using the underlying pagination.

Offending line:
https://github.com/google/adk-python/blob/a16e3cc67e1cb391228ba78662547672404ae550/src/google/adk/sessions/vertex_ai_session_service.py#L218

**Environment Details:**

 - ADK Library Version (pip show google-adk): 1.23.0
 - Desktop OS:** MacOS
 - Python Version (python -V): 3.11

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: (e.g., gemini-2.5-pro) N/A

---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

**Regression:**

> Did this work in a previous version of ADK? If so, which one?

Not aware of this

**Minimal Reproduction Code:**
Requires installing UV, the script itself is standalone. Fill in `GOOGLE_PROJECT_ID`, `LOCATION` and `AGENT_ENGINE_ID`
Please provide a code snippet or a link to a Gist/repo that isolates the issue.
```python
#!/usr/bin/env -S uv run --script
# /// script
[...]

PR #4435, opened 2026-02-10
fix(sessions): use async iteration for VertexAiSessionService.list_sessions pagination

anmolg1997:

- Closes: #4302

**Problem:**

`VertexAiSessionService.list_sessions()` only returns the first ~100 sessions. The `sessions_iterator` from `api_client.agent_engines.sessions.list()` is an `AsyncPager` — it implements `__aiter__`/`__anext__` for fetching subsequent pages, but the code uses a plain `for` loop which only calls `__iter__`/`__next__`, so it never fetches beyond the first page.

**Solution:**

Changed `for api_session in sessions_iterator` to `async for api_session in sessions_iterator` so the `AsyncPager` actually paginates. Updated the test mock to return an `AsyncIterableList` (supports both sync and async iteration) instead of a bare list, so the tests properly simulate real `AsyncPager` behaviour.

### Testing Plan

**Unit Tests:**

```
$ pytest tests/unittests/sessions/
115 passed, 1 warning in 2.25s
```

The existing `test_list_sessions`, `test_list_sessions_with_pagination`, and `test_list_sessions_all_users` all continue to pass with the updated mock.

### DIFF
--- a/tests/unittests/sessions/test_vertex_ai_session_service.py
+++ b/tests/unittests/sessions/test_vertex_ai_session_service.py
@@ -267,10 +267,7 @@ async def to_async_iterator(data):
 class AsyncIterableList(list):
   """A list that also supports async iteration, mimicking AsyncPager behavior."""
 
-  def __aiter__(self):
-    return self._async_iter()
-
-  async def _async_iter(self):
+  async def __aiter__(self):
     for item in self:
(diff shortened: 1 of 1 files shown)


anmolg1997:
I have signed the CLA - @googlebot please verify.

PR #4469, opened 2026-02-12
fix: recurse into object properties in _update_type_string for Anthropic adapter

encounter:
## Summary

`_update_type_string` lowercases Gemini-style type strings (e.g. `STRING` → `string`) for Anthropic API compatibility. It already recursed into `items` (arrays) and `items.properties` (arrays of objects), but not into top-level `properties` of object types. This caused tool schemas with nested object parameters (e.g. Pydantic `BaseModel` subclasses used directly as function parameters) to produce invalid JSON Schema with uppercase type strings, resulting in Anthropic API 400 errors:

```
{'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'tools.1.custom.input_schema: JSON schema is invalid. It must match JSON Schema draft 2020-12 (https://json-schema.org/draft/2020-12). Learn more about tool use at https://docs.claude.com/en/docs/tool-use.'}, 'request_id': 'x'}
```

The fix recurses into `properties` at every level, which also subsumes the existing `items.properties` handling.

## Reproduction

Use a Pydantic model with nested objects as a `FunctionTool` parameter with the `Claude` model class:

```python
class Address(BaseModel):
    city: str = ""
    state: str = ""

class Profile(BaseModel):
    name: str = ""
    address: Address = Field(default_factory=Address)

async def update_profile(profile: Profile, tool_context: ToolContext) -> dict:
    ...

root_agent = LlmAgent(
    model=Claude(model="claude-opus-4-5@20251101"),
    tools=[FunctionTool(update_profile)],
    ...
)
```

Before this fix, `address.city` and `address.state` retain `"STRING"` (uppercase) in the schema sent to Anthropic, which rejects it.

## Test plan

- [x] Added `function_with_nested_object_parameter` test case to `test_function_declaration_to_tool_param` — verifies a 3-level deep object schema (parameter → profile → address → city/state) is fully lowercased
- [x] Confirmed the new test fails without the fix and passes with it
- [x] All 18 existing tests in `test_anthropic_llm.py` continue to pass


rohityan:
Hi @encounter, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:
Hi @GWeale  , can you please review this. LGTM

encounter:
It looks like this was merged in ab4b736807dabee65659486a68135d9f1530834c. Thanks!

PR #4501, opened 2026-02-15
Keep the temp: state keys in the in-memory context session

### LINKED ISSUE #3047: Temp prefixed state key not available on after_agent_callback
**Describe the bug**
The documentation states that `state` variables prefixed with `temp:` should be available during the entire invocation process:

> temp: Prefix (Temporary Invocation State):
> 
> Scope: Specific to the current invocation (the entire process from an agent receiving user input to generating the final output for that input).

However, if the agent `output_key ` is set with a `temp:` prefixed key, this key is not present anymore in the state during `after_agent_callback`

**To Reproduce**
The following code works as expected
```
 from google.adk.agents.callback_context import CallbackContext
 
 
 def _callback_test(callback_context: CallbackContext) -> types.Content | None:
     print(callback_context.state["test_output"])
     return None
 
 
 root_agent = LlmAgent(
     name="test_agent",
     model="gemini-2.5-flash",
     instruction="You are a helpful assistant that can generate a 10 words random text.",
     output_key="test_output",
     after_agent_callback=_callback_test,
 )
```

But the following code fails, because it can't find the key in the `after_agent_callback` execution
```
 from google.adk.agents.callback_context import CallbackContext
 
 
 def _callback_test(callback_context: CallbackContext) -> types.Content | None:
     print(callback_context.state["temp:test_output"])
     return None
 
 
 root_agent = LlmAgent(
     name="test_agent",
     model="gemini-2.5-flash",
     instruction="You are a helpful assistant that can generate a 10 words random text.",
     output_key="temp:test_output",
     after_agent_callback=_callback_test,
 )
```

**Expected behavior**
Temp keys should remain available during the entire invocation process

 - OS: windows, mac
 - Python version: 3.13
 - ADK version: 1.15.1

 **Model Information:**
 - Are you using LiteLLM: No

### LINKED ISSUE #3399: Agent runner cannot pass temp_state via state_delta; session service ignores temp state during invocation
**Describe the bug**
When passing a state_delta to the agent runner, I need to include a temp_state. However, in the current ADK flow, the session service ignores the provided temp_state when creating the invocation and injecting events. As a result, temp-only state cannot be applied for a single turn and is effectively dropped.

**To Reproduce** 
1. Create an agent invocation through the agent runner with state_delta that includes a temp_state payload intended only for the current turn.
2. Observe that the session service builds the invocation and injects events without applying the provided temp_state.
3. The agent runs as if temp_state was never provided.

**Impact**
1. Cannot apply ephemeral, per-turn overrides or sandboxed state adjustments.
2. Forces callers to mutate persistent state for transient needs, which is undesirable and error-prone.

ashubham:
This enables downstream consumers to read temp: state values when set during invocation start.

- Closes: #3047 
- Closes: #3399 

**Problem:**
When `temp:` state values are set during invocation start (`run_async`), these get stripped out for downstream consumers. Making the utility of these variables marginal as stated in the above issues.

**Solution:**
Do not strip the `temp:` keys from the in memory `session.state`. These are still stripped from the `state_delta` persisted in the session store.

### Testing Plan

**Unit Tests:**

```
========================================================================= 4281 passed, 2068 warnings in 50.93s =========================================================================
```

- When calling Runner `run_async`, pass a `temp:test` state_delta.
- Read the value of `temp:test` inside a lifecycle callback like `before_agent_callback`.
- See, that the value is the same as passed when calling the runner.

### Checklist


rohityan:
Hi @ashubham , Thank you for your contribution! We appreciate you taking the time to submit this pull request.

We cannot merge temp: keys directly into session.state. The session.state dictionary represents the permanent, long-term memory of the session. When we remove `if key.startswith(State.TEMP_PREFIX): continue`, any backend database saving the session will permanently write these temporary variables to disk, which is exactly what the temp: prefix is designed to prevent."

ashubham:
@ryanaiagent Thank you for the review! I believe there's a misunderstanding about what gets persisted.

**The temp: keys are NOT persisted to the database.** Here's the flow in `append_event()`:

```python
# Step 1: Update in-memory session.state with ALL keys (including temp:)
self._update_session_state(session, event)
# Step 2: Remove temp: keys from event.actions.state_delta
event = self._trim_temp_delta_state(event)
# Step 3: Append the cleaned event (temp: keys already removed)
session.events.append(event)
```

**Key distinction:**
- `session.state` (in-memory): Contains temp: keys during the invocation so callbacks can access them
- `event.actions.state_delta` (persisted): Cleaned of temp: keys via `_trim_temp_delta_state()` before appending to `session.events`

Only the cleaned events are persisted to the database. When sessions are loaded, `session.state` is reconstructed from the cleaned event deltas, so temp: keys never reach disk.

The in-memory presence of temp: keys is intentional and temporary, they only exist during the current invocation.

GWeale:
Thanks for working through this and for the detailed write up. This behavior is already in main, where BaseSessionService.append_event applies temp: keys to the in-memory session via _apply_temp_state before trimming them from the persisted event (sessions/base_session_service.py), so they stay readable within the invocation but are never saved; closing as already resolved.

Issue #4506, opened 2026-02-16
runner: prevent duplicate user event append on invocation retry with same invocation_id


davidahmann:
## Problem
Retrying/resuming with the same `invocation_id` and same user message can append duplicate user events to the session timeline. That weakens invocation idempotency and can trigger duplicate downstream effects.

## Expected behavior
If the same user event for the same invocation is already persisted, retry should be a no-op for message append.

## Scope (Phase 1)
- Add duplicate-event guard before appending user message event.
- Add runner-level regression test for repeated resume/retry with same invocation ID + message.

## Acceptance criteria
- Duplicate retry does not append a second identical user event.
- Existing non-duplicate message append behavior remains unchanged.
- Regression test covers repeated retry path.

llalitkumarrr:
Hello @davidahmann ,

We would like to analyze this issue in detail. Could you please let us know which version of `google-adk` you are using? It would also be very helpful if you could provide reproduction steps and a minimal reproducible code snippet to assist in our analysis.

davidahmann:
@llalitkumarrr  Thanks for the follow-up. I reproduced this on a released build.

Environment:
- google-adk: `1.25.0`
- Python: `3.13.7`
- OS: `macOS arm64`

Minimal repro:

```python
import asyncio
from google.adk.agents.base_agent import BaseAgent
from google.adk.agents.invocation_context import InvocationContext
from google.adk.apps.app import App, ResumabilityConfig
from google.adk.events.event import Event
from google.adk.runners import Runner
from google.adk.sessions.in_memory_session_service import InMemorySessionService
from google.adk.artifacts.in_memory_artifact_service import InMemoryArtifactService
from google.genai import types

class StubAgent(BaseAgent):
    async def _run_async_impl(self, invocation_context: InvocationContext):
        yield Event(
            invocation_id=invocation_context.invocation_id,
            author=self.name,
            content=types.Content(role="model", parts=[types.Part(text="ok")]),
        )

async def main():
    ss = InMemorySessionService()
    app = App(
        name="app",
        root_agent=StubAgent(name="stub", sub_agents=[]),
        resumability_config=ResumabilityConfig(is_resumable=True),
    )
    runner = Runner(app=app, session_service=ss, artifact_service=InMemoryArtifactService())
    session = await ss.create_session(app_name="app", user_id="u")
    msg = types.Content(role="user", parts=[types.Part(text="hello")])

    # Simulate first attempt already persisted for invocation_id inv-1
[...]

PR #4507, opened 2026-02-16
runner: avoid duplicate user event append on invocation retry

davidahmann:
## Problem
Retry/resume calls with the same `invocation_id` and same user message could append duplicate user events to session history.

## What changed
- Added a duplicate-user-event guard in `Runner._append_new_message_to_session` keyed by invocation id + content + state delta.
- Skips append when the same user event already exists for that invocation.
- Added runner tests for duplicate retry no-op and non-duplicate append behavior.

## Validation
- `uv sync --extra test`
- `uvx pyink --config pyproject.toml src/google/adk/runners.py tests/unittests/test_runners.py`
- `uv run pytest tests/unittests/test_runners.py`
- `uv run pytest tests/unittests/runners/test_resume_invocation.py`

Refs #4506

### DIFF
--- a/src/google/adk/memory/vertex_ai_memory_bank_service.py
+++ b/src/google/adk/memory/vertex_ai_memory_bank_service.py
@@ -65,6 +65,11 @@ _CREATE_MEMORY_CONFIG_FALLBACK_KEYS = frozenset({
     'wait_for_completion',
 })
 
+_ENABLE_CONSOLIDATION_KEY = 'enable_consolidation'
+# Vertex docs for GenerateMemoriesRequest.DirectMemoriesSource allow
+# at most 5 direct_memories per request.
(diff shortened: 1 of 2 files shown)


davidahmann:
Implemented an invocation-scoped dedupe guard for user-message append in retry/resume paths.

Scope in this PR:
- Added duplicate detection by `(invocation_id, user content, state_delta)`.
- Skips append when a matching user event already exists.
- Added tests for duplicate retry no-op and non-duplicate behavior.

This contribution was informed by patterns from Gait: https://github.com/davidahmann/gait

rohityan:
Hi @davidahmann , Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Please fix the formatting errors. You can use autoformat.sh. Also please update the branch to main.

Issue #4595, opened 2026-02-23
Enforce deterministic persisted-event counts after run_async


davidahmann:
Runner/session tests can misreport event totals when assertions rely on stale session snapshots, reducing confidence in replay behavior.

Current behavior is insufficient because post-invocation assertions sometimes count in-memory state that is no longer authoritative for persisted session events.

Why now: session replay and event-contract stability are core runtime guarantees and this failure pattern is recurring.

Claim-to-codepath map:
- Invocation persistence flow in src/google/adk/runners.py.
- Session storage/read paths in src/google/adk/sessions.
- Assertions and fixtures in tests/unittests.

Requested behavior: tests re-fetch session state before counting events and demonstrate deterministic totals across repeated runs.

## Evidence Packet
- Commit under test: aa779205
- Runtime environment:
  - OS: Darwin 25.3.0 arm64
  - Python 3.14.0
  - go version go1.25.7 darwin/arm64
  - Node: v22.19.0
- Minimal repro:
  1. Follow the repo's current local validation/CI-triage path.
  2. Observe the behavior described above for this issue.
- Expected behavior: deterministic, documented behavior matching acceptance criteria below.
- Actual behavior: behavior can be inconsistent for the scenario described above.

## Acceptance Criteria
- Implement the smallest code change that addresses the proven pain point.
- Add focused regression test coverage for the changed behavior.
- Ensure backward compatibility expectations are explicit in tests/docs.

llalitkumarrr:
Hello @davidahmann ,

Could you please elaborate on this issue to help understand this a bit better?

PR #4596, opened 2026-02-23
tests(runners): require fresh session fetch for persisted event counts

davidahmann:
## Problem
Persisted-event assertions can drift when tests rely on stale session snapshots after run_async invocations.

## Why now
This is active operator friction in current workflows and needs deterministic behavior for repeatable triage/replay.

## What changed
Added a regression test that demonstrates stale snapshot behavior and asserts counts using a freshly fetched persisted session.

## Validation
- .venv/bin/pytest tests/unittests/runners/test_persisted_event_counts.py -q (pass)

Refs #4595

### DIFF
--- a/tests/unittests/runners/test_persisted_event_counts.py
+++ b/tests/unittests/runners/test_persisted_event_counts.py
@@ -1,0 +1,56 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


davidahmann:
Added a regression test that demonstrates stale session snapshots after run_async and enforces persisted-event assertions against a freshly fetched session state.

This contribution was informed by patterns from Wrkr: https://github.com/Clyra-AI/wrkr

PR #4618, opened 2026-02-25
fix: temp-scoped state now visible to subsequent agents in same invocation

### LINKED ISSUE #4564: Output key doesn't work with temp state scope
**Describe the Bug:**
I am running a Sequential Agent. I want to save the output of first agent as output_key with temporary scope (valid for that invocation). However, when I use the following code -
`output_key='temp:output_key'`
I get no value in state. I checked that that its probably due to `_trim_temp_delta_state` function in BaseSessionService. 

**Steps to Reproduce:**
1. Create a sequential agent with subagents having output keys
2. prefix output key with temp

**Expected Behavior:**
For a SequentialAgent, the lifecycle of invocation starts from the first agent till the last agent. Therefore the 'temp' prefixed state should be available across the agents. But right now its available only for the start of an (sub)agent till that agent produces response.

**Observed Behavior:**
temp prefixed state not transferred across agents

**Environment Details:**

 - ADK Library Version: 1.24.1
 - Desktop OS:** MacOS

**Model Information:**

 - Are you using LiteLLM: Yes
 - Which model is being used: Claude 4.5

```python
a1 = Agent(
    model=LiteLlm(),
    name="a2",
    static_instruction=get_instrcutions(),
    output_schema=InputData,
    output_key='temp:output_key',
)

a2 = Agent(
    model=LiteLlm(),
    name="a2",
    instruction= get_instructions,
)

workflow = SequentialAgent(
    name="workflow_agent",
    sub_agents=[a1, a2]
)
```

 - Always (100%)

stakeswky:
## Summary

Fixes #4564

When using `output_key` with a `temp:` prefix (e.g. `output_key='temp:result'`) in a `SequentialAgent`, the output was silently lost. Agent-2 could never read the temp state written by agent-1.

## Root Cause

Two issues in `append_event`:

1. `_trim_temp_delta_state()` removed temp keys from the event delta **before** `_update_session_state()` could apply them to the in-memory session
2. `_update_session_state()` also explicitly skipped `temp:`-prefixed keys

```python
# Before (broken ordering):
async def append_event(self, session, event):
    event = self._trim_temp_delta_state(event)   # temp keys gone!
    self._update_session_state(session, event)    # nothing to apply
```

## Fix

Introduce `_apply_temp_state()` which writes temp-scoped keys to the in-memory `session.state` **before** the event delta is trimmed:

```python
# After:
async def append_event(self, session, event):
    self._apply_temp_state(session, event)        # temp keys → session.state
    event = self._trim_temp_delta_state(event)    # temp keys removed from delta
    self._update_session_state(session, event)    # non-temp keys applied
```

This ensures:
- ✅ Temp state is available to subsequent agents within the same invocation
- ✅ Temp state is still stripped from event deltas (not persisted to storage)
- ✅ All three session services (InMemory, Database, SQLite) behave consistently

## Files Changed

- `src/google/adk/sessions/base_session_service.py`: Added `_apply_temp_state()`, reordered `append_event` logic, removed temp-skip in `_update_session_state`
- `src/google/adk/sessions/database_session_service.py`: Added `_apply_temp_state()` call before trim
- `src/google/adk/sessions/sqlite_session_service.py`: Added `_apply_temp_state()` call before trim
- `tests/unittests/sessions/test_session_service.py`: Updated existing test + added new test for sequential agent scenario

## Testing

All 67 session service tests pass across InMemory, Database, and SQLite backends.


rohityan:
Hi @stakeswky , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

xuanyang15:
@DeanChensj Could you please help review?

PR #4622, opened 2026-02-25
Feat/Issue Monitoring Agent

rohityan:

- Closes: N/A
- Related: N/A

**Problem:**
Open-source repositories frequently suffer from SEO spam, unsolicited promotional links, and irrelevant 3rd-party endorsements in issue comments. Manually monitoring, reading, and moderating these comments is a time-consuming burden for maintainers.

**Solution:**
Implemented the `adk_issue_monitoring_agent`, an automated moderation bot built using the Google Agent Development Kit (ADK) that monitors recently updated issues for spam. 

Key features and cost-optimizations include:
* **Zero-Waste LLM Pre-filtering:** The agent fetches comments via REST API and pre-filters them in Python. It completely ignores comments from maintainers, standard `[bot]` accounts, and the official `adk-bot`, ensuring Gemini LLM tokens are never wasted on safe comments.
* **Token Bloat Prevention:** Uses Regex to strip out Markdown code blocks (replacing them with `[CODE BLOCK REMOVED]`) and truncates unusually long text/logs to 1500 characters, preventing spammers from hiding links in massive stack traces while keeping API costs near zero.
* **Anti-Double-Posting (Idempotency):** The bot reads the comment history for its own signature (`🚨 **Automated Spam Detection Alert** 🚨`). If it has already flagged an issue, it instantly skips it to prevent infinite feedback loops.
* **Dual-Mode GitHub Actions:** Includes an `issue-monitor.yml` workflow that runs a lightweight "Daily Sweep" (only checking issues updated in the last 24 hours) via cron, and supports a manual `workflow_dispatch` checkbox to perform a Day 1 "Deep Clean" of the entire repository.

### Testing Plan

**Unit Tests:**

_Note: The agent relies on the core ADK `Agent` and `InMemoryRunner` components which are covered by the existing test suite. No core framework tests were broken._

I rigorously tested the agent locally on a fork of the repository with Issues enabled. 

**Setup:**
1. Configured a local `.env` file with `INITIAL_FULL_SCAN=true` and `OWNER` set to my fork.
2. Created a test issue and populated it with 10 different types of spam (SEO links, crypto airdrops, fake tech support) using a non-maintainer account.
3. Included "False Positive" test cases (e.g., beginner questions, massive safe code blocks, safe external documentation links).

**Results:**
* **Detection:** The agent successfully identified all 10 spam variations, applied the `spam` label, and posted the warning alert tagging maintainers.
* **False Positives:** The agent correctly ignored the beginner questions and safe documentation links.
* **Code Stripping:** Verified via logs that massive JSON/YAML blocks were successfully regex-stripped before being passed to the LLM.
* **Idempotency:** Re-ran the script immediately after a successful flag; the bot detected its own alert signature and successfully skipped the issue without double-posting or hitting the LLM.
* **Formatting:** Ran `isort` and `pyink` to ensure all code adheres strictly to the Google Python Style Guide.

### Checklist

### Additional context

This agent is located in `contributing/samples/adk_issue_monitoring_agent/` and includes a comprehensive `README.md` detailing how repository maintainers can configure, deploy, and customize the bot using GitHub Actions.


rohityan:

.github/workflows/issue-monitor.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Run Issue Monitoring Agent
+        env:
+          GITHUB_TOKEN: ${{ secrets.ADK_TRIAGE_AGENT }}
+          GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.event.repository.name }}
+          CONCURRENCY_LIMIT: 3
+          INITIAL_FULL_SCAN: ${{ github.event.inputs.full_scan == 'true' }}
+        run: python -m spam_sweeper_agent.main
```
xuanyang15: Shouldn't this be `python -m adk_issue_monitoring_agent.main`?

.github/workflows/issue-monitor.yml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install requests google-adk python-dotenv
+
+      - name: Run Issue Monitoring Agent
+        env:
```
xuanyang15, line 54: Do we miss `PYTHONPATH`, otherwise, it may fail with `adk_issue_monitoring_agent` module not being found.

contributing/samples/adk_issue_monitoring_agent/main.py
```diff
@@ -0,0 +1,203 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+import logging
+import re
+import time
+from typing import List
```
xuanyang15: We prefer to use built-in list and `|`

contributing/samples/adk_issue_monitoring_agent/utils.py
```diff
@@ -0,0 +1,178 @@
...
+from adk_issue_monitoring_agent.settings import INITIAL_FULL_SCAN
+from adk_issue_monitoring_agent.settings import SPAM_LABEL_NAME
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+
+logger = logging.getLogger("google_adk." + __name__)
+
+_api_call_count = 0
+_counter_lock = threading.Lock()
```
xuanyang15: It looks we are using `asyncio.gather` which is concurrent but not multi threadking, so `threading.Lock` might not work.

contributing/samples/adk_issue_monitoring_agent/utils.py
```diff
@@ -0,0 +1,178 @@
...
+  global _api_call_count
+  with _counter_lock:
+    _api_call_count += 1
+
+
+retry_strategy = Retry(
+    total=6,
+    backoff_factor=2,
+    status_forcelist=[429, 500, 502, 503, 504],
+    allowed_methods=["GET", "POST", "PATCH", "DELETE"],
```
xuanyang15: `POST` is not idempotent, so we probably need to be careful for retrying POST calls to avoid duplications.

Maybe we should add some checking logic before POST calls, e.g

In flag_issue_as_spam, we can check 
- if an issue is already labeled and commented, we skip
- if an issue is labeled but not commented, we comment
- if an issue is not labeled nor commented, we label and comment

Similarly, for other POST calls if needed.

contributing/samples/adk_issue_monitoring_agent/main.py
rohityan: resolved

.github/workflows/issue-monitor.yml
rohityan: resolved

.github/workflows/issue-monitor.yml
rohityan: Corrected.

contributing/samples/adk_issue_monitoring_agent/utils.py
rohityan: resolved

rohityan:
Hi @xuanyang15 , addressed all your comments. Let me know if you need any more changes.

contributing/samples/adk_issue_monitoring_agent/utils.py
xuanyang15: Still seeing the old code, could you please double check your change is uploaded?

contributing/samples/adk_issue_monitoring_agent/main.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+
+    # 5. Format prompt and invoke LLM
+    compiled_comments = "\n".join(user_comments)
+    prompt_text = (
+        "Please review the following text for issue"
+        f" #{issue_number}:\n\n{compiled_comments}"
+    )
+
+    runner = InMemoryRunner(agent=root_agent, app_name=APP_NAME)
```
xuanyang15: This will create a new runner for each issue. Shouldn't reuse the runner for all the issues? We can just use new sessions to make the processing of each issue independant.

contributing/samples/adk_issue_monitoring_agent/agent.py
```diff
@@ -0,0 +1,115 @@
...
+    return f.read()
+
+
+PROMPT_TEMPLATE = load_prompt_template("PROMPT_INSTRUCTION.txt")
+
+# --- Tools ---
+
+
+def flag_issue_as_spam(
+    item_number: int, detection_reason: str
```
xuanyang15, line 47: detection_reason comes from LLM, which may be markdown, if we put it directly in Github comments, Github will render it, is it something we want? should we escape or add some safeguard?

contributing/samples/adk_issue_monitoring_agent/agent.py
```diff
@@ -0,0 +1,115 @@
...
+  alert_body = (
+      f"{BOT_ALERT_SIGNATURE}\n"
+      "@maintainers, a suspected spam comment was detected in this thread.\n\n"
+      f"**Reason:** {detection_reason}"
+  )
+
+  try:
+    # 1. Fetch current state to check what actions are actually needed
+    issue = get_issue_details(OWNER, REPO, item_number)
+    comments = get_issue_comments(OWNER, REPO, item_number)
```
xuanyang15, line 78: nit: `get_issue_details` and `get_issue_comments` are called twice for each issue, once in main.py, once here, is there a way to avoid duplicated calls so that we can reduce the API usage.

Please feel free to keep it as is if there is not clean way to achieve it!

contributing/samples/adk_issue_monitoring_agent/utils.py
```diff
@@ -0,0 +1,174 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from typing import Any
+from typing import Dict
+from typing import List
```
xuanyang15: Are these still being used? Since we prefer to use built in dict and list, can we remove these?

rohityan: removed.

contributing/samples/adk_issue_monitoring_agent/main.py
rohityan: resolved

contributing/samples/adk_issue_monitoring_agent/agent.py
rohityan: added a lru_cache.

contributing/samples/adk_issue_monitoring_agent/agent.py
rohityan: I sanitized the LLM's detection_reason and wrapped it in a ```text block to prevent GitHub from rendering accidental @mentions or clickable spam links.

rohityan:
Hi @xuanyang15 , PTAL
1. Instantiated runner once.
2. sanitized llm detection_reason to ensure github doesn't render clickable spam links.
3. added lru_cache to avoid duplicated calls.
4. removed unnecessary imports.

contributing/samples/adk_issue_monitoring_agent/agent.py
xuanyang15: Thanks for the change! I think the lru_ache might bringing some new issues:
1. lru_cache caches object references, not copies. Both get_issue_details and get_issue_comments return mutable dict/list. Any downstream mutation silently corrupts the cache for all future callers.

2. Idempotency check is defeated. e.g. process_single_issue calls get_issue_comments() first (result cached), then LLM triggers flag_issue_as_spam, which calls get_issue_comments() again — cache hit, stale data. The BOT_ALERT_SIGNATURE check never sees the bot's own alert, making the idempotency guard a no-op.

rohityan: Thanks for catching this. Reverted back for now. I can't think of any clean way to achieve this. Do you have any suggestions?
However after the initial run, the agent will only check the new issues or issues with updates. So in-spite of duplicate calls the token usage shouldn't be high.

rohityan:
Hi @xuanyang15 , removed the lru_cache.

xuanyang15:
Thanks! It is imported for internal rewiew.

PR #4630, opened 2026-02-25
fix: handle Pydantic models in _safe_json_serialize for tool tracing

### LINKED ISSUE #4629: _safe_json_serialize drops Pydantic model tool results as '<not serializable>' in traces
## Description

`_safe_json_serialize` in `src/google/adk/telemetry/tracing.py` silently replaces Pydantic `BaseModel` instances with the string `'<not serializable>'` when serializing tool responses for tracing spans.

This affects **any tool that returns a Pydantic model** rather than a plain `dict` — the tool executes correctly, but the traced output is lost.

## Root Cause

The serializer uses:

```python
json.dumps(obj, ensure_ascii=False, default=lambda o: '<not serializable>')
```

When a tool returns a Pydantic model, ADK wraps it as `{'result': <BaseModel>}` in `__build_response_event` (functions.py line 798-799). `json.dumps` can't serialize `BaseModel` natively, so the `default` lambda fires and replaces it.

## Reproduction

```python
import json
import asyncio
from google.adk.telemetry.tracing import _safe_json_serialize


class SearchResults(BaseModel):
    query: str
    total: int
    items: list[str]


async def search_tool(query: str) -> SearchResults:
    """A tool that returns a Pydantic model."""
    return SearchResults(query=query, total=2, items=["doc1", "doc2"])


result = asyncio.run(search_tool(query="test"))

# ADK wraps non-dict results (functions.py __build_response_event)
tool_response = {"result": result}

# ADK traces the response (tracing.py trace_tool_call)
serialized = _safe_json_serialize(tool_response)
print(serialized)
# Output: {"result": "<not serializable>"}

# Expected: {"result": {"query": "test", "total": 2, "items": ["doc1", "doc2"]}}
```

## Suggested Fix

`BaseModel` is already imported in `tracing.py` (line 60) and pydantic is already a dependency. The fix is minimal:

```python
def _safe_json_serialize(obj) -> str:
  def _default(o):
    if isinstance(o, BaseModel):
      return o.model_dump(mode="json")
    return '<not serializable>'

  try:
    return json.dumps(obj, ensure_ascii=False, default=_default)
  except (TypeError, OverflowError):
    return '<not serializable>'
```

This is consistent with how the rest of the file already handles Pydantic models (e.g. `_serialize_content` at line 470 calls `content.model_dump()`).

## Impact

Every tool returning a Pydantic model has its trace output silently dropped. This affects observability integrations (Langfuse, Phoenix, etc.) that consume these spans. The tool itself works fine — only the trace is broken.

## Related Issues

- #1540 — `AnyUrl` not serializable (MCP tools, same root cause)
- #2457 — `date` not serializable (BigQuery tools, same root cause)
- #586, #3288 — other type-specific serialization failures

All of these are symptoms of the same gap: `_safe_json_serialize` only handles JSON-native types.

## Environment

- ADK version: 1.25.1 (also confirmed on main branch)
- Python: 3.12
- Pydantic: 2.x

aperepel:
## Summary

`_safe_json_serialize` in `telemetry/tracing.py` silently replaces Pydantic `BaseModel` tool return values with `'<not serializable>'` in trace spans, because `json.dumps` cannot serialize Pydantic models natively.

This PR updates the `default` handler to call `model_dump(mode='json')` on `BaseModel` instances before falling back — consistent with how other functions in the same file already handle Pydantic objects (e.g., `_serialize_content` calls `content.model_dump()`).

### Reproduction

```python
from google.adk.telemetry.tracing import _safe_json_serialize

class SearchResults(BaseModel):
    query: str
    total: int
    items: list[str]

# Simulates ADK wrapping a non-dict tool result
tool_response = {"result": SearchResults(query="test", total=2, items=["a", "b"])}

print(_safe_json_serialize(tool_response))
# Before: {"result": "<not serializable>"}
# After:  {"result": {"query": "test", "total": 2, "items": ["a", "b"]}}
```

### Changes

- **`src/google/adk/telemetry/tracing.py`**: Replace the `default=lambda o: '<not serializable>'` with a handler that checks `isinstance(o, BaseModel)` and calls `o.model_dump(mode='json')`. Uses a lazy import of `BaseModel` to avoid import-time overhead.
- **`tests/unittests/telemetry/test_spans.py`**: Add 5 unit tests covering plain dicts, nested Pydantic models, top-level models, and non-serializable fallback.

### Test plan

- [x] All 33 existing telemetry tests pass
- [x] 5 new `_safe_json_serialize` tests pass
- [x] Non-Pydantic fallback behavior preserved

Fixes #4629

### DIFF
--- a/src/google/adk/telemetry/tracing.py
+++ b/src/google/adk/telemetry/tracing.py
@@ -109,19 +109,24 @@ logger = logging.getLogger('google_adk.' + __name__)
 def _safe_json_serialize(obj) -> str:
   """Convert any Python object to a JSON-serializable type or string.
 
+  Handles Pydantic BaseModel instances (common as tool return types)
+  by calling model_dump() before JSON encoding.
+
   Args:
     obj: The object to serialize.
 
   Returns:
-    The JSON-serialized object string or <non-serializable> if the object cannot be serialized.
+    The JSON-serialized object string or <not serializable> if the object
+    cannot be serialized.
   """
+  def _default(o: Any) -> Any:
+    if isinstance(o, BaseModel):
+      return o.model_dump(mode='json')
+    return '<not serializable>'
 
   try:
-    # Try direct JSON serialization first
-    return json.dumps(
-        obj, ensure_ascii=False, default=lambda o: '<not serializable>'
-    )
-  except (TypeError, OverflowError):
+    return json.dumps(obj, ensure_ascii=False, default=_default)
+  except (TypeError, OverflowError, ValueError):
     return '<not serializable>'
 
 
--- a/tests/unittests/telemetry/test_spans.py
+++ b/tests/unittests/telemetry/test_spans.py
@@ -1175,3 +1175,58 @@ async def test_generate_content_span_with_experimental_semconv(
   assert attributes[GEN_AI_AGENT_NAME] == invocation_context.agent.name
   assert GEN_AI_CONVERSATION_ID in attributes
   assert attributes[GEN_AI_CONVERSATION_ID] == invocation_context.session.id
+
+
+# ---------------------------------------------------------------------------
+# _safe_json_serialize tests
+# ---------------------------------------------------------------------------
+
+from google.adk.telemetry.tracing import _safe_json_serialize
+from pydantic import BaseModel as PydanticBaseModel
+
+
+class _SampleToolResult(PydanticBaseModel):
+  query: str
+  total: int
+  items: list[str] = []
+
+
+class _NestedModel(PydanticBaseModel):
+  inner: _SampleToolResult
+
+
+def test_safe_json_serialize_plain_dict():
+  """Plain dicts serialize normally."""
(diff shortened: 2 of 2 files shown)


rohityan:
Hi @aperepel ,Thank you for your contribution through this pull request!  This PR has merge conflicts that require changes from your end. Could you please rebase your branch with the latest main branch to address these? Once this is complete, please let us know so we can proceed with the review.

aperepel:
Rebased onto the latest main and resolved merge conflicts. Also addressed the review suggestion — added `ValueError` to the exception handling in `_safe_json_serialize`. Additionally removed the redundant lazy import of `BaseModel` since it's now imported at the top level of the module.

GWeale:
Thanks for the contribution. The _safe_json_serialize helper this PR patches has since been removed from telemetry/tracing.py, where tool-response tracing now serializes the response event via its Pydantic model dump, so this change no longer applies and I'm closing it as already resolved.

Issue #4639, opened 2026-02-27
Auto-create BigQuery views when BQ Agent Analytics plugin creates a new table


caohy1988:
## Summary

When the `BigQueryAgentAnalyticsPlugin` creates a new `agent_events` table (via `_ensure_schema_exists()`), it should automatically create per-event-type BigQuery views that unnest the generic table into typed, queryable columns. Currently users must manually create these views or use the external [BigQuery Agent Analytics SDK](https://github.com/haiyuan-eng-google/BigQuery-Agent-Analytics-SDK) `ViewManager`.

## Motivation

The `agent_events` table stores all event types in a single generic schema with JSON `content` and `attributes` columns. Querying specific event types (e.g., "show me all LLM latencies" or "find tool errors") requires users to write complex JSON extraction SQL every time. Per-event-type views solve this by providing pre-built, typed columns for each event category.

The [BigQuery Agent Analytics SDK](https://github.com/haiyuan-eng-google/BigQuery-Agent-Analytics-SDK/blob/main/SDK.md#16-bigquery-view-management) already implements this via a `ViewManager` class that:

- Creates **18 per-event-type views** (one for each event type: `USER_MESSAGE_RECEIVED`, `LLM_REQUEST`, `LLM_RESPONSE`, `TOOL_STARTING`, `TOOL_COMPLETED`, `AGENT_STARTING`, `AGENT_COMPLETED`, HITL events, etc.)
- Each view filters by `event_type` and unnests JSON fields into typed columns while preserving standard identity headers (`timestamp`, `agent`, `session_id`, `invocation_id`, `user_id`, `trace_id`, etc.)
- Provides `create_all_views()` for bulk creation, `create_view(event_type)` for individual views, and `get_view_sql(event_type)` for SQL inspection
- Uses a SQL template pattern with per-event-type column definitions (`_EVENT_VIEW_DEFS` + `_VIEW_SQL_TEMPLATE`)

### Option A: Integrated into the plugin (recommended)

Add view creation as an optional step in `BigQueryAgentAnalyticsPlugin._ensure_schema_exists()`:

1. After successfully creating the `agent_events` table, automatically create all per-event-type views in the same dataset
2. Add a config flag `create_views: bool = True` to `BigQueryLoggerConfig` to allow opting out
3. View creation should be idempotent — use `CREATE OR REPLACE VIEW` so it's safe to run multiple times
4. Views should be updated when schema upgrades occur (new columns added to existing event types)

### Option B: Separate utility method

[...]

Issue #4636, opened 2026-02-26
BigQueryAgentAnalyticsPlugin Deadlock on Linux Multiprocessing


evekhm:
**Component:** [BigQueryAgentAnalyticsPlugin](file:///usr/local/google/home/evekhm/projects/adk/agent-operations/.venv/lib/python3.13/site-packages/google/adk/plugins/bigquery_agent_analytics_plugin.py#1571-2553)
**Severity:** High (Breaks common multiprocessing patterns on Linux)
**Environment:** Linux (where `multiprocessing` defaults to `fork`), Python 3.10+
## Summary
The [BigQueryAgentAnalyticsPlugin](https://github.com/google/adk-python/tree/main/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#1571-2553) causes deadlocks and `TimeoutError`s when used with `concurrent.futures.ProcessPoolExecutor` on Linux.

This occurs because the plugin initializes a `BigQueryWriteAsyncClient` (and its underlying gRPC channel) immediately in its [__init__](https://github.com/google/adk-python/tree/main/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#1103-1113) method. The standard Python `grpc` library is **not fork-safe**. When a process forks (default start method on Linux), the child process inherits the parent's memory, including the already-open (and now broken) gRPC channel, leading to hangs or timeouts when the child tries to use it.

## Root Cause Analysis
1.  **Eager Initialization**: `BigQueryAgentAnalyticsPlugin.__init__` creates `self.write_client = BigQueryWriteAsyncClient(...)`. This establishes a gRPC channel in the parent process.
2.  **Fork Safety**: On Linux, `multiprocessing.Process` (and `ProcessPoolExecutor`) defaults to `fork()`. The child process inherits the parent's address space.
3.  **gRPC Deadlock**: The inherited gRPC channel is in a bad state in the child process. Any attempt to use it (even just waiting for connectivity) results in a deadlock or `TimeoutError` (after 30s).

## Why It Works on macOS / Windows
-   **macOS / Windows**: The default multiprocessing start method is `spawn`.
[...]

PR #4640, opened 2026-02-27
fix: add fork-safety and auto-create analytics views to BQ plugin

caohy1988:
## Summary
- **Fork-safety (#4636):** Adds PID tracking to `BigQueryAgentAnalyticsPlugin` so that forked child processes detect stale gRPC channels and re-initialize instead of deadlocking. Uses the standard `os.getpid()` pattern (same as SQLAlchemy, gRPC-python).
- **Auto-create views (#4639):** After ensuring the `agent_events` table exists, automatically creates 15 per-event-type BigQuery views (`v_llm_request`, `v_tool_completed`, etc.) that unnest JSON columns into typed, queryable columns. Controlled by `BigQueryLoggerConfig.create_views` (default `True`), idempotent via `CREATE OR REPLACE VIEW`.
- **Trace-ID continuity & o11y alignment (#4645):** Fixes trace_id fracture between early events (USER_MESSAGE_RECEIVED, INVOCATION_STARTING) and later events (AGENT_STARTING onwards) when no ambient OTel span exists. Also aligns BQ rows with Cloud Trace span IDs when o11y is active.
- **Span-ID consistency under ambient OTel (#4640 review):** Fixes `*_STARTING` / `*_COMPLETED` events producing mismatched span IDs when an ambient OTel span is active. Completion callbacks now check for ambient spans and defer to `_resolve_ids` Layer 2 instead of overriding with plugin-synthetic IDs.
- **Stack leak safety (#4640 review):** Adds `TraceManager.clear_stack()` and makes `ensure_invocation_span()` clear stale records from *different* invocations, preventing span stack leaks across invocations. Uses `_active_invocation_id_ctx` to distinguish stale leak vs same-invocation re-entry.
- **Root agent name staleness fix:** `init_trace()` now refreshes `_root_agent_name_ctx` unconditionally on each invocation (previously set-once-on-None). `after_run_callback` resets it alongside other invocation cleanup.
- **Exception-safe cleanup:** `after_run_callback` uses `try/finally` to guarantee invocation state (`clear_stack`, `_active_invocation_id_ctx`, `_root_agent_name_ctx`) is always reset, even if `_log_event` raises.
- **Bonus: `on_tool_error_callback` span fix:** Previously discarded the span_id from `pop_span()`, causing TOOL_ERROR events to get the wrong span. Now captures and uses the popped span_id.

### Trace-ID fix details (#4645)

**Root cause:** When OTel is configured but the Runner's ambient span is absent (Agent Engine, custom runners), early events fell back to `invocation_id` while `push_span("agent")` created a new OTel span with a different hex trace_id.

**Fix:**
- `TraceManager.ensure_invocation_span()` — called in `on_user_message_callback` and `before_run_callback` before any events fire. Attaches the ambient span if valid, or creates a new root span. Idempotent within the same invocation via `_active_invocation_id_ctx`.
- `push_span()` — now creates child spans under existing stack parent via `trace.set_span_in_context()`, so all spans share one trace_id.
- `_resolve_ids()` (replaces `_resolve_span_ids()`) — 3-layer priority for trace_id/span_id/parent_span_id:
  1. `EventData` overrides (post-pop callbacks like `after_run_callback`)
  2. Ambient OTel span (aligns BQ rows with Cloud Trace when o11y is active)
  3. Plugin's internal span stack (fallback for no-ambient paths)
- `EventData.trace_id_override` — captured before popping the invocation span in `after_run_callback`, ensuring INVOCATION_COMPLETED shares the same trace_id.

**Result when o11y is enabled:** BQ rows use the same trace_id and span_id as the framework's Cloud Trace spans (invocation, invoke_agent, call_llm, execute_tool).

### Span-ID consistency fix details

**Root cause:** `*_STARTING` events used ambient framework span IDs via `_resolve_ids` Layer 2, while `*_COMPLETED` events bypassed Layer 2 by passing explicit `span_id_override` / `parent_span_id_override` from the plugin's internal popped span. This meant a single lifecycle (e.g. AGENT_STARTING → AGENT_COMPLETED) could mix ambient framework IDs and plugin synthetic IDs.

**Fix:** Six completion callbacks (`after_agent`, `after_model`, `on_model_error`, `after_tool`, `on_tool_error`, `after_run`) now check `trace.get_current_span().get_span_context().is_valid` and pass `None` overrides when ambient exists, letting `_resolve_ids` Layer 2 use the framework's ambient span — keeping STARTING/COMPLETED pairs consistent.

**Span-ID resolution contract:**
- When OTel is active: BQ rows use the same trace/span/parent IDs as Cloud Trace. STARTING and COMPLETED events in the same lifecycle share the same span_id.
- When OTel is not active: BQ rows use the plugin's internal span stack. STARTING gets the current top-of-stack; COMPLETED gets the popped span.

### Stack leak safety details

**Root cause:** `_span_records_ctx` is not keyed by invocation and relies on balanced push/pop. If `after_run_callback` is skipped (e.g. exception in generator, asyncio cancellation), stale records leak into the next invocation.

**Fix:**
- `_active_invocation_id_ctx` — new contextvar tracking the invocation that owns the current stack. Allows `ensure_invocation_span()` to distinguish same-invocation re-entry (idempotent no-op) from stale leak (clear and re-init).
- `TraceManager.clear_stack()` — ends all owned spans and resets the stack to `[]`.
- `ensure_invocation_span()` — idempotent within the same invocation; clears stale records only when the invocation_id differs.
- `after_run_callback` — uses `try/finally` to guarantee `clear_stack()`, `_active_invocation_id_ctx`, and `_root_agent_name_ctx` are always reset, even if `_log_event` raises (since `_safe_callback` would swallow the exception silently).

### Known limitation

Trace context storage (`_span_records_ctx`, `_active_invocation_id_ctx`, `_root_agent_name_ctx`) is module-global, not plugin-instance-scoped. This is safe in practice because: (1) `PluginManager` enforces name-uniqueness, preventing duplicate BQ plugins on the same Runner; (2) concurrent asyncio tasks get isolated contextvar copies. Acknowledged in code comment.

## Test plan
- [x] `TestForkSafety` (4 tests): PID mismatch triggers re-init, same PID skips reset, `_reset_runtime_state` clears fields, `__getstate__` zeroes PID
- [x] `TestAnalyticsViews` (8 tests): views created on new/existing tables, disabled when `create_views=False`, errors logged not raised, SQL correctness, config default, explicit async method, startup failure
- [x] `TestTraceIdContinuity` (5 tests): no-ambient span continuity, ambient span continuity, cross-turn isolation, INVOCATION_COMPLETED continuity, full callback-chain integration
- [x] `TestResolveIds` (7 tests): plugin stack defaults, span_id/parent_span_id overrides, ambient OTel priority, override-beats-ambient, ambient root no self-parent, ambient span used for completed events
- [x] `TestSpanIdConsistency` (3 tests): STARTING/COMPLETED same span with ambient, plugin span without ambient, tool error captures span_id
- [x] `TestStackLeakSafety` (6 tests): ensure_invocation_span clears stale records, clear_stack ends owned spans, after_run_callback clears remaining stack, next invocation clean after incomplete previous, ensure_invocation_span idempotent within same invocation, on_user_message → before_run same trace_id regression test
- [x] `TestRootAgentNameAcrossInvocations` (1 test): two invocations with different root agents log correct names
- [x] `TestAfterRunCleanupExceptionSafety` (1 test): cleanup runs even when _log_event raises
- [x] All 173 tests pass

### DIFF
--- a/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
+++ b/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
@@ -39,6 +39,13 @@ from typing import TYPE_CHECKING
 import uuid
 import weakref
 
+from google.adk.agents.callback_context import CallbackContext
+from google.adk.models.llm_request import LlmRequest
+from google.adk.models.llm_response import LlmResponse
+from google.adk.plugins.base_plugin import BasePlugin
+from google.adk.tools.base_tool import BaseTool
+from google.adk.tools.tool_context import ToolContext
+from google.adk.version import __version__
 from google.api_core import client_options
 from google.api_core.exceptions import InternalServerError
 from google.api_core.exceptions import ServiceUnavailable
@@ -55,16 +62,8 @@ from google.genai import types
 from opentelemetry import trace
 import pyarrow as pa
 
-from ..agents.callback_context import CallbackContext
-from ..models.llm_request import LlmRequest
-from ..models.llm_response import LlmResponse
-from ..tools.base_tool import BaseTool
-from ..tools.tool_context import ToolContext
-from ..version import __version__
-from .base_plugin import BasePlugin
-
 if TYPE_CHECKING:
-  from ..agents.invocation_context import InvocationContext
+  from google.adk.agents.invocation_context import InvocationContext
 
 logger: logging.Logger = logging.getLogger("google_adk." + __name__)
 tracer = trace.get_tracer(


caohy1988:
ERROR: Found 10 breaking change(s) in API surface.
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(span_id_override): Positional parameter was moved
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(parent_span_id_override): Positional parameter was moved
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(latency_ms): Positional parameter was moved
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(time_to_first_token_ms): Positional parameter was moved
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(model): Positional parameter was moved
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(model_version): Positional parameter was moved
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(usage_metadata): Positional parameter was moved
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(status): Positional parameter was moved
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(error_message): Positional parameter was moved
src/google/adk/plugins/bigquery_agent_analytics_plugin.py:0: EventData.__init__(extra_attributes): Positional parameter was moved

caohy1988:
Good catch — the API surface checker is correct.

**Root cause**: I added `trace_id_override` as the first field in the `EventData` dataclass, which shifted all 10 existing positional parameters (`span_id_override`, `parent_span_id_override`, `latency_ms`, etc.).

**Fix** (commit b5bdc30f): Moved `trace_id_override` to the end of the field list (after `extra_attributes`), preserving the original positional parameter order. All fields have defaults so keyword usage (which is how the plugin itself and all tests use `EventData`) is unaffected.

All 161 unit tests pass after the fix.

PR #4642, opened 2026-02-27
fix: prevent state_delta overwrite on function_response-only events

### LINKED ISSUE #3178: Overridden `state_delta` by empty string in `after_tool_callback` after `skip_summarization`
👋 Hi everyone, thank you in advance for taking your time to review this issue!

**Describe the bug**

When updating the session state using the `state_delta` field within an `EventActions` object (returned by a custom `after_tool_callback`), the custom state update is lost if the `Event` contains a `function_response` but no `text` content field. This issue specifically occurs when `skip_summarization` is set to `True`. As a result, even though the `state_delta` field is correctly set in the emitted event, the state is not updated and the field appears as an empty string in the final state.

**To Reproduce**

Given a `FunctionTool` wrapping the following function:

```python
async def custom_tool_function(input: str) -> dict[str, any]:
    ...
    return {
        "result": result,
        "status": "success",
    }
```

And a `callback` function:

```python
def custom_after_tool_callback(
    tool: base_tool.BaseTool,
    args: typing.Dict[str, typing.Any],
    tool_context: tool_context.ToolContext,
    tool_response: typing.Dict[str, typing.Any],
) -> None:
    ...
    tool_context.actions.skip_summarization = True
    tool_context.actions.state_delta[output_key] = result
    return None
```

Steps to reproduce:

 1. Call the custom `tool` and trigger the callback.
 2. Set `skip_summarization = True` and update `state_delta` with custom tool output.
 3. Emit the event, confirming that the `state_delta` is included correctly.
 4. Check the final state (or event before updating the state), where the expected value in `state_delta` is an empty string.

**Observed behavior**

The emitted event initially contains the correct `state_delta`, but the final state includes the `state_delta` with an empty string as the value for the respective `output_key`, despite the `state_delta` being correctly set in the event. The `state_delta` is not stored as expected.

Here is the current code in the flow:

_Event Emission (Expected behavior)_

Initially, the event contains the correct `state_delta`:

```
Event(
  content=Content(
    parts=[
      Part(
        function_response=FunctionResponse(
          id='call_K74WBIwpMhj35PJIOJk7ILbQ',
          name='<... function ...>',
          response={
            'result': [<... items ...>],
            'status': 'success'
          }
        )
      ),
    ],
    role='user'
  ), 
  actions=EventActions(
    skip_summarization=True, 
    state_delta={'<... function name ...>': [<... items ...>]}, 
  ), 
  id='29c25193-2f5f-4886-8d3b-de6356cd245d', 
  timestamp=1760448942.311269
)
```

_Final State (Observed behavior)_

The final state incorrectly shows an empty string for the `state_delta`:

```
Event(
  content=Content(
    parts=[
      Part(
        function_response=FunctionResponse(
          id='call_K74WBIwpMhj35PJIOJk7ILbQ',
          name='<... function ...>',
          response={
            'result': [<... items ...>],
            'status': 'success'
          }
        )
      ),
    ],
    role='user'
  ), 
  actions=EventActions(
    skip_summarization=True, 
    state_delta={'<... function name ...>': ''}, 
  ), 
  id='29c25193-2f5f-4886-8d3b-de6356cd245d', 
  timestamp=1760448942.311269
)
```

_Methods Involved_

`is_final_response()` currently used to check if the event is a final response (in [code](https://github.com/google/adk-python/blob/d82c4921404d638eda31ac42788ffb14d54854e7/src/google/adk/events/event.py#L82)):

```python
def is_final_response(self) -> bool:
    """Returns whether the event is the final response of an agent."""
    if self.actions.skip_summarization or self.long_running_tool_ids:
        return True
    ...
```

`__maybe_save_output_to_state()` used to save the model output to state (in [code](https://github.com/google/adk-python/blob/d82c4921404d638eda31ac42788ffb14d54854e7/src/google/adk/agents/llm_agent.py#L663)):

```python
def __maybe_save_output_to_state(self, event: Event):
    """Saves the model output to state if needed."""
    if event.author != self.name:
        logger.debug('Skipping output save for agent %s: event authored by %s', self.name, event.author)
        return
    
    if (
        self.output_key
        and event.is_final_response()
        and event.content
        and event.content.parts
    ):
        result = ''.join(
            part.text
            for part in event.content.parts
            if part.text and not part.thought
        )
        if self.output_schema:
            if not result.strip():
                return
            result = self.output_schema.model_validate_json(result).model_dump(
                exclude_none=True
            )
        event.actions.state_delta[self.output_key] = result
```

As seen in the code, the `__maybe_save_output_to_state` method looks for a `text` field within the event content and uses it to determine the final result. However, in cases where `skip_summarization` is `True` and no text part exists, the method sets the state_delta to an empty string. This is the case here, as we only have a `function_response`.

**Expected behavior**

The state update should persist in the final state with the correct value in `state_delta`. If the `state_delta` is correctly set in the emitted event, it should be stored properly in the final state, without being overridden by an empty string. This should not only work with a `text` content, but also work with a `function_response` in the case of a use of `skip_summerization`.

**Workaround**

Interestingly, a workaround that seems to solve the issue is by setting an `output_schema` on the `LlmAgent`:

While this workaround does seem to correctly store the state, this should technically not work according to the [documentation](https://google.github.io/adk-docs/agents/llm-agents/#structuring-data-input_schema-output_schema-output_key), which states that `output_schema` should not be set when using `tools`. Despite this, the workaround successfully resolves the issue by triggering the expected behavior.

**Potential solution**

The issue seems to stem from the method `__maybe_save_output_to_state`, which checks for `event.is_final_response()` and attempts to extract a text part from the event. Since `skip_summarization` is set to `True`, the `Event` is marked as a final response, but when no text content is available, the result defaults to an empty string.

A potential solution is to modify the `__maybe_save_output_to_state` method to handle cases where there is no text field but a valid `state_delta` exists. The method could check if `state_delta` is populated in the event and save the values to the state directly, bypassing the text extraction logic.

```python
def __maybe_save_output_to_state(self, event: Event):
    """Saves the model output to state if needed."""
    if event.author != self.name:
        logger.debug('Skipping output save for agent %s: event authored by %s', self.name, event.author)
        return
    
    # Check if state_delta is populated, even if there is no text content
    if event.actions.state_delta:
        for key, value in event.actions.state_delta.items():
            self.state[key] = value
        return

    if (
        self.output_key
        and event.is_final_response()
        and event.content
        and event.content.parts
    ):
        result = ''.join(
            part.text
            for part in event.content.parts
            if part.text and not part.thought
        )
        if self.output_schema:
            if not result.strip():
                return
            result = self.output_schema.model_validate_json(result).model_dump(
                exclude_none=True
            )
        event.actions.state_delta[self.output_key] = result
```

This way, if `state_delta` is set, the method will prioritize saving it directly, even if no text content is present.

**Desktop**

- OS: macOS Sequoia 15.6.1 (24G90)
- Python version: Python 3.13.8
- ADK version: 1.16.0

**Model Information**

- Are you using LiteLLM: Yes
- Which model is being used (e.g., gemini-2.5-pro): azure/gpt-4.1

👉 I hope the problem description and inputs help to figure out the issues and a potential resolution. If you're interested, I can also create a PR to fix this issue based on your feedbacks and opinions!

yuvrajangadsingh:
Fixes #3178

When `skip_summarization=True` and an event only has `function_response` parts (no text), `__maybe_save_output_to_state` joins zero text parts into `""` and overwrites whatever the `after_tool_callback` already set in `state_delta`.

**Root cause:** `is_final_response()` returns `True` for `skip_summarization` events, so the method enters the save block. But there are no text parts to extract, so `result = ""`. Without `output_schema`, the empty string gets written directly to `state_delta[output_key]`, clobbering the callback's value.

**Fix:** skip writing to `state_delta` when `result` is empty and no `output_schema` is configured. This preserves callback-set values on function_response-only events while keeping existing behavior for text-based responses unchanged.

**Testing:** added a test that sets `state_delta` via callback on a function_response-only event with `skip_summarization=True`, then verifies the value isn't overwritten. All 406 agent tests pass.

### DIFF
--- a/src/google/adk/agents/llm_agent.py
+++ b/src/google/adk/agents/llm_agent.py
@@ -856,6 +856,12 @@ class LlmAgent(BaseAgent):
         if not result.strip():
           return
         result = validate_schema(self.output_schema, result)
+      elif not result:
+        # No text parts found and no output_schema. Skip to avoid
+        # overwriting state_delta values already set by callbacks
+        # (e.g. after_tool_callback with skip_summarization on
+        # function_response-only events).
+        return
       event.actions.state_delta[self.output_key] = result
 
   @model_validator(mode='after')
--- a/tests/unittests/agents/test_llm_agent_output_save.py
+++ b/tests/unittests/agents/test_llm_agent_output_save.py
@@ -276,3 +276,34 @@ class TestLlmAgentOutputSave:
     # ASSERT: Because the method should return early, the state_delta
     # should remain empty.
     assert len(event.actions.state_delta) == 0
+
+  def test_maybe_save_output_to_state_skips_function_response_only_event(self):
+    """Test that state_delta set by callback is not overwritten when event
+    only has function_response parts and no text."""
+    agent = LlmAgent(name="test_agent", output_key="result")
+
+    # Simulate a function_response-only event (no text parts)
+    parts = [
+        types.Part(
+            function_response=types.FunctionResponse(
+                name="my_tool",
+                response={"status": "success", "data": [1, 2, 3]},
+            )
+        )
+    ]
+    content = types.Content(role="user", parts=parts)
+
+    event = Event(
+        invocation_id="test_invocation",
+        author="test_agent",
+        content=content,
+        actions=EventActions(
+            skip_summarization=True,
+            state_delta={"result": [1, 2, 3]},
+        ),
+    )
+
+    agent._LlmAgent__maybe_save_output_to_state(event)
+
+    # The callback-set value should be preserved, not overwritten with ""
+    assert event.actions.state_delta["result"] == [1, 2, 3]


rohityan:
Hi @yuvrajangadsingh , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

yuvrajangadsingh:
hey @xuanyang15, just checking in on this one. @rohityan tagged you for review about 9 days ago. small fix for state_delta getting overwritten on function_response-only events. let me know if anything needs changing.

yuvrajangadsingh:
@rohityan rebased onto latest main, clean rebase. issue is tagged help wanted + needs review. this prevents state_delta from getting overwritten by empty dicts on function_response-only events.

yuvrajangadsingh:
@rohityan @xuanyang15 @Jacksunwei @aneeshgarg this has been open for over a month now. xuanyang15 was tagged twice, no response. all feedback addressed, rebased on main, ci green. issue #3178 is tagged help wanted + needs review. can someone take a look? small fix, prevents state_delta from getting overwritten on function_response-only events.

yuvrajangadsingh:
@xuanyang15 any update on this one? CI green, ready for review.

yuvrajangadsingh:
@GWeale @DeanChensj since you've both been active in src/google/adk/sessions/ recently (the rewind state-preservation fix + the raw_event work), would either of you have a moment to look at this? small fix preventing state_delta from getting overwritten by empty dicts on function_response-only events. fixes #3178, ci green, addressed all gemini bot review.

@xuanyang15 was tagged 4 times since feb but hasn't been able to get to it.

xuanyang15:
@yuvrajangadsingh Sorry for the delay! It didn't catch my attention because neither the PR nor the issue was assigned to me. I just added me as the assignee, will work on it as soon as possible.

PR #4652, opened 2026-02-28
fix(cli): filter non-agent directories from list_agents

### LINKED ISSUE #4647: list_agents() returns non-agent directories in /list-apps endpoint
## 🔴 Required Information

**Describe the Bug:**
The `/list-apps` endpoint returns directories that are not valid agents. Any subdirectory inside the agents directory (e.g. `tmp/`, `data/`, `utils/`) is returned in the response, even when it contains no agent definition (`root_agent.yaml`, `agent.py`, or `__init__.py`). This affects both the ADK web UI (where non-agent directories appear in the agent selector) and any production deployment depending on this API to list available agents.

**Steps to Reproduce:**
1. Create a project directory with a valid agent and some non-agent subdirectories:
```
my_project/
├── my_agent/
│   ├── __init__.py
│   └── agent.py       # valid agent
├── utils/              # not an agent
├── data/               # not an agent
└── tmp/                # not an agent
```
2. Run `adk web .` from `my_project/`
3. Open the web UI
4. Observe that `utils`, `data`, and `tmp` appear in the agent selector alongside `my_agent`

**Expected Behavior:**
Only directories containing a valid agent definition should appear in the agent list.

**Observed Behavior:**
All non-hidden subdirectories are listed regardless of whether they contain an agent. Selecting a non-agent directory results in a loading error.

**Environment Details:**

- ADK Library Version: 1.26.0
- Desktop OS: macOS (Darwin 25.2.0)
- Python Version: 3.12.10

**Model Information:**

- Are you using LiteLLM: N/A
- Which model is being used: N/A (issue is in agent discovery, not model interaction)

---

## 🟡 Optional Information

**Regression:**
No, this has been the behavior since `list_agents()` was introduced. PR #3430 (closing #3429) added a `list_agents_detailed()` endpoint that filters properly by loading each agent, but the web UI frontend never adopted it and still calls the plain `/list-apps` endpoint.

**Logs:**
N/A - no error on the backend, the non-agent directories are simply returned in the API response.

**Screenshots / Video:**
(non-agent directories listed) ( from the adk web)
Image

**Additional Context:**
The root cause is in `AgentLoader.list_agents()` which only filters by `os.path.isdir()`, hidden directory prefix, and `__pycache__`. It does not verify the directory contains a recognized agent structure (`root_agent.yaml`, `agent.py`, or `__init__.py`).

Related:
- Issue #3429
- PR #3430

```python
from pathlib import Path
from google.adk.cli.utils.agent_loader import AgentLoader
import tempfile

with tempfile.TemporaryDirectory() as tmp:
 (Path(tmp) / "real_agent").mkdir()
 (Path(tmp) / "real_agent" / "__init__.py").write_text("from google.adk.agents.base_agent import BaseAgent\nclass A(BaseAgent):\n  def __init__(self): super().__init__(name='a')\nroot_agent = A()")
 (Path(tmp) / "not_an_agent").mkdir()

 loader = AgentLoader(tmp)
 print(loader.list_agents())
 # Returns: ['not_an_agent', 'real_agent']
 # Expected: ['real_agent']
```

 - Always (100%)

giulio-leone:
## Problem

`list_agents()` returns all non-hidden subdirectories from the agents directory, including those that contain no agent definition (e.g. `utils/`, `data/`, `tmp/`). These appear in the web UI agent selector and `/list-apps` endpoint, and fail to load when selected.

## Root Cause

The directory filter only checks `isdir()`, `not startswith('.')`, and `!= '__pycache__'` — it never validates that the directory actually contains an agent definition.

## Fix

Add a `_is_agent_dir()` static method that checks for the presence of `root_agent.yaml`, `agent.py`, or `__init__.py` — the same files that `_determine_agent_language()` already relies on — and use it as an additional filter in `list_agents()`.

## Test

286 CLI tests pass, 2 consecutive clean runs.

Fixes #4647

### DIFF
--- a/src/google/adk/cli/utils/agent_loader.py
+++ b/src/google/adk/cli/utils/agent_loader.py
@@ -332,6 +332,15 @@ class AgentLoader(BaseAgentLoader):
     return agent_or_app
 
   @override
+  @staticmethod
+  def _is_agent_dir(path: Path) -> bool:
+    """Check if a directory contains a valid agent definition."""
+    return (
+        (path / 'root_agent.yaml').exists()
+        or (path / 'agent.py').exists()
+        or (path / '__init__.py').exists()
+    )
+
   def list_agents(self) -> list[str]:
     """Lists all agents available in the agent loader (sorted alphabetically)."""
     base_path = Path.cwd() / self.agents_dir
@@ -341,6 +350,7 @@ class AgentLoader(BaseAgentLoader):
         if os.path.isdir(os.path.join(base_path, x))
         and not x.startswith(".")
         and x != "__pycache__"
+        and self._is_agent_dir(base_path / x)
     ]
     agent_names.sort()
     return agent_names


giulio-leone:
Closing — CLA not yet signed. Will resubmit when ready.

Issue #764, opened 2025-05-16
unnecessary excessive DB calls?


mohammad-passage:
** Please make sure you read the contribution guide and file the issues in the rigth place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
We ran an ADK agent over 3500 objects and that took our database down.
Looking at queries it seems this query is causing it:

```
UPDATE app_states SET state=$1::JSONB, update_time=now() WHERE app_states.app_name = $2
```
I don't ever see a case that we actually want state for apps yet this is causing huge load on the DB.  Also I don't have any state with prefix `app:` so I was not expecting this DB all at all.

**To Reproduce**
N/A

**Expected behavior**
N/A
**Screenshots**
![Image](https://github.com/user-attachments/assets/d7a5a77c-8ee6-481f-a03d-146ea8a2e893)

**Desktop (please complete the following information):**
N/A
**Additional context**
N//a

seanzhou1023:
@mohammad-passage thanks for reporting the issue. Are you able to provide some agent codes and sample user queries that we can reproduce ?

mohammad-passage:
Our code has no app: state so I guess any random agent has same issue that creates an empty app state and suffers from hot key update at  scale.

DeanChensj:
Thanks for raising this issue. It will be fixed when https://github.com/google/adk-python/commit/d83ee1a6f20a52df9c30aacf9fd0e5013d98f32d is included in next release.

PR #4656, opened 2026-02-28
fix(sessions): skip unnecessary FOR UPDATE lock on app/user state rows

### LINKED ISSUE #4655: DatabaseSessionService.append_event: unnecessary FOR UPDATE lock on app_states/user_states when no state delta exists
## Description

`DatabaseSessionService.append_event()` unconditionally acquires `SELECT ... FOR UPDATE` on both `app_states` and `user_states` tables, even when the event carries no state delta for those scopes.

This was partially addressed by #764, which fixed the unnecessary `UPDATE` (merge) side. However, the `SELECT ... FOR UPDATE` (row-level lock acquisition) still happens unconditionally.

## Problem

In `database_session_service.py` (lines 553-578 in v1.26.0), `_select_required_state` is called with `use_row_level_locking=use_row_level_locking` for both `app_states` and `user_states`, regardless of whether the event actually has `app:` or `user:` prefixed state deltas.

```python
storage_app_state = await _select_required_state(
    sql_session=sql_session,
    state_model=schema.StorageAppState,
    predicates=(schema.StorageAppState.app_name == session.app_name,),
    use_row_level_locking=use_row_level_locking,  # Always True on PostgreSQL
    ...
)
```

Since `app_states` is keyed by `app_name` alone, **all concurrent `append_event` calls within the same app serialize on this single row lock**, even if none of them have `app:` state deltas. In practice, the vast majority of events only carry session-scoped state, making this lock contention unnecessary.

## Impact

For applications with moderate concurrency (e.g., multiple agents within the same app writing events simultaneously), this creates a significant bottleneck:

- All `append_event` calls for the same `app_name` are serialized by the `FOR UPDATE` lock on `app_states`
- The `user_states` lock similarly serializes all calls for the same `(app_name, user_id)` pair
- This happens even when the event has zero state delta (e.g., plain conversation events)

## Suggested Fix

Only acquire `FOR UPDATE` when the event actually has a delta for the corresponding scope:

```python
# Pre-analyze deltas before the transaction
has_app_delta = bool(state_deltas.get("app")) if state_deltas else False
has_user_delta = bool(state_deltas.get("user")) if state_deltas else False

# Then in the transaction:
storage_app_state = await _select_required_state(
    sql_session=sql_session,
    state_model=schema.StorageAppState,
    predicates=(schema.StorageAppState.app_name == session.app_name,),
    use_row_level_locking=use_row_level_locking and has_app_delta,
    ...
)
storage_user_state = await _select_required_state(
    sql_session=sql_session,
    state_model=schema.StorageUserState,
    predicates=(...),
    use_row_level_locking=use_row_level_locking and has_user_delta,
    ...
)
```

The state delta can be extracted before the DB transaction using `extract_state_delta()` since it only reads from the event object and has no side effects.

## Environment

- Database: PostgreSQL (with `asyncpg`)
- Python: 3.12

giulio-leone:
## Summary

Fixes #4655

`DatabaseSessionService.append_event()` unconditionally acquires `SELECT ... FOR UPDATE` on both `app_states` and `user_states` tables, even when the event carries no state delta for those scopes.

## Problem

Since `app_states` is keyed by `app_name` alone, **all concurrent `append_event` calls within the same app serialize on this single row lock**, even when they only carry session-scoped state (the vast majority of events). The `user_states` lock similarly serializes all calls for the same `(app_name, user_id)` pair.

This was partially addressed by #764, which fixed the unnecessary `UPDATE` (merge) side. However, the `SELECT ... FOR UPDATE` (row-level lock acquisition) still happened unconditionally.

## Fix

Pre-analyze the event's `state_delta` before entering the transaction to determine which scopes actually need write locks:

```python
# Only acquire FOR UPDATE when the event actually modifies that scope
has_app_delta = bool(state_deltas.get("app"))
has_user_delta = bool(state_deltas.get("user"))

storage_app_state = await _select_required_state(
    ...,
    use_row_level_locking=use_row_level_locking and has_app_delta,
)
storage_user_state = await _select_required_state(
    ...,
    use_row_level_locking=use_row_level_locking and has_user_delta,
)
```

The state rows are still fetched (needed for the stale-timestamp reload path), but without the `FOR UPDATE` clause when there's nothing to write.

This also eliminates a redundant call to `extract_state_delta()` — the deltas are now computed once and reused.

## Impact

- Events with no `app:`/`user:` state deltas no longer block on the shared row lock
- Concurrent `append_event` throughput improves significantly for apps with moderate concurrency
- No behavioral change — the same state merges happen when deltas are present

### DIFF
--- a/src/google/adk/sessions/database_session_service.py
+++ b/src/google/adk/sessions/database_session_service.py
@@ -554,15 +554,13 @@ class DatabaseSessionService(BaseSessionService):
         # write locks. Most events carry only session-scoped state (or no
         # state at all), so acquiring FOR UPDATE on app_states / user_states
         # unnecessarily serializes all concurrent append_event calls.
-        has_app_delta = False
-        has_user_delta = False
-        state_deltas = None
-        if event.actions and event.actions.state_delta:
-          state_deltas = _session_util.extract_state_delta(
-              event.actions.state_delta
-          )
-          has_app_delta = bool(state_deltas.get("app"))
-          has_user_delta = bool(state_deltas.get("user"))
+        state_deltas = (
+            _session_util.extract_state_delta(event.actions.state_delta)
+            if event.actions and event.actions.state_delta
+            else None
+        )
+        has_app_delta = bool(state_deltas and state_deltas.get("app"))
+        has_user_delta = bool(state_deltas and state_deltas.get("user"))
 
         storage_app_state = await _select_required_state(
             sql_session=sql_session,


giulio-leone:
Closing — CLA not yet signed. Will resubmit when ready.

PR #4661, opened 2026-02-28
fix: skip unnecessary FOR UPDATE locks in append_event when no app/user state delta

### LINKED ISSUE #4655: DatabaseSessionService.append_event: unnecessary FOR UPDATE lock on app_states/user_states when no state delta exists

yuvrajangadsingh:
`DatabaseSessionService.append_event()` acquires `SELECT ... FOR UPDATE` on both `app_states` and `user_states` unconditionally, even when the event carries no delta for those scopes.

Since `app_states` is keyed by `app_name` alone, this means all concurrent `append_event` calls within the same app serialize on that single row lock. Most events only carry session-scoped state, so this lock contention is unnecessary.

This was partially addressed by #764 which fixed the unnecessary UPDATE (merge) side, but the SELECT ... FOR UPDATE (row-level lock acquisition) still happens unconditionally.

The fix pre-analyzes the event's state delta before the transaction using `extract_state_delta()` (which only reads from the event object, no side effects) and only passes `use_row_level_locking=True` for scopes that actually have pending writes.

Added 5 tests covering all combinations: no delta, session-only delta, app-only delta, user-only delta, and both app+user deltas. All 69 session tests pass.

Fixes #4655

### DIFF
--- a/src/google/adk/sessions/database_session_service.py
+++ b/src/google/adk/sessions/database_session_service.py
@@ -536,12 +536,13 @@ class DatabaseSessionService(BaseSessionService):
     # FOR UPDATE locks on rows that will actually be written.
     # The result is reused later to avoid calling extract_state_delta twice.
     state_deltas = None
+    has_app_delta, has_user_delta = False, False
     if event.actions and event.actions.state_delta:
       state_deltas = _session_util.extract_state_delta(
           event.actions.state_delta
       )
-    has_app_delta = bool(state_deltas and state_deltas.get("app"))
-    has_user_delta = bool(state_deltas and state_deltas.get("user"))
+      has_app_delta = bool(state_deltas.get("app"))
+      has_user_delta = bool(state_deltas.get("user"))
 
     async with self._with_session_lock(
         app_name=session.app_name,


yuvrajangadsingh:
addressed the review feedback in 7c416ae6. extracted a `lock_spy_harness` fixture and collapsed all 5 tests into a single parametrized test.

/gemini review

yuvrajangadsingh:
addressed the second review comment in 9ec1187d. now reusing the pre-analyzed `state_deltas` variable instead of calling `extract_state_delta` twice.

/gemini review

yuvrajangadsingh:
done in adb51fb4, moved the bool checks inside the if block.

/gemini review

yuvrajangadsingh:
keeping the `event.actions and event.actions.state_delta` guard as-is. this pattern is used consistently across the codebase (in_memory_session_service.py, sqlite_session_service.py, bigquery_agent_analytics_plugin.py) and changing it here would be inconsistent. the `event.actions` check is defensive and i'd rather match the existing convention.

same reasoning for `.get()` vs direct access, both work fine but `.get()` is more defensive for a dict that's being read outside its creation scope.

/gemini review

yuvrajangadsingh:
@rusherman implemented the fix you suggested in #4655. let me know if you have any feedback.

@xuanyang15 @ryanaiagent would appreciate a review when you get a chance.

rohityan:
Hi @yuvrajangadsingh , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Closing this since it has been fixed in this commit [d61846f](https://github.com/google/adk-python/commit/d61846f6c6dd5e357abb0e30eaf61fe27896ae6a).

PR #4664, opened 2026-02-28
fix: validate and coerce function tool argument types

### LINKED ISSUE #4612: ADK doesn't enforce argument types in function calling
## 🔴 Required Information

**Describe the Bug:**

ADK doesn't seem to check and enforce argument types in function calls. I'm very surprised.

**Steps to Reproduce:**

Code:

```py

def set_num(num: int) -> dict:
    print("DEBUG:", type(num), num)
    return {"status": "success"}

root_agent = Agent(
    model='gemini-3-flash-preview',
    name='root_agent',
    instruction="the user is debugging something. do whatever they instruct",
    tools=[set_num],
)
```

Test run:

```
Running agent root_agent, type exit to exit.
[user]: set num to 3.14159. I know it only accept int. but I'd like to see the error.
DEBUG: <class 'int'> 3
DEBUG: <class 'float'> 3.14159
[root_agent]: It appears that the system accepted the value `3.14159` without returning an error, despite the parameter being defined as an integer. This suggests that the value may have been automatically rounded or truncated by the underlying implementation.
[user]: set num to "foobar"
DEBUG: <class 'str'> foobar
[root_agent]: The tool unexpectedly returned `success` even when passed the string `"foobar"`. It seems the parameter type enforcement for `num` is not currently working as expected in the underlying system.
[user]: set num to null.       
DEBUG: <class 'NoneType'> None
[root_agent]: The system also returned `success` when `num` was set to `null`. It appears that the tool is accepting various types without validation errors.
```

**Expected Behavior:**

ADK enforces that invalid arguments don't go through, as an average LLM also believes:

Image

**Observed Behavior:**

Invalid arguments that have different types than the function signature requirements are accepted.

**Environment Details:**

 - Desktop OS: Linux
 - Python Version (python -V): 3.13.12

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: gemini-3-flash-preview

yuvrajangadsingh:
Fixes #4612

`_preprocess_args` currently only handles Pydantic model conversion but passes all other types through without any validation. This means a function declared with `num: int` happily accepts `"foobar"`, `3.14159`, or `None` at runtime.

This PR extends `_preprocess_args` to validate and coerce all annotated argument types using `pydantic.TypeAdapter`. When coercion fails, the validation error is returned to the LLM so it can self-correct and retry (same pattern as the existing missing-mandatory-args check).

**What changed:**

- `_preprocess_args` now returns `(args, validation_errors)` tuple
- Uses `pydantic.TypeAdapter.validate_python()` for all non-Pydantic-BaseModel annotated params
- Compatible types get coerced automatically (e.g. `"42"` -> `42` for `int`, `"red"` -> `Color.RED` for enum)
- Incompatible types produce a validation error returned to the LLM
- Existing Pydantic BaseModel handling is unchanged (graceful failure with warning log)
- All 3 call sites updated: `FunctionTool.run_async`, `CrewAITool.run_async`, sync tool path in `functions.py`

**Before (no validation):**
```python
def set_num(num: int):
    print(type(num), num)

# LLM sends num=3.14159 -> passes through as float, no error
# LLM sends num="foobar" -> passes through as string, no error
```

**After (validates and coerces):**
```python
# LLM sends num="42" -> coerced to int 42, function runs
# LLM sends num="foobar" -> error returned to LLM:
#   "Parameter 'num': expected type '<class 'int'>', validation error: ..."
# LLM can then retry with correct types
```

**Tests:** 17 new tests covering coercion (str->int, str->float, str->enum, list coercion, Optional handling, bool coercion), validation errors (invalid str for int, None for required int, multiple param errors), and run_async integration (error returned to LLM, valid coercion invokes function, enum validation). All 1776 existing tests continue to pass.


yuvrajangadsingh:
@cuihaoleo this implements the TypeAdapter-based validation you suggested in the issue comments. validation errors go back to the LLM so it can self-correct (the LangChain approach you mentioned). existing Pydantic model behavior is unchanged.

@llalitkumarrr @xuanyang15 would appreciate a review when you get a chance.

yuvrajangadsingh:
@gemini-code-assist addressed both suggestions in 08c9f557:

1. narrowed `except Exception` to `except (TypeError, NameError)` for the TypeAdapter fallback
2. extracted `_build_validation_error_response` helper on FunctionTool, all 3 call sites now use it

/gemini review

yuvrajangadsingh:
@gemini-code-assist addressed all 3 suggestions in 2d5b34fb:

1. pydantic model None check now uses `is_optional` too, consistent with the other types
2. `(TypeError, NameError)` fallback now logs a warning instead of silent skip
3. strengthened multi-param test: asserts exact error count (3) and checks all param names. pydantic rejects int->str in lax mode too, so all 3 params fail.

/gemini review

yuvrajangadsingh:
@gemini-code-assist applied in 75071217. error messages now show `int` instead of `<class 'int'>`.

/gemini review

yuvrajangadsingh:
@gemini-code-assist addressed in 6ed9ba18:

1. **test assertion (int->str)**: pushed back on this one. `pydantic.TypeAdapter(str).validate_python(123)` actually raises `ValidationError` in pydantic v2, it does NOT coerce int to str. verified locally. the test with 3 errors is correct.

2. **deferred logging**: applied, using `%s` formatting instead of f-string.

/gemini review

yuvrajangadsingh:
@gemini-code-assist the `get_origin(param.annotation) is Union` check is pre-existing code from before this PR, not something I introduced. and in practice it doesn't cause issues here because TypeAdapter handles `int | None` natively: `TypeAdapter(int | None).validate_python(None)` returns `None` correctly, no unwrapping needed.

fixing the Union vs UnionType handling across the whole method would be a good follow-up but it's out of scope for this PR since it affects the pre-existing Pydantic model path too.

/gemini review

yuvrajangadsingh:
@gemini-code-assist applied in e35c1308. hoisted the Optional None check before the Pydantic branch so it covers both paths in one place.

/gemini review

yuvrajangadsingh:
@gemini-code-assist the sync/async duplication in `_call_tool_in_thread_pool` is pre-existing code, not something introduced by this PR. I only updated the existing call site to handle the new tuple return from `_preprocess_args`.

refactoring the entire sync tool execution path to go through `run_async` is a valid improvement but it's a separate concern with its own risk surface. out of scope here.

/gemini review

yuvrajangadsingh:
@gemini-code-assist applied in cb122062. added `types.UnionType` handling for Python 3.10+ `T | None` syntax. the Optional unwrap now works for both `Optional[T]` and `T | None`.

/gemini review

yuvrajangadsingh:
@gemini-code-assist pushing back on both:

1. **Pydantic model error handling**: the graceful-failure behavior for Pydantic models is pre-existing and intentional. changing it to return validation errors would be a behavioral change for existing users who rely on the pass-through behavior. out of scope for this PR.

2. **verbose ValidationError**: the full error string from pydantic is actually useful for the LLM since it includes the exact field paths and expected types. truncating to just `msg` would lose context. the LLM can parse multi-line text fine.

/gemini review

yuvrajangadsingh:
@gemini-code-assist this is incorrect. pydantic v2 (2.12.5) does NOT coerce int to str in lax mode:

```
>>> pydantic.TypeAdapter(str).validate_python(123)
pydantic_core._pydantic_core.ValidationError: 1 validation error for str
  Input should be a valid string [type=string_type, input_value=123, input_type=int]
```

the test asserting 3 errors is correct. all three params fail validation: int->str, "not_a_number"->int, "not_a_bool"->bool.

/gemini review

yuvrajangadsingh:
@gemini-code-assist these are all repeat suggestions that have been addressed or explicitly declined in previous rounds:

- **int->str coercion**: already disproved with actual pydantic 2.12.5 output. `TypeAdapter(str).validate_python(123)` raises `ValidationError`. the test is correct.
- **Pydantic model error handling**: intentionally preserving pre-existing behavior, out of scope.
- **verbose ValidationError**: full error string is more useful for LLM self-correction.

no further changes needed on these. waiting for human review.

yuvrajangadsingh:
@gemini-code-assist reconsidered the Pydantic model inconsistency point and agreed. fb7827d3 removes the special case so TypeAdapter handles all types uniformly. also skips framework-managed params (`_ignore_params`) to avoid validating injected values like `credential=None`.

/gemini review

yuvrajangadsingh:
@gemini-code-assist applied in 3b62f632. TypeAdapter instances are now cached per target_type on the FunctionTool instance, so they're only created once per param type across invocations.

/gemini review

yuvrajangadsingh:
@gemini-code-assist good call. applied in 8ac8cb95. removed the entire Optional unwrap block since TypeAdapter handles `Optional[T]` and `T | None` natively. also dropped the now-unused `sys`, `get_args`, `get_origin`, and `_UNION_TYPES` imports.

/gemini review

yuvrajangadsingh:
@gemini-code-assist good catch. applied in 91e5a9dc. cache lookup now uses try/except KeyError+TypeError so unhashable types like `Union[int, str]` still get validated, just not cached.

/gemini review

yuvrajangadsingh:
@gemini-code-assist pushing back on this one. `TypeError`/`NameError` from TypeAdapter means a broken type annotation (forward reference, etc), which is a developer bug, not an LLM mistake. returning it as a validation error to the LLM would just confuse it since it can't fix the tool definition. logging a warning is the right call here, it surfaces the issue for the developer without polluting the LLM's error feedback loop.

/gemini review

yuvrajangadsingh:
@gemini-code-assist applied in ba51ee07. split the cache except into separate `TypeError` (unhashable, skip cache) and `KeyError` (cache miss, create and store) handlers.

/gemini review

yuvrajangadsingh:
@gemini-code-assist this suggestion reverses the previous review's suggestion (which asked to split them apart). going to keep the current version with separate handlers since that's what the previous round asked for. no changes needed here.

waiting for human review.

yuvrajangadsingh:
@gemini-code-assist to be clear, this is now going in circles. the 09:11 review asked me to split `(KeyError, TypeError)` into separate handlers. the 09:15 review asks me to combine them back. i'm keeping the split version since it's more explicit. no further changes on this.

@llalitkumarrr @cuihaoleo @xuanyang15 this PR is ready for human review. gemini has gone through ~8 review rounds and all actionable feedback has been addressed. the remaining suggestions are either contradictory (split vs combine exception handlers) or already declined with reasoning (logging vs error for developer-side TypeError/NameError). would appreciate a look when you get a chance.

rohityan:
Hi @yuvrajangadsingh, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

yuvrajangadsingh:
@rohityan done, ran autoformat.sh and rebased on upstream main. pyink reformatted a few line wraps in function_tool.py.

yuvrajangadsingh:
@rohityan rebased onto latest main. had one conflict in crewai_tool.py since it got refactored to integrations, resolved cleanly. the core changes to FunctionTool._preprocess_args and the test suite are all intact.

yuvrajangadsingh:
@rohityan @llalitkumarrr @wuliang229 @cuihaoleo @xuanyang15 formatting was fixed same day rohityan asked, rebased on main. been 4 weeks with no code review. ci green, tests cover all the validation edge cases. this implements the TypeAdapter approach cuihaoleo suggested in #4612. would be great to get eyes on it.

yuvrajangadsingh:
@rohityan @xuanyang15 bump on this one too, rebased and formatting fixed 5 weeks ago. CI green, all review comments addressed. happy to make changes if anything needs adjusting.

yuvrajangadsingh:
@rohityan ran autoformat back in dc50106a (mar 5). re-ran today on the latest branch and no further changes needed. ready for review whenever you have a moment.

@llalitkumarrr @cuihaoleo @xuanyang15 still hoping for a human look on this. summary: validates and coerces function tool argument types via pydantic.TypeAdapter so the LLM gets self-correctable feedback when it sends bad types. 8 rounds of gemini bot review addressed, all CI green.

yuvrajangadsingh:
Rebased onto latest main and resolved the semantic conflict in `flows/llm_flows/functions.py`.

The conflict was in the new thread-pool sync FunctionTool path: upstream still expects `_preprocess_args(args)` to return args only, while this PR needed validation errors returned to the LLM. I preserved upstream's current structure and split into two methods: `_preprocess_args` keeps its single-return signature for backward compatibility, and a new `_preprocess_args_with_validation` returns the tuple for the validated execution paths (`run_async`, the thread-pool sync path, and the crewai wrapper).

Net effect: existing callers untouched, blast radius limited to the validated paths only.

@seanzhougoogle @GWeale could one of you sanity-check this resolution or assign the right owner? This fixes #4612 and keeps invalid typed tool args from reaching user functions.

kkj333:
Hi maintainers — bumping this with a concrete ADK 2.0 GA regression we filed separately.

**Related issue:** #5799 — `FunctionTool` does not convert `Union[BaseModel, BaseModel]` at runtime (`Unexpected entity type: <class 'dict'>`). Repro is in the issue; also recorded in `contributing/samples/tools/pydantic_argument/tests/test_create_company.json`.

On current `main`, only `Optional[SingleBaseModel]` is converted in `_preprocess_args`. ADK 2.0 made this much more visible because `JSON_SCHEMA_FOR_FUNC_DECL` is now default-on (tool declaration succeeds, but runtime still passes dicts through).

I checked this PR branch locally:
- `_preprocess_args_with_validation` + `pydantic.TypeAdapter` correctly converts `Union[UserProfile, CompanyProfile]` from JSON dicts (both user and company payloads).
- Existing arg-validation tests pass on the branch.

**Suggestions before merge (if helpful):**
1. Add an explicit regression test for `Union[BaseModel, BaseModel]` (not just primitives/enums/Optional) — would close #5799 cleanly.
2. On latest `main`, `CrewaiTool` lives under `integrations/crewai/crewai_tool.py`; the shim at `tools/crewai_tool.py` is deprecation-only. Worth confirming validation changes land on the integrations path after rebase.

This PR has been open since Feb with CI green (CLA/header) and extensive Gemini review rounds — would appreciate a human review when someone has bandwidth. Happy to help test a rebased branch or review follow-ups.

Thanks @yuvrajangadsingh for the work on #4612 — this looks like the right fix direction for the 2.0 tool typing gaps.

yuvrajangadsingh:
thanks for the validation @kkj333. @xuanyang15 / @GWeale, when you get a chance, would appreciate a look here. happy to address any review feedback or rebase if needed.

fluxehub:
Is there any progress on this PR? Would be great to see this!

yuvrajangadsingh:
Rebased onto latest main. The type-coercion that landed on main since I opened this already handles the reported cases, so I've refocused this PR: it now refactors that logic to use `pydantic.TypeAdapter` instead of the manual `Union`/`get_origin`/`issubclass` branching.

The upside is TypeAdapter collapses ~80 lines to ~20 and picks up enums, nested containers, and primitive coercion (int/float/str/bool from strings) uniformly, with validation errors surfaced back to the model instead of silently keeping the original value. The added tests in `test_function_tool_pydantic.py` cover model/Optional/Union/list/primitive/error-capture.

Totally happy to close this if you'd rather keep the current approach, just offering it as a cleanup since I was already in here.

Issue #501, opened 2025-05-01
Crash while exiting loop agent in adk web


ThomasSilloway:
**Describe the bug**

When a sub-agent within a `google.adk.agents.LoopAgent` signals termination using `tool_context.actions.escalate = True`, the loop terminates, but an `opentelemetry.context` error (`ValueError: <Token ...> was created in a different Context`) occurs immediately after the sub-agent's `GeneratorExit`. This appears to be an issue with OpenTelemetry context handling during the abrupt termination of the agent's async generator, triggered by the ADK escalation mechanism.

**To Reproduce**

1.  Install the Google ADK (`google-adk`) v0.3
2.  Create a `LoopAgent` that runs a sub-agent (e.g., an `LlmAgent` as shown in the code below).
3.  Define a tool for the sub-agent that accepts `tool_context: ToolContext` as an argument (see `process_backlog_file` example below).
4.  Implement logic within the tool to set `tool_context.actions.escalate = True` under a specific condition (e.g., when a backlog file is empty).
5.  Run the `LoopAgent` (e.g., via `adk web` or programmatically) and trigger the condition in the sub-agent's tool that sets `escalate = True`.
6.  Observe the application logs for the `ValueError: ... was created in a different Context` traceback immediately following the `GeneratorExit`.

**Expected behavior**

The `LoopAgent` should terminate cleanly when `escalate` is set to `True` by the sub-agent's tool, without any subsequent OpenTelemetry context detachment errors in the traceback.

**Screenshots**

(Not applicable, traceback provided below)

**Desktop (please complete the following information):**

* **OS:** Windows 
* **Python version(python -V):** Python 3.13.2
* **ADK version(pip show google-adk):**  0.3.0

**Additional context**

* This issue appears related to known context propagation challenges within `opentelemetry-python` when dealing with async generators and abrupt exits, particularly involving `contextvars`. See related discussion: [open-telemetry/opentelemetry-python#2606](https://github.com/open-telemetry/opentelemetry-python/issues/2606)
* The core ADK escalation mechanism (`tool_context.actions.escalate = True`) seems to function correctly in triggering the loop termination, but the subsequent cleanup phase within the ADK's flow hits this OpenTelemetry context issue.

**Traceback:**
[...]

PR #4717, opened 2026-03-05
fix: ensure call_llm spans are always ended in multi-agent scenarios

### LINKED ISSUE #4715: [Bug] call_llm spans not exported in multi-agent setups due to GeneratorExit breaking span.end() in _call_llm_async
## Description

In multi-agent setups using `transfer_to_agent`, the `call_llm` spans for parent agents are created but never exported to Cloud Trace / OpenTelemetry backends. Only sub-agent spans are correctly exported.

The spans are referenced (via `parent_span_id`) by child spans, confirming they are created — but the span bodies themselves are absent from the trace backend (orphan spans).

## Root Cause

`_call_llm_async` in `base_llm_flow.py` uses `tracer.start_as_current_span('call_llm')` as a context manager around an `async for` loop that yields responses:

```python
async def _call_llm_with_tracing():
    with tracer.start_as_current_span('call_llm') as span:
        ...
        async for llm_response in agen:
            yield llm_response  # GeneratorExit raised here when transfer_to_agent occurs
```

When the LLM returns `transfer_to_agent`, the sub-agent runs (can take 10+ seconds), then the AsyncGenerator is closed, raising `GeneratorExit` inside the `with` block.

Inside OTel's `start_as_current_span` `finally` block:

```python
finally:
    context.detach(token)  # raises ValueError: Token was created in a different Context
    span.end()             # NEVER REACHED
```

Because `context.detach()` raises `ValueError` (the contextvars token is invalid after the async context switch caused by sub-agent execution), `span.end()` is never called. Spans that are never ended are never exported.

## Relationship to Existing Issues

This is the same root cause as #501 and #1670 (GeneratorExit + OTel contextvars in AsyncGenerator), both of which were fixed in `base_agent.py` (commit 955f477). However, `base_llm_flow.py:_call_llm_async` was not updated with the same fix.

## Steps to Reproduce

1. Deploy a multi-agent system to Vertex AI Agent Engine with OpenTelemetry tracing enabled
2. Configure agents with `transfer_to_agent` tool usage
3. Query the parent agent so it transfers to a sub-agent
4. Check Cloud Trace or BigQuery linked dataset for `call_llm` spans

**Expected:** `call_llm` spans for all agents appear in the trace
**Actual:** Only sub-agent `call_llm` spans appear; parent agent `call_llm` spans are missing (but referenced as `parent_span_id` by child spans)

## Proposed Fix

Replace `start_as_current_span` with explicit span management to ensure `span.end()` is always called even when `GeneratorExit` is raised:

```python
span = tracer.start_span('call_llm')
ctx = trace.set_span_in_context(span)
token = context.attach(ctx)
try:
    async for llm_response in agen:
        yield llm_response
finally:
    try:
        context.detach(token)
    except ValueError:
        pass
    span.end()  # Always called, even on GeneratorExit
```

## Environment

- ADK version: 1.23.0 (also reproduced on later versions)
- Deployment: Vertex AI Agent Engine
- Python: 3.11
- Related: opentelemetry-python#2606 (upstream OTel issue with AsyncGenerator context propagation)

weiguangli-io:
## Summary

Fixes #4715

In multi-agent setups using `transfer_to_agent`, the `call_llm` tracing spans for parent agents are created but never exported to OpenTelemetry backends. This happens because `_call_llm_with_tracing()` in `base_llm_flow.py` uses `tracer.start_as_current_span('call_llm')` as a context manager around an async generator that yields responses. When the LLM returns `transfer_to_agent`, the sub-agent runs (potentially for 10+ seconds), then the async generator is closed, raising `GeneratorExit`. Inside the OTel context manager's `finally` block, `context.detach(token)` raises `ValueError` (the contextvars token is stale after the async context switch), which prevents `span.end()` from ever being called. Spans that are never ended are never exported.

This is the same root cause as #501 and #1670 (previously fixed in `base_agent.py`), but `base_llm_flow.py` was not updated with the same fix pattern.

### Changes

- **`src/google/adk/flows/llm_flows/base_llm_flow.py`**: Replace `tracer.start_as_current_span('call_llm')` context manager with explicit span lifecycle management (`tracer.start_span()` + manual `context.attach()`/`context.detach()`) wrapped in a `try/finally` that catches the `ValueError` from `detach()` and always calls `span.end()`.

- **`tests/unittests/telemetry/test_functional.py`**: Update the `span_exporter` test fixture to also monkeypatch `start_span` (in addition to the existing `start_as_current_span`), so that the `call_llm` spans created by the new code path are properly captured by the in-memory span exporter.

## Test Plan

- [x] All 56 telemetry tests pass (`tests/unittests/telemetry/`)
- [x] All 357 llm_flows tests pass (`tests/unittests/flows/llm_flows/`)
- [x] `test_tracer_start_as_current_span` functional test validates that `call_llm` spans are correctly exported
- [x] `test_exception_preserves_attributes` confirms span attributes are preserved on errors

### DIFF
--- a/contributing/samples/gepa/experiment.py
+++ b/contributing/samples/gepa/experiment.py
@@ -44,6 +44,7 @@ from tau_bench.run import display_metrics
 from tau_bench.types import EnvRunResult
 from tau_bench.types import RunConfig
 import tau_bench_agent as tau_bench_agent_lib
+import utils
 
 
 def run_tau_bench_rollouts(
--- a/contributing/samples/gepa/run_experiment.py
+++ b/contributing/samples/gepa/run_experiment.py
@@ -26,6 +26,7 @@ from absl import flags
 import experiment
 import gepa_utils
 from google.genai import types
+import utils
 
 _OUTPUT_DIR = flags.DEFINE_string(
     'output_dir',
--- a/src/google/adk/flows/llm_flows/base_llm_flow.py
+++ b/src/google/adk/flows/llm_flows/base_llm_flow.py
@@ -24,6 +24,7 @@ from typing import TYPE_CHECKING
 
 from google.adk.platform import time as platform_time
 from google.genai import types
+from opentelemetry import context as otel_context
 from opentelemetry import trace
 from websockets.exceptions import ConnectionClosed
 from websockets.exceptions import ConnectionClosedOK
@@ -41,6 +42,7 @@ from ...events.event import Event
 from ...models.base_llm_connection import BaseLlmConnection
 from ...models.llm_request import LlmRequest
 from ...models.llm_response import LlmResponse
+
 from ...telemetry import tracing
 from ...telemetry.tracing import trace_call_llm
 from ...telemetry.tracing import trace_send_data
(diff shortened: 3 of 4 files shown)


rohityan:

rohityan:
Hi @wuliang229 , can you please review this.

weiguangli-io:
Friendly ping — CI is green. Would appreciate a review when you get a chance. Thanks!

haranrk:
Hey @weiguangli-io, thanks for the PR! I'm closing this since call_llm already ends on GeneratorExit due to context.detach() swallowing exceptions (opentelemetry/context/__init__.py:147) and span.end() running in an outer finally (trace/__init__.py:636). If #4715 is still reproducible, let's get a fresh repro targeting span export or flush timing.

Issue #4777, opened 2026-03-10
`FunctionTool` schema missing required fields for Pydantic BaseModel parameters


gautamvarmadatla:
**Describe the Bug:**
When a `FunctionTool` accepts a Pydantic `BaseModel` parameter, the generated function declaration schema does not populate the nested model schema’s `required` field. As a result, the LLM may treat required model fields as optional and omit them when invoking the tool.

When this happens, `model_validate(...)` fails internally, but the exception is silently caught. The tool is still invoked with the original raw `dict` instead of the expected model instance. This defers the failure to the tool implementation, where it may later surface as a downstream runtime error.

**Steps to Reproduce:**

Create a script with following repro and run it 

``` python 
from pydantic import BaseModel
from google.adk.tools.function_tool import FunctionTool

class SearchRequest(BaseModel):
      query: str           # required
      max_results: int     # required
      filter: str = ""     # optional (has default)

def search(request: SearchRequest) -> list:
      """Search for documents."""
      return []

tool = FunctionTool(search)
decl = tool._get_declaration()
inner_schema = decl.parameters.properties["request"]

print("properties:", list(inner_schema.properties.keys()))
print("required:  ", inner_schema.required)
```

**Expected Behavior:**
The generated schema for a `BaseModel` parameter should include a required field listing all model fields that have no default value. This allows the LLM to correctly identify which fields are mandatory and always include them when invoking the tool.

**Observed Behavior:**
Running the above script produces the following output:

```
properties: ['query', 'max_results', 'filter']
required:   None
```

**Environment Details:**

 - ADK Library Version (pip show google-adk): 1.26.0
 - Desktop OS : Windows
 - Python Version (python -V) : 3.10

**How often has this issue occurred?:**
 - Always (100%)

surajksharma07:
@gautamvarmadatla Confirmed the root cause — in `_function_parameter_parse_util.py` around line 392–403, when iterating a Pydantic BaseModel's `model_fields` to build the nested schema, `schema.required` is never populated (and field defaults aren't forwarded either), so the LLM sees all fields as optional.

As a workaround, you can patch `_parse_schema_from_parameter` to use `field_info.is_required()` and set `schema.required` before returning — could you give that a try and let us know if it resolves the issue on your end?

Your linked PR #4778 looks like the right fix for this; once you've confirmed the workaround holds up, it would be great to get that PR tested thoroughly before pushing it forward.

gautamvarmadatla:
[...]

PR #4778, opened 2026-03-10
fix: populate `required` for Pydantic `BaseModel` parameters in `FunctionTool`

gautamvarmadatla:
Closes: #4777

**Problem:**
When a `FunctionTool` accepts a Pydantic `BaseModel` parameter, the generated function declaration schema does not set the nested model schema’s `required` field. Because of this, the LLM may treat required fields as optional and omit them when calling the tool. If that happens, `model_validate(...)` fails internally, the exception is caught, and the function receives a raw `dict` instead of the expected model instance. This can later cause a runtime error.

**Solution:**
In the `BaseModel` branch of `_parse_schema_from_parameter` in `_function_parameter_parse_util.py`, compute required fields using `field_info.is_required()` after building `schema.properties`, and assign them to `schema.required`. This correctly handles required fields, defaults, `Optional` fields, and `default_factory` fields. Since the same logic is used recursively, nested `BaseModel` parameters are fixed as well.

### Testing Plan
`tests/unittests/tools/test_build_function_declaration.py`
- `test_basemodel_required_fields`  
  Verifies that `required` is populated only with fields that have no default (`query`, `max_results`) and excludes fields with defaults (`filter`).
- `test_basemodel_all_optional_fields_no_required`  
  Verifies that `required` is not set when all `BaseModel` fields have defaults.
- `test_nested_basemodel_required_fields`  
  Verifies that `required` is set correctly at every nesting level for nested `BaseModel` parameters.

**Unit Tests:**

- [X] I have added or updated unit tests for my change.
- [X] All unit tests pass locally.

Create and run the following script:

```python
from google.adk.tools.function_tool import FunctionTool


class SearchRequest(BaseModel):
    query: str
    max_results: int
    filter: str = ""


def search(request: SearchRequest) -> list:
    """Search for documents."""
    return []


tool = FunctionTool(search)
decl = tool._get_declaration()
inner = decl.parameters.properties["request"]

print("properties:", list(inner.properties.keys()))
print("required:  ", inner.required)
````

### Before fix

```text
properties: ['query', 'max_results', 'filter']
required:   None
```

### After fix

```text
properties: ['query', 'max_results', 'filter']
required:   ['query', 'max_results']
```

### Checklist

### DIFF
--- a/src/google/adk/a2a/converters/long_running_functions.py
+++ b/src/google/adk/a2a/converters/long_running_functions.py
@@ -74,7 +74,10 @@ class LongRunningFunctions:
     for part in event.content.parts:
       should_remove = False
       if part.function_call:
-        if part.function_call.id in event.long_running_tool_ids:
+        if (
+            event.long_running_tool_ids
+            and part.function_call.id in event.long_running_tool_ids
+        ):
           if not event.partial:
             self._parts.append(part)
             self._long_running_tool_ids.add(part.function_call.id)
--- a/src/google/adk/a2a/utils/agent_to_a2a.py
+++ b/src/google/adk/a2a/utils/agent_to_a2a.py
@@ -14,7 +14,10 @@
 
 from __future__ import annotations
 
+from contextlib import asynccontextmanager
 import logging
+from typing import AsyncIterator
+from typing import Callable
 from typing import Optional
 from typing import Union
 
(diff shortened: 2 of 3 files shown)


rohityan:
Hi @gautamvarmadatla , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:
Hi @wuliang229 , can you please review this.

PR #4799, opened 2026-03-12
fix: set required fields in FunctionDeclaration json_schema fallback path

### LINKED ISSUE #4798: Bug: required fields lost in FunctionDeclaration when parameters_json_schema fallback is used
## Bug Description

When `FunctionTool` wraps a Python function that includes parameter types causing `_parse_schema_from_parameter` to raise `ValueError` (e.g. `list[str] | None`), the generated `FunctionDeclaration` has `required: None` — even for mandatory (no-default) parameters.

This happens because `from_function_with_options()` in `_automatic_function_calling_util.py` has two code paths:

1. **Primary path** (`parameters_properties`): calls `_get_required_fields()` ✅
2. **Fallback path** (`parameters_json_schema`): **never calls `_get_required_fields()`** ❌

The fallback is triggered when any parameter has a complex union type like `list[str] | None` that `_parse_schema_from_parameter` can't handle. Since the `except ValueError` catches the error for *all* parameters (not just the complex one), the entire function's schema is rebuilt via the fallback — and `required` is silently dropped for every parameter.

## Impact

The LLM sees all parameters as optional and may omit required ones. We observed this in production: Gemini Flash omitted the mandatory `query` parameter when calling a tool, because the schema had no `required` field.

## Reproduction

```python
"""Minimal repro — tested with google-adk 1.26.0, Python 3.12"""

from google.adk.tools._automatic_function_calling_util import (
    build_function_declaration,
)

# Case 1: simple types → primary path → required IS set
def simple_tool(
    query: str,
    mode: str = "default",
    verbose: bool = False,
) -> str:
    """A tool with simple parameter types."""
    return query

# Case 2: list[str] | None → fallback path → required is LOST
def complex_tool(
    query: str,
    mode: str = "default",
    tags: list[str] | None = None,
) -> str:
    """A tool where one param has a complex union type."""
    return query

decl_simple = build_function_declaration(simple_tool)
decl_complex = build_function_declaration(complex_tool)

print("Case 1 (simple types):", decl_simple.parameters.required)
# → ['query'] ✅

print("Case 2 (list[str] | None):", decl_complex.parameters.required)
# → None ❌  (expected ['query'])
```

## Root Cause

In [`src/google/adk/tools/_automatic_function_calling_util.py`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/_automatic_function_calling_util.py), `from_function_with_options()`:

```python
  if parameters_properties:
    declaration.parameters = types.Schema(
        type='OBJECT',
        properties=parameters_properties,
    )
    declaration.parameters.required = (                    # ✅ set here
        _function_parameter_parse_util._get_required_fields(
            declaration.parameters
        )
    )
  elif parameters_json_schema:
    declaration.parameters = types.Schema(
        type='OBJECT',
        properties=parameters_json_schema,
    )
    # ❌ required is never set in this branch
```

## Suggested Fix

Add the same `_get_required_fields()` call to the `elif` branch:

```python
  elif parameters_json_schema:
    declaration.parameters = types.Schema(
        type='OBJECT',
        properties=parameters_json_schema,
    )
    declaration.parameters.required = (
        _function_parameter_parse_util._get_required_fields(
            declaration.parameters
        )
    )
```

## Environment

- google-adk version: 1.26.0
- Python version: 3.12
- OS: macOS

aperepel:
## Summary

- Fix `from_function_with_options()` to set `required` fields when the `parameters_json_schema` fallback path is used
- Add regression test covering the bug

## Problem

When a `FunctionTool` wraps a function with parameter types that cause `_parse_schema_from_parameter` to raise `ValueError` (e.g. `list[str] | None`), `from_function_with_options()` falls into the `parameters_json_schema` code path. This path never set `required` on the generated `FunctionDeclaration`, causing LLMs to treat **all** parameters as optional — even mandatory ones with no default value.

We observed this in production: Gemini Flash omitted a mandatory `query` parameter when calling a tool because the schema had `required: None`.

### Root cause

The primary path (`parameters_properties`) calls `_get_required_fields()`, but the fallback path (`parameters_json_schema`) did not:

```python
if parameters_properties:
    # ... sets required ✅
elif parameters_json_schema:
    # ... never set required ❌
```

### Fix

Compute `required` directly from `inspect.signature(func)` in the fallback path — parameters with no default are required. Using `_get_required_fields()` doesn't work here because it relies on `nullable`/`default` metadata on Schema objects, which isn't preserved through the `_generate_json_schema_for_parameter` → `Schema.model_validate()` round-trip.

## Test plan

- [x] New test: `test_from_function_with_options_required_fields_with_complex_union` — verifies `query` (no default) is in `required` while `mode` (has default) and `tags` (`list[str] | None = None`) are not
- [x] All 16 tests in `test_from_function_with_options.py` pass
- [x] All 81 tests in related tool test files pass

Fixes #4798


xuanyang15:
Thanks for creating this PR! Will proceed with https://github.com/google/adk-python/pull/5000.

PR #4800, opened 2026-03-12
fix: extract URL fragment key=value pairs as query params in _prepare_request_params

### LINKED ISSUE #4598: ApplicationIntegrationToolset ExecuteConnection requests return HTTP 400 (triggerId fragment not sent)
# ApplicationIntegrationToolset ExecuteConnection requests return HTTP 400 (triggerId fragment not sent)

## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient triaging. Requests without complete information may be rejected / deprioritized. If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**
When using `ApplicationIntegrationToolset` with an Integration Connectors **connection** (e.g. Salesforce) and entity operations (GET/LIST), the underlying POST to `integrations.googleapis.com/.../ExecuteConnection:execute` returns **HTTP 400 "Request contains an invalid argument."** The root cause is that the tool builds the URL with the trigger action in the **fragment** (e.g. `?triggerId=api_trigger/ExecuteConnection#get_EntityName`). Per HTTP semantics, the fragment is never sent to the server; only path and query string are. So the ExecuteConnection API receives a request that is missing the full triggerId (the part after `#` is lost), and rejects it with 400.

**Steps to Reproduce:**
1. Install ADK: `pip install "google-adk[a2a]>=1.25.1,<2.0.0"` (or `uv sync` in a project that depends on it).
2. Create a GCP Integration Connectors connection (e.g. Salesforce) in a project/region, with at least one entity that supports GET and LIST (e.g. a custom object or any entity).
3. Create an agent that uses `ApplicationIntegrationToolset` with that connection and `entity_operations={"YourEntity": ["GET", "LIST"]}`.
4. Run the agent (e.g. via `Runner.run_async`) and send a user message that triggers a GET or LIST tool call.
5. Observe: the tool’s POST to `ExecuteConnection:execute` returns HTTP 400. In logs, the request URL shown by the HTTP client is `.../ExecuteConnection:execute#get_EntityName` (fragment only, no `triggerId` in query), while the RestApiTool debug log may show `?triggerId=api_trigger/ExecuteConnection#get_EntityName` — the fragment is not sent to the server.

**Expected Behavior:**
The ExecuteConnection request should succeed (HTTP 200) and return the connector payload. The same request works when sent manually or via a custom tool that puts the full triggerId in the **query string** (e.g. `triggerId=api_trigger/ExecuteConnection%23get_EntityName` or by sending only the base triggerId if the API accepts it in the body).

**Observed Behavior:**
- Every GET and LIST tool call from ApplicationIntegrationToolset results in HTTP 400 from `integrations.googleapis.com`.
- Response body: `{"error": {"code": 400, "message": "Request contains an invalid argument.", "status": "INVALID_ARGUMENT"}}`.
- In httpx logs, the actual request URL is `.../integrations/ExecuteConnection:execute#get_EntityName` (no query string with `triggerId`), so the server does not receive the full trigger identifier.

**Environment Details:**

- ADK Library Version (pip show google-adk): **1.25.1** (constraint: `google-adk[a2a]>=1.25.1,<2.0.0`)
- Desktop OS: **macOS** (darwin 25.2.0)
- Python Version (python -V): **Python 3.12.x**

**Model Information:**

- Are you using LiteLLM: **No**
- Which model is being used: **gemini-2.5-flash** (Vertex AI)

---

## 🟡 Optional Information

**Regression:**
Unknown. Same behavior observed with ADK 1.25.1. There is a unit test in adk-python for extracting embedded query params and fragments from the path (`test_prepare_request_params_extracts_embedded_query_params` in `test_rest_api_tool.py`, regression for issue #4555); the ApplicationIntegrationToolset path may not be passing the built URL through that logic, or the triggerId may be embedded in a way that still ends up in the fragment when the request is sent.

**Logs:**
Relevant snippet showing the mismatch between intended URL and actual request:

```text
[INFO] httpx: HTTP Request: POST https://integrations.googleapis.com/v2/projects/.../locations/.../integrations/ExecuteConnection:execute#get_EntityName "HTTP/1.1 400 Bad Request"
[DEBUG] google_adk...rest_api_tool: API Response: POST https://integrations.googleapis.com/.../ExecuteConnection:execute?triggerId=api_trigger/ExecuteConnection#get_EntityName - Status: 400
```

The first line is what the HTTP client actually sent (URL has only fragment, no `triggerId` query). The second is what the RestApiTool reports (includes `triggerId` and fragment). The fragment is not sent over the wire, so the server receives an invalid or incomplete request.

**Screenshots / Video:**
N/A

**Additional Context:**
- The same GCP connection and entity work when calling ExecuteConnection directly (e.g. with `requests.post(url, json=payload)` where the URL is built with `?triggerId=api_trigger/ExecuteConnection#get_EntityName` — in that case, the fragment is still not sent, but the server may accept the base triggerId and route using the request body; in our testing, a **custom tool** that sends `triggerId=api_trigger/ExecuteConnection` in the query string (no fragment) returns 200).
- So the fix could be: ensure the full triggerId (including the part that AIT puts after `#`) is sent in the **query string** (e.g. as a single parameter with `#` encoded as `%23`), not in the URL fragment, when building the request in the RestApiTool / OpenAPI path used by ApplicationIntegrationToolset.

Requires a GCP project with an Integration Connectors connection (e.g. Salesforce). Replace `YOUR_PROJECT`, `YOUR_REGION`, `YOUR_CONNECTION_NAME`, and entity name as needed.

```python
# agent.py
import os
from google.adk.tools.application_integration_tool import ApplicationIntegrationToolset

load_dotenv()

project = os.getenv("SALESFORCE_CONNECTION_PROJECT_ID")
location = os.getenv("SALESFORCE_CONNECTION_REGION")
connection = os.getenv("SALESFORCE_CONNECTION_NAME")

toolset = ApplicationIntegrationToolset(
    project=project,
    location=location,
    connection=connection,
    entity_operations={"YourEntity": ["GET", "LIST"]},  # use your connector's entity API name
    tool_instructions="Use to get or list records.",
)

root_agent = Agent(
    name="test_coordinator",
    description="Test agent for connector.",
    instruction="Use the tools to retrieve the requested record. For GET use only the bare record ID.",
    model="gemini-2.5-flash",
    tools=[toolset],
)
```

Run with ADK Runner and send a message that triggers a GET (e.g. "Get the record with ID <record_id>"). The tool call will return 400.

- **Always (100%)**

saiprasanth-git:
Closes #4598

### What's the problem?

When `_prepare_request_params` builds the final request URL, it uses `urlparse` to split the URL into its components. The existing code was already smart enough to pull any key=value pairs out of the query string and move them into `query_params` (so httpx doesn't lose them when it sets the `params` arg). But it was completely ignoring the fragment part of the URL.

This matters because `ApplicationIntegrationToolset` sometimes generates endpoint URLs like:
```
https://integrations.googleapis.com/v2/.../execute?triggerId=api_trigger/Foo#httpMethod=POST
```
The fragment (`httpMethod=POST`) was being silently dropped, which caused the downstream API to return HTTP 400.

### What changed?

Instead of only looping over `parsed_url.query`, I refactored the extraction into a single loop that processes both `parsed_url.query` and `parsed_url.fragment`:

```python
if parsed_url.query or parsed_url.fragment:
    # Extract parameters from query and fragment
    for part in (parsed_url.query, parsed_url.fragment):
        if part:
            for key, values in parse_qs(part).items():
                query_params.setdefault(
                    key,
                    values[0] if len(values) == 1 else values
                )
    # Rebuild URL without query and fragment
    url = urlunparse(parsed_url._replace(query="", fragment=""))
```

Using `setdefault` makes sure any explicitly passed query params always take priority over whatever was embedded in the URL.

### Testing

Added `test_prepare_request_params_fragment_params_become_query_params` which uses a realistic Google Cloud integration URL containing both a query string param and a fragment param. Verifies both end up in `query_params` and the final URL has no leftover `?` or `#`.


saiprasanth-git:
@googlebot I signed it

saiprasanth-git:
@googlebot check

saiprasanth-git:
@googlebot I signed the CLA. Please re-verify.

rohityan:
Hi @saiprasanth-git , Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Can you please fix the failing unit tests, mypy-diff tests and formatting errors.

saiprasanth-git:
Hi @rohityan, I've addressed the feedback:

1. **Syntax error fixed** — removed the redundant outer `if parsed_url.query or parsed_url.fragment:` guard that had no indented body, which was causing the mypy syntax error on line 384.

Could you please approve the workflow runs so the remaining checks (pyink, unit tests, mypy-diff) can run? Thank you!

rohityan:
Hi @saiprasanth-git , you can use autoformat.sh to fix the formatting errors.

PR #4810, opened 2026-03-13
fix(sessions): use read-only transactions for get/list operations

### LINKED ISSUE #4771: DatabaseSessionService: get_session / list_sessions use read-write transactions unnecessarily, causing RetryAborted on Spanner
## Summary

`DatabaseSessionService.get_session()` and `list_sessions()` are pure read operations, but they execute inside SQLAlchemy's default read-write transactions. On Cloud Spanner, this means they use OCC read-write transactions instead of read-only snapshots, which causes `RetryAborted` errors when another request is concurrently writing to the same session via `append_event()`.

## Observed behavior

When two requests for the same session overlap (e.g. a new turn arrives while the previous turn is still writing events), the `get_session()` read in the new request can fail with `RetryAborted` from `google.cloud.spanner_dbapi.exceptions`. This is a transient Spanner OCC conflict — the read-write transaction's read set intersects with the concurrent `append_event()` write.

This is surprising because `get_session()` never writes — it only executes `SELECT` statements.

### Real-world impact

In production workloads with rapid successive turns on the same session, the `get_session()` call fails with `RetryAborted`, causing the turn to start with missing session state. Application-level retry works around this, but it shouldn't be necessary for a read-only operation.

## Root cause

`DatabaseSessionService` uses a single `async_sessionmaker` for all operations:

```python
# database_session_service.py, line ~143
self.database_session_factory: async_sessionmaker[DatabaseSessionFactory] = (
    async_sessionmaker(bind=self.db_engine, expire_on_commit=False)
)
```

Both `get_session()` and `list_sessions()` use `async with self.database_session_factory() as sql_session:` which opens a read-write transaction. On Spanner, read-write transactions participate in OCC and can abort when conflicting with concurrent writes.

The `spanner_dbapi` connection already supports `read_only=True` (see `google.cloud.spanner_dbapi.connection`, line ~78), which switches to snapshot reads that:
- Never abort due to OCC conflicts
- Have lower latency (no commit phase)
- Don't hold locks

## Proposed fix

Create a separate `async_sessionmaker` with `read_only=True` for read-only operations, and use it in `get_session()` and `list_sessions()`. This would only affect the Spanner dialect — other databases (PostgreSQL, MySQL, SQLite) would see no behavioral difference since they don't have Spanner's OCC semantics.

One approach:

```python
# In __init__:
self.database_session_factory = async_sessionmaker(bind=self.db_engine, expire_on_commit=False)

# For Spanner, create a read-only engine/session factory
# Option A: execution_options on the session
# Option B: separate engine with connect_args={"read_only": True}
```

The cleanest path may be passing `execution_options` or using SQLAlchemy's `Session.execute()` with connection-level options, but the exact mechanism depends on how `sqlalchemy-spanner` propagates `read_only` to the underlying `spanner_dbapi` connection.

## Environment

- ADK version: 1.23.x
- Backend: Spanner (`spanner+spanner:///...`)
- SQLAlchemy-Spanner: 1.14+

## Related

- #4751 — OCC corruption from concurrent `append_event` (same underlying Spanner concurrency theme)
- #4655 — unnecessary `FOR UPDATE` lock on app/user states

giulio-leone:
## Summary

Fixes #4771

`DatabaseSessionService.get_session()` and `list_sessions()` are pure-read operations but currently open regular **read-write transactions** via `_rollback_on_exception_session()`. On **Cloud Spanner** this causes `RetryAborted` errors when a concurrent write commits during the read, because Spanner's OCC (Optimistic Concurrency Control) layer sees the read-write transaction as conflicting.

## Root Cause

Spanner uses OCC for read-write transactions. Even a `SELECT`-only transaction opened in read-write mode will be aborted if a concurrent write touches the same data. Since reads never mutate data, they should use read-only transactions which are not subject to OCC aborts.

## Fix

Added a `_readonly_session()` context manager that:
- Acquires the underlying connection and sets `execution_options(postgresql_readonly=True)`
- This option is also respected by the `sqlalchemy-spanner` dialect to open read-only transactions
- Never commits (reads don't need commits), only rolls back on exception
- Reuses the existing `database_session_factory` and connection pool

Switched `get_session()` and `list_sessions()` to use `_readonly_session()`.

Write operations (`create_session`, `delete_session`, `append_event`) continue to use `_rollback_on_exception_session()` with explicit commits — unchanged.

## Testing

All 128 session tests pass. Full suite: **4725 passed, 0 failures**, 0 regressions.

### DIFF
--- a/src/google/adk/a2a/converters/to_adk_event.py
+++ b/src/google/adk/a2a/converters/to_adk_event.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable
+import json
 import logging
 from typing import Any
 from typing import List
@@ -28,9 +29,11 @@ from a2a.types import TaskArtifactUpdateEvent
 from a2a.types import TaskState
 from a2a.types import TaskStatusUpdateEvent
 from google.genai import types as genai_types
+from pydantic import ValidationError
 
 from ...agents.invocation_context import InvocationContext
 from ...events.event import Event
+from ...events.event_actions import EventActions
 from ..experimental import a2a_experimental
 from .part_converter import A2A_DATA_PART_METADATA_IS_LONG_RUNNING_KEY
 from .part_converter import A2APartToGenAIPartConverter
@@ -171,11 +174,15 @@ def _create_event(
     output_parts: List[genai_types.Part],
     invocation_context: Optional[InvocationContext],
     author: Optional[str],
+    actions: Optional[EventActions] = None,
     long_running_function_ids: Optional[set[str]] = None,
     partial: bool = False,
 ) -> Optional[Event]:
   """Creates an ADK event from parts and metadata."""
-  if not output_parts:
+  event_actions = actions or EventActions()
+  if not output_parts and not event_actions.model_dump(
+      exclude_none=True, exclude_defaults=True
+  ):
     return None
 
   event = Event(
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @giulio-leone , Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

giulio-leone:
@rohityan Thanks for the heads-up! I'll get the Google CLA signed. Will follow up once it's done.

giulio-leone:
I have signed the Google CLA. Could you please re-run the CLA check? Thank you!

giulio-leone:
Hi @rohityan — the CLA is now signed and passing ✅ (it was a `Co-authored-by` trailer issue that has been resolved). Ready for review!

src/google/adk/sessions/database_session_service.py
```diff
@@ -213,6 +213,26 @@ async def _rollback_on_exception_session(
...
+    """Yields a read-only database session for pure SELECT operations.
+
+    On Spanner this avoids OCC read-write transactions (which can trigger
+    RetryAborted when conflicting with concurrent writes).  On PostgreSQL
+    the session is marked read-only.  On other dialects the behaviour is
+    equivalent to a regular session but the transaction is never committed.
+    """
+    async with self.database_session_factory() as sql_session:
+      conn = await sql_session.connection()
+      await conn.execution_options(postgresql_readonly=True)
```
rohityan: In SQLAlchemy, `conn.execution_options()` is synchronous. Will raise a TypeError. 
You should use execution_options={"read_only": True} for spanner

giulio-leone:
This branch has now been rebased to current `main` and no longer carries any delta (`HEAD == 4b677e73`, `0 ahead / 0 behind`).

I verified the absorbed fix locally from that refreshed head:
- `pyink --check src/google/adk/sessions/database_session_service.py tests/unittests/sessions/test_session_service.py`
- `isort --check-only src/google/adk/sessions/database_session_service.py tests/unittests/sessions/test_session_service.py`
- `pytest tests/unittests/sessions/test_session_service.py -q -k "read_only_engine or uses_read_only_factory"` → `4 passed` on both passes
- direct runtime proof showed `get_session()` and `list_sessions()` both route through `_rollback_on_exception_session(read_only=True)` (`read_only_calls == [true, true]`)

So the original contributor-side fix is already present upstream and this PR is effectively superseded/absorbed.

PR #4815, opened 2026-03-13
feat: add turn_id to Event for grouping streaming chunks by LLM call

ferponse:
## Summary

- Adds an optional `turn_id` field to `Event` that groups all streaming chunks belonging to the same LLM call
- Within a single invocation the LLM may be called multiple times (text → tool call → text). Each call produces events that share the same `turn_id`, while `id` changes on every yield
- Consumers can use this field to trivially separate streaming chunks into message bubbles without brittle heuristics based on event type transitions

## Problem

When using `runner.run_async()` with `StreamingMode.SSE`, there is no way to distinguish which partial streaming chunks belong to which LLM response turn. The `invocation_id` is shared across all events in the invocation, and `id` changes on every yield. The only workaround is observing transition patterns (partial→function_call→function_response), which is fragile.

## Solution

A new `turn_id: Optional[str]` field on `Event`:
- Generated once per LLM call in `_run_one_step_async`
- Remains the same across all partial chunks and the final aggregated event of that call
- Changes when a new LLM call starts (next iteration of the `while True` loop in `run_async`)
- Also set in the BIDI (live streaming) flow for consistency
- `None` by default, so existing code is unaffected

## Changes

| File | Change |
|------|--------|
| `src/google/adk/events/event.py` | Added `turn_id` field with docstring |
| `src/google/adk/flows/llm_flows/base_llm_flow.py` | Generate and assign `turn_id` when creating `model_response_event` (SSE + BIDI flows) |
| `tests/unittests/flows/llm_flows/test_turn_id.py` | 3 new tests covering partial chunks grouping, final response, and multi-turn separation |

## Test plan

- [x] `test_partial_chunks_share_same_turn_id` — partial chunks from one LLM call share the same `turn_id`
- [x] `test_turn_id_present_on_final_response` — a single final response carries a `turn_id`
- [x] `test_different_llm_calls_get_different_turn_ids` — events from separate LLM calls (text → tool → text) have different `turn_id`s
- [x] All 443 existing flow + runner tests pass unchanged

### DIFF
--- a/src/google/adk/events/event.py
+++ b/src/google/adk/events/event.py
@@ -67,6 +67,18 @@ class Event(LlmResponse):
   conversation history.
   """
 
+  turn_id: Optional[int] = None
+  """Groups all streaming chunks that belong to the same LLM call.
+
+  Within a single invocation the LLM may be called multiple times (e.g. text →
+  tool call → text). Each call is a "turn" and every event produced by that
+  call shares the same ``turn_id``, while ``id`` changes on every yield.
+  Consumers can use this field to trivially group partial chunks by turn.
+
+  The value is a 1-based counter that increments with each LLM call inside
+  ``run_async``.
+  """
+
   # The following are computed fields.
   # Do not assign the ID. It will be assigned by the session.
   id: str = ''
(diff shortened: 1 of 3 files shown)


ferponse:
Closing in favour of a new PR with improved implementation (turn_id as integer counter instead of UUID).

PR #4819, opened 2026-03-13
feat: add turn_id to Event for grouping streaming chunks by LLM call

ferponse:
## Summary

- Adds an optional `turn_id: int` field to `Event` that groups all streaming chunks belonging to the same LLM call
- `turn_id` is a 1-based counter that increments with each LLM call inside `run_async`, making it easy to identify turn boundaries (turn 1, turn 2, …)
- Consumers can use this field to trivially separate streaming chunks into message bubbles without brittle heuristics based on event type transitions

## Problem

When using `runner.run_async()` with `StreamingMode.SSE`, there is no way to distinguish which partial streaming chunks belong to which LLM response turn. The `invocation_id` is shared across all events in the invocation, and `id` changes on every yield. The only workaround is observing transition patterns (partial → function_call → function_response), which is fragile.

## Solution

A new `turn_id: Optional[int]` field on `Event`:
- 1-based counter incremented with each LLM call in `run_async`
- Remains the same across all partial chunks and the final aggregated event of that call
- Increments when a new LLM call starts (next iteration of the `while True` loop)
- Scoped to the SSE streaming flow only; BIDI (live) flow is not affected
- `None` by default, so existing code is unaffected

## Changes

| File | Change |
|------|--------|
| `src/google/adk/events/event.py` | Added `turn_id: Optional[int]` field with docstring |
| `src/google/adk/flows/llm_flows/base_llm_flow.py` | Counter in `run_async` loop, passed to `_run_one_step_async` as `Optional[int] = None` |
| `tests/unittests/flows/llm_flows/test_turn_id.py` | 3 new tests |

## Test plan

- [x] `test_partial_chunks_share_same_turn_id` — partial chunks from one LLM call share `turn_id=1`
- [x] `test_turn_id_present_on_final_response` — a single final response carries `turn_id=1`
- [x] `test_different_llm_calls_get_different_turn_ids` — events from separate LLM calls (text → tool → text) get `turn_id` 1 and 2
- [x] All 108 existing flow + runner tests pass unchanged

### DIFF
--- a/src/google/adk/events/event.py
+++ b/src/google/adk/events/event.py
@@ -67,6 +67,18 @@ class Event(LlmResponse):
   conversation history.
   """
 
+  turn_id: Optional[int] = None
+  """Groups all streaming chunks that belong to the same LLM call.
+
+  Within a single invocation the LLM may be called multiple times (e.g. text →
+  tool call → text). Each call is a "turn" and every event produced by that
+  call shares the same ``turn_id``, while ``id`` changes on every yield.
+  Consumers can use this field to trivially group partial chunks by turn.
+
+  The value is a 1-based counter that increments with each LLM call inside
+  ``run_async``.
+  """
+
   # The following are computed fields.
   # Do not assign the ID. It will be assigned by the session.
   id: str = ''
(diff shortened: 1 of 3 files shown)


ferponse:
Closing in favour of a new PR with review feedback applied and associated issue #4820.

PR #4828, opened 2026-03-13
fix: strip empty text parts in streaming to prevent skipped tool execution

akashbangad:
## Summary

Some Gemini models (e.g. `gemini-3.1-pro-preview`, `gemini-3-flash-preview`) return `function_call` + `text=""` in the same streaming response when `stream=True`. The empty text part causes ADK to treat it as the model's final answer, preventing the second LLM call that should happen after tool execution. The result: the tool runs correctly but the agent never generates a text response — the user sees nothing.

**Root cause**: In `StreamingResponseAggregator.process_response()`:
- **Non-progressive path**: `parts[0].text` is falsy for `""`, so the empty-text chunk falls through without being marked `partial`. It gets yielded as a non-partial response, and `is_final_response()` returns `True` — the agent loop stops.
- **Progressive path**: `if part.text:` skips `text=""` (falsy), but the part falls into the `else` branch and gets added to `_parts_sequence` as an "other non-text part".

With `stream=False`, the same models return ONLY the `function_call` (no empty text), so everything works correctly.

## Changes

- **Progressive path**: Added explicit `continue` to skip parts where `text is not None and text == ''` before they reach the else branch
- **Non-progressive path**: Filter out empty text parts from `llm_response.content.parts` before the existing text-accumulation logic
- **Tests**: Added 2 test cases covering both paths — verifying empty text parts are stripped when they appear alongside function calls

## Reproduction

```python
from google.adk.agents.run_config import RunConfig, StreamingMode

# Any agent with tools + StreamingMode.SSE + a Gemini preview model
agent = Agent(name="test", model="gemini-3.1-pro-preview", tools=[some_tool])
runner = Runner(agent=agent, app_name="t", session_service=InMemorySessionService())
# ...
    user_id="u", session_id="s", new_message=msg,
    run_config=RunConfig(streaming_mode=StreamingMode.SSE)
):
    # Tool is called correctly, but model never generates text response
    # Last event has content.parts=[Part(text="")] — empty final answer
```

## Test plan

- [x] All 14 existing `test_streaming_utils.py` tests pass
- [x] All 8 `test_progressive_sse_streaming.py` tests pass
- [x] All 33 `tests/unittests/streaming/` tests pass
- [x] New test: `test_empty_text_with_function_call_non_progressive` — verifies empty text stripped in non-progressive mode
- [x] New test: `test_empty_text_with_function_call_progressive` — verifies empty text not accumulated in progressive mode


rohityan:
Hi @akashbangad, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

akashbangad:
> Hi @akashbangad, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

Done

akashbangad:
Related: I've opened #4982 which addresses the broader version of this bug  models returning empty content after tool execution. PR #4828 fixes the Gemini-specific empty text part variant in the streaming aggregator, while #4982 adds a retry mechanism in `BaseLlmFlow.run_async()` that catches all empty response cases (including Claude's `parts: []`) regardless of model.

rohityan:
Hi @akashbangad , closing this and will track #4982

PR #4845, opened 2026-03-16
fix(runners): preserve initial session state keys during rewind

### LINKED ISSUE #4839: _compute_state_delta_for_rewind nullifies initial session state keys not present in event state_deltas
## Summary

`_compute_state_delta_for_rewind` nullifies session state keys that were set via `create_session(state={...})` but never appear in any event's `state_delta`. After `rewind_async()`, these keys are set to `None` in the session state.

## Root Cause

`_compute_state_delta_for_rewind` (runners.py L615-647) rebuilds `state_at_rewind_point` by replaying only **event** `state_delta` entries:

```python
state_at_rewind_point: dict[str, Any] = {}
for i in range(rewind_event_index):
    if session.events[i].actions.state_delta:
        for k, v in session.events[i].actions.state_delta.items():
            if k.startswith('app:') or k.startswith('user:'):
                continue
            # ...
            state_at_rewind_point[k] = v
```

Then it nullifies any key in `current_state` that isn't in `state_at_rewind_point`:

```python
for key in current_state:
    if key.startswith('app:') or key.startswith('user:'):
        continue
    if key not in state_at_rewind_point:
        rewind_state_delta[key] = None  # <-- nullified
```

Keys set via `create_session(state={"my_key": "my_value"})` are present in `session.state` (via `current_state`) but **never** appear in any event's `state_delta` — they were set as initial state, not through events. The algorithm sees them in `current_state` but not in `state_at_rewind_point`, and emits `my_key: None`.

Only `app:` and `user:` prefixed keys are protected from this nullification. Any other initial state key is affected.

## Reproduction

```python
from google.adk.events import Event, EventActions

# Create session with initial state
session = await svc.create_session(
    app_name="test_app",
    user_id="user",
    state={"my_key": "important_value"},  # set via create_session
)

# Simulate an invocation by appending an event
ev = Event(
    invocation_id="inv_A",
    author="model",
    actions=EventActions(state_delta={"other_key": "foo"}),
)
await svc.append_event(session, ev)

# Simulate a second invocation
ev2 = Event(
    invocation_id="inv_B",
    author="model",
    actions=EventActions(state_delta={"other_key": "bar"}),
)
await svc.append_event(session, ev2)

# Rewind before inv_B
agent = LlmAgent(name="test_app", model="gemini-2.5-flash")
runner = Runner(agent=agent, session_service=svc)
await runner.rewind_async(
    user_id="user",
    session_id=session.id,
    rewind_before_invocation_id="inv_B",
)

# Re-fetch session
rewound = await svc.get_session(
    app_name="test_app", user_id="user", session_id=session.id
)

print(rewound.state.get("my_key"))    # None — was "important_value"
print(rewound.state.get("other_key")) # "foo" — correctly restored
```

## Expected Behavior

After `rewind_async()`, `my_key` should still be `"important_value"` — it was set before any invocation and should be unaffected by rewinding invocations.

## Actual Behavior

`my_key` is `None` after rewind because `_compute_state_delta_for_rewind` doesn't account for initial session state.

## Impact

This affects any integration that sets initial session state via `create_session(state={...})`. In our case, `ag-ui-adk` sets `_ag_ui_thread_id` and `_ag_ui_user_id` as initial state — after rewind, session lookup by these keys fails because they're nullified.

## Suggested Fix

Include initial session state when building `state_at_rewind_point`. For example, seed it with the session's initial state before replaying event deltas:

```python
async def _compute_state_delta_for_rewind(
    self, session: Session, rewind_event_index: int
) -> dict[str, Any]:
    # Seed with initial state (keys set via create_session)
    state_at_rewind_point: dict[str, Any] = dict(session.initial_state or {})
    for i in range(rewind_event_index):
        # ... existing replay logic
```

If `initial_state` is not persisted on the `Session` object, an alternative is to treat keys present in `current_state` but absent from **all** event `state_delta` entries (across the entire session, not just up to the rewind point) as initial state that should be preserved.

## Workaround

After `rewind_async()`, append a synthetic event that re-sets the nullified keys:

```python
Event(
    invocation_id="",  # empty — skipped by checkpoint enumeration
    author="system",
    actions=EventActions(state_delta=saved_initial_state),
)
```

## Environment

- `google-adk`: 1.26.x
- `ag-ui-adk`: 0.5.1
- Python 3.13

saiprasanth-git:
Closes #4839

## Problem

`_compute_state_delta_for_rewind` in `runners.py` incorrectly nullifies session state keys that were set via `create_session(state={...})`. These "initial state" keys never appear in any event's `state_delta`, so they were absent from `state_at_rewind_point` and incorrectly set to `None` after `rewind_async()`.

## Root Cause

Step 2 of the rewind state delta computation sets `rewind_state_delta[key] = None` for any key in `current_state` not found in `state_at_rewind_point`. This over-eagerly nullifies keys that were set as initial session state, not through events.

## Fix

Before computing the delta, collect `keys_ever_in_event_deltas` — the set of all keys that appeared in **any** event's `state_delta` throughout the entire session.

In step 2, only nullify keys that:
1. Are NOT in `state_at_rewind_point` (were not present at rewind point), AND  
2. DID appear in at least one event's `state_delta` (were set via events, not initial state)

Keys that never appeared in any event delta are treated as "initial state" (set via `create_session`) and are preserved through rewinds.

## Test

Reproduction from issue #4839:
```python
# After rewind_async(), my_key should still be "important_value"
print(rewound.state.get("my_key"))  # Was: None, Now: "important_value"
print(rewound.state.get("other_key"))  # "foo" (correctly restored)
```


saiprasanth-git:
Thanks @gemini-code-assist for the review! Applied the suggested fix in commit 4182ab6 — replaced the imperative loop with a set comprehension and moved `rewind_state_delta = {}` outside the loop to avoid re-initialization on every iteration and prevent a potential `NameError` when no events have state deltas.

rohityan:
Hi @saiprasanth-git , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

GWeale:
half of this landed. the fragment fix in rest_api_tool is on main as 0cfe43b6, from your own #4800, with your test. not in a release yet.

the rewind half didn't — that code moved to sessions/_rewind_utils.py, so the diff has nothing to apply to, and #4839 stays open. your diagnosis was right. closing; #7118 is redoing that half against the new module, and a review there would help more than a rebase here.

PR #4906, opened 2026-03-20
fix: detect MCP session disconnect to avoid ~5min hang on HTTP errors

### LINKED ISSUE #4901: ADK agent hangs for ~5 minutes when MCP server returns non-2xx HTTP response
**Related issue:** https://github.com/google/adk-python/issues/4162

## Description

When an MCP server (or a proxy/gateway in front of one) returns a non-2xx HTTP response (e.g., 403 Forbidden), the ADK agent hangs for approximately 5 minutes before timing out. The MCP Python SDK's streamable HTTP transport crashes in a background `TaskGroup`, but the crash never propagates to the `send_request()` caller -- it remains blocked on an `anyio` memory stream that will never receive data.

The `on_tool_error_callback` never fires because `send_request()` never raises an exception during the hang period. After the 5-minute `sse_read_timeout` expires, it raises `McpError(REQUEST_TIMEOUT)` rather than the original HTTP error.

## Environment

- **google-adk**: 1.27.2
- **mcp** (Python SDK): 1.26.0
- **Python**: 3.12
- **OS**: Linux (Debian 12 / GKE)
- **Model**: `gemini-3-flash-preview` (via Vertex AI Agent Engine)

## Steps to Reproduce

1. Set up any MCP server (or HTTP endpoint) that returns HTTP 403 for certain requests.
2. Create an ADK agent with `McpToolset` pointing to that server.
3. Invoke a tool that triggers the 403 response.
4. Observe the agent hangs for ~5 minutes with no response.

### Minimal reproduction

```python
"""Minimal reproduction: ADK agent hangs on MCP HTTP 403."""

from google.adk.tools import McpToolset
from google.adk.tools.mcp_tool.mcp_session_manager import StreamableHTTPConnectionParams

# Point to any MCP server behind a proxy that returns HTTP 403
# for certain tool calls (e.g., an authorization gateway)
MCP_URL = "https://example.com/mcp"  # returns 403 on blocked tools

agent = Agent(
    model="gemini-3-flash-preview",
    name="repro_agent",
    instruction="Call the 'blocked_tool' tool.",
    tools=[
        McpToolset(
            connection_params=StreamableHTTPConnectionParams(url=MCP_URL),
        ),
    ],
)
```

When the agent calls a tool and the server responds with HTTP 403, the agent hangs for ~5 minutes instead of reporting the error.

## Observed Behavior

- The agent UI/client keeps spinning with no response for ~5 minutes.
- Server logs show: `Error on session runner task: unhandled errors in a TaskGroup (1 sub-exception)`
- After ~5 minutes, the request times out with `McpError(REQUEST_TIMEOUT)`.
- `on_tool_error_callback` is **never invoked** during the hang.

## Expected Behavior

- The agent should immediately surface the HTTP error (or a user-friendly message) back to the LLM/user.
- `on_tool_error_callback` should be invoked with the HTTP error.
- The agent should not hang.

## Root Cause Analysis

Full code path trace through `mcp` SDK and ADK:

### 1. Tool call initiation
`mcp_tool.py:376` -> `session.call_tool("tool_name")` -> `session.send_request()` writes a JSON-RPC message to `write_stream`, then blocks waiting on `response_stream_reader.receive()` (`session.py:292`) with a ~5-minute timeout (`sse_read_timeout`).

### 2. Transport sends HTTP POST
`streamable_http.py:569` -> `post_writer` spawns `handle_request_async` via `tg.start_soon()` inside a `TaskGroup` (line ~647).

### 3. HTTP error raised
`streamable_http.py:358` -> `response.raise_for_status()` raises `httpx.HTTPStatusError` on the 403 response.

### 4. TaskGroup crashes
The `TaskGroup` at line ~647 catches the exception, wraps it in an `ExceptionGroup`, and crashes. The `finally` block closes `read_stream_writer` (line ~576).

### 5. Error logged but not propagated to caller
`session_context.py:189-191` -> catches `BaseException`, logs `"Error on session runner task"`, and re-raises. But this is in a background task -- it doesn't reach `send_request()`.

### 6. send_request() hangs
**Meanwhile**, `send_request()` at `session.py:292` is still blocked on `response_stream_reader.receive()`. This is a **per-request** `anyio` memory stream. The transport crashed before writing any response to it. The stream never receives data, so `send_request()` hangs until the `sse_read_timeout` (~5 minutes) expires.

### 7. on_tool_error_callback never fires
Because `send_request()` doesn't raise during the hang, the ADK error callback path is never triggered. When it finally times out, it raises `McpError(REQUEST_TIMEOUT)` -- losing the original HTTP 403 context entirely.

## Why `on_tool_error_callback` cannot work for HTTP-level errors

The error occurs in a background task (the transport's `TaskGroup`). The `send_request()` call waiting for a response doesn't receive the exception -- it's blocked on a memory stream that will never be written to. The two execution contexts (background transport task vs. foreground request) are not connected for error propagation.

## Suggested Fix

The `mcp` SDK's `StreamableHTTPTransport` should propagate HTTP errors from background tasks to the waiting `send_request()` call. Options:

1. **Write an error sentinel to the per-request memory stream** before closing it, so `send_request()` can raise immediately with the original error.
2. **Cancel the pending request** when the transport's TaskGroup crashes, instead of letting it hang until timeout.
3. **Catch `httpx.HTTPStatusError` in `handle_request_async`** and convert it to a JSON-RPC error response written to the response stream.

## Workaround

If you control the proxy/gateway returning the non-2xx response, return **HTTP 200 with a JSON-RPC response containing `CallToolResult(isError=true)`** instead. This keeps the error at the MCP protocol level where the SDK handles it correctly:

```json
{
  "jsonrpc": "2.0",
  "id": "<request-id>",
  "result": {
    "content": [{"type": "text", "text": "Tool call denied by authorization policy."}],
    "isError": true
  }
}
```

The MCP SDK parses this as a normal tool result, no exceptions are raised, and the LLM sees the denial message directly.

Ratnaditya-J:

- Closes: #4901

**Problem:**
When an MCP server (or a proxy in front of one) returns a non-2xx HTTP response (e.g. 403), the agent hangs for ~5 minutes. The transport crashes in a background `TaskGroup` but the error never propagates to the `send_request()` call — it just sits there blocked on an anyio memory stream that'll never get data. Eventually `sse_read_timeout` fires and you get a generic `REQUEST_TIMEOUT` instead of the actual HTTP error.

**Solution:**
Race the `session.call_tool()` against a lightweight health check that polls the session's stream state every 500ms. If the streams close (which happens when the transport crashes), we raise a `ConnectionError` immediately with a descriptive message instead of waiting for the 5-minute timeout.

The health check uses `MCPSessionManager._is_session_disconnected()` which already exists and checks `_read_stream._closed` / `_write_stream._closed`. The `retry_on_errors` decorator then kicks in and retries once (creating a fresh session), which is the expected behavior for connection failures.

This is an ADK-level workaround for the upstream MCP SDK issue where `StreamableHTTPTransport` doesn't propagate HTTP errors from background tasks to pending `send_request()` calls.

### Testing Plan

**Unit Tests:**

Added two tests:
- `test_call_tool_with_health_check_disconnected_session` — simulates the transport crash by closing streams while `call_tool` is hanging, verifies `ConnectionError` is raised within ~1s instead of hanging
- `test_call_tool_with_health_check_success` — verifies normal tool calls still work fine with the health check running alongside

```
44 passed in 3.30s
```

### Checklist

### Additional context

Ideally the MCP Python SDK itself would propagate HTTP errors to pending requests (see the suggested fixes in #4901), but this workaround gives fast failure at the ADK level without waiting for an upstream fix.


rohityan:
Hi @Ratnaditya-J , Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

Ratnaditya-J:
Thanks, @rohityan. I have already signed the CLA and unsure why its not reflected in your system. Check the attached image. 
Screenshot 2026-03-20 at 1 22 20 PM

Ratnaditya-J:
@rohityan  looks like the cla check has passed now.

GWeale:
Thanks for digging into this hang. Main already detects the transport crash and fails fast instead of waiting out the sse_read_timeout, via SessionContext._run_guarded in src/google/adk/tools/mcp_tool/session_context.py (on by default), so we are closing this as already-resolved.

PR #4912, opened 2026-03-20
feat: add event_callback parameter to AgentTool for sub-agent observability

### LINKED ISSUE #4905: Feature: AgentTool should support an event callback for sub-agent observability
## Problem

`AgentTool` creates an isolated `Runner` internally and consumes all intermediate sub-agent events without propagating them to the parent. The only output visible to the caller is the final result text. This makes it impossible to observe what a sub-agent is doing while it runs — no tool calls, no thoughts, no state deltas, no intermediate text.

For applications that use `AgentTool` to compose deep agent hierarchies, this is a significant observability gap. Users see a long pause with no output until the sub-agent finishes.

## Current behavior (`agent_tool.py`, `run_async`, ~line 248-258)

```python
last_content = None
async with Aclosing(
    runner.run_async(
        user_id=session.user_id, session_id=session.id, new_message=content
    )
) as agen:
    async for event in agen:
        # Forward state delta to parent session.
        if event.actions.state_delta:
            tool_context.state.update(event.actions.state_delta)
        if event.content:
            last_content = event.content
```

Every intermediate event is consumed silently. Only `state_delta` forwarding and `last_content` capture happen. The events are discarded.

## Proposed change

Add an optional `event_callback` parameter to `AgentTool.__init__` that, when provided, is called with each intermediate event during `run_async`. This is fully backwards-compatible — the default is `None` (no callback, identical to current behavior).

```python
class AgentTool(BaseTool):
    def __init__(
        self,
        agent: BaseAgent,
        skip_summarization: bool = False,
        *,
        include_plugins: bool = True,
        event_callback: Callable[[Event], None] | None = None,  # NEW
    ):
        self.agent = agent
        self.skip_summarization = skip_summarization
        self.include_plugins = include_plugins
        self.event_callback = event_callback
        super().__init__(name=agent.name, description=agent.description)
```

And in `run_async`, after the existing state_delta/content handling:

```python
async for event in agen:
    if event.actions.state_delta:
        tool_context.state.update(event.actions.state_delta)
    if event.content:
        last_content = event.content

    # NEW: forward intermediate events to callback
    if self.event_callback is not None:
        self.event_callback(event)
```

## Use case

We build a CLI tool that orchestrates multi-agent pipelines using ADK. Our agent hierarchy is 3-4 levels deep (coordinator → board manager → implementation manager → coding pipeline). Without event propagation, the user sees nothing for minutes while deeply nested agents work. With the callback, we can render intermediate events in real time with proper hierarchy indentation.

We currently work around this with a custom `ObservableAgentTool` subclass that overrides `run_async` entirely, duplicating ~50 lines of internal ADK code. This is fragile and breaks when ADK internals change. A first-class callback parameter would make this unnecessary.

## Alternatives considered

- **Yielding events to the parent Runner**: This would change the semantics of tool execution (tools return a result, not a stream). The callback approach is less invasive.
- **Plugin hooks**: The existing `on_event_callback` plugin hook only fires for events yielded by the top-level Runner, not for events consumed inside `AgentTool`. A plugin-based approach could also work but would require deeper changes.
- **Subclassing** (current workaround): Works but duplicates internal code and is fragile across versions.

## Environment

- Python 3.12

brucearctor:
## Description

Closes #4905

`AgentTool` creates an isolated `Runner` internally and consumes all intermediate sub-agent events without propagating them to the parent. This makes it impossible to observe what a sub-agent is doing while it runs.

This PR adds an optional `event_callback` parameter to `AgentTool.__init__` that, when provided, is called with each intermediate event during `run_async`. This is fully backwards-compatible — the default is `None` (no callback, identical to current behavior).

### `src/google/adk/tools/agent_tool.py`
- Added `Callable` import
- Added optional `event_callback: Callable[..., None] | None = None` keyword argument to `__init__`
- In `run_async`, after existing `state_delta`/content handling, calls `self.event_callback(event)` if set

### Unit Tests (3 new tests added)
- `test_event_callback_receives_intermediate_events` — verifies callback collects events from tool_agent
- `test_event_callback_none_by_default` — verifies backward compatibility (no callback = same behavior)
- `test_event_callback_with_state_delta` — verifies callback works alongside state forwarding

### Results
```
5013 passed, 1950 warnings in 123.13s
```

Full unit test suite passes with zero failures.


rohityan:
Hi @brucearctor, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

xuanyang15:
https://github.com/google/adk-python/issues/4905 is closed.

brucearctor:
@xuanyang15  or @rohityan  -- any comments, after review?  I only see that this was closed?

xuanyang15:
Hi @brucearctor, sorry, I should provide more context. As discussed in https://github.com/google/adk-python/issues/4905, since the existing plugin interface works, we don't plan to proceed with this PR.

PR #4919, opened 2026-03-20
fix(telemetry): prevent spurious OTel context detach errors when async generators are cancelled

### LINKED ISSUE #4894: OpenTelemetry context detach errors when BaseAgent wrapper exits inner agent's run_async() early
## Describe the bug

When a `BaseAgent` subclass wraps another agent and exits early from iterating over `self.main_agent.run_async(ctx)` (via `return` or `break` inside `async for`), multiple `Failed to detach context` errors are logged by OpenTelemetry. This happens because `base_agent.run_async()` wraps the generator yield in `tracer.start_as_current_span()`, and when `GeneratorExit` propagates through the generator chain, the OTel context tokens cannot be detached — they were created in a different `contextvars.Context`.

**google-adk version**: 1.27.2
**Python version**: 3.13
**opentelemetry-api version**: 1.33.0

This appears to be the same root cause as #423 and #1670 (both closed), but with a distinct and reproducible trigger path.

## To Reproduce

Create a `BaseAgent` subclass that wraps another agent and conditionally exits early:

```python
from google.adk.agents import BaseAgent, LlmAgent
from typing_extensions import override

class WrapperAgent(BaseAgent):
    main_agent: BaseAgent

    @override
    async def _run_async_impl(self, ctx) -> AsyncGenerator[Event, None]:
        async for event in self.main_agent.run_async(ctx):
            if should_block(event):
                yield create_redirect_event(ctx)
                return  # <-- triggers GeneratorExit on inner generator chain
            yield event
```

When `should_block()` returns `True` and the wrapper does `return` inside the `async for` loop, the inner generator (`main_agent.run_async(ctx)`) is abandoned. Python sends `GeneratorExit` into the generator chain, which unwinds through `base_agent.run_async()`:

```
base_agent.run_async()
  → tracer.start_as_current_span('invoke_agent ...')  # attaches OTel context token
    → Aclosing(self._run_async_impl(ctx))
      → async for event in agen:
        → yield event  # GeneratorExit thrown here
    # __exit__ tries to detach token → ValueError
```

## Error output

```
ERROR:opentelemetry.context:Failed to detach context
  File ".../opentelemetry/trace/__init__.py", line 589, in use_span
    yield span
  File ".../opentelemetry/sdk/trace/__init__.py", line 1105, in start_as_current_span
    yield span
  File ".../google/adk/agents/base_agent.py", line 304, in run_async
    yield event
GeneratorExit

During handling of the above exception, another exception occurred:

  File ".../opentelemetry/context/__init__.py", line 155, in detach
    _RUNTIME_CONTEXT.detach(token)
  File ".../opentelemetry/context/contextvars_context.py", line 53, in detach
    self._current_context.reset(token)
ValueError: <Token ...> was created in a different Context
```

This repeats for each nesting level in the agent/tracing hierarchy (4 errors for a typical pipeline).

## Root cause analysis

`base_agent.run_async()` at line 294 uses a synchronous context manager (`tracer.start_as_current_span()`) wrapping an async generator's `yield` point:

```python
with tracer.start_as_current_span(f'invoke_agent {self.name}') as span:
    # ...
    async with Aclosing(self._run_async_impl(ctx)) as agen:
        async for event in agen:
            yield event  # <-- GeneratorExit enters here
```

The `start_as_current_span` context manager calls `context.attach()` on `__enter__` (storing a `contextvars.Token`) and `context.detach(token)` on `__exit__`. When `GeneratorExit` is thrown into the generator, `__exit__` fires, but the `ContextVar.reset(token)` fails because the token was created in a different `contextvars.Context` (the async context has shifted between generator suspension and cleanup).

## Workaround

**Application-level**: Suppress the OTel context logger:
```python
logging.getLogger("opentelemetry.context").setLevel(logging.CRITICAL)
```

**Wrapper agent level**: Use `break` instead of `return` inside the `async for` loop, and yield the redirect event after the loop exits. This ensures the inner generator's cleanup (including OTel span detach) completes before the wrapper yields:

```python
@override
async def _run_async_impl(self, ctx) -> AsyncGenerator[Event, None]:
    redirect_event = None
    async for event in self.main_agent.run_async(ctx):
        if should_block(event):
            redirect_event = create_redirect_event(ctx)
            break  # inner generator closed here, in same async context
        yield event
    if redirect_event is not None:
        yield redirect_event
```

**Note**: The `break` workaround may or may not resolve the issue depending on whether Python's `async for` cleanup after `break` preserves the `contextvars.Context`. In our testing, it reduces but may not fully eliminate the errors.

## Suggested fix

Consider wrapping the tracing in `base_agent.run_async()` with `GeneratorExit`-safe cleanup:

```python
async def run_async(self, parent_context):
    with tracer.start_as_current_span(f'invoke_agent {self.name}') as span:
        ctx = self._create_invocation_context(parent_context)
        tracing.trace_agent_invocation(span, self, ctx)
        try:
            if event := await self._handle_before_agent_callback(ctx):
                yield event
            if ctx.end_invocation:
                return
            async with Aclosing(self._run_async_impl(ctx)) as agen:
                async for event in agen:
                    yield event
            if ctx.end_invocation:
                return
            if event := await self._handle_after_agent_callback(ctx):
                yield event
        except GeneratorExit:
            # Suppress OTel context detach errors during generator cleanup
            span.end()
            return
```

Or use an async-generator-safe tracing wrapper that doesn't rely on `contextvars.Token` for cleanup.

## Related issues

- #423 (closed) — same error, different trigger (parallel tool downloads)
- #1670 (closed) — same error, SSE streaming context
- #3607 (closed) — same error, Human-in-the-Loop tool confirmation
- #2792 — request to disable/extend internal OTel tracing

mmphego:
## Summary

Fixes #4894

Eliminates the noisy `ERROR [opentelemetry.context] Failed to detach context` log that appears whenever an in-flight agent run is interrupted mid-generation (e.g. a guardrail that blocks a request after the LLM has already started streaming).

## Root Cause

`tracer.start_as_current_span()` returns a context manager that stores a `contextvars.Token` on enter and calls `ContextVar.reset(token)` on exit. A `Token` can **only** be reset in the same `contextvars.Context` object that produced it.

When an async generator using `start_as_current_span()` is closed from a different `asyncio.Task` — which happens when asyncio's asyncgen finalizer hook schedules `aclose()` via `call_soon` in the event-loop's **base context** — `reset(token)` raises `ValueError`.

`otel_context.detach()` catches this `ValueError` internally and logs it at ERROR level before returning, so the error is invisible to callers but still appears in logs.

## Affected code paths

| File | Location | Span |
|---|---|---|
| `tracing.py` | `_use_native_generate_content_span_stable_semconv` | `generate_content {model}` |
| `tracing.py` | `_use_native_generate_content_span` (experimental semconv) | `generate_content {model}` |
| `base_agent.py` | `run_async` | `invoke_agent {name}` |
| `base_agent.py` | `run_live` | `invoke_agent {name}` |

## Fix

Replace `with tracer.start_as_current_span(...)` in async generator bodies with **manual span management**:

```python
span = tracer.start_span(name)
token = otel_context.attach(trace.set_span_in_context(span))
try:
    yield span
finally:
    _safe_detach(token)
    if span.is_recording():
        span.end()
```

`_safe_detach()` calls `_RUNTIME_CONTEXT.detach()` directly (bypassing OTel's logging wrapper) and silently absorbs `ValueError` when cleanup runs in a different context.

**Span data is fully preserved** (attributes, timing, parent relationships). Only the `contextvars` state restoration is skipped for generators that are being discarded — an acceptable trade-off.

## Reproducer

```python
import asyncio
from opentelemetry import context as otel_context, trace
from opentelemetry.context import _RUNTIME_CONTEXT
from opentelemetry.sdk.trace import TracerProvider

trace.set_tracer_provider(TracerProvider())
tracer = trace.get_tracer("test")

async def gen_with_span():
    with tracer.start_as_current_span("test_span"):
        yield "event"

async def main():
    gen = gen_with_span()
    await gen.__anext__()
    # Closing from a different Task reproduces the error:
    await asyncio.create_task(gen.aclose())  # logs ERROR: Failed to detach context

asyncio.run(main())
```

## Testing

Reproducer and fix verification tests added in the linked issue. Key test cases:

- `test_should_reproduce_error_when_generator_closed_from_different_task` — confirms the bug exists with the raw `start_as_current_span` pattern
- `test_should_not_error_with_adk_safe_span_pattern` — confirms the `_safe_detach` pattern eliminates the error even under cross-context close

## Checklist

- [x] Bug reproduced with a minimal test case
- [x] Fix verified: no ERROR logs after patching
- [x] Span data preserved (manual span management is functionally equivalent)
- [x] Both `run_async` and `run_live` paths fixed
- [x] Both stable and experimental semconv paths fixed


rohityan:
Hi @mmphego , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the formatting errors

mmphego:
@rohityan please run the workflows

mmphego:
Bump!!

RKest:
Hi @mmphego

Thank you for your contribution.

We're aware of this issue, and unfortunately this is a core limitation of how ContextVars & Python Async Generators interplay with each other.

The solution is to wrap all async generators in `contextlib.aclosing`, but we sometimes miss a few or, sometimes user code which interacts with ADK doesn't warp their Async Generators.

I fear replacing the OTel error message with a custom `logger.debug`, would only hurt discoverability of the error especially because it can also result in wrong span hierarchies, hence I don't think this should be merged in

rohityan:
Hi @mmphego , closing this since a requested clarification from @jawoszek in #4894 is pending.

Issue #3765, opened 2025-11-30
Task id not respected while sending a2a send message request


AneshkaGoyal:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
When making call to a2a remote agent, the first send message request does not contain any context id or task id which is relevant. When the remote agent responds with a task that has task id and context id and task state as Input required, the subsequent send message request sends the context id but not the task id back even though it has it.

**To Reproduce**
the first send message request does not contain any context id or task id which is relevant. When the remote agent responds with a task that has task id and context id and task state as Input required, the subsequent send message request sends the context id but not the task id back even though it has it.

**Expected behavior**
Expected behaviour is to have task id sent along with context id.

**Proposed code changes are as below**

```
def _construct_message_parts_from_session(
      self, ctx: InvocationContext
  ) -> tuple[list[A2APart], dict[str, Any], str]:
    """Construct A2A message parts from session events.

    Args:
      ctx: The invocation context

    Returns:
      List of A2A parts extracted from session events, context ID, task ID
    """
    message_parts: list[A2APart] = []
    context_id = None
    task_id = None
    for event in reversed(ctx.session.events):
      if _is_other_agent_reply(self.name, event):
        event = _present_other_agent_message(event)
      elif event.author == self.name:
        # stop on content generated by current a2a agent given it should already
        # be in remote session
        if event.custom_metadata:
          context_id = (
              event.custom_metadata.get(A2A_METADATA_PREFIX + "context_id")
              if event.custom_metadata
              else None
          )
          task_id = (
              event.custom_metadata.get(A2A_METADATA_PREFIX + "task_id")
              if event.custom_metadata
              else None
          )
        break

      if not event.content or not event.content.parts:
        continue

      for part in event.content.parts:

        converted_part = self._genai_part_converter(part)
        if converted_part:
          message_parts.append(converted_part)
        else:
          logger.warning("Failed to convert part to A2A format: %s", part)

    return message_parts[::-1], context_id, task_id
```

```
async def _run_async_impl(
      self, ctx: InvocationContext
  ) -> AsyncGenerator[Event, None]:
    """Core implementation for async agent execution."""
    try:
      await self._ensure_resolved()
    except Exception as e:
      yield Event(
          author=self.name,
          error_message=f"Failed to initialize remote A2A agent: {e}",
          invocation_id=ctx.invocation_id,
[...]

PR #4944, opened 2026-03-21
fix: RemoteA2aAgent fails to send taskId in multi-turn scenarios

### LINKED ISSUE #4934: RemoteA2AAgent fails to send taskId in multi-turn scenarios
## 🔴 Required Information

**Describe the Bug:**
RemoteA2AAgent only sends contextId and doesn't send the associated taskId for multi-turn A2A scenarios.

**Steps to Reproduce:**
1. Create an agent with a RemoteA2AAgent as show below. You can use currency agent from a2a-samples repository.
```

currency_agent = RemoteA2aAgent(
    name="currency_agent",
    agent_card="http://localhost:10999/.well-known/agent-card.json",
)

root_agent = Agent(
    model='gemini-2.5-flash',
    name='root_agent',
    description='A helpful assistant for user questions.',
    instruction='Helpful assistant to answer user queries. Use currency_agent for any currency related queries from the user',
    sub_agents=[currency_agent]
)
```
2. Launch `adk web` and start a new conversation with message 'How much is 10 USD?'.
3. A2A agent responds back with message 'Provide the currency to convert to' and task.status='input-required'.
4. Send the message 'CAD' as user's second turn.

**Expected Behavior:**
ADK should invoke the remote agent with task_id and context_id for the second turn as the task has not terminated.

**Observed Behavior:**
ADK only sends the context_id in the ADK payload. This results in unexpected agent behavior and results in creation of new tasks for every turn and stale task records get created.

---

## 🟡 Optional Information
https://github.com/google/adk-python/blob/4b677e73b939f5a13269abd9ba9fe65e4b78d7f6/src/google/adk/agents/remote_a2a_agent.py#L643 only sets the context_id.

Ideally RemoteA2AAgent implementation should include the taskId when the task status is in paused awaiting user input such as 'input-required', 'auth-required' states.

**Regression:**
No

 - Always (100%)

shriramThakare3:
- Closes: #4934
- Related: #3765

**Problem:**
`RemoteA2aAgent` only sends `context_id` on subsequent turns, never `task_id`, causing the remote server to spawn a new task every turn instead of resuming the paused one. A multi-turn flow (e.g. `input-required` → user replies) always creates stale orphaned tasks on the remote agent.

**Solution:**
In `_construct_message_parts_from_session`, after reading `context_id` from `custom_metadata`, also read the stored task state from the previous response metadata. Forward `task_id` only when the task is still open (`input-required` or `auth-required`). For terminal states (`completed`, `failed`, `cancelled`) `task_id` stays `None` so the server correctly starts a fresh task under the same `context_id`.

### Testing Plan

**Unit Tests:**

- `test_multiturn_sends_task_id_when_input_required` — verifies `task_id` is included in the second-turn request when task state is `input-required`
- `test_multiturn_sends_task_id_when_auth_required` — same for `auth-required`
- `test_first_turn_omits_task_id` — verifies `task_id` is `None` on the first turn
- `test_completed_task_omits_task_id` — verifies `task_id` is not forwarded when previous task state is `completed`

1. Set up a `RemoteA2aAgent` pointing at the currency agent from `a2a-samples`
2. Launch `adk web` and send `How much is 10 USD?`
3. Agent responds with `Provide the currency to convert to` and `task.status=input-required`
4. Send `CAD` as the second turn
5. Verify the outgoing A2A request contains both `contextId` and `taskId` matching the first turn's task
6. Verify the remote agent returns the conversion result without creating a new task

### Checklist

### Additional context
The same root cause was previously reported in #3765. That issue proposed a fix but it was not fully applied — `task_id` was still never populated because the state check was missing.


rohityan:
Hi @shriramThakare3 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing unit tests , mypy-diff tests and formatting errors

shriramThakare3:
hello, @rohityan 
made changes to fix failing tests

src/google/adk/agents/remote_a2a_agent.py
```diff
@@ -400,6 +401,14 @@ def _construct_message_parts_from_session(
...
           metadata = event.custom_metadata
           context_id = metadata.get(A2A_METADATA_PREFIX + "context_id")
+          response_meta = metadata.get(A2A_METADATA_PREFIX + "response", {})
+          task_state = None
+          if isinstance(response_meta, dict):
+            status = response_meta.get("status", {})
+            if isinstance(status, dict):
+              task_state = status.get("state")
+          if task_state in ("input-required", "auth-required"):
+            task_id = metadata.get(A2A_METADATA_PREFIX + "task_id")
```
guglielmo-san: The issue with this implementation is that it is not possible to define if the new user message is a follow up to the input-required event, or a request for a new task -> in this case the task_id should not be set

shriramThakare3: The remote agent owns the task,it knows if the task_id is still valid or not. If the user starts a new request, the remote agent can handle it. Forwarding task_id on input-required is the right default behavior. If more control is needed, that can be a separate improvement.

GWeale:
RemoteA2aAgent moved into the a2a package and agents/remote_a2a_agent.py is a re-export shim, so every hunk targets code that's gone. part of this is on main: task_id is forwarded when resuming from a function response. the general path still sends only context_id, so #4934 stays open. a fresh pull request against the a2a module is welcome.

PR #3839, opened 2025-12-05
feat: Add support for custom task store in to_a2a function

secprog:
This update introduces an optional task_store parameter to the to_a2a function, allowing users to provide a custom task store instance. If not specified, the function defaults to using InMemoryTaskStore. Additionally, unit tests have been added to verify the functionality with a custom task store.

Replaces my old closed PR https://github.com/google/adk-python/pull/3639.

PR #4970, opened 2026-03-24
feat(a2a): add a2a_task_store and a2a_push_config_store params to get_fast_api_app

### LINKED ISSUE #4971: get_fast_api_app does not allow injecting a custom A2A task store
## Summary

`get_fast_api_app` unconditionally creates an `InMemoryTaskStore` and `InMemoryPushNotificationConfigStore` when `a2a=True`. There is no way for callers to provide their own store implementations.

This becomes a problem in production setups where:
- Multiple replicas need a **shared** task store so A2A push callbacks can be routed back to the right server
- Operators want task state to **survive restarts** (e.g. using `DatabaseTaskStore` backed by SQLite or Postgres from `a2a-sdk`)

## Current behaviour

```python
# Inside get_fast_api_app — no way to override
a2a_task_store = InMemoryTaskStore()
```

## Expected behaviour

Callers should be able to pass a custom store:

```python
from a2a.server.tasks import DatabaseTaskStore
from sqlalchemy.ext.asyncio import create_async_engine

engine = create_async_engine("sqlite+aiosqlite:///tasks.db")
app = get_fast_api_app(
    agents_dir="agents/",
    web=True,
    a2a=True,
    a2a_task_store=DatabaseTaskStore(engine),
)
```

## Notes

The lower-level `to_a2a()` helper already supports this via the `task_store` parameter added in #3839. This issue tracks extending the same capability to `get_fast_api_app`.

A fix is proposed in #4970.

STHITAPRAJNAS:
Closes #4971.

Follows the same pattern as #3839 (`to_a2a`), but for the higher-level `get_fast_api_app` entry point.

Currently both stores are unconditionally created as in-memory instances inside the function, with no way for callers to inject their own. This makes it impossible to use a persistent backend (e.g. `DatabaseTaskStore` backed by SQLite or Postgres) without monkey-patching.

The fix adds two optional kwargs:

```python
def get_fast_api_app(
    ...
    a2a_task_store: Optional[Any] = None,
    a2a_push_config_store: Optional[Any] = None,
    ...
)
```

When `None` (the default), the existing `InMemory*` instances are created as before, so there is no change in behaviour for existing callers.

Both stores are now initialised once before the per-agent loop rather than `push_config_store` being re-created on every iteration, which was likely unintentional in the original code.

Two tests added:
- `test_a2a_uses_in_memory_task_store_by_default` — default path unchanged
- `test_a2a_custom_task_store_is_used` — custom store forwarded, `InMemoryTaskStore` not constructed

### DIFF
--- a/src/google/adk/cli/fast_api.py
+++ b/src/google/adk/cli/fast_api.py
@@ -131,22 +131,11 @@ def get_fast_api_app(
     web: Whether to enable the web UI and serve its assets.
     a2a: Whether to enable Agent-to-Agent (A2A) protocol support.
     a2a_task_store: Optional A2A TaskStore instance. Defaults to
-      InMemoryTaskStore when a2a=True. Pass a DatabaseTaskStore (from the
-      a2a-sdk) for persistence across server restarts and horizontal replicas.
-      Example::
-
-          from a2a.server.tasks import DatabaseTaskStore
-          from sqlalchemy.ext.asyncio import create_async_engine
-
-          engine = create_async_engine("postgresql+asyncpg://user:pw@host/db")
-          app = get_fast_api_app(
-              agents_dir="agents/", web=True, a2a=True,
-              a2a_task_store=DatabaseTaskStore(engine),
-          )
-
-    a2a_push_config_store: Optional A2A PushNotificationConfigStore instance.
-      Defaults to InMemoryPushNotificationConfigStore when a2a=True. Pass a
-      DatabasePushNotificationConfigStore for persistence across restarts.
+      InMemoryTaskStore. Pass a custom store (e.g. DatabaseTaskStore) to
+      persist task state across restarts or share it across replicas.
+    a2a_push_config_store: Optional A2A PushNotificationConfigStore.
+      Defaults to InMemoryPushNotificationConfigStore. Pass a custom store
+      for persistence across restarts.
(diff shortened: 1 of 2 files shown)


STHITAPRAJNAS:
Linking issue: Fixes #4971

CLA is now signed and passing. The linked issue (#4971) tracks the missing `a2a_task_store` / `a2a_push_config_store` params in `get_fast_api_app`. This PR implements that feature.

STHITAPRAJNAS:
Hi @rohityan — CLA is signed and the linked issue (#4971) is in place. All checks are passing. Could you assign a reviewer or take a look when you get a chance? Thanks!

STHITAPRAJNAS:
Hi @rohityan — just following up here. The clarification requested by the triage bot has been addressed (issue #4971 linked, CLA signed), but the `request clarification` label is still on. Could you clear that and assign a reviewer when you get a chance? Happy to make any adjustments if there's something else needed.

STHITAPRAJNAS:
Closing this as the underlying need has been addressed upstream. PR #5597 ("feat(a2a): add support for persistent task stores") was merged via commit `cd78d87b`, which adds persistent task store support to `get_fast_api_app` referencing the same issue #4971. The implementation went via `task_store_uri` rather than direct object injection, but the core gap is now filled. Thanks for looking into this.

PR #4978, opened 2026-03-24
fix: include rewind_before_invocation_id in VertexAiSessionService.append_event

### LINKED ISSUE #4930: VertexAiSessionService.append_event drops rewind_before_invocation_id, breaking rewind filtering on subsequent reads
## 🔴 Required Information

**Describe the Bug:**

`VertexAiSessionService.append_event` does not serialize `EventActions.rewind_before_invocation_id` when writing events to the Vertex AI API. The field is omitted from the `config['actions']` dict (lines 261–274 of `vertex_ai_session_service.py`). On subsequent reads, `_from_api_event` never reconstructs it, so the field is permanently lost.

This means ADK's own rewind filtering in `contents.py` (lines 432–449) — which depends on `event.actions.rewind_before_invocation_id` — silently becomes a no-op for any session backed by `VertexAiSessionService`. After a rewind, the rewound events are not filtered out of the LLM context on subsequent agent invocations.

**Steps to Reproduce:**

1. Create an ADK application using `VertexAiSessionService` (i.e. `agentengine://` session URI)
2. Run a few agent invocations to build up session events
3. Call `runner.rewind_async(user_id=..., session_id=..., rewind_before_invocation_id="<target_invocation_id>")`
4. Verify: `runner.rewind_async` creates an event with `actions.rewind_before_invocation_id` set and calls `session_service.append_event` (runners.py:672–684) — this succeeds
5. Call `session_service.get_session(...)` to reload the session
6. Inspect the last event (the rewind event): `session.events[-1].actions.rewind_before_invocation_id` is `None`
7. Trigger another agent invocation — the LLM context in `contents.py` includes the events that should have been annulled by the rewind

**Expected Behavior:**

After calling `rewind_async`, reloading the session should preserve `rewind_before_invocation_id` on the rewind event. The rewind filtering in `contents.py:432-449` should correctly annul the targeted events from the LLM context.

**Observed Behavior:**

`rewind_before_invocation_id` is silently dropped during `append_event` serialization. On the next `get_session` call, the rewind event has `actions.rewind_before_invocation_id = None`. The rewind filtering in `contents.py` has no effect, so the LLM sees the full un-rewound event history.

**Root Cause:**

In `vertex_ai_session_service.py`, `append_event` manually constructs the `config['actions']` dict (lines 261–274) with an explicit allowlist of fields:

```python
config['actions'] = {
    'skip_summarization': event.actions.skip_summarization,
    'state_delta': event.actions.state_delta,
    'artifact_delta': event.actions.artifact_delta,
    'transfer_agent': event.actions.transfer_to_agent,
    'escalate': event.actions.escalate,
    'requested_auth_configs': { ... },
    # TODO: add requested_tool_confirmations, agent_state once
    # they are available in the API.
}
```

`rewind_before_invocation_id` is not in this allowlist.

Similarly, `_from_api_event` (lines 362–437) reconstructs `EventActions` via `actions.model_dump()` → `EventActions.model_validate()`, which would round-trip the field correctly **if** it were persisted — but since it was never written, it's absent from the API response.

**Note:** The `_compaction` field had the same problem and was solved by stashing it in `custom_metadata` (see lines 296–307 of `append_event` and lines 378–387 of `_from_api_event`). A similar approach could work for `rewind_before_invocation_id`, or the field could simply be added to the `config['actions']` allowlist.

**Environment Details:**

 - ADK Library Version: 1.27.1
 - Desktop OS: macOS (darwin 24.6.0)
 - Python Version: 3.13

**Model Information:**

 - Are you using LiteLLM: No

---

## 🟡 Optional Information

 - Always

yuting0624:
## Summary

Fixes #4930

`VertexAiSessionService.append_event()` was not including `rewind_before_invocation_id` in the `actions` dict sent to the Vertex AI Agent Engine API. This caused rewind events to lose their target invocation ID when persisted, making session rewind non-functional for Agent Engine deployments.

## Changes

- **`src/google/adk/sessions/vertex_ai_session_service.py`**: Added `rewind_before_invocation_id` to the actions dict in `append_event()` (3 lines)
- **`tests/unittests/sessions/test_vertex_ai_session_service.py`**: Added `test_append_event_with_rewind` to verify the field round-trips correctly through append and get

## Root Cause

The `actions` dict construction in `append_event()` explicitly maps each field from `EventActions` to the API payload, but `rewind_before_invocation_id` was omitted. Other fields like `transfer_to_agent`, `state_delta`, `skip_summarization`, etc. were all present.

### DIFF
--- a/src/google/adk/sessions/vertex_ai_session_service.py
+++ b/src/google/adk/sessions/vertex_ai_session_service.py
@@ -282,6 +282,9 @@ class VertexAiSessionService(BaseSessionService):
               k: json.loads(v.model_dump_json(exclude_none=True, by_alias=True))
               for k, v in event.actions.requested_auth_configs.items()
           },
+          'rewind_before_invocation_id': (
+              event.actions.rewind_before_invocation_id
+          ),
           # TODO: add requested_tool_confirmations, agent_state once
           # they are available in the API.
           # Note: compaction is stored via event_metadata.custom_metadata.
--- a/src/google/adk/sessions/vertex_ai_session_service.py
+++ b/src/google/adk/sessions/vertex_ai_session_service.py
@@ -43,6 +43,7 @@ from .session import Session
 logger = logging.getLogger('google_adk.' + __name__)
 
 _COMPACTION_CUSTOM_METADATA_KEY = '_compaction'
+_REWIND_CUSTOM_METADATA_KEY = '_rewind_before_invocation_id'
 _USAGE_METADATA_CUSTOM_METADATA_KEY = '_usage_metadata'
 
 
@@ -282,12 +283,10 @@ class VertexAiSessionService(BaseSessionService):
               k: json.loads(v.model_dump_json(exclude_none=True, by_alias=True))
               for k, v in event.actions.requested_auth_configs.items()
           },
-          'rewind_before_invocation_id': (
-              event.actions.rewind_before_invocation_id
-          ),
           # TODO: add requested_tool_confirmations, agent_state once
           # they are available in the API.
-          # Note: compaction is stored via event_metadata.custom_metadata.
+          # Note: compaction and rewind_before_invocation_id are stored via
+          # event_metadata.custom_metadata.
       }
     if event.error_code:
       config['error_code'] = event.error_code
@@ -322,6 +321,16 @@ class VertexAiSessionService(BaseSessionService):
           metadata_dict,
           key=_COMPACTION_CUSTOM_METADATA_KEY,
           value=compaction_dict,
+      )
+    # Store rewind_before_invocation_id in custom_metadata since the Vertex AI
+    # service does not yet support the field in EventActions.
+    # TODO: Stop writing to custom_metadata once the Vertex AI service
+    # supports rewind_before_invocation_id natively in EventActions.
+    if event.actions and event.actions.rewind_before_invocation_id:
+      _set_internal_custom_metadata(
+          metadata_dict,
+          key=_REWIND_CUSTOM_METADATA_KEY,
+          value=event.actions.rewind_before_invocation_id,
       )
     # Store usage_metadata in custom_metadata since the Vertex AI service
     # does not persist it in EventMetadata.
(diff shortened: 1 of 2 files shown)


lucasbarzotto-axonify:
Hi @yuting0624, does this fix works?

I tried it and got the following error:

```
Error rewinding session 137352779248173056: 1 validation error for _AppendAgentEngineSessionEventRequestParameters

config.actions.rewind_before_invocation_id

Extra inputs are not permitted [type=extra_forbidden, input_value='p-146156f2-8f8e-42ae-af2a-75d00a48130c', input_type=str]

For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
```

Looking at [this](https://github.com/googleapis/python-aiplatform/blob/da663c0d000deeb5663d53334abc1d997f3575b5/vertexai/_genai/types/common.py#L11346), it looks like the `rewind_before_invocation_id` field is not allowed?

Should we use `custom_metadata` instead? The same way it's done for event compaction?

```
    # Store compaction data in custom_metadata since the Vertex AI service
    # does not yet support the compaction field.
    # TODO: Stop writing to custom_metadata once the Vertex AI service
    # supports the compaction field natively in EventActions.
    if event.actions and event.actions.compaction:
      compaction_dict = event.actions.compaction.model_dump(
          exclude_none=True, mode='json'
      )
      _set_internal_custom_metadata(
          metadata_dict,
          key=_COMPACTION_CUSTOM_METADATA_KEY,
          value=compaction_dict,
      )
```

yuting0624:
Thanks for catching this @lucasbarzotto-axonify! You're absolutely right — the Vertex AI SDK doesn't support `rewind_before_invocation_id` in `EventActions` yet.

I've pushed a fix that stores it in `custom_metadata` following the same pattern as `compaction`:
- **Write path:** `rewind_before_invocation_id` is now stored via `_set_internal_custom_metadata()` instead of directly in the `actions` dict
- **Read path:** Extracted from `custom_metadata` and restored into `EventActions` on retrieval

All 25 session service tests pass. The Pydantic validation error should no longer occur.

DeanChensj:
Thanks for the contribution!

Instead of introducing a new custom piggyback field (or extending the schema with a new custom parameter for this), we should leverage the underlying SDK's existing raw_event (or raw/extra data) field. This keeps the ADK schema cleaner and natively aligns with how the SDK already passes through arbitrary underlying event properties.

Could you update the implementation to map this data using the SDK's raw event field instead? Let us know if you need any pointers on where it's exposed!

PR #4979, opened 2026-03-24
fix: use load_artifact instead of file_data in rewind artifact restoration

### LINKED ISSUE #4932: Runner._compute_artifact_delta_for_rewind saves artifacts with "file_data", which GcsArtifactService and FileArtifactService don't support
## 🔴 Required Information

**Describe the Bug:**

`Runner._compute_artifact_delta_for_rewind` (runners.py:755–765) restores historical artifact versions by constructing a `types.Part(file_data=types.FileData(file_uri=...))` and passing it to `artifact_service.save_artifact`. However, `GcsArtifactService._save_artifact` explicitly raises `NotImplementedError` when it receives a `file_data` part (gcs_artifact_service.py:232–236), and `FileArtifactService` has no `file_data` handling path at all.

This means `rewind_async` crashes at runtime for any application using `GcsArtifactService` or `FileArtifactService` when the session contains artifacts that changed after the rewind point. Only `InMemoryArtifactService` works, since it stores the `Part` object as-is without attempting to extract bytes.

**Steps to Reproduce:**

1. Create an ADK application using `GcsArtifactService` (or `FileArtifactService`)
2. Run agent invocations that produce artifacts (e.g. save an image via `tool_context.save_artifact`)
3. Run additional invocations that modify or create new versions of the same artifact
4. Call `runner.rewind_async(user_id=..., session_id=..., rewind_before_invocation_id="<invocation_before_artifact_change>")`
5. Observe `NotImplementedError` from `GcsArtifactService`

```
NotImplementedError: Saving artifact with file_data is not supported yet in GcsArtifactService.
```

**Expected Behavior:**

`rewind_async` should successfully restore artifacts to their historical version regardless of which `ArtifactService` implementation is used. Since `load_artifact` already returns artifact content as `inline_data` (it reads the actual bytes from storage), the rewind code should use `load_artifact` to read the historical version and then `save_artifact` with `inline_data` — which all artifact service implementations support.

**Observed Behavior:**

`rewind_async` raises `NotImplementedError` when any artifact needs restoration and the artifact service is `GcsArtifactService`:

```
NotImplementedError: Saving artifact with file_data is not supported yet in GcsArtifactService.
```

The crash occurs at `runners.py:766` when calling `self.artifact_service.save_artifact(... artifact=artifact ...)` with the `file_data` part constructed at line 765.

**Root Cause:**

In `runners.py:755–765`, `_compute_artifact_delta_for_rewind` constructs an internal `artifact://` URI and wraps it in `file_data`:

```python
artifact_uri = artifact_util.get_artifact_uri(
    app_name=self.app_name,
    user_id=session.user_id,
    session_id=session.id,
    filename=filename,
    version=vt,
)
artifact = types.Part(file_data=types.FileData(file_uri=artifact_uri))
await self.artifact_service.save_artifact(
    app_name=self.app_name,
    user_id=session.user_id,
    session_id=session.id,
    filename=filename,
    artifact=artifact,
)
```

But `GcsArtifactService._save_artifact` (gcs_artifact_service.py:232–236) rejects `file_data`:

```python
elif artifact.file_data:
    raise NotImplementedError(
        "Saving artifact with file_data is not supported yet in"
        " GcsArtifactService."
    )
```

And `FileArtifactService._save_artifact` only handles `inline_data` and `text` — a `file_data` part falls through to the "Artifact must have either inline_data or text" validation error.

**Suggested Fix:**

Instead of constructing a `file_data` reference, the rewind code should `load_artifact` at the target version (which returns the actual bytes as `inline_data`) and then `save_artifact` with that `inline_data` part. This works with all artifact service implementations:

```python
# Instead of constructing a file_data reference:
loaded = await self.artifact_service.load_artifact(
    app_name=self.app_name,
    user_id=session.user_id,
    session_id=session.id,
    filename=filename,
    version=vt,
)
if loaded and loaded.inline_data:
    artifact = types.Part(
        inline_data=types.Blob(
            mime_type=loaded.inline_data.mime_type,
            data=loaded.inline_data.data,
        )
    )
else:
    # Fallback: artifact couldn't be loaded, mark as empty
    artifact = types.Part(
        inline_data=types.Blob(mime_type="application/octet-stream", data=b"")
    )
```

Note: the code already uses `inline_data` for the "artifact did not exist at rewind point" case (runners.py:750–753) — the fix just extends this approach to the "restore existing version" case.

**Environment Details:**

 - ADK Library Version: 1.27.1
 - Desktop OS: macOS (darwin 24.6.0)
 - Python Version: 3.13

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: gemini-2.5-pro

---

## 🟡 Optional Information

 - Always (100%) — occurs whenever `rewind_async` needs to restore an artifact and the artifact service is `GcsArtifactService` or `FileArtifactService`

yuting0624:
## Summary

Fixes #4932

`Runner._compute_artifact_delta_for_rewind` constructed a `file_data` reference to restore artifacts to their historical version during rewind. However:

- **`GcsArtifactService`** raises `NotImplementedError` for `file_data` parts
- **`FileArtifactService`** has no `file_data` handling path (falls through to validation error)

This caused `rewind_async` to crash for any application not using `InMemoryArtifactService`.

## Fix

Use `load_artifact(version=target_version)` to read the historical version's actual bytes, then `save_artifact` with `inline_data` — which all artifact service implementations support.

This is consistent with how the "artifact did not exist at rewind point" case already works (lines 750–753 use `inline_data`).

## Changes

- **`src/google/adk/runners.py`**: Replace `file_data` reference construction with `load_artifact` + `inline_data` pattern (+18/-3 lines)

### DIFF
--- a/src/google/adk/runners.py
+++ b/src/google/adk/runners.py
@@ -765,14 +765,15 @@ class Runner:
             version=vt,
         )
         if loaded and loaded.inline_data:
-          artifact = types.Part(
-              inline_data=types.Blob(
-                  mime_type=loaded.inline_data.mime_type,
-                  data=loaded.inline_data.data,
-              )
-          )
+          artifact = loaded
         else:
           # Fallback: artifact couldn't be loaded, mark as empty.
+          logger.warning(
+              'Could not load artifact %s version %s for rewind;'
+              ' replacing with empty blob.',
+              filename,
+              vt,
+          )
           artifact = types.Part(
               inline_data=types.Blob(
                   mime_type='application/octet-stream', data=b''
--- a/src/google/adk/runners.py
+++ b/src/google/adk/runners.py
@@ -754,15 +754,30 @@ class Runner:
         )
       else:
         # Artifact version changed after rewind point. Restore to version at
-        # rewind point.
-        artifact_uri = artifact_util.get_artifact_uri(
+        # rewind point by loading the actual bytes (inline_data) rather than
+        # constructing a file_data reference, since GcsArtifactService and
+        # FileArtifactService do not support saving file_data parts.
+        loaded = await self.artifact_service.load_artifact(
             app_name=self.app_name,
             user_id=session.user_id,
             session_id=session.id,
             filename=filename,
             version=vt,
         )
-        artifact = types.Part(file_data=types.FileData(file_uri=artifact_uri))
+        if loaded and loaded.inline_data:
+          artifact = types.Part(
+              inline_data=types.Blob(
+                  mime_type=loaded.inline_data.mime_type,
+                  data=loaded.inline_data.data,
+              )
+          )
+        else:
+          # Fallback: artifact couldn't be loaded, mark as empty.
+          artifact = types.Part(
+              inline_data=types.Blob(
+                  mime_type='application/octet-stream', data=b''
+              )
+          )
       await self.artifact_service.save_artifact(
           app_name=self.app_name,
           user_id=session.user_id,


src/google/adk/runners.py
```diff
@@ -754,15 +754,30 @@ async def _compute_artifact_delta_for_rewind(
...
             version=vt,
         )
-        artifact = types.Part(file_data=types.FileData(file_uri=artifact_uri))
+        if loaded and loaded.inline_data:
+          artifact = types.Part(
+              inline_data=types.Blob(
+                  mime_type=loaded.inline_data.mime_type,
+                  data=loaded.inline_data.data,
+              )
+          )
```
lucasbarzotto-axonify: (minor) `load_artifact` already returns a `types.Part` with `inline_data` populated, so the loaded part could be passed directly to `save_artifact` without re-wrapping into a new `types.Part`

src/google/adk/runners.py
```diff
@@ -754,15 +754,30 @@ async def _compute_artifact_delta_for_rewind(
...
+        else:
+          # Fallback: artifact couldn't be loaded, mark as empty.
+          artifact = types.Part(
+              inline_data=types.Blob(
+                  mime_type='application/octet-stream', data=b''
+              )
+          )
```
lucasbarzotto-axonify, line 781: Minor suggestion: it might be worth logging a warning when the fallback is hit. A silent empty-blob replacement could be hard to debug if artifact storage is inconsistent

src/google/adk/runners.py
yuting0624: Good catch! Updated to pass the loaded Part directly to save_artifact without re-wrapping. Much cleaner, thanks!

src/google/adk/runners.py
yuting0624: Agreed, added a logger.warning with the filename and version so it is easy to trace. Both suggestions addressed in the latest push.

GWeale:
Thanks for diagnosing this and for the clean fix! The same change landed on main in commit c3d50db ("Use artifact_service.load_artifact during rewind", released in v1.32.0), which closed #4932 — rewind now loads artifact bytes via artifact_service.load_artifact instead of building a file_data part, exactly as you proposed. Closing as already resolved.

Issue #1273, opened 2025-06-10
Support tool-calling for LLMs with coding skills


nileshtrivedi:
Even if the LLM (for eg: Gemma3 hosted by Gemini) does not support standard tool-calling spec, it can be accomplished by sending the function signatures in the system prompt and post-processing the LLM response. LiteLLM does this with a flag (`add_function_to_prompt`) and has [documented it here](https://docs.litellm.ai/docs/completion/function_call#function-calling-for-models-wout-function-calling-support).

This would be great to have in ADK as it will become possible to use models like Gemma3-27b for Agentic tasks.

selcukgun:
Thank you for sharing this!

kwargs for our LiteLlm constructor are passed as-is to litellm generate method: https://github.com/google/adk-python/blob/0c4054200fd50041f0dce4b1c8e56292b99a8ea8/src/google/adk/models/lite_llm.py#L630

The following should work without any change on the ADK side.
```python
litellm.add_function_to_prompt = True 
... 
root_agent = Agent(
    model=LiteLlm(model='YOUR_MODEL', functions=functions),
...
```

please give it a try.

nileshtrivedi:
It's [currently broken in LiteLLM](https://github.com/BerriAI/litellm/issues/10313) which is why I submitted this feature request for ADK to support this directly for Gemma etc without LiteLLM.

genquan9:
After some experiments, it seems that the support of add_function_to_prompt not working well yet.

btw, there is an example to preprocess prompt and post process function calls in https://github.com/GoogleCloudPlatform/vertex-ai-samples/blob/main/notebooks/community/model_garden/model_garden_integration_with_agent.ipynb.

klateefa:
Hi @nileshtrivedi  and thanks to everyone who participated in this thread.

We're currently doing some housekeeping on our issue tracker. Given the rapid pace of development and the number of updates since this was last active, there's a good chance this issue has been resolved in a newer version. To help us keep our backlog focused on current issues, we are closing this as stale.

If you're still facing this problem, could you please update to the very latest version of the library and confirm the issue still exists? If it does, please comment here, and we will be happy to reopen this issue and investigate further.

Thanks for your contribution and understanding!

PR #4985, opened 2026-03-24
fix: remove global add_function_to_prompt — breaks native tool calling (Groq, OpenAI)

vitas:
## Problem

`_ensure_litellm_imported()` sets `litellm.add_function_to_prompt = True` globally at import time (line 188). This forces ALL models through LiteLLM's text-based tool calling path — tool definitions are injected into the system prompt as text instead of being passed as the `tools` parameter.

Models that support native function calling (Groq, OpenAI, Anthropic) then output XML-style function tags instead of proper `tool_calls` JSON:

```
<function=run_command {"command": "kubectl get pods -n demo"} </function>
```

Groq rejects with:
```json
{"error":{"message":"Failed to call a function. See 'failed_generation'","code":"tool_use_failed"}}
```

## Proof

Direct `litellm.completion()` inside the same environment **without** the global flag returns proper `tool_calls` JSON with `finish_reason: "tool_calls"`:

```python
# Works (direct LiteLLM, no global flag):
litellm.completion(model="groq/llama-3.3-70b-versatile", tools=[...])
# → finish_reason: "tool_calls", proper JSON ✓

# Fails (via ADK, which sets the global flag):
Agent(model=LiteLlm(model="groq/llama-3.3-70b-versatile"), tools=[...])
# → XML tags, Groq 400 ✗
```

## Fix

Remove the global `litellm.add_function_to_prompt = True`. Models that need text-based tool calling (e.g., some Ollama models without native support) can opt in per-instance:

```python
LiteLlm(model="ollama/qwen2", add_function_to_prompt=True)
```

This kwarg flows through `_additional_args` into `acompletion()`, so per-model opt-in already works.

## Impact

- **Fixes:** All models with native tool calling (Groq, OpenAI, Anthropic) through `LiteLlm`
- **No regression for models without native tool calling:** They can pass `add_function_to_prompt=True` explicitly
- **One line removed**, comment added explaining why

## References

- huggingface/smolagents#1119 — same XML tag issue across frameworks
- BerriAI/litellm#11001 — Groq tool calling format issues
- Introduced in f33e0903 for #1273 (support models without native tool calling)


rohityan:
Hi @vitas , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

rohityan:
Hi @vitas , closing this since this is no longer needed as it has been already addressed.

PR #4807, opened 2026-03-13
fix: serialize non-standard FunctionResponse dicts in AnthropicLlm

giulio-leone:
## Summary

Fixes #4779

`part_to_message_block()` only handled `function_response` dicts containing `"content"` or `"result"` keys. Any other dict structure (e.g. SkillToolset returning `{"skill_name": ..., "instructions": ...}`) fell through to an empty string, causing Claude to never see the tool output.

## Changes

- **`anthropic_llm.py`**: Added `json.dumps()` fallback after the existing `"content"` / `"result"` checks, so non-standard response dicts are serialized to JSON text instead of being silently dropped.
- **`test_anthropic_llm.py`**: Added regression test verifying that a SkillToolset-style response dict is round-tripped correctly through `part_to_message_block()`.

## Testing

- All 18 `test_anthropic_llm.py` tests pass (including the new one)
- Full unit test suite: 4623 passed, 0 regressions

PR #4997, opened 2026-03-25
fix: serialize non-standard FunctionResponse dicts in AnthropicLlm

### LINKED ISSUE #4779: AnthropicLlm.part_to_message_block() drops non-standard FunctionResponse dicts (breaks SkillToolset with Claude)`
**Describe the Bug:**
`AnthropicLlm.part_to_message_block()` only handles `FunctionResponse.response` dicts that contain a `"content"` or `"result"` key. When `SkillToolset.load_skill()` returns a response with keys `skill_name`, `instructions`, `frontmatter`, the method falls through and returns an empty string. This means Claude models never see the skill instructions after calling `load_skill`.

**Steps to Reproduce:**
1. Create an agent with `AnthropicLlm` (any Claude model)
2. Add a `SkillToolset` with at least one skill
3. Ask the agent to use the skill
4. The agent calls `load_skill` but receives an empty tool response — it cannot follow the skill instructions

**Expected Behavior:**
The tool response should contain the full skill instructions returned by `load_skill`.

**Observed Behavior:**
The tool response is empty. The agent sees no skill instructions.

### Root cause

In `anthropic_llm.py`, `part_to_message_block()` checks for `response.get("content")` and `response.get("result")` but doesn't handle other dict structures. The `load_skill` tool returns `{"skill_name": ..., "instructions": ..., "frontmatter": ...}` which matches neither key.

**Environment Details:**
- google-adk: main branch (post-1.26.0)
- Python: 3.13

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: claude sonnet 4.6 (Other models as well)

**Additional Context:**

### Suggested fix

Add a fallback that serializes arbitrary response dicts:

```python
# In part_to_message_block(), after existing content/result checks:
else:
    content = json.dumps(response_data)
```

This ensures any FunctionResponse dict is passed to the model, regardless of key names.

 - Always (100%)

giulio-leone:
## Summary

Fixes #4779

`part_to_message_block()` only handled `function_response` dicts containing `"content"` or `"result"` keys. Any other dict structure (e.g. SkillToolset returning `{"skill_name": ..., "instructions": ...}`) fell through to an empty string, causing Claude to never see the tool output.

## Changes

- **`anthropic_llm.py`**: Added `json.dumps()` fallback after the existing `"content"` / `"result"` checks, so non-standard response dicts are serialized to JSON text instead of being silently dropped.
- **`test_anthropic_llm.py`**: Added regression test verifying that a SkillToolset-style response dict is round-tripped correctly through `part_to_message_block()`.

## Testing

- All 18 `test_anthropic_llm.py` tests pass (including the new one)
- Full unit test suite: 4623 passed, 0 regressions

> ⚠️ This reopens #4807 which was accidentally closed due to fork deletion.

### DIFF
--- a/tests/unittests/models/test_anthropic_llm.py
+++ b/tests/unittests/models/test_anthropic_llm.py
@@ -764,6 +764,17 @@ def test_part_to_message_block_with_non_standard_response():
   )
   part.function_response.id = "test_skill_id"
 
+  result = part_to_message_block(part)
+
+  assert isinstance(result, dict)
+  assert result["tool_use_id"] == "test_skill_id"
+  assert result["type"] == "tool_result"
+  assert not result["is_error"]
+  parsed = json.loads(result["content"])
+  assert parsed["skill_name"] == "search_docs"
+  assert parsed["instructions"] == "Use the search API to find documents."
+  assert parsed["frontmatter"] == {"version": "1.0"}
+
 
 def test_part_to_message_block_with_pdf_document():
   """Test that part_to_message_block handles PDF document parts."""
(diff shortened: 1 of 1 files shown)


rohityan:
Hi @giulio-leone, Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Closing this since the fix has already been merged in https://github.com/google/adk-python/commit/c26d35916e1b6cd12a412516381c6fcbf867bcee

PR #4812, opened 2026-03-13
fix: populate required fields in FunctionDeclaration json_schema fallback

giulio-leone:
## Summary

Fixes #4798 — `required` fields lost in `FunctionDeclaration` when the `parameters_json_schema` fallback path is used.

## Root Cause

`from_function_with_options()` has two code paths for building function declarations:

1. **Primary path** (`parameters_properties`): calls `_get_required_fields()` ✅  
2. **Fallback path** (`parameters_json_schema`): triggered when `_parse_schema_from_parameter` raises `ValueError` for complex union types (e.g. `list[str] | None`) — **never called `_get_required_fields()`** ❌

Additionally, the fallback path didn't propagate `default` values from `inspect.Parameter` to the generated `Schema`, so `_get_required_fields()` (which checks `schema.default is None`) would treat defaulted parameters as required.

## Impact

The LLM sees all parameters as optional and may omit required ones. Observed in production: Gemini Flash omitted the mandatory `query` parameter when calling a tool, because the schema had no `required` field.

## Fix

Three changes to the `elif parameters_json_schema` branch in `from_function_with_options()`:

| Change | Purpose |
|---|---|
| Added `_get_required_fields()` call | Populates `required` field (mirrors primary path) |
| Propagate non-None defaults to `schema.default` | Ensures `_get_required_fields()` correctly excludes defaulted params |
| Set `schema.nullable=True` for `None`-default params | Ensures `_get_required_fields()` correctly excludes nullable params |

## Test

Added regression test `test_required_fields_set_in_json_schema_fallback` that verifies:
- `query` (no default) → required ✅
- `mode` (default='default') → not required ✅  
- `tags: list[str] | None = None` → not required ✅

Full test suite: **4726 passed, 0 failures**

PR #5000, opened 2026-03-25
fix: populate required fields in FunctionDeclaration json_schema fallback

### LINKED ISSUE #4798: Bug: required fields lost in FunctionDeclaration when parameters_json_schema fallback is used

giulio-leone:
## Summary

Fixes #4798 — `required` fields lost in `FunctionDeclaration` when the `parameters_json_schema` fallback path is used.

## Root Cause

`from_function_with_options()` has two code paths for building function declarations:

1. **Primary path** (`parameters_properties`): calls `_get_required_fields()` ✅  
2. **Fallback path** (`parameters_json_schema`): triggered when `_parse_schema_from_parameter` raises `ValueError` for complex union types (e.g. `list[str] | None`) — **never called `_get_required_fields()`** ❌

Additionally, the fallback path didn't propagate `default` values from `inspect.Parameter` to the generated `Schema`, so `_get_required_fields()` (which checks `schema.default is None`) would treat defaulted parameters as required.

## Impact

The LLM sees all parameters as optional and may omit required ones. Observed in production: Gemini Flash omitted the mandatory `query` parameter when calling a tool, because the schema had no `required` field.

## Fix

Three changes to the `elif parameters_json_schema` branch in `from_function_with_options()`:

| Change | Purpose |
|---|---|
| Added `_get_required_fields()` call | Populates `required` field (mirrors primary path) |
| Propagate non-None defaults to `schema.default` | Ensures `_get_required_fields()` correctly excludes defaulted params |
| Set `schema.nullable=True` for `None`-default params | Ensures `_get_required_fields()` correctly excludes nullable params |

## Test

Added regression test `test_required_fields_set_in_json_schema_fallback` that verifies:
- `query` (no default) → required ✅
- `mode` (default='default') → not required ✅  
- `tags: list[str] | None = None` → not required ✅

Full test suite: **4726 passed, 0 failures**

> ⚠️ This reopens #4812 which was accidentally closed due to fork deletion.

### DIFF
--- a/tests/unittests/tools/test_from_function_with_options.py
+++ b/tests/unittests/tools/test_from_function_with_options.py
@@ -334,7 +334,7 @@ def test_required_fields_set_in_json_schema_fallback():
 
   def complex_tool(
       query: str,
-      mode: str = "default",
+      mode: str = 'default',
       tags: list[str] | None = None,
   ) -> str:
     """A tool where one param has a complex union type."""


rohityan:
Hi @giulio-leone , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

giulio-leone:
Thanks @rohityan. Quick maintainer note on the current head: the effective PR checks are now `14 successful, 0 failing, 1 cancelled` (`gh pr checks` on the current branch state).

The cancelled entry is from the superseded mypy matrix run; the latest `mypy-diff` reruns are green on the active head, and the full unit-test matrix is green as well.

Happy to adjust anything else if helpful during review.

xuanyang15:
Started internal review process

Issue #3525, opened 2025-11-12
Empty model message after tool call


CommonGibbon:
**Describe the bug**
After calling a tool, the final model message in a session will randomly have no content parts. 
For example:
Event(model_version='gemini-2.5-flash-lite', content=Content(
  role='model'
)

**To Reproduce**
Please share a minimal code and data to reproduce your problem.
1. Install google-adk
2. Run:
```
###############################################
# SETUP CODE
from typing import Any, Dict

from google.adk.agents import LlmAgent
from google.adk.models.google_llm import Gemini
from google.adk.sessions import InMemorySessionService
from google.adk.runners import Runner
from google.adk.tools.tool_context import ToolContext
from google.genai import types


retry_config = types.HttpRetryOptions(
    attempts=5,  # Maximum retry attempts
    exp_base=7,  # Delay multiplier
    initial_delay=1,
    http_status_codes=[429, 500, 503, 504],  # Retry on these HTTP errors
)

USER_NAME_SCOPE_LEVELS = ("temp", "user", "app")

# This demonstrates how tools can read from session state.
def retrieve_userinfo(tool_context: ToolContext) -> Dict[str, Any]:
    """
    Tool to retrieve user name and country from session state.
    """
    # Read from session state
    user_name = tool_context.state.get("user:name", "Username not found")
    country = tool_context.state.get("user:country", "Country not found")

    return {"status": "success", "user_name": user_name, "country": country}


    # Configuration
APP_NAME = "default"
USER_ID = "default"

# Create an agent with session state tools
root_agent = LlmAgent(
    model=Gemini(model="gemini-2.5-flash-lite", retry_options=retry_config),
    name="text_chat_bot",
    description="""A text chatbot.
    Tools for managing user context:
    * To fetch username and country when required use `retrieve_userinfo` tool.
    """,
    tools=[retrieve_userinfo],  # Provide the tools to the agent
)

# Set up session service and runner
session_service = InMemorySessionService()
runner = Runner(agent=root_agent, session_service=session_service, app_name="default")

# Inject user info into session state
session = await runner.session_service.create_session(
    app_name=APP_NAME,
    user_id=USER_ID,
    state = {"user:name": "Will", "user:country":"USA"}
)
idx=0 # we'll use this below
###############################################
# RUN CODE 
# You may need to run the following code multiple times in order to reproduce the issue. 
# I've built in session cycling so you don't accidentally contaminate new attempts
[...]

PR #5006, opened 2026-03-26
fix: retry with resume message when model returns empty response

gurjot-05:
## Bug

Some models (notably Gemini 2.5 Pro/Flash) intermittently return empty content (`parts: []`, `candidatesTokenCount: 0`, `finishReason: STOP`) after processing tool results. This is especially common under concurrent load and with streaming + thinking enabled.

ADK's `is_final_response()` treats this as a valid completed turn because it only checks for the **absence** of function calls — not the **presence** of actual content. The agent loop stops and the user sees nothing.

**Observed with:**
- Gemini 2.5 Pro via `vertex_ai` — after tool execution in orchestrator agents
- Intermittent: ~15-29% failure rate under parallel load (7+ concurrent sessions)
- Affects both streaming (SSE + thoughts) and non-streaming modes

**Example session showing the bug:**
```
Event: agent calls nth_research → completes successfully
Event: agent calls review_research → completes successfully (status: "approved")
Event: agent should call draft_response next, but returns parts: [] ← EMPTY
Agent loop ends. User sees nothing.
```

Related: #3525

## Root Cause

There are **three** distinct failure modes, all leading to the same silent halt:

### Mode 1: Non-streaming empty response
In `BaseLlmFlow.run_async()`:
```python
if not last_event or last_event.is_final_response() or last_event.partial:
    break
```
An event with `parts: []` passes `is_final_response()` — no function calls, no function responses, not partial — so the loop breaks silently.

### Mode 2: Streaming + thinking, no final response yielded
With streaming + thinking enabled, the LiteLLM adapter yields thought chunks as `partial=True`. When `finish_reason=stop` arrives with no text content and no accumulated text/reasoning:
- `lite_llm.py` finalization requires `(text or reasoning_parts)` which is False
- **No final aggregated response is ever yielded** from the streaming generator
- `last_event` is either `None` or `partial=True` from a thought chunk
- The loop breaks on `not last_event` or `last_event.partial` — **before** any retry logic

### Mode 3: Model returns content=None
When the model returns `content=None`, `_postprocess_async` filters it out (returns without yielding if no content and no error). `last_event` stays `None`, and the loop breaks.

## Fix

Two-layer fix addressing the root cause and adding defense-in-depth:

### Layer 1: lite_llm.py — Ensure streaming always yields a final response

Added a fallback after the streaming loop: if the model produced no meaningful output at all (no text, no reasoning, no tool calls), yield an explicit empty non-partial `LlmResponse` so downstream retry logic can detect and handle it.

This converts Mode 2 into Mode 1, making it catchable by the retry logic.

### Layer 2: base_llm_flow.py — Unified retry for all empty response modes

Restructured `run_async()` to check for empty responses **before** normal termination conditions. The retry now handles all three cases:

1. **`not last_event`** — no events yielded at all (Mode 3)
2. **`last_event.partial` with no meaningful content** — streaming thought-only chunks, no final response (Mode 2, defense-in-depth)
3. **`last_event.is_final_response()` with no meaningful content** — non-streaming empty response (Mode 1)

When an empty response is detected:
- A **resume message** is injected: *"Your previous response was empty. Please resume execution from where you left off."*
- This gives the model context about why it's being re-prompted, improving recovery rate vs. silent retry
- `_MAX_EMPTY_RESPONSE_RETRIES = 2` prevents infinite loops

### False positive prevention

- **Partial events with real content** (normal streaming): `_has_meaningful_content()` returns True — no retry
- **Agent transfer events**: Author check (`last_event.author == agent.name`) prevents retrying events from other agents
- **Thought-only parts**: Treated as non-meaningful (they have no user-visible content)

### Comprehensive scenario tests (test_empty_response_all_scenarios.py): 12 passed

| # | Scenario | Expected | Result |
|---|----------|----------|--------|
| 1 | Non-streaming empty parts:[] then recovery | 1 retry, good response | PASS |
| 2 | Thought-only final response | 1 retry | PASS |
| 3a | No events at all (last_event=None) | Retries up to max | PASS |
| 3b | Partial + empty content | 1 retry, then recovery | PASS |
| 3b' | Partial + thought-only content | 1 retry, then recovery | PASS |
| 4 | Partial WITH real text content | No retry (normal break) | PASS |
| 5 | Empty exhausts max retries | Exactly 2 retries, then stop | PASS |
| 6 | Empty, Empty, Good (multi-retry recovery) | 2 retries, then good | PASS |
| 7 | LiteLLM fallback response is non-partial | is_final_response()=True | PASS |
| 8 | Whitespace-only text | 1 retry | PASS |
| 9 | Function call is meaningful | Not retried | PASS |
| 10 | Partial empty then partial with content | 1 retry, then normal break | PASS |

## Test plan

- [x] `pytest tests/unittests/flows/llm_flows/` — 384 passed, 0 failed
- [x] All 12 original PR tests pass
- [x] All 12 comprehensive scenario tests pass
- [x] Agent transfer tests not affected (author check prevents false positives)
- [x] Verified with production logs from real orchestrator agent running Gemini 2.5 Pro — both non-streaming and streaming+thinking failure modes are covered


rohityan:
Hi @gurjot-05 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

gurjot-05:
Thank you @rohityan I look forward to your feedback. Let me know if there’s anything else I can provide in the meantime.

shai-prashant:
@gurjot-05, please rebase the branch.

gurjot-05:
Any updates? @rohityan @GWeale

tomasz321 (approved):

gurjot-05:
Hi @rohityan @GWeale 
Wanted to know if there's any updates on this PR or have these issues already been addressed in the newer releases?

DeanChensj:
Hi @gurjot-05, thank you very much for putting this together, and especially for the detailed root-cause analysis breaking down the three failure modes (including the streaming + thinking finalization edge case in `lite_llm.py`).

After reviewing the implementation and testing how it interacts with `HEAD`, we’ve decided not to merge this PR in its current form and will close it for now. Here is the technical context:

1. **Why retrying in `BaseLlmFlow.run_async()` (Event layer) is problematic**:
   - `BaseLlmFlow.run_async()` yields `Event` objects directly to `Runner.run_async()`, which immediately persists every non-partial (`partial=False`) event to `session.events`.
   - In **SSE streaming mode**, yielding an empty fallback `partial=False` event causes `Runner` to append an empty `model` turn (`parts: []`) to `session.events` before the retry loop continues. On the retry attempt, sending a conversation history ending with an empty model turn (or consecutive model turns) causes downstream LLM APIs to reject the request with `400 Bad Request`.
   - Similarly, mutating `session.events` to inject a persistent user resume message pollutes the user-visible session transcript on every transient model glitch.
2. **Interaction with `MODEL_RETURNED_NO_CONTENT` at `HEAD`**:
   - `HEAD` recently introduced explicit `error_code = "MODEL_RETURNED_NO_CONTENT"` surfacing in `_model_response_finalizer.py` / `_postprocess_async()`. Buffering and suppressing candidate events inside `BaseLlmFlow.run_async()` creates complex interactions with legitimate model error events (such as `SAFETY` or `RESOURCE_EXHAUSTED`), which should never be suppressed or retried as empty turns.

### How we plan to address this instead
We agree that intermittent empty / thought-only responses after tool calls need better handling. Rather than orchestrating retries at the `Event` stream level in `BaseLlmFlow.run_async()`, the clean architectural layer for this is **inside `_call_llm_async()` / `_model_response_finalizer.py` (at the `LlmResponse` level, before the final `partial=False` `Event` is yielded and persisted by `Runner`)**:
- Since `Runner` never persists `partial=True` streaming chunks to `session.events`, detecting an empty or thought-only terminal `LlmResponse` (`finish_reason=STOP`) *before* yielding the final `partial=False` response allows retrying (with an ephemeral nudge on `llm_request.contents`) without ever polluting `session.events` in either SSE or non-SSE mode.
- We will also extend `MODEL_RETURNED_NO_CONTENT` detection to cover thought-only / whitespace-only turns and LiteLLM streaming finalization.

Thanks again for investigating this issue and highlighting the `lite_llm.py` streaming+thinking edge case — we will incorporate that insight into the model-layer fix!

PR #5007, opened 2026-03-26
fix: session resumption reconnection loop never iterates

### LINKED ISSUE #4996: Session resumption reconnection loop in run_live() never iterates
## 🔴 Required Information

**Describe the Bug:**

The session resumption reconnection loop in `base_llm_flow.py`'s `run_live()` never iterates. The `while True:` loop at line 481 and the handle-injection code at lines 484–496 are unreachable because both exception handlers (lines 595 and 600) re-raise unconditionally. Additionally, `goAway` messages from the server are silently dropped at the connection layer, preventing proactive reconnection before the server terminates the connection.

The net effect is that session resumption handles are received and stored (via `_receive_from_model` → `invocation_context.live_session_resumption_handle`) but never used. After the ~10-minute connection lifetime expires, `run_live()` terminates with an error instead of reconnecting with the saved handle.

**Steps to Reproduce:**

1. Configure a live agent with `RunConfig(session_resumption=types.SessionResumptionConfig(transparent=True))`
2. Start a `run_live()` session with a live-only model (e.g. `gemini-2.5-flash-live`)
3. Verify that `SessionResumptionUpdate` messages are received (visible at DEBUG level: "Update session resumption handle: ...")
4. Trigger any WebSocket disconnect, either:
   - Wait ~10 minutes for the server's connection lifetime limit, or
   - Simulate a network interruption
5. Observe that `run_live()` raises an error instead of reconnecting with the saved handle
6. The log message "Attempting to reconnect (Attempt N)..." (line 485) is never emitted

**Expected Behavior:**

When the WebSocket connection closes (after goAway or session timeout):
1. The `while True:` loop should catch the exception and continue
2. On the next iteration, the saved `live_session_resumption_handle` should be injected into the setup config (the code at lines 484–496 already does this)
3. A new connection should be opened with the handle, restoring the session without replaying history
4. `run_live()` should continue yielding events seamlessly

**Observed Behavior:**

The connection dies and `run_live()` raises an exception. The reconnection loop never gets a second iteration. There are two disconnect paths, both broken:

**Path A: Disconnect during receive (most common):**
The genai SDK's `_receive()` (`live.py:538`) catches `ConnectionClosed` from the websockets library and converts it to `APIError`:
```python
# google/genai/live.py:538-545
except ConnectionClosed as e:
    if e.rcvd:
        code = e.rcvd.code
        reason = e.rcvd.reason
    else:
        code = 1006
        reason = ...
    errors.APIError.raise_error(code, reason, None)
```
This `APIError` propagates through `_receive_from_model` and hits `except Exception` at line 600, which re-raises. The `except (ConnectionClosed, ConnectionClosedOK)` at line 595 never fires on this path because the exception is `APIError`, not `ConnectionClosed`.

**Path B: Disconnect during send:**
The genai SDK's send methods (`send_client_content`, `send_tool_response`, etc.) call `self._ws.send()` without catching `ConnectionClosed`. The raw `ConnectionClosed` propagates through the `send_task`'s `finally` cleanup and hits `except (ConnectionClosed, ConnectionClosedOK)` at line 595, which also re-raises.

**Separately, `goAway` is unhandled:**
`gemini_llm_connection.py`'s `receive()` method processes `message.server_content`, `message.tool_call`, `message.usage_metadata`, and `message.session_resumption_update` but has no `if message.go_away:` check. The `LiveServerGoAway` event (sent ~60s before server termination) is silently dropped, preventing proactive reconnection.

**Environment Details:**

- ADK Library Version: 1.27.2
- Desktop OS: macOS (Darwin 24.6.0)
- Python Version: 3.12

**Model Information:**

- Are you using LiteLLM: No
- Which model is being used: gemini-live-2.5-flash-native-audio

---

## 🟡 Optional Information

**Regression:**

No, the reconnection plumbing was added in PR #2270 (fixing #2179) but the exception handling was never updated to actually loop. The handle injection code at lines 484–496 has been unreachable since it was introduced.

**Logs:**

When the connection times out, the following is logged:
```text
ERROR    An unexpected error occurred in live flow: 1000 None. connection closed
```
The "Attempting to reconnect" log message at line 485 is never reached.

**Additional Context:**

The code structure suggests the intent was for clean WebSocket closes to end the `_receive_from_model` generator naturally (the comment at line 596 says "when the session timeout, it will just close and not throw exception"), allowing the `while True:` loop to iterate. However, the genai SDK's `_receive()` converts ALL `ConnectionClosed` exceptions (including `ConnectionClosedOK` / code 1000) into `APIError`, so the "clean end" path never occurs.

`_receive_from_model` does have an `except ConnectionClosedOK: pass` at line 744, which would enable the clean-end path but `APIError` is not `ConnectionClosedOK`, so it doesn't fire for connection closes that come through `_receive()`.

**Suggested fix:**

There are three changes needed:

1. **Fix the reconnection loop** In `run_live()`, catch the genai SDK's `APIError` wrapping `ConnectionClosed` and `continue` instead of `raise` when a session resumption handle is available:
```python
except (ConnectionClosed, ConnectionClosedOK) as e:
    if invocation_context.live_session_resumption_handle:
        logger.info('Connection closed (%s), reconnecting with session handle.', e)
        continue
    raise
except Exception as e:
    if (
        invocation_context.live_session_resumption_handle
        and isinstance(e, APIError)
    ):
        logger.info('Connection lost (%s), reconnecting with session handle.', e)
        continue
    raise
```

2. **Skip `send_history` on reconnection** When reconnecting with a handle, the server already has the session context. Add a guard at line 503:
```python
if llm_request.contents and not invocation_context.live_session_resumption_handle:
    await llm_connection.send_history(llm_request.contents)
```

3. **Handle `goAway`** In `gemini_llm_connection.py`'s `receive()`, surface `message.go_away` so the flow layer can proactively reconnect before the server terminates the connection.

**Related issues / PRs:**

- #2179 "Session resumption not working" (closed, partially fixed by #2270)
- PR #2270 Added handle injection plumbing (but didn't fix exception handling)
- PR #4358 Expose `live_session_resumption_update` in events (open)

- Always (100%) the reconnection loop has never successfully iterated.

brucearctor:
## Summary

Fixes #4996 — the `run_live()` reconnection loop was unreachable due to unconditional re-raising of exceptions, redundant history transmission on reconnection, and ignored `goAway` server messages.

### 1. Exception handlers continue instead of raise (`base_llm_flow.py`)
- `ConnectionClosed` / `ConnectionClosedOK` handler now `continue`s the `while True` loop when `live_session_resumption_handle` is present
- Added `APIError` handling (genai SDK wraps `ConnectionClosed` as `APIError`) with the same continue-on-handle logic

### 2. Skip `send_history` on reconnection (`base_llm_flow.py`)
- Added guard: `if llm_request.contents and not invocation_context.live_session_resumption_handle`
- Server already has the session context via the resumption handle — no need to resend

### 3. Surface `go_away` messages (`gemini_llm_connection.py` + `llm_response.py`)
- `GeminiLlmConnection.receive()` now detects `message.go_away` and yields it as `LlmResponse.go_away`
- Added `go_away: Optional[LiveServerGoAway]` field to `LlmResponse`
- Enables proactive reconnection ~60s before server terminates the connection

## Testing

| Suite | Tests | Status |
|---|---|---|
| `test_gemini_llm_connection.py` | 27 (incl. new `test_receive_go_away`) | ✅ |
| `test_run_live_reconnection.py` | 7 (new) | ✅ |
| `tests/unittests/flows/llm_flows/` (all) | 364 | ✅ |

### New reconnection tests cover:
- Loop continues on `ConnectionClosedOK` / `APIError` when resumption handle exists
- Exceptions propagate without handle (preserves old behavior)
- Non-`APIError` exceptions always propagate, even with handle
- `send_history` skipped with handle, called without

### DIFF
--- a/src/google/adk/flows/llm_flows/base_llm_flow.py
+++ b/src/google/adk/flows/llm_flows/base_llm_flow.py
@@ -610,7 +610,10 @@ class BaseLlmFlow(ABC):
                 'Max reconnection attempts (%d) reached, giving up.',
                 MAX_RECONNECT_ATTEMPTS,
             )
-            raise
+            raise ConnectionError(
+                'Live session reconnection failed after'
+                f' {MAX_RECONNECT_ATTEMPTS} attempts'
+            ) from e
           delay = min(
               RECONNECT_BASE_DELAY * (2 ** (attempt - 1)),
               RECONNECT_MAX_DELAY,
@@ -637,7 +640,10 @@ class BaseLlmFlow(ABC):
                 'Max reconnection attempts (%d) reached, giving up.',
                 MAX_RECONNECT_ATTEMPTS,
             )
-            raise
+            raise ConnectionError(
+                'Live session reconnection failed after'
+                f' {MAX_RECONNECT_ATTEMPTS} attempts'
+            ) from e
           delay = min(
               RECONNECT_BASE_DELAY * (2 ** (attempt - 1)),
               RECONNECT_MAX_DELAY,
--- a/tests/unittests/flows/llm_flows/test_run_live_reconnection.py
+++ b/tests/unittests/flows/llm_flows/test_run_live_reconnection.py
@@ -212,7 +212,7 @@ async def test_raises_after_max_retries_connection_closed():
     with mock.patch.object(
         type(agent.canonical_model), 'connect', side_effect=mock_connect
     ):
-      with pytest.raises(ConnectionClosedOK):
+      with pytest.raises(ConnectionError, match='reconnection failed after'):
         async with Aclosing(flow.run_live(ctx)) as agen:
           async for event in agen:
             pass
@@ -238,7 +238,7 @@ async def test_raises_after_max_retries_api_error():
     with mock.patch.object(
         type(agent.canonical_model), 'connect', side_effect=mock_connect
     ):
-      with pytest.raises(genai_errors.APIError):
+      with pytest.raises(ConnectionError, match='reconnection failed after'):
         async with Aclosing(flow.run_live(ctx)) as agen:
           async for event in agen:
             pass


src/google/adk/flows/llm_flows/base_llm_flow.py
```diff
@@ -593,11 +596,22 @@ async def run_live(
             except asyncio.CancelledError:
               pass
       except (ConnectionClosed, ConnectionClosedOK) as e:
-        # when the session timeout, it will just close and not throw exception.
-        # so this is for bad cases
+        if invocation_context.live_session_resumption_handle:
+          logger.info(
+              'Connection closed (%s), reconnecting with session handle.', e
+          )
+          continue
```
rohityan, line 630: This is great for brief network glitches. How do you think this would behave if the server was down for a few minutes?

src/google/adk/flows/llm_flows/base_llm_flow.py
```diff
@@ -500,8 +501,10 @@ async def run_live(
             invocation_context.agent.name,
         )
         async with llm.connect(llm_request) as llm_connection:
-          if llm_request.contents:
+          if llm_request.contents and not invocation_context.live_session_resumption_handle:
```
rohityan, line 510: What happens if the resumption handle is rejected by the server?

src/google/adk/flows/llm_flows/base_llm_flow.py
brucearctor: Good question! Right now the loop relies on the inherent connection timeout from llm.connect() to throttle reconnection attempts. For brief glitches that's sufficient, but you're right that a multi-minute outage would result in aggressive retries.

Makes me think:
* Add exponential backoff with jitter (e.g., 1s → 2s → 4s → ... capped at ~30s) between reconnection attempts
AND
* Add a max retry count and raise after N failures


I'll add that to this PR.  Or, would you suggest other?

src/google/adk/flows/llm_flows/base_llm_flow.py
brucearctor: Great point. Looking at the Gemini Live API behavior: if the handle is rejected, the server sends back a `session_resumption_update` with `resumable=False` (or simply doesn't echo back a new_handle). It does not silently drop the context.

However, to be safe, we could:

Clear the handle on rejection: When we receive a `session_resumption_update` where resumable is False, clear `live_session_resumption_handle` so the next reconnection falls back to sending full history
Add a fallback: If the connection succeeds but no `session_resumption_update` arrives within a timeout, assume the handle was rejected and resend history
I think option (1) is already partially covered by the existing `_receive_from_model` logic that updates the handle from `session_resumption_update`. 

Want me to verify the server behavior and add an explicit guard for the rejection case?

wuliang229:
@brucearctor thanks for the PR. We decided to go with https://github.com/google/adk-python/commit/6b1600fbf53bcf634c5fe4793f02921bc0b75125.

PR #5020, opened 2026-03-27
fix(sessions): truncate error_message in v0 StorageEvent to guard against VARCHAR(255) overflow

### LINKED ISSUE #4993: [Bug] DatabaseSessionService (v0 schema): StringDataRightTruncationError on PostgreSQL when error_message exceeds VARCHAR(255)
## What happened

Running `DatabaseSessionService` with PostgreSQL on a v0 schema database. When Gemini returns a `MALFORMED_FUNCTION_CALL` with a large payload (tool call containing a JSON object), `append_event()` throws:

```
asyncpg.exceptions.StringDataRightTruncationError: value too long for type character varying(255)
```

The SSE stream terminates, the user gets no response, and the session state from that turn is lost entirely. No warning, no graceful fallback.

## Environment

- google-adk: **1.26.0**
- Database: **PostgreSQL** via asyncpg
- Model: gemini-2.5-flash (Vertex AI)
- DB schema: **v0** (has `actions` column, no `event_data`)

## Stack trace

```
adk_web_server.py:1616  event_generator
  runners.py:833        _exec_with_plugin
    → session_service.append_event()
      database_session_service.py:627  sql_session.add(StorageEvent.from_event(...))
      database_session_service.py:629  await sql_session.commit()  ← crash here

sqlalchemy.exc.DBAPIError: value too long for type character varying(255)
[INSERT INTO events (..., error_code, error_message, ...)]
[error_code='MALFORMED_FUNCTION_CALL',
 error_message="Malformed function call: print(default_api.update_state(
   key='event:context', value={'menu_type': 'tea_break', 'guest_count': 50,
   ...large dict...}))"]
```

## Why this happens

`schemas/v0.py` currently defines `error_message` as `Text` (unlimited), which is correct. But ADK never ALTERs existing columns `create_all()` is additive only. So any database that was first created when `error_message` was `VARCHAR(255)` keeps that old constraint forever, even after upgrading the package.

There's also no truncation anywhere in the write path. `StorageEvent.from_event()` copies the value directly:

```python
# v0.py line 282–283
error_code=event.error_code,
error_message=event.error_message,  # raw, no length guard
```

`MALFORMED_FUNCTION_CALL` messages are unbounded in practice when the model echoes back a malformed tool call with a large JSON body, the message easily exceeds 255 chars.

Note: v1 schema doesn't have this issue because everything goes into a single `event_data JSONB` column.

## Suggested fix

**1. Add a length guard in `StorageEvent.from_event()` (v0.py)**

```python
error_message=(
    event.error_message[:65000] + "...[truncated]"
    if event.error_message and len(event.error_message) > 65000
    else event.error_message
),
```

**2. Catch and recover in `append_event()` instead of crashing**

If the INSERT fails due to a column size violation, retry with a truncated message so at least the session state survives:

```python
try:
    sql_session.add(schema.StorageEvent.from_event(session, event))
    await sql_session.commit()
except DBAPIError as e:
    if "value too long" in str(e) or "StringDataRightTruncationError" in str(e):
        await sql_session.rollback()
        truncated = event.model_copy(
            update={"error_message": (event.error_message or "")[:200] + "...[truncated]"}
        )
        sql_session.add(schema.StorageEvent.from_event(session, truncated))
        await sql_session.commit()
        logger.warning(
            "error_message truncated to fit DB column constraint. "
            "Run `ALTER TABLE events ALTER COLUMN error_message TYPE TEXT` "
            "or migrate to v1 schema to resolve permanently."
        )
    else:
        raise
```

**3. Mention this in the upgrade/migration docs**

Users upgrading ADK on an existing PostgreSQL database should know to run:

```sql
ALTER TABLE events ALTER COLUMN error_message TYPE TEXT;
```

Or use `adk migrate session` to move to v1 fully.

## Workaround (for anyone hitting this now)

```sql
ALTER TABLE events ALTER COLUMN error_message TYPE TEXT;
```

Or migrate to v1 schema using `adk migrate session`.

STHITAPRAJNAS:
Fixes #4993.

## Problem

`DatabaseSessionService` crashes on PostgreSQL databases that were created with an older ADK version where `error_message` was a `VARCHAR(255)` column. `create_all()` is additive-only — it never `ALTER`s existing columns — so those databases keep the narrow column indefinitely even after upgrading the package.

When a model returns a `MALFORMED_FUNCTION_CALL` error whose payload echoes back a large JSON body, the unbounded `error_message` string exceeds 255 characters and hits:

```
asyncpg.exceptions.StringDataRightTruncationError: value too long for type character varying(255)
```

This terminates the SSE stream entirely with no fallback and no warning to the caller.

## Fix

Add a write-path guard in `StorageEvent.from_event()` (v0 schema only). If `error_message` exceeds the legacy column width it is clipped to 241 characters and suffixed with `"...[truncated]"`, keeping the total at exactly 255. Messages within the limit are passed through unchanged.

The guard lives in the schema layer rather than the service layer so it is applied regardless of the call site, and it is scoped to v0 only because v1 stores everything in a `JSONB event_data` column and is not affected.

Two constants are introduced at module level with a comment explaining the invariant, so the magic number is traceable without digging into the issue history.

## Permanent fix

Users on PostgreSQL with a legacy schema can resolve this at the DB level with:

```sql
ALTER TABLE events ALTER COLUMN error_message TYPE TEXT;
```

Or migrate to v1 fully via `adk migrate session`.

## Testing

Added two unit tests to `test_v0_storage_event.py`:
- `test_from_event_truncates_error_message_exceeding_varchar255` — a 300-char message is clamped to ≤ 255 chars and ends with the truncation suffix
- `test_from_event_preserves_short_error_message` — a short message is stored verbatim

### DIFF
--- a/tests/unittests/sessions/test_v0_storage_event.py
+++ b/tests/unittests/sessions/test_v0_storage_event.py
@@ -125,3 +125,24 @@ def test_from_event_preserves_short_error_message():
   storage_event = StorageEvent.from_event(session, event)
 
   assert storage_event.error_message == short_error
+
+
+def test_from_event_with_none_error_message():
+  session = Session(
+      app_name="app",
+      user_id="user",
+      id="session_id",
+      state={},
+      events=[],
+      last_update_time=0.0,
+  )
+  event = Event(
+      id="event_id",
+      invocation_id="inv_id",
+      author="agent",
+      timestamp=1.0,
+  )
+
+  storage_event = StorageEvent.from_event(session, event)
+
+  assert storage_event.error_message is None


JiwaniZakir (commented):
The test `test_from_event_truncates_error_message_exceeding_varchar255` asserts `len(...) <= _LEGACY_ERROR_MSG_MAX_LEN`, but given the deterministic truncation logic in `from_event`, this should be a strict equality check (`== _LEGACY_ERROR_MSG_MAX_LEN`) — the weaker assertion would pass even if the truncation were accidentally off by several characters. There's also no test case for `error_message=None`, which exercises the falsy branch of the condition in `from_event`; while the logic looks correct, an explicit test would document the intent and prevent regressions. Additionally, the test file imports the private symbols `_LEGACY_ERROR_MSG_MAX_LEN` and `_TRUNCATION_SUFFIX` directly — if these ever need to be refactored or inlined, the test becomes a coupling point; consider asserting the concrete expected string (`"x" * 241 + "...[truncated]"`) or at least the exact final length (`255`) to reduce that dependency.

STHITAPRAJNAS:
Thanks for the thorough review @JiwaniZakir — all three points are valid, addressed in 9353950:

- **Strict equality**: changed `<= _LEGACY_ERROR_MSG_MAX_LEN` to `== "x" * 241 + "...[truncated]"` — the full concrete expected string, which also removes the need for the private symbol import
- **Private symbol coupling**: dropped the `_LEGACY_ERROR_MSG_MAX_LEN` and `_TRUNCATION_SUFFIX` imports; the test now asserts the exact output string directly
- **`None` branch**: added `test_from_event_with_none_error_message` to document and protect the falsy path through the guard

rohityan:
Hi @STHITAPRAJNAS, Thank you for your contribution! We appreciate you taking the time to submit this pull request.

src/google/adk/sessions/schemas/v0.py
rohityan, line 314: Before the truncation happens, adding a logger.warning ensures the developer can still find the full, original payload in the server logs for debugging.

src/google/adk/sessions/schemas/v0.py
```diff
@@ -298,7 +305,15 @@ def from_event(cls, session: Session, event: Event) -> StorageEvent:
         partial=event.partial,
         turn_complete=event.turn_complete,
         error_code=event.error_code,
```
rohityan, line 327: What happens if the model returns a long text in the `error_code`? Shouldn't it have the same truncation logic?

JiwaniZakir:
The truncation approach is reasonable as a defensive guard, but since this is superseded by #5062, it's worth checking whether that PR takes a different approach (e.g., migrating the column to `TEXT`) — if so, this guard may still be useful as a belt-and-suspenders measure for deployments that can't run migrations immediately. The 241 + `"...[truncated]"` = 255 boundary logic looks correct.

JiwaniZakir:
Good points both. Adding a `logger.warning` with the full original `error_message` before truncation makes sense — that so it's surfaced without requiring a debugger. On `error_code`: the existing schema defines it as `VARCHAR(50)`, but in practice the values are short enum-style strings (e.g. `"MALFORMED_FUNCTION_CALL"`), so overflow there is unlikely; that said, applying the same defensive truncation for consistency is reasonable and I can add it in the same pass.

STHITAPRAJNAS:
Thanks @JiwaniZakir for the follow-up.

**On PR #5062:** I checked — #5062 is `feat(to_a2a): add service_url support to to_a2a`, which is unrelated to the database schema. There is no competing fix for this issue, so this PR stands on its own.

**On `logger.warning`:** Good call. I have extracted the truncation into a pre-flight block before the constructor and added a `WARNING` log that includes the event id, the length limit, and the full original `error_message`. This way the content is always recoverable from logs without needing a debugger. Updated in the latest push (8d65e92). The test now also asserts the warning is emitted and contains the full original text.

STHITAPRAJNAS:
Hi @rohityan — all review feedback from @JiwaniZakir has been addressed (Mar 31 push). Checks are passing. Could you take a final look when you get a chance? Thanks!

STHITAPRAJNAS:
Hi @rohityan — all the points @JiwaniZakir raised in the Mar 27 review have been addressed (strict equality assertion, `logger.warning` with the full message before truncation, and the `None` edge case test). The fix itself is a one-liner guard against `VARCHAR(255)` overflow on older databases. Would appreciate a final look when you get a chance.

haranrk:
@STHITAPRAJNAS Thanks for your contribution! The issue this PR addresses has already been fixed in https://github.com/google/adk-python/commit/feea8cac433b57b6e71a1745c0af487d00be84d6, so I'll close this issue.

PR #5023, opened 2026-03-27
fix: raise ValueError for unsupported MIME types in file_data URI path

### LINKED ISSUE #5022: LiteLLM: file_uri with unknown or generic MIME type causes cryptic downstream ValueError
## Summary

When a Part with file_data.file_uri has no determinable MIME type (no extension, no display_name, no explicit mime_type field), the library falls back to application/octet-stream. This produces a cryptic ValueError deep inside LiteLLM rather than a clear error at the ADK layer.

The same problem occurs when the caller explicitly sets mime_type to application/octet-stream. Both paths reach the same failure point in LiteLLM with no useful message.

## Root Cause

The constant _DEFAULT_MIME_TYPE = "application/octet-stream" was added to satisfy LiteLLM's Vertex AI backend requirement for a non-empty format field in file objects. However application/octet-stream is not in _SUPPORTED_FILE_CONTENT_MIME_TYPES and is not a type any backend can meaningfully process.

This creates an inconsistency between the two content paths:

The inline_data path raises ValueError immediately for unsupported MIME types (fail-fast behavior).
The file_data path silently falls back to application/octet-stream and fails later with a cryptic error.

GcsArtifactService generates URIs like gs://bucket/artifact/0 with no extension and no MIME type set, making ADK's own artifact system the primary trigger for this fallback. The library raises an error caused by a value it generated itself.

## Steps to Reproduce

```python

part = types.Part(
    file_data=types.FileData(
    )
)
# Passing this part to a LiteLlm model call results in a cryptic
# ValueError from LiteLLM internals with no guidance on how to fix it.
```

## Expected Behavior

A clear ValueError at the ADK layer with an actionable message telling the user to set file_data.mime_type explicitly on the Part.

## Actual Behavior

A cryptic ValueError originating from inside LiteLLM with no reference to mime_type or how to resolve it.

## Additional Context

The same failure occurs when mime_type is explicitly set to application/octet-stream by the caller. The type is semantically equivalent to an unknown type and no backend in this pipeline can use it. For OpenRouter specifically, LiteLLM strips the format field entirely before the request goes out so the value never reaches the provider regardless.

The fix aligns the file_data path with the existing fail-fast behavior of the inline_data path.

Raman369AI:
Fixes #5022

## Problem

When a `Part` with `file_data.file_uri` has no determinable MIME type, the library fell back to `application/octet-stream` via `_DEFAULT_MIME_TYPE`. This value then propagated to LiteLLM which raised a cryptic internal `ValueError` with no guidance for the user.

The same failure occurred when the caller explicitly set `mime_type = "application/octet-stream"` on the Part. Both cases reach the same failure point.

There was also an inconsistency between the two content paths:

- The `inline_data` path raises `ValueError` immediately for unsupported MIME types
- The `file_data` path silently used a fallback and failed later with a cryptic message

`GcsArtifactService` generates URIs like `gs://bucket/artifact/0` with no extension and no MIME type, making ADK's own artifact system the primary trigger for this fallback.

## Fix

Removes `_DEFAULT_MIME_TYPE` and raises `ValueError` early with an actionable message when the resolved MIME type is either unknown or `application/octet-stream`. This aligns the `file_data` path with the existing fail-fast behavior of the `inline_data` path.

The logic order is also corrected so providers that always produce a text fallback (anthropic, non-Gemini Vertex AI) and OpenAI/Azure HTTP media URLs are handled before the MIME type guard, keeping those paths unaffected.

## Changes

- `src/google/adk/models/lite_llm.py`: remove `_DEFAULT_MIME_TYPE`, restructure file_uri handling block, raise `ValueError` for missing or generic MIME types
- `tests/unittests/models/test_litellm.py`: update two existing tests to assert the new `ValueError`, add one new test covering explicit `application/octet-stream`

## Testing

```
pytest tests/unittests/models/test_litellm.py
241 passed, 5 errors (pre-existing, missing pytest-mock fixture)
```

Format verified with `pyink`. mypy error count unchanged from main (26).

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -234,10 +234,6 @@ def _get_provider_from_model(model: str) -> str:
   return ""
 
 
-# Default MIME type when none can be inferred
-_DEFAULT_MIME_TYPE = "application/octet-stream"
-
-
 def _infer_mime_type_from_uri(uri: str) -> Optional[str]:
   """Attempts to infer MIME type from a URI's path extension.
 
@@ -1103,33 +1099,33 @@ async def _get_content(
         })
         continue
 
-      # Determine MIME type: use explicit value, infer from URI, or use default.
+      # Resolve MIME type early: needed before the media-URL shortcut below,
+      # which must run before the generic text-fallback check. The raise is
+      # deferred until after all early-continue paths so that providers which
+      # always fall back to text (anthropic, non-Gemini Vertex AI) are never
+      # asked for a MIME type they cannot supply.
       mime_type = part.file_data.mime_type
       if not mime_type:
         mime_type = _infer_mime_type_from_uri(part.file_data.file_uri)
       if not mime_type and part.file_data.display_name:
         guessed_mime_type, _ = mimetypes.guess_type(part.file_data.display_name)
         mime_type = guessed_mime_type
-      if not mime_type:
-        # LiteLLM's Vertex AI backend requires format for GCS URIs.
-        mime_type = _DEFAULT_MIME_TYPE
-        logger.debug(
-            "Could not determine MIME type for file_uri %s, using default: %s",
-            part.file_data.file_uri,
-            mime_type,
-        )
-      mime_type = _normalize_mime_type(mime_type)
+      if mime_type:
+        mime_type = _normalize_mime_type(mime_type)
 
+      # For OpenAI/Azure: HTTP media URLs (image, video, audio) are sent as
+      # typed URL blocks and must be handled before the generic text fallback.
       if provider in _FILE_ID_REQUIRED_PROVIDERS and _is_http_url(
           part.file_data.file_uri
       ):
-        url_content_type = _media_url_content_type(mime_type)
-        if url_content_type:
-          content_objects.append({
-              "type": url_content_type,
-              url_content_type: {"url": part.file_data.file_uri},
-          })
-          continue
+        if mime_type:
+          url_content_type = _media_url_content_type(mime_type)
+          if url_content_type:
+            content_objects.append({
+                "type": url_content_type,
+                url_content_type: {"url": part.file_data.file_uri},
+            })
+            continue
 
       if _requires_file_uri_fallback(provider, model, part.file_data.file_uri):
         logger.debug(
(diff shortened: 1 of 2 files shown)


rohityan:

rohityan:
Hi @wukath, can you please review this.

Raman369AI:
> @Raman369AI
> One thing to confirm before this is merge-ready: do the unit tests exercise both branches —
> (1) a file_uri with no inferrable MIME type where resolution falls through to None, and
> (2) an explicit mime_type=\"application/octet-stream\" passed in by an upstream caller?

Following up on @surajksharma07's review — yes, both branches are now covered.

Introduced `test_content_to_message_param_user_message_file_uri_explicit_octet_stream` to exercise branch (2) at the `_content_to_message_param` level, alongside the existing tests that covered branch (1). All four relevant tests pass:

**Branch (1) — MIME resolves to `None`:**
- `test_content_to_message_param_user_message_file_uri_without_mime_type`
- `test_get_content_file_uri_default_mime_type`

**Branch (2) — explicit `application/octet-stream` from caller:**
- `test_content_to_message_param_user_message_file_uri_explicit_octet_stream` *(new)*
- `test_get_content_file_uri_explicit_octet_stream_raises`

```
============================= test session starts ==============================
platform linux -- Python 3.12.3, pytest-9.0.2, pluggy-1.6.0

collected 247 items / 243 deselected / 4 selected

tests/unittests/models/test_litellm.py::test_content_to_message_param_user_message_file_uri_without_mime_type PASSED [ 25%]
tests/unittests/models/test_litellm.py::test_content_to_message_param_user_message_file_uri_explicit_octet_stream PASSED [ 50%]
tests/unittests/models/test_litellm.py::test_get_content_file_uri_default_mime_type PASSED [ 75%]
tests/unittests/models/test_litellm.py::test_get_content_file_uri_explicit_octet_stream_raises PASSED [100%]

====================== 4 passed, 243 deselected in 2.43s =======================
```

Raman369AI:
Hi @wukath - picking up from @surajksharma07's last review pass on the issue.

@surajksharma07 confirmed the guard positioning is correct and asked whether both `ValueError` branches are covered by tests. That's addressed in the latest commit (`6a7b6a37`):

- Branch (1) - MIME resolves to `None` (no extension, no `display_name`, no explicit `mime_type`): covered by `test_content_to_message_param_user_message_file_uri_without_mime_type` and `test_get_content_file_uri_default_mime_type`
- Branch (2) - explicit `mime_type="application/octet-stream"` from an upstream caller: covered by `test_content_to_message_param_user_message_file_uri_explicit_octet_stream` *(new)* and `test_get_content_file_uri_explicit_octet_stream_raises`

All four pass cleanly. Would appreciate your sign-off when you get a chance - happy to adjust anything on your end.

Raman369AI:
Rebased onto `main` and CI is green. @wukath, gentle ping when you have a moment - happy to address any further feedback.

Issue #5027, opened 2026-03-27
Feature Request: Allow passing a custom `Client` to `Gemini`


naresh-gn:
### Is your feature request related to a specific problem?

When building a multi-agent system with `LlmAgent` and `Gemini`, each agent may need a different model hosted in a different Vertex AI location. For example:

- **Agent 1** uses `gemini-3.1-pro-preview`, which is only available in the `global` location
- **Agent 2** uses `gemini-2.5-flash`, which is available in `us-central1`

Currently, `Gemini` creates its own `Client` internally via the `api_client` cached property ([google_llm.py L273](https://github.com/google/adk-python/blob/main/src/google/adk/models/google_llm.py)), which reads project and location from environment variables (`GOOGLE_CLOUD_PROJECT`, `GOOGLE_CLOUD_LOCATION`). Since environment variables are process-global, there is no way to configure different locations (or projects) for different agents within the same process.

Previously it was possible to construct a `google.genai.Client` with per-agent settings and pass it into `Gemini(client=client)`, but this is no longer supported — `Gemini` is a Pydantic `BaseModel` that only accepts `model`, `base_url`, `speech_config`, `use_interactions_api`, and `retry_options`.

This makes it **impossible to run a multi-agent system where agents target different Vertex AI locations or projects**.

### Describe the Solution You'd Like

Allow `Gemini` to accept an optional `client: Optional[Client] = None` parameter. When provided, `Gemini` should use the supplied client instead of constructing its own in the `api_client` cached property.

**Proposed API:**

```python
from google.genai import Client
from google.adk.models.google_llm import Gemini
from google.adk.agents import LlmAgent

client_global = Client(vertexai=True, project="my-project", location="global")
client_us = Client(vertexai=True, project="my-project", location="us-central1")

agent_1 = LlmAgent(
    name="Agent1",
    model=Gemini(client=client_global, model="gemini-3.1-pro-preview"),
    instruction="...",
)

agent_2 = LlmAgent(
    name="Agent2",
    model=Gemini(client=client_us, model="gemini-2.5-flash"),
    instruction="...",
)
```

If `client` is not provided, behavior stays exactly as it is today (auto-create from env vars).

### Impact on your work

[...]

PR #5035, opened 2026-03-28
feat: support injecting custom LLM clients into Gemini and AnthropicLlm

brucearctor:
Enable injecting pre-configured LLM clients into Gemini and AnthropicLlm models to support multi-agent systems with distinct configurations. Addresses #5027.

## Testing Plan

I have verified these changes with the following:

### Automated Tests
- **New Unit Tests**: Added  which specifically mocks the  and  clients to verify they are correctly consumed by the cached properties.
- **Regression Testing**: Ran all 86 existing unit tests in  and  to ensure legacy behavior via environment variables is preserved.

### Manual Verification
- **Pydantic Validation**: Verified model initialization with actual  and  objects to ensure no runtime Pydantic resolution errors ().

### Verification Command
```bash
uv run pytest tests/unittests/models/test_custom_client.py
uv run pytest tests/unittests/models/test_google_llm.py tests/unittests/models/test_anthropic_llm.py
```


src/google/adk/models/google_llm.py
```diff
@@ -302,6 +310,9 @@ def api_client(self) -> Client:
     Returns:
       The api client.
     """
+    if self.client:
```
rohityan, line 313: Shouldn't this logic be applied to `_live_api_client` as well?

src/google/adk/models/google_llm.py
```diff
@@ -26,6 +26,7 @@
 from typing import TYPE_CHECKING
 from typing import Union
 
+from google.genai import Client
```
rohityan: Please move this import back inside the method or use TYPE_CHECKING.

src/google/adk/models/google_llm.py
```diff
@@ -91,6 +92,13 @@ class Gemini(BaseLlm):
 
   model: str = 'gemini-2.5-flash'
 
+  client: Optional[Client] = None
```
rohityan: Since the client is a live network connection and not just data, Pydantic will fail or produce errors if it tries to serialize it during a model_dump() call.

src/google/adk/models/apigee_llm.py
```diff
@@ -220,6 +220,9 @@ def api_client(self) -> Client:
     Returns:
       The api client.
     """
+    if self.client:
```
rohityan, line 225: You are missing client field declaration to the ApigeeLlm class definition.

rohityan:
Hi @brucearctor , Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Please address the review comments.

src/google/adk/models/google_llm.py
brucearctor: I'll double check.  I thought I moved this out for a reason - but has been awhile.

brucearctor:
> Hi @brucearctor , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please address the review comments.

looking into them

brucearctor:
I think got these comments addressed.  While I hate using Any type, to address your comments/suggestions, I failed to find a workaround.

naresh-gn:
@rohityan any update on this PR?

PR #5052, opened 2026-03-29
fix: emit additional_properties with value type schema in dict branch

### LINKED ISSUE #4868: Output_Schema + Tool with structured arg [List of object] causes blank tool input [{}] and infinite loop
U**Description:**
When for an LLMAgent output_schema is defined and the agent has a tool with at least one argument typed as a list or object structure, the tool receives blank/empty values for that argument. After that, the agent keeps invoking the same tool repeatedly and appears to run indefinitely.

**Steps to Reproduce:**
 1 Create an Agent
 2 Define an output_schema using a Pydantic model
 3 Add a tool that accepts at least one structured arg of type list[dict[str, str]]
 4 Run the agent with a prompt that should cause it to call the tool with populated values
 5 Observe that the tool is invoked with blank/empty values and then continues being called repeatedly instead of the agent completing normally

**Expected Behavior:**
The agent should construct and pass proper values to the tool argument, execute the tool normally, and then return a final response that conforms to the defined output_schema.

**Observed Behavior:**
The tool is called with blank/empty values for the structured argument. After that, the agent continues calling the tool repeatedly and does not terminate normally, appearing to enter an infinite loop.

**Environment Details:**

 - ADK Library Version (pip show google-adk): Reproduced across versions 1.23.0 to 1.27.2
 - Desktop OS: Windows

**Model Information:**

 - Are you using LiteLLM: No

**Screenshots:**
---

## 🟡 Optional Information

**Regression:**
I tested this across multiple ADK versions from 1.23.0 to 1.27.2, and the issue reproduces in all of them. I have not found a version where this works correctly in my setup.

```python
from google.adk.tools import ToolContext
import random

load_dotenv()

class AgentResponse(BaseModel):
    """Structured agent output with the main response separated from facts."""
    response: str = Field(
        description="Friendly user-facing summary of the fruit basket total."
    )
    facts: list[str] | None = Field(
        description="Exactly 3 short and fun fruit facts related to the selected fruits."
    )

def fruits_to_cost(tool_context: ToolContext, fruits: list[dict[str, str]]) -> str:
    """
    Takes a list of fruit dicts and returns a total cost.
    If qty or cost-per-piece is missing, a random value is generated.
    The agent can build this fruit list even from a short user request.
    Example user request: "Make me a colorful tropical fruit basket"
    Example tool input: [{"name": "Mango", "qty": "3", "cost-per-piece": "1.80"}, {"name": "Pineapple"}, {"name": "Kiwi"}]
    Returns: "Estimated basket total for Mango x3 @ $1.80, Pineapple x4 @ $2.10, Kiwi x2 @ $0.90 is $15.00"
    """
    if not fruits:
        return "No fruits were provided, so the total is $0.00"

    basket_lines: list[str] = []
    total_cost = 0.0

    for fruit in fruits:
        name = fruit.get("name", "Unknown Fruit")
        qty = int(fruit.get("qty", str(random.randint(1, 6))))
        cost_per_piece = float(
            fruit.get("cost-per-piece", f"{random.uniform(0.4, 2.5):.2f}")
        )
        line_total = qty * cost_per_piece
        total_cost += line_total
        basket_lines.append(f"{name} x{qty} @ ${cost_per_piece:.2f}")

    return (
        f"Estimated basket total for ( {', '.join(basket_lines)} ) is ${total_cost:.2f}"
    )


root_agent = Agent(
    model='gemini-2.5-flash',
    name='root_agent',
    description=("Dynamic fruit basket cost calculator"),
    instruction=(
        "## Role\n"
        "You are responsible for creating a lively fruit basket estimate. "
        "The user may give only a short request, such as a mood, color, occasion, or just one fruit name. "
        "In that case, infer a suitable fruit basket yourself and call the tool 'fruits_to_cost' with a list of fruit dictionaries. "
        "Each fruit can include 'name', and may optionally include 'qty' and 'cost-per-piece'. "
        "If some values are missing, the tool will generate them dynamically, so do not ask the user for extra details unless the request is truly unclear. "
        "Choose fruits that match the user intent and make the basket feel vivid and practical. "
        "Then return a cheerful summary in 'response' and exactly 3 engaging fruit facts in 'facts'."
    ),
    tools=[fruits_to_cost],
    output_schema=AgentResponse
)
```

 - Always (100%)

weiguangli-io:
## Summary

Fixes #4868

The `dict` origin branch in `_parse_schema_from_parameter` previously returned `Schema(type=OBJECT)` without any value type information for parameterized dict types like `dict[str, str]`. This caused Gemini to receive a bare `{"type": "object"}` with no field constraints, leading to:

- Empty/blank tool arguments (`[{}]`)
- Infinite retry loops when combined with `output_schema`

### Changes

- **`_function_parameter_parse_util.py`**: When the dict type has 2 type args (key and value), parse the value type and set `additional_properties` on the schema. For example, `dict[str, str]` now emits `{"type": "OBJECT", "additionalProperties": {"type": "STRING"}}` instead of just `{"type": "OBJECT"}`.

- **`test_build_function_declaration.py`**: Added tests for `dict[str, str]`, `dict[str, int]`, and `list[dict[str, str]]` to verify `additional_properties` is correctly set. Updated existing `test_list` to also assert on `additional_properties` for the dict items.

### How it works

The fix follows the same pattern used by the `list` branch, which recursively calls `_parse_schema_from_parameter` on the element type to build `schema.items`. Similarly, the dict branch now recursively calls `_parse_schema_from_parameter` on the value type (args[1]) to build `schema.additional_properties`.

## Test plan

- [x] Existing tests pass (40/40 in `test_build_function_declaration.py`)
- [x] New tests added for `dict[str, str]`, `dict[str, int]`, and `list[dict[str, str]]`
- [ ] Manual verification with the reproduction script from the issue

### DIFF
--- a/src/google/adk/tools/_function_parameter_parse_util.py
+++ b/src/google/adk/tools/_function_parameter_parse_util.py
@@ -294,6 +294,17 @@ def _parse_schema_from_parameter(
     args = get_args(param.annotation)
     if origin is dict:
       schema.type = types.Type.OBJECT
+      if len(args) == 2:
+        value_type = args[1]
+        schema.additional_properties = _parse_schema_from_parameter(
+            variant,
+            inspect.Parameter(
+                'value',
+                inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                annotation=value_type,
+            ),
+            func_name,
+        )
       if param.default is not inspect.Parameter.empty:
         if not _is_default_value_compatible(param.default, param.annotation):
           raise ValueError(default_value_error_msg)
--- a/tests/unittests/tools/test_build_function_declaration.py
+++ b/tests/unittests/tools/test_build_function_declaration.py
@@ -102,6 +102,53 @@ def test_dict_input():
   assert function_decl.name == 'simple_function'
   assert function_decl.parameters.type == 'OBJECT'
   assert function_decl.parameters.properties['input_str'].type == 'OBJECT'
+  assert (
+      function_decl.parameters.properties[
+          'input_str'
+      ].additional_properties.type
+      == 'STRING'
+  )
+
+
+def test_dict_input_with_int_values():
+  def simple_function(input_str: dict[str, int]) -> str:
+    return {'result': input_str}
+
+  function_decl = _automatic_function_calling_util.build_function_declaration(
+      func=simple_function
+  )
+
+  assert function_decl.name == 'simple_function'
+  assert function_decl.parameters.type == 'OBJECT'
+  assert function_decl.parameters.properties['input_str'].type == 'OBJECT'
+  assert (
+      function_decl.parameters.properties[
+          'input_str'
+      ].additional_properties.type
+      == 'INTEGER'
+  )
+
+
+def test_list_of_dict_input():
+  """Test list[dict[str, str]] emits proper schema with additional_properties."""
+
+  def simple_function(fruits: list[dict[str, str]]) -> str:
+    return str(fruits)
+
+  function_decl = _automatic_function_calling_util.build_function_declaration(
+      func=simple_function
+  )
+
+  assert function_decl.name == 'simple_function'
+  assert function_decl.parameters.type == 'OBJECT'
+  assert function_decl.parameters.properties['fruits'].type == 'ARRAY'
+  assert function_decl.parameters.properties['fruits'].items.type == 'OBJECT'
+  assert (
+      function_decl.parameters.properties[
+          'fruits'
+      ].items.additional_properties.type
+      == 'STRING'
+  )
 
 
 def test_basemodel_input():
@@ -279,6 +326,12 @@ def test_list():
   assert function_decl.parameters.properties['input_str'].items.type == 'STRING'
   assert function_decl.parameters.properties['input_dir'].type == 'ARRAY'
   assert function_decl.parameters.properties['input_dir'].items.type == 'OBJECT'
+  assert (
+      function_decl.parameters.properties[
+          'input_dir'
+      ].items.additional_properties.type
+      == 'STRING'
+  )
 
 
 def test_enums():


rohityan:
Hi @weiguangli-io , Thank you for your contribution! We appreciate you taking the time to submit this pull request.
The [`_sanitize_schema_formats_for_gemini`](https://github.com/google/adk-python/blob/f97367381e820c75ad16d4ce7ee27c0f9929c81d/src/google/adk/tools/_gemini_schema_util.py#L167) would likely delete the `additional_properties` added to `_parse_schema_from_parameter`. 
Can you upload some screenshots/logs verifying the end to end testing and it won't get deleted.

weiguangli-io:
@rohityan Thanks for the thorough review! I've investigated and here's the analysis:

**`_sanitize_schema_formats_for_gemini` does intentionally strip `additional_properties`** (line 167 in `_gemini_schema_util.py`), but this operates on a **separate code path** from the changes in this PR:

1. **`_parse_schema_from_parameter`** (this PR's path) — constructs `types.Schema` objects directly. When `build_function_declaration()` is called, it returns `types.Schema` objects with `additional_properties` correctly set. This is the path used for automatic function calling.

2. **`_to_gemini_schema` → `_sanitize_schema_formats_for_gemini`** — operates on JSON dict representations and strips `additional_properties` before sending to Gemini API, because Gemini rejects it. This is intentional and there's even an existing test (`test_gemini_schema_util.py` line 554) verifying this behavior.

**These are independent paths.** This PR adds `additional_properties` in path (1), which is correct — the `types.Schema` object carries the full schema information for automatic function calling. The Gemini sanitization in path (2) is a separate concern that only applies when converting schemas to Gemini's wire format.

**Test evidence:**
- All 3 new tests in `test_build_function_declaration.py` verify `additional_properties` is correctly set on the `types.Schema` objects: `dict[str, str]` → `STRING`, `dict[str, int]` → `INTEGER`, `list[dict[str, str]]` → nested `STRING`.
- The existing `test_gemini_schema_util.py` tests continue to pass, confirming no interference.

The behavior is consistent with how `_parse_schema_from_parameter` already handles other `types.Schema` fields that may later be filtered in the Gemini path.

Issue #4525, opened 2026-02-18
`set_model_response` error responses are treated as final output, bypassing ReflectAndRetryToolPlugin retry


chaerrypick01:
## Description

When `SetModelResponseTool` raises a `ValidationError` (e.g., LLM passes `null` for required string fields), the `ReflectAndRetryToolPlugin` correctly catches the error via `on_tool_error_callback` and returns a `ToolFailureResponse`. However, `get_structured_model_response()` in `_output_schema_processor.py` only checks the tool **name** (`set_model_response`) without inspecting whether the response is an error or a valid result. This causes the error response to be treated as the **final model output**, completely bypassing the plugin's retry mechanism.

## Steps to Reproduce

1. Create an `LlmAgent` with `output_schema` (e.g., required `str` fields) and at least one tool
2. Use a non-Vertex AI model (e.g., LiteLLM) so `can_use_output_schema_with_tools()` returns `False` and `SetModelResponseTool` is injected
3. Enable `ReflectAndRetryToolPlugin` with `max_retries=3`
4. Trigger the agent where the LLM calls `set_model_response` with `null` for a required string field

## Expected Behavior

`ReflectAndRetryToolPlugin` should retry the `set_model_response` call up to `max_retries` times, giving the LLM a chance to correct its output.

## Actual Behavior

The `ToolFailureResponse` (with `response_type: "ERROR_HANDLED_BY_REFLECT_AND_RETRY_PLUGIN"`) is immediately treated as the final model response. No retry occurs.

## Root Cause

In `base_llm_flow.py` (lines ~649-659 and ~695-705):

```python
yield function_response_event

if json_response := _output_schema_processor.get_structured_model_response(
    function_response_event
):
    final_event = _output_schema_processor.create_final_model_response_event(
        invocation_context, json_response
    )
    yield final_event  # Error response becomes final output
```

And in `_output_schema_processor.py` (lines 111-114):

```python
for func_response in function_response_event.get_function_responses():
    if func_response.name == 'set_model_response':
        return json.dumps(func_response.response, ensure_ascii=False)  # No validation
```

`get_structured_model_response()` matches only on `func_response.name == 'set_model_response'` and returns the response as-is, without checking if it's a valid schema result or an error/reflection response from a plugin.

## Suggested Fix

[...]

Issue #3413, opened 2025-11-05
Agent with output_schema + tools enters infinite loop ADK v1.18.0


tommyhutcheson:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
  When an Agent is configured with both output_schema and tools in v1.18.0, the agent repeatedly calls the tool indefinitely instead of using the tool's response to generate the final structured output.

**To Reproduce**
AI generated code demonstrating the issue

dependencies installed:
```
requires-python = ">=3.12"
dependencies = [
    "backoff>=2.2.1",
    "cloudpickle>=3.1.1",
    "google-adk[eval]>=1.18.0",
    "google-cloud-aiplatform[adk,agent-engines]>=1.120.0",
    "google-cloud-secret-manager>=2.16.0",
    "google-cloud-bigquery>=3.25.0",
    "google-cloud-run>=0.11.0",
    "google-cloud-logging>=3.11.0",
]
```
```python
"""

Minimal reproduction case for ADK infinite loop issue.

This test demonstrates the problem where an agent with both output_schema and tools
may infinite loop when the tool is called repeatedly instead of producing final output.

Issue: When an agent has both output_schema and tools configured, it may repeatedly
call the tool instead of using the tool's response to generate the final output.
"""

import asyncio
import logging
import os
from typing import Any

import vertexai
from google.adk.agents import Agent
from google.adk.tools import FunctionTool, ToolContext
from pydantic import BaseModel, Field
from vertexai.preview import reasoning_engines

logging.basicConfig(
    level=logging.INFO,
    format="[%(asctime)s] %(levelname)s [%(name)s:%(lineno)d] - %(message)s",
)
logger = logging.getLogger(__name__)


class SimpleResult(BaseModel):
    """Simple output schema with just two fields."""

    status: bool = Field(description="Whether the validation was successful")
    message: str = Field(description="A message explaining the result")


def simple_validate(tool_context: ToolContext) -> dict[str, Any]:
    """
    A simple validation tool that returns data matching SimpleResult schema.

    Returns:
        Dict with 'status' (bool) and 'message' (str) fields
    """
    [logger.info](http://logger.info/)("🔧 simple_validate tool called")
    result = {"status": True, "message": "Validation successful"}
    [logger.info](http://logger.info/)(f"🔧 simple_validate returning: {result}")
    return result


test_agent = Agent(
[...]

PR #5057, opened 2026-03-30
fix: output_schema + tools infinite loop on Gemini 2.x

### LINKED ISSUE #5054: [BUG]: output_schema=str combined with tools causes infinite call_llm/execute_tools loop
## Description

Using bare `str` type as `output_schema` on an `Agent` that also has `tools` causes an infinite `call_llm → execute_tools → call_llm` loop. The LLM repeatedly calls tools instead of producing the final structured response via `set_model_response`.

This does not happen with `BaseModel` subclasses as `output_schema`, only with primitive types like `str`.

## Reproduction

```python
from google.adk import Agent, Workflow, Event

def plan_visualization(approach: str, technology: str, key_elements: list[str]) -> str:
    """Plan a visualization."""
    return f"Plan: {approach} using {technology}"

def query_data(query: str) -> list[dict]:
    """Query the database."""
    return [{"q": "Q1", "revenue": 100}]

# This agent loops infinitely:
plan_agent = Agent(
    name="plan_agent",
    model="gemini-2.5-pro",
    instruction="Call plan_visualization with approach, technology, and key elements.",
    tools=[plan_visualization, query_data],
    output_schema=str,       # <-- bare str causes the loop
    mode="single_turn",
)

# Use in a Workflow or with Runner — agent calls tools repeatedly
# without ever settling on a final string response
```

## Expected Behavior

Either:
1. `output_schema=str` works correctly with tools — the LLM calls tools, then uses `set_model_response` to return a plain string, OR
2. ADK raises a validation error at agent creation time: `"output_schema must be a BaseModel subclass when tools are specified"`

## Actual Behavior

The agent enters an infinite loop:
1. LLM generates a `function_call` (e.g., `plan_visualization`)
2. ADK executes the tool and returns the result
3. LLM generates another `function_call` instead of calling `set_model_response`
4. Repeat endlessly — burns API credits silently

## Analysis

ADK auto-adds `set_model_response` when both `output_schema` and `tools` are specified ([documented here](https://github.com/google/adk-python/blob/v2.0.0a1/contributing/samples/output_schema_with_tools/README.md)). With `BaseModel` schemas, the model correctly calls `set_model_response` with structured JSON. With bare `str`, the `set_model_response` schema may be malformed or confusing to the model, causing it to fall back to repeated tool calls.

## Workaround

Remove `output_schema` from agents that have tools. Use other mechanisms to suppress unwanted text output if needed.

## Related Issues

- #3413 — Agent with output_schema + tools enters infinite loop (model regression, fixed)
- #3969 — Output Schema consistently ignored by LLM Agent with tools
- #4868 — output_schema + Tool with structured arg causes infinite loop

## Environment

- `google-adk` 2.0.0a2
- Model: `gemini-2.5-pro` (also tested with `gemini-2.5-flash`)
- Python 3.14

> Context for this issue was developed with AI assistance (Claude).

vietnamesekid:
## what's going on

Using `output_schema` together with `tools` on Gemini 2.5 Flash / 2.5 Pro / 2.0 Flash via Vertex AI causes an infinite loop. The agent just keeps calling tools over and over and never actually returns the structured response.

Ran into this while building a multi-agent setup where a sub-agent has both `output_schema` and `tools`, then gets wrapped as an `AgentTool` for a parent agent. The sub-agent logs were just endless cycles of "Sending out request" / "Response received" every ~2 seconds.

## root cause

`can_use_output_schema_with_tools()` in `output_schema_utils.py` uses `is_gemini_2_or_above()`, which returns `True` for all Gemini 2.x. This tells ADK to set `response_schema` directly on the API request alongside `tools`.

Problem is Gemini 2.x doesn't actually support this combo. The model receives both constraints and can't reconcile them, so it just keeps generating tool calls without ever producing structured output. According to [Google's structured output docs](https://ai.google.dev/gemini-api/docs/structured-output), this is only supported on Gemini 3 series (preview).

When `can_use_output_schema_with_tools()` returns `False`, ADK uses the `SetModelResponseTool` workaround instead, which works perfectly model treats it as a regular function call to deliver the structured output.

## the fix

Added `is_gemini_3_or_above()` to `model_name_utils.py` and swapped it in `output_schema_utils.py`. Pretty straightforward:

- Gemini 2.x → uses `SetModelResponseTool` workaround (reliable, already battle tested)
- Gemini 3.x → uses native `response_schema` + `tools` (officially supported)

No other callers of `is_gemini_2_or_above` are affected those are for unrelated features (URL context, RAG retrieval, code executor).

## related issues

Likely fixes #5054, #4868, #4525, and the regression from #3413.

## test plan

- [x] All existing tests pass (59/59)
- [x] Updated `test_output_schema_utils.py` Gemini 2.x on Vertex AI now correctly returns `False`
- [x] Added Gemini 3.x test cases that return `True`
- [x] Added `TestIsGemini3OrAbove` class in `test_model_name_utils.py`
- [x] Downstream tests in `test_basic_processor.py` and `test_output_schema_processor.py` unaffected (they mock the function)


rohityan:
Hi @vietnamesekid ,Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Can you please fix the failing mypy-diff tests and also formatting errors. You can use autoformat.sh to fix formatting errors.

vietnamesekid:
> Hi @vietnamesekid ,Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing mypy-diff tests and also formatting errors. You can use autoformat.sh to fix formatting errors.

Thanks @rohityan! I've pushed fixes for both the mypy and formatting issues. Note that these were pre-existing in the codebase, not introduced by my PR, but happy to fix them here

surfai:
Thanks @nino-robotfutures-co for the model comparison data on #5054 and @surajksharma07 for tracing the issue to `_OutputSchemaRequestProcessor` and pointing us in the right direction. Dropping findings here as requested.

*This comment was developed with AI assistance (Claude).*

---

### Context

This PR correctly fixes which models use `SetModelResponseTool` vs native `response_schema + tools`. That's necessary — but the underlying problem is broader: even when `SetModelResponseTool` is active, flash models can still ignore it and loop indefinitely.

@nino-robotfutures-co's testing on #5054 quantifies this:

| Model | output_schema + tools | Tool loop? | Structured output produced? |
|---|---|---|---|
| gemini-2.5-flash | HalOutput (BaseModel) | Yes — loops on `check_backend_health` ~15x, eventually breaks out | Yes, after loop |
| gemini-3-flash-preview | HalOutput (BaseModel) | Yes — 47 model calls, never terminates within 300s | No |
| **gemini-3.1-pro-preview** | HalOutput (BaseModel) | **No — 1 model call** | **Yes, first attempt** |

Same ADK code, same `SetModelResponseTool`, different models, different outcomes. The loop termination is non-deterministic — it depends entirely on whether the LLM chooses to call `set_model_response` vs other tools on any given turn. A prompt instruction improves the probability but can never guarantee it. Passing 10 times doesn't prove it won't fail on attempt 11.

**The core insight: non-deterministic routing will always be probabilistically right... and probabilistically wrong. The fix needs a deterministic escape path, not just a better prompt.**

---

### Findings

Four complementary changes that, together with this PR's model-detection fix, make the `output_schema + tools` path robust across all models.

#### 1. The instruction is identical for all schema types — the tool signature is the real differentiator

`_output_schema_processor.py:57-63` injects the same instruction regardless of schema type:

```python
instruction = (
    'IMPORTANT: You have access to other tools, but you must provide '
    'your final response using the set_model_response tool with the '
    'required structured format. After using any other tools needed '
    'to complete the task, always call set_model_response with your '
    'final answer in the specified schema format.'
)
```

What *does* differ by type is the **`set_model_response` tool signature** the LLM sees (from `set_model_response_tool.py:69-99`):
- `BaseModel` → `set_model_response(*, summary: str, score: int)` — distinct named fields
- `str` → `set_model_response(*, response: str)` — single generic field

Flash models treat the trivial single-field signature as redundant ("I can just respond with text") rather than recognizing it as a required termination mechanism. The named fields in the BaseModel case act as a stronger signal.

**Proposed fix — type-aware instruction:**

```python
# _output_schema_processor.py:56-64
from ...utils._schema_utils import is_basemodel_schema

if is_basemodel_schema(agent.output_schema):
    instruction = (
        'After completing any needed tool calls, provide your final '
        'response by calling set_model_response with the required fields.'
    )
else:
    instruction = (
        'CRITICAL: You MUST call the set_model_response tool exactly once '
        'as your FINAL action. Do NOT respond with plain text. Do NOT call '
        'any other tools after set_model_response. This is the ONLY way '
        'to complete the task.'
    )
```

#### 2. No safety net when the model ignores the instruction — a 2-layer deterministic guard

The stricter instruction (finding 1) reduces how often the model loops, but it's still probabilistic. A deterministic safety net complements it — two layers, both driven by a single round counter computed early in `run_async()`:

```python
# _output_schema_processor.py — at the top of run_async(), after the early-return guard
_MAX_TOOL_ROUNDS = 25

tool_rounds = sum(
    1 for e in invocation_context._get_events(
        current_invocation=True, current_branch=True
    )
    if e.get_function_responses()
)
```

**Layer 2 (checked first): Hard cutoff on round N** — last-resort termination if the forced `tool_choice` still doesn't produce output:

```python
# _output_schema_processor.py — immediately after computing tool_rounds
if tool_rounds >= _MAX_TOOL_ROUNDS:
    logger.error(
        'Tool execution reached %d rounds without producing structured '
        'output via set_model_response. Breaking loop to prevent '
        'runaway API costs.',
        tool_rounds,
    )
    invocation_context.end_invocation = True
    return
```

**Layer 1: Force `tool_choice` on round N-1** — constrain the LLM to *only* `set_model_response`, guaranteeing a valid structured response:

```python
# _output_schema_processor.py — after appending tools/instruction
if tool_rounds >= _MAX_TOOL_ROUNDS - 1:
    llm_request.config.tool_config = types.ToolConfig(
        function_calling_config=types.FunctionCallingConfig(
            mode=types.FunctionCallingConfigMode.ANY,
            allowed_function_names=['set_model_response'],
        )
    )
```

The type-aware instruction reduces how often the guard fires. The `tool_choice` enforcement ensures valid output when the model won't comply. The hard cutoff prevents runaway costs in all cases. All three layers together make the termination **deterministic** rather than prompt-probabilistic.

#### 3. Missing `return` after successful `set_model_response`

Separate from the infinite loop — when `set_model_response` IS successfully called in `base_llm_flow.py:1083-1093`, the `final_event` is yielded but execution falls through to `transfer_to_agent` (line 1094) and any subsequent processing:

```python
# base_llm_flow.py:1083-1093 — current code
      if json_response := _output_schema_processor.get_structured_model_response(
          function_response_event
      ):
        final_event = (
            _output_schema_processor.create_final_model_response_event(
                invocation_context, json_response
            )
        )
        yield final_event
      transfer_to_agent = function_response_event.actions.transfer_to_agent  # falls through
```

**Proposed fix:**

```python
        yield final_event
        return  # Structured output produced — terminate processing
```

#### 4. Two architectural patterns that avoid this class of problem entirely

For context on why the deterministic guard works — two patterns that reframe the problem:

**Pattern A: Two-phase execution.** Phase 1: model calls tools freely (no output schema). Phase 2: model is re-invoked with full history but *only* the schema tool, no other tools. The model can't confuse "call a tool" with "produce the response" because those actions never compete in the same LLM call.

**Pattern B: Always-forced tool selection.** Register the schema as a tool alongside real tools, but set `tool_choice=any` so the model *must* pick a tool every turn. It either calls a real tool (keep working) or calls the schema tool (finish). The model can't silently ignore the schema tool because it's forced to pick *something*.

The `tool_choice` guard in finding 2 is a lightweight version of Pattern B, applied as a safety net on round N-1 rather than as the primary mechanism.

---

### Summary for reviewer

This PR's model-detection fix is necessary and correct — Gemini 2.x should not use native `response_schema + tools`. But the `SetModelResponseTool` fallback path it activates has its own failure mode: flash models can still ignore `set_model_response` and loop indefinitely.

Findings 1-3 harden that fallback path with a layered defense:

| # | Change | Type | Scope |
|---|---|---|---|
| 1 | Type-aware instruction in `_output_schema_processor.py` | Probabilistic improvement | Reduces loop frequency for primitive schemas (`str`, `int`) |
| 2 | `tool_choice` enforcement on round N-1 | **Deterministic guard** | Forces valid structured output before cutoff |
| 3 | Hard cutoff on round N | **Deterministic guard** | Prevents runaway API costs in all cases |
| 4 | `return` after `yield final_event` in `base_llm_flow.py:1093` | Bug fix | Skips unnecessary `transfer_to_agent` processing after success |

Without findings 2-3, the path this PR activates for Gemini 2.x remains vulnerable to the same infinite loop on flash models — just through `SetModelResponseTool` instead of native `response_schema`. The model-detection fix and the deterministic guard are complementary; ideally both ship together.

vietnamesekid:
Thanks @surfai for the thorough analysis. I agree findings 1-3 are valuable improvements. I'd prefer to keep this PR focused on the model detection fix and handle those as a separate PR. That said, if you and @rohityan prefer, I'm happy to include them here. What do you think?

surfai:
Hope our discussion does not distract.  Certainly.  Thank you!!

vietnamesekid:
I've opened #5091 as a follow-up to implement the findings from @surfai's analysis. It adds type-aware instructions, a deterministic tool_choice guard, a hard cutoff, and the early return fix. All tests passing locally (unit + integration on both GOOGLE_AI and Vertex AI).

wuliang229:
@vietnamesekid did you use [gemini api](https://ai.google.dev/gemini-api/docs) or Vertex API to call the models? The behavior is slightly different on these platforms.

vietnamesekid:
> @vietnamesekid did you use [gemini api](https://ai.google.dev/gemini-api/docs) or Vertex API to call the models? The behavior is slightly different on these platforms.

I enabled Vertex

wuliang229:
@vietnamesekid did you observe that 2.5 in Vertex is not working? Since you linked a [documentation](https://ai.google.dev/gemini-api/docs/structured-output#structured_outputs_with_tools) of the Gemini API, I'm wondering have you tested it manually? It'd be great if you could provide a code snippet to reproduce.

GWeale:
closing as stale, the patch can't apply any more. the model-id version buckets were removed so is_gemini_2_or_above is gone, and the check moved into models/_capabilities.gemini_output_schema_and_tools. #5054 still tracks the report, but the repro @wuliang229 asked for is what a version gate needs first, since this path is vertex-only and the vertex docs still recommend structured output together with tools.

PR #5062, opened 2026-03-30
feat(to_a2a): add service_url support to to_a2a

### LINKED ISSUE #5060: to_a2a(): accept a service_url string instead of separate host/port/protocol params
## Feature request

`to_a2a()` currently requires `host`, `port`, and `protocol` as separate parameters:

```python
app = to_a2a(
    root_agent,
    host="localhost",
    port=8002,
    protocol="http",
    runner=_create_runner,
    agent_card=agent_card,
)
```

In practice, the service URL is typically available as a single string (e.g. `SERVICE_URL=http://localhost:8002` for local dev, or `https://my-agent.europe-west1.run.app` on Cloud Run). This forces every consumer to manually parse the URL with `urlparse()` to extract the three components:

```python
parsed = urlparse(settings.SERVICE_URL)
protocol = parsed.scheme or "http"
host = parsed.hostname or "localhost"
port = parsed.port or (443 if protocol == "https" else 80)
```

## Proposal

Accept an optional `service_url` parameter as an alternative to `host`/`port`/`protocol`:

```python
# New — single URL string
app = to_a2a(root_agent, service_url="http://localhost:8002", ...)

# Existing — still works
app = to_a2a(root_agent, host="localhost", port=8002, protocol="http", ...)
```

This would reduce boilerplate for Cloud Run, Kubernetes, and any deployment where the URL comes from an environment variable.

kaligautier:
## Summary
- add an optional `service_url` parameter to `to_a2a()` and use it for the published agent card URL
- preserve the existing `host`/`port`/`protocol` path as the fallback behavior and document precedence
- cover the new API with unit tests and update the A2A sample documentation

## Root cause
`to_a2a()` only accepted split host/port/protocol inputs even though the generated agent card ultimately needed a single service URL string, forcing callers to parse environment-provided URLs themselves.

## Testing
- `.venv/bin/python -m pytest tests/unittests/a2a/utils/test_agent_to_a2a.py`

Fixes #5060

### DIFF
--- a/contributing/samples/a2a_root/README.md
+++ b/contributing/samples/a2a_root/README.md
@@ -39,6 +39,7 @@ The A2A Root sample consists of:
 
 ### 4. **Simple Deployment Pattern**
 - Uses the `to_a2a()` utility to convert a standard ADK agent to an A2A service
+- Publishes the agent card URL with a single `service_url` string
 - Minimal configuration required for remote agent deployment
 
 ## Setup and Usage
--- a/contributing/samples/a2a_root/remote_a2a/hello_world/agent.py
+++ b/contributing/samples/a2a_root/remote_a2a/hello_world/agent.py
@@ -108,4 +108,4 @@ root_agent = Agent(
     ),
 )
 
-a2a_app = to_a2a(root_agent, port=8001)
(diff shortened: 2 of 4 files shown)


kaligautier:
@rohityan @xuanyang15

PR #5063, opened 2026-03-30
 feat: add input validation and type hints to BaseTool name

Pranav334:
This change improves the robustness of the BaseTool class by ensuring that tool names are compatible with LLM function-calling requirements.

Key Changes:
- Added explicit `str` type hints to `name` and `description` in the `BaseTool.__init__` method.
- Implemented a validation check that raises a `ValueError` if a tool name contains spaces.

Reasoning:
Most LLMs (including Gemini) expect function names to follow standard identifier rules (no spaces). Providing a tool with a space in the name can cause silent failures or unexpected behavior during the "Reason-Act" loop. This fix catches the error at instantiation time, providing a clear guidance to the developer to use underscores instead.

Test Coverage:
- Verified that `BaseTool(name="valid_name", ...)` passes.
- Verified that `BaseTool(name="invalid name", ...)` raises ValueError.


src/google/adk/tools/base_tool.py
```diff
@@ -68,11 +68,19 @@ class BaseTool(ABC):
...
       self,
       *,
-      name,
-      description,
+      name: str, # Added type hint for clarity
+      description: str,
       is_long_running: bool = False,
       custom_metadata: Optional[dict[str, Any]] = None,
   ):
+    # -CONTRIBUTION START ---
```
rohityan, line 76: Pls remove these comments

src/google/adk/tools/base_tool.py
```diff
@@ -68,11 +68,19 @@ class BaseTool(ABC):
...
+    if " " in name:
```
rohityan, line 77: add a regex to include all prohibited naming conventions.

rohityan:
Hi @Pranav334 , Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Pls address the review comments and fix the failing workflow tests later.

src/google/adk/tools/base_tool.py
Pranav334: Ok Rohit I will look into it Thanks

src/google/adk/tools/base_tool.py
Pranav334: Okay

Pranav334:
Hi @rohityan 
I have updated the PR to remove the comments and added identifier-based name validation as requested.

DeanChensj:
Enforcing strict no-whitespace validation in BaseTool.__init__ is a breaking change that breaks existing tools and third-party wrappers across the ecosystem. Function naming constraints should be sanitized or validated when converting tools to specific LLM schema declarations (e.g. Gemini FunctionDeclaration) rather than breaking base tool instantiation. Closing this PR.

Issue #4145, opened 2026-01-13
LongRunningFunctionTool + A2A Protocol Task Completion Issue


secprog:
# ADK Bug Report: LongRunningFunctionTool + A2A Protocol Task Completion Issue

## Summary
Tasks using `LongRunningFunctionTool` in A2A protocol never transition to "completed" state, causing frontend clients to wait indefinitely despite the tool function completing successfully.

## Environment
- **ADK Version:** Latest from current branch 
- **A2A SDK Version:** >=0.3.11
- **Python Version:** 3.12
- **LiteLLM Version:** 1.80.16
- **Pydantic Version:** 2.12.5

## Problem Description
When agents use `LongRunningFunctionTool` for async operations, the A2A protocol task state machine gets stuck in "working" state. The tool function executes successfully and returns results, but the task never receives the completion signal, leaving clients waiting indefinitely.

### Expected Behavior
1. Client sends A2A request to agent
2. Agent creates task with state "submitted"
3. Agent calls `LongRunningFunctionTool` function, state changes to "working"
4. Function completes and returns result
5. Task state transitions to "completed"
6. Client receives final response with task completion

### Actual Behavior
1. Client sends A2A request to agent ✅
2. Agent creates task with state "submitted" ✅
3. Agent calls `LongRunningFunctionTool` function, state changes to "working" ✅
4. Function completes and returns result ✅
5. **Task state NEVER transitions to "completed"** ❌
6. **Client never receives final response, waits indefinitely** ❌

### 1. Agent Setup (vision_agent.py)
```python
from google.adk.tools.long_running_tool import LongRunningFunctionTool
from google.adk.agents import Agent

# Async function that takes ~10-30 seconds
async def run_core_analysis_parallel(user_id: str, session_id: str, filename: str) -> str:
    # Perform analysis with LLM calls
    result = await some_async_operation()
    return json.dumps(result)

visual_analysis_agent = Agent(
    name="visual_analyzer",
    model=LiteLlm(model="openai/gpt-5.1"),
    instruction="Analyze images using tools",
    tools=[
        LongRunningFunctionTool(func=run_core_analysis_parallel),
    ],
    planner=plan_re_act_planner.PlanReActPlanner(),
)

app = to_a2a(
    agent=root_agent,
    app=App(
        resumability_config=ResumabilityConfig(is_resumable=True),
        session_service=DatabaseSessionService(db_url=db_url),
        artifact_service=FileArtifactService(root_dir="./artifacts"),
    ),
[...]

Issue #3986, opened 2025-12-20
Long-Running Tool FunctionResponse Routing Bypasses Custom Agent Orchestration


selimseker:
## Summary

When a sub-agent within a custom orchestrator agent calls a `LongRunningFunctionTool`, the `FunctionResponse` is routed directly to the sub-agent, bypassing the parent custom agent's `_run_async_impl` method entirely. This prevents custom agents from implementing proper workflow orchestration and continuation logic after long-running tool pauses.

## Our Architecture

We have a custom orchestrator agent (`DomainAgent`) that manages multi-step workflows:

```
DomainAgent (custom orchestrator)
├── Router Agent (LlmAgent)
├── Planner Agent (LlmAgent)
└── Step Executor Agent (LlmAgent) ← calls long-running tool
```

**Flow:**
1. `DomainAgent._run_async_impl()` orchestrates the workflow
2. Calls router → planner → executes multiple steps sequentially
3. Step Executor agent calls `ask_clarifications` (LongRunningFunctionTool)
4. Frontend collects user input and sends `FunctionResponse`
5. **Problem**: Runner routes directly to Step Executor, never re-enters `DomainAgent._run_async_impl()`

## The Issue

In `runners.py`, the `_find_agent_to_run()` method (line 1003-1055) determines which agent should handle the next message:

```python
def _find_agent_to_run(self, session: Session, root_agent: BaseAgent) -> BaseAgent:
    # If the last event is a function response, send to agent that made the function call
    event = find_matching_function_call(session.events)
    if event and event.author:
      return root_agent.find_agent(event.author)  # ← Returns sub-agent directly
    # ... rest of logic
```

**What happens:**
1. Step Executor calls long-running tool → event.author = "step_executor"
2. Frontend sends FunctionResponse via `run_async(invocation_id=..., new_message=FunctionResponse)`
3. `_find_agent_to_run()` finds the function call event with author="step_executor"
4. Returns Step Executor agent directly
5. Runner calls `step_executor.run_async()` 
6. **DomainAgent's orchestration logic is completely bypassed**

## Expected Behavior

The `DomainAgent` should be able to:
1. Detect that a FunctionResponse is being processed
2. Update workflow state
3. Continue orchestration (e.g., move to next workflow step)
4. Properly manage the entire workflow lifecycle

## Questions

1. **Is this the intended behavior?** Should long-running tools always route directly to the calling agent?
[...]

PR #5072, opened 2026-03-30
fix(flows): resume long-running tools after matching responses

### LINKED ISSUE #5064: LongRunningFunctionTool resume fails: unresolved pause check + streaming ID mismatch
## Description

`LongRunningFunctionTool` resume is broken when using streaming (`stream=True`, the default). After a user sends a `functionResponse` to resume a paused tool, the agent immediately returns without calling the LLM — effectively ending the run instead of continuing.

There are two distinct bugs that compound to cause this:

**Bug 1: `_run_one_step_async` doesn't check if a pause has been resolved**

In `base_llm_flow.py` around line 792, the pause check looks at the last two events:

```python
if (
    invocation_context.is_resumable
    and events
    and len(events) > 1
    and (
        invocation_context.should_pause_invocation(events[-1])
        or invocation_context.should_pause_invocation(events[-2])
    )
):
    return
```

When a `functionResponse` is appended to the session (resolving the pause), `events[-2]` is still the original event with `long_running_tool_ids` set. `should_pause_invocation(events[-2])` returns `True`, so the method returns early — even though the pause was already resolved by the `functionResponse` in `events[-1]`.

**Expected behavior:** The check should verify whether the paused tool IDs have corresponding `functionResponse` entries elsewhere in the event list. If all paused IDs are resolved, execution should continue.

**Bug 2: Streaming causes function call ID mismatch between partial and stored events**

`_finalize_model_response_event` (line 80-113) creates a **new** `Event` via `Event.model_validate()` and then calls `populate_client_function_call_id`, which assigns fresh `adk-*` UUIDs each time it's called.

With streaming enabled, `_finalize_model_response_event` is called for **every chunk** — partial events and the final non-partial event each get **different** `adk-*` IDs. However:

- **Partial events** are sent via SSE to the client but are **not** stored in the session (`_should_append_event` returns `False` for partial events)
- **Non-partial events** are stored in the session

If the frontend captures the `functionCall.id` from a partial SSE event (which arrives first and has `longRunningToolIds` set), that ID won't match the ID stored in the session's non-partial event. When the client sends a `functionResponse` with the partial event's ID, `find_event_by_function_call_id` fails with:

```
ValueError: Function call event not found for function response id: adk-XXXX
```

**Expected behavior:** Function call IDs should be stable across partial and non-partial events for the same LLM response, OR only non-partial events should have `longRunningToolIds` set.

## Steps to Reproduce

1. Create an agent with `LongRunningFunctionTool` and `ResumabilityConfig(is_resumable=True)`
2. Use streaming (the default)
3. Trigger the long-running tool — agent pauses correctly
4. Send a `functionResponse` to resume
5. **Bug 1:** Agent returns immediately without continuing (if you work around Bug 2)
6. **Bug 2:** `ValueError: Function call event not found` because the ID from the SSE partial event doesn't match the stored non-partial event

## Workarounds

**For Bug 1:** Patch `_run_one_step_async` to collect all `functionResponse` IDs across events and check if paused tool IDs have been resolved:

```python
if (
    invocation_context.is_resumable
    and events
    and len(events) > 1
):
    all_response_ids = set()
    for evt in events:
        for fr in evt.get_function_responses():
            if fr.id:
                all_response_ids.add(fr.id)

    has_unresolved_pause = False
    for check_evt in (events[-1], events[-2]):
        if invocation_context.should_pause_invocation(check_evt):
            paused_ids = check_evt.long_running_tool_ids or set()
            if paused_ids - all_response_ids:
                has_unresolved_pause = True
                break

    if has_unresolved_pause:
        return
```

**For Bug 2:** On the client side, only capture `longRunningToolIds` from **non-partial** events (`event.partial !== true`). Partial events have unstable IDs that won't match what's stored in the session.

## Environment

- `google-adk` version: 1.27.3
- Python: 3.13
- Model: gemini-2.5-flash (streaming enabled)
- Using: `LongRunningFunctionTool`, `ResumabilityConfig`, `AgentTool` (sub-agents)

## Related Issues

- #3986 — Long-Running Tool FunctionResponse routing
- #4145 — LongRunningFunctionTool + A2A task completion

zeel2104:

- Closes: #5064
- Related: #3986, #4145

Problem:
`LongRunningFunctionTool` resume could fail in resumable flows when streaming was enabled. There were two compounding issues:
1. Resume logic could still treat an invocation as paused even after a matching `functionResponse` had already resolved the long-running tool call.
2. Streaming partial and final model events could end up with different ADK-generated function call IDs, causing resume-time lookup of the original function call event to fail.

Solution:
This change fixes both sides of the resume path:
1. It adds unresolved-pause detection that checks long-running tool calls against all matching `functionResponse` IDs in the current invocation branch before deciding to stop execution.
2. It preserves previously assigned function call IDs across streaming partial and final model response events so the client-visible ID remains stable and matches the event persisted in session history.

### Testing Plan

I verified the change with targeted unit tests covering both parts of the fix:

resumable invocation logic now continues when a long-running tool call has a matching functionResponse
streaming finalization preserves function call IDs across partial and final events

Commands run

python -m pytest tests\unittests\agents\test_invocation_context.py tests\unittests\flows\llm_flows\test_base_llm_flow.py -q --basetemp=.pytest_tmp
python -m pytest tests\unittests\agents\test_invocation_context.py -q -k "has_unresolved_long_running_tool_calls" --basetemp=.pytest_tmp
python -m pytest tests\unittests\flows\llm_flows\test_base_llm_flow.py -q -k "preserves_function_call_ids" --basetemp=.pytest_tmp

**Unit Tests:**

Passed locally:

```bash
python -m pytest tests\unittests\agents\test_invocation_context.py tests\unittests\flows\llm_flows\test_base_llm_flow.py -q --basetemp=.pytest_tmp

Manual E2E tests were not run. I validated the behavior through targeted unit tests that cover the unresolved pause check and streaming function call ID stability, but I did not run a full end-to-end resumable streaming flow through the CLI/web/API with a live client resume sequence.


### Checklist

### Additional context

This fix is intentionally narrow:

resume behavior now only remains paused when a long-running tool call is still unresolved
streaming function call IDs are preserved across partial and final events so resume routing remains stable
The local test setup on Windows required installing the package with uv pip install -e . plus pytest dependencies directly, because the full test extra currently pulls a dependency chain that includes lancedb, which does not have a compatible wheel in this environment.

### CODE REVIEW

**Comment by rohityan** [comment]
Hi @zeel2104 , Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Can you please fix the failing tests before we can proceed with the review

**Review by tottenjordan — APPROVED** [review]
Tested this against our production `interactive_creative` agent (3 sequential `LongRunningFunctionTool` checkpoints with streaming enabled, `ResumabilityConfig(is_resumable=True)`).

## Verification

- **Installed PR branch** (`zeel2104/adk-python@fix-5064-long-running-resume`) — installs as `google-adk==1.28.0`
- **All 4 patches confirmed present:**
  1. `InvocationContext.has_unresolved_long_running_tool_calls()` replaces old `events[-2:]` check
  2. `preserve_existing_function_call_ids()` added to `functions.py`
  3. `_finalize_model_response_event` calls `preserve_existing_function_call_ids` before `populate_client_function_call_id`
  4. `LlmAgent._run_async_impl` uses `has_unresolved_long_running_tool_calls`
- **3 upstream unit tests pass:** `test_has_unresolved_long_running_tool_calls_with_matching_response`, `test_has_unresolved_long_running_tool_calls_without_matching_response`, `test_finalize_model_response_event_preserves_function_call_ids`
- **125 downstream project tests pass** with no regressions

## Review of the fix

**Bug 1 fix** — Clean approach. Collecting all `functionResponse` IDs across the event list and checking against `long_running_tool_ids` is correct and handles the general case (not just the last 2 events). Using the same method in both `llm_agent.py` and `base_llm_flow.py` eliminates the inconsistency.

**Bug 2 fix** — `preserve_existing_function_call_ids` correctly carries forward IDs by matching on function name and only filling in missing IDs (`if current_function_call.id: continue`). Calling it before `populate_client_function_call_id` ensures existing IDs are preserved and only truly new calls get fresh UUIDs.

LGTM — this fixes both issues cleanly with minimal surface area. Thanks @zeel2104!

**Comment by tottenjordan** [comment]
@zeel2104 I dug into the 3 CI failures and opened a PR against your branch with fixes: https://github.com/zeel2104/adk-python/pull/1

All 3 issues are in `has_unresolved_long_running_tool_calls()` in `invocation_context.py`:

**1. pyink** — method signature needs to be on one line per pyink rules.

**2. mypy** — `event.long_running_tool_ids` is `set[str] | None`, needs a null guard before `in`.

**3. 5 test failures** — the `function_response_ids` comprehension counts ALL function_responses as resolutions, but when a `LongRunningFunctionTool` executes, it auto-generates a function_response with `author=<agent_name>`. This makes the method think the long-running call is already resolved, so the agent doesn't pause.

Fix: filter to `event.author == 'user'` so only actual user resume responses count as resolutions:

```python
function_response_ids = {
    function_response.id
    for event in events
    for function_response in event.get_function_responses()
    if function_response.id and event.author == 'user'
}
```

After these fixes: **5,158 tests pass** (including the 5 previously failing pause/resume tests + your 3 new tests).

tottenjordan:
@zeel2104 The CLA check is failing because my commit had a `Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>` trailer — the CLA bot requires all authors/co-authors to have signed the CLA.

I've force-pushed an amended commit to my branch (`tottenjordan:fix-5064-long-running-resume-ci-fixes`) with the co-author line removed. To pick up the clean version, you can reset and re-merge:

```bash
git reset --hard a7b2b73  # back to before the merge
git pull https://github.com/tottenjordan/adk-python.git fix-5064-long-running-resume-ci-fixes
git push --force
```

Or just squash everything when merging to main — that would also resolve the CLA issue.

zeel2104:
@tottenjordan 
Updated the branch to remove the CLA-blocking & pulled clean CI follow-up fix. The branch is ready for maintainer workflow approval / review

tottenjordan:
@rohityan - have we addressed all the tests needed for approval?

rohityan:
Hi @zeel2104 , Please fix formatting errors by running autoformat.sh

zeel2104:
@rohityan 
I reran formatting locally and verified `pyink` passes. The remaining workflows now appear to be awaiting maintainer approval.

xuanyang15:
Hi @zeel2104, Thanks for creating this PR. The issue 2 (Streaming partial and final model events could end up with different ADK-generated function call IDs, causing resume-time lookup of the original function call event to fail.) should already be fixed by [commit/fe4181718d104843b974417c59203ed8a7b15255](https://github.com/google/adk-python/commit/fe4181718d104843b974417c59203ed8a7b15255),  do you mind removing the change for issue 2 and focus on issue 1 in this PR?

zeel2104:
Hi @xuanyang15, thanks for pointing that out. I’ve removed the issue 2 changes from this PR and kept it focused on issue 1 only.

The PR now only updates long-running tool resume handling so matching user functionResponse IDs correctly unblock resumable flows, with targeted tests covering the resolved and unresolved cases.

xuanyang15:
@zeel2104 Sorry for the late reply! ADK was migrating to 2.0 (and one of the internal reviewer was suggesting to hold this on for 2.0 because there is similar logic in 2.0), since the main branch has changed to 2.0 now, do you mind helping rebase your PR to 2.0?

xuanyang15:
Thanks a lot for this fix, @zeel2104 — the `has_unresolved_long_running_tool_calls`
approach (collecting all `functionResponse` ids and checking each paused call) is
clean and could be more robust than what currently sits in `main`.

Since this PR was opened (v1.27.3), the resume path has been substantially
rewritten (v2.5.0), and the underlying bug from #5064 is now fixed on `main`:
`_run_one_step_async` in `base_llm_flow.py` performs a function-call/response id
resolution check before pausing, and the streaming function-call id is now stable
across partial/final events. As a result this PR no longer applies cleanly (it's
currently marked as conflicting).

Given the bug is resolved and this branch has drifted out of sync, I'm going to
close this PR. The one thing worth carrying forward is the "scan all events"
idea here vs. the current 2-event window (see `base_llm_flow.py:995-997`) — if
you're up for a fresh PR that swaps the window for that approach, it'd be a
welcome hardening. Really appreciate the contribution and the thorough testing on
the issue thread.

PR #5091, opened 2026-04-01
fix: harden SetModelResponseTool fallback to prevent infinite loops

vietnamesekid:
## Summary

This is a follow-up to #5057. It improves the fallback behavior of `SetModelResponseTool` to avoid infinite loops, especially when flash models (like gemini-2.5-flash and gemini-3-flash) ignore `set_model_response` and keep calling other tools.

The changes come from the investigation and discussion in #5054.

### Changes

- Stronger instructions for simple types (`_output_schema_processor.py`): primitive schemas like `str` and `int` are easy for models to ignore, so we now give clearer guidance in these cases.
- Deterministic tool restriction: on the second-to-last round (N-1), we force the model to only call `set_model_response` using `tool_config`, so we can guarantee structured output.
- Hard cutoff: at round N (`_MAX_TOOL_ROUNDS=25`), we stop execution completely to avoid runaway loops and unnecessary API usage.
- Early return after success (`base_llm_flow.py`): once `set_model_response` succeeds, we skip extra steps like `transfer_to_agent`.

## Test plan

- All unit tests pass (19/19), including 6 new tests
- All integration tests pass (4/4), covering BaseModel and `str` schemas across GOOGLE_AI and Vertex AI
- No regressions in existing tests
- Formatting and lint checks pass

## Related

- Builds on #5057
- Addresses remaining infinite loop issues from #5054, #4868, and #4525


rohityan:
Hi @vietnamesekid , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

vietnamesekid:
Any updates on this @rohityan ?

PR #5111, opened 2026-04-02
fix(artifacts): validate user_id and session_id in FileArtifactService path construction

### LINKED ISSUE #5110: FileArtifactService: user_id and session_id not validated for path traversal
`_resolve_scoped_artifact_path` validates the filename against traversal, but `_base_root` and `_session_artifacts_dir` use `user_id` and `session_id` directly in `Path()` construction without the same check.

A `user_id` or `session_id` containing `../` segments builds a scope root outside `root_dir`. The filename guard then validates against the already-escaped scope root, so it doesn't catch the escape.

`GcsArtifactService` and `InMemoryArtifactService` use string keys, not filesystem paths - only `FileArtifactService` is affected.

vnykmshr:
Fixes #5110.

`_base_root` and `_session_artifacts_dir` pass `user_id` and `session_id` directly into `Path()` construction. The existing `_resolve_scoped_artifact_path` guards the filename, but the scope root itself is built from these unvalidated values - so traversal segments escape `root_dir` before the filename check runs.

Adds `_validate_path_segment`, using the same `resolve(strict=False)` + `relative_to()` pattern from `_resolve_scoped_artifact_path`, and applies it in both functions.

## Testing plan

7 new test cases: user_id traversal (3 parametrized), session_id traversal (3 parametrized), delete-path traversal (confirms `shutil.rmtree` unreachable with crafted session_id). FILE service only - GCS and InMemory are not affected.

```text
47 passed (artifacts)
5165 passed, 1 skipped (full suite)
```


vnykmshr:
Superseded by cbcb5e6. Thanks for the review flag on Apr 12 @surajksharma07

Issue #5169, opened 2026-04-06
bug: StreamingMode.SSE breaks Interactions API function calling — interaction_id not propagated to session events


gamepop:
## Description

When using `Gemini(use_interactions_api=True)` with `RunConfig(streaming_mode=StreamingMode.SSE)`, the ReAct loop (function_call → execute tool → function_result → resume) fails because `interaction_id` from the first LLM response is not saved to session events. The same code works with `StreamingMode.NONE`.

## ADK Version

- google-adk: 1.28.1
- google-genai: 1.70.0
- Python: 3.14.2

## Reproduction

```python
from google.adk.agents import LlmAgent
from google.adk.models import Gemini
from google.adk.runners import InMemoryRunner
from google.adk.agents.run_config import RunConfig, StreamingMode
from google.genai import types
import asyncio

def get_weather(city: str) -> dict:
    """Get weather for a city."""
    return {"city": city, "temp": 72, "condition": "Sunny"}

agent = LlmAgent(
    model=Gemini(model="gemini-2.5-flash", use_interactions_api=True),
    name="test_agent",
    instruction="Use get_weather when asked about weather. Be concise.",
    tools=[get_weather],
)

async def test():
    runner = InMemoryRunner(agent=agent, app_name="test")
    session = await runner.session_service.create_session(
        user_id="user1", app_name="test"
    )

    # WORKS
    # run_config = RunConfig(streaming_mode=StreamingMode.NONE)

    # FAILS
    run_config = RunConfig(streaming_mode=StreamingMode.SSE)

    async for event in runner.run_async(
        user_id="user1",
        session_id=session.id,
        new_message=types.Content(
            role="user", parts=[types.Part(text="Weather in Tokyo?")]
        ),
        run_config=run_config,
    ):
        if event.get_function_calls():
            for fc in event.get_function_calls():
                print(f"Tool: {fc.name}({fc.args})")
        if (event.content and event.content.parts
            and not event.partial and event.author != "user"):
            for p in event.content.parts:
                if p.text:
                    print(f"Response: {p.text[:100]}")

asyncio.run(test())
```

## Expected Behavior

With `StreamingMode.SSE`, the agent should call `get_weather`, receive the result, and stream the text response with partial events — same behavior as `StreamingMode.NONE` but with token-level streaming.

## Actual Behavior

The function_call is generated and the tool executes, but the function_result round-trip fails:

```
google.genai._interactions.BadRequestError: event: error
[...]

PR #5177, opened 2026-04-06
 fix interactions SSE ids for function calls 

kshirajahere:
## What changed
- preserve interaction chain IDs in Interactions API SSE mode by reading the SDK's actual streaming fields (`interaction.id`, `interaction_id`, `event_id`) instead of only `event.id`
- treat `interaction.complete` as the final full interaction event during streaming conversion
- add regression coverage for `interaction.complete` handling and for SSE function-call streams that must retain the interaction ID before the tool round-trip

## Why
Issue #5169 reports that `StreamingMode.SSE` breaks Interactions API function calling because the second model call cannot find the previous `interaction_id` in session history.

The root cause is that ADK's SSE adapter was keyed to older/generic event shapes (`event.id` and `event_type == "interaction"`), while `google-genai` 1.70.0 emits `interaction.start`, `interaction.complete`, and `interaction.status_update` events with IDs stored on `interaction.id`, `interaction_id`, and `event_id`.

That mismatch meant streamed function-call events were persisted without an `interaction_id`, so the next round of the ReAct loop could not chain the interaction and failed with `invalid_request`.

## Impact
- restores function-calling flows for `Gemini(use_interactions_api=True)` when SSE streaming is enabled
- keeps the interaction chain intact in session events so follow-up tool/result turns can resume correctly
- improves compatibility with the current GenAI SDK SSE event schema without changing non-streaming behavior

## Validation
- `.venv\\Scripts\\pytest tests\\unittests\\models\\test_interactions_utils.py tests\\unittests\\flows\\llm_flows\\test_interactions_processor.py tests\\unittests\\models\\test_google_llm.py -q`
- `.venv\\Scripts\\pyink --check --config pyproject.toml src\\google\\adk\\models\\interactions_utils.py tests\\unittests\\models\\test_interactions_utils.py`
- `.venv\\Scripts\\isort --check-only src\\google\\adk\\models\\interactions_utils.py tests\\unittests\\models\\test_interactions_utils.py`
- `.venv\\Scripts\\pylint --rcfile=pylintrc src\\google\\adk\\models\\interactions_utils.py`


kshirajahere:
I validated the latest head `a36d829` locally and the changed-file formatting checks plus `mypy-diff` against current `main` are clean. The SSE regression tests also pass locally (`128 passed` across the touched test files).
Can u please approve the workflow for tests
@rohityan

rohityan:
Hi @kshirajahere , Thank you for your contribution through this pull request!  This PR has merge conflicts that require changes from your end. Could you please rebase your branch with the latest main branch to address these? Once this is complete, please let us know so we can proceed with the review.

rohityan:
Hi @kshirajahere , closing this since it has been fixed in this commit https://github.com/google/adk-python/commit/9a1930407a4eff67093ea9f14292f1931631a661

PR #5181, opened 2026-04-07
fix: preserve JSON-native types in A2A metadata serialization

### LINKED ISSUE #5183: A2A _serialize_value() corrupts JSON-native metadata types to strings
## Bug Description

`_serialize_value()` in `from_adk_event.py` falls through to `str()` for JSON-native Python types (`dict`, `list`, `bool`, `int`, `float`), corrupting structured metadata values into their string representations during A2A event conversion.

## Steps to Reproduce

```python
from google.adk.a2a.converters.from_adk_event import _serialize_value

metadata = {"retry": True, "max_attempts": 3, "tags": ["alpha", "beta"]}
result = _serialize_value(metadata)
print(type(result), result)
# Actual: <class 'str'> "{'retry': True, 'max_attempts': 3, 'tags': ['alpha', 'beta']}"
# Expected: <class 'dict'> {'retry': True, 'max_attempts': 3, 'tags': ['alpha', 'beta']}
```

## Expected Behavior

JSON-native types (`dict`, `list`, `str`, `bool`, `int`, `float`) should be preserved as-is since they are already valid JSON values, not converted to strings.

## Actual Behavior

All non-Pydantic values that aren't `None` are passed through `str()`, which corrupts structured data into Python repr strings.

YizukiAme:
Fixes #5183

## Problem

`_serialize_value()` in `from_adk_event.py` falls through to `str()` for JSON-native Python types (`dict`, `list`, `bool`, `int`, `float`), corrupting structured metadata values into their string representations during A2A event conversion.

For example, a `custom_metadata` dict like `{"retry": True, "max_attempts": 3}` gets serialized as `"{'retry': True, 'max_attempts': 3}"` — a Python repr string instead of a proper JSON object.

## Fix

- **Preserve JSON-native types directly**: `dict`, `list`, `str`, `bool`, `int`, `float` are now returned as-is since they are already valid JSON values.
- **JSON normalization fallback**: For other JSON-serializable objects (e.g., custom classes with `__dict__`), normalize via `json.loads(json.dumps(value))` before falling back to `str()`.
- **Added regression test** for structured `custom_metadata` round-trip to ensure values are not corrupted to strings.

## Logs After Fix

```
$ uv run pytest tests/unittests/a2a/converters/test_event_round_trip.py -v

tests/unittests/a2a/converters/test_event_round_trip.py::test_custom_metadata_structured_roundtrip PASSED

========== 143 passed, 124 warnings in 9.12s ==========
```

## Testing Plan

```bash
uv run pytest tests/unittests/a2a/converters/ -v
```

All 143 A2A converter tests pass. The new `test_custom_metadata_structured_roundtrip` test verifies that `dict`, `list`, `bool`, `int`, and `float` values survive the `_serialize_value()` → metadata round-trip without being corrupted to strings.

### DIFF
--- a/src/google/adk/a2a/converters/from_adk_event.py
+++ b/src/google/adk/a2a/converters/from_adk_event.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 from collections.abc import Callable
 from datetime import datetime
 from datetime import timezone
+import json
 import logging
 from typing import Any
 from typing import Dict
@@ -265,6 +266,14 @@ def _serialize_value(value: Any) -> Optional[Any]:
     except Exception as e:
       logger.warning("Failed to serialize Pydantic model, falling back: %s", e)
       return str(value)
+
+  if isinstance(value, (dict, list, str, bool, int, float)):
+    return value
+
+  try:
+    return json.loads(json.dumps(value))
+  except (TypeError, ValueError):
+    pass
 
   return str(value)
 
--- a/src/google/adk/a2a/converters/from_adk_event.py
+++ b/src/google/adk/a2a/converters/from_adk_event.py
@@ -267,11 +267,24 @@ def _serialize_value(value: Any) -> Optional[Any]:
       logger.warning("Failed to serialize Pydantic model, falling back: %s", e)
       return str(value)
 
-  if isinstance(value, (dict, list, str, bool, int, float)):
+  # JSON-native leaf types — return as-is
+  if isinstance(value, (str, bool, int, float)):
     return value
 
+  # Containers — recurse so nested non-serializable values are handled
+  if isinstance(value, dict):
+    return {str(k): _serialize_value(v) for k, v in value.items()}
+
+  if isinstance(value, (list, tuple)):
+    return [_serialize_value(item) for item in value]
+
+  # Common Python types with no JSON equivalent
+  if isinstance(value, (set, frozenset)):
+    return [_serialize_value(item) for item in sorted(value, key=str)]
+
+  # Other objects — try JSON normalization, then str() fallback
   try:
-    return json.loads(json.dumps(value))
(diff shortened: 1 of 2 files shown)


src/google/adk/a2a/converters/from_adk_event.py
```diff
@@ -266,6 +267,14 @@ def _serialize_value(value: Any) -> Optional[Any]:
       logger.warning("Failed to serialize Pydantic model, falling back: %s", e)
       return str(value)
 
+  if isinstance(value, (dict, list, str, bool, int, float)):
```
rohityan: What happens when a non serializable object is inside the dictionary like a (datetime, set)?

rohityan:
Hi @YizukiAme , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you address my review comment.

src/google/adk/a2a/converters/from_adk_event.py
YizukiAme: Good catch! Updated to recursive serialization. dict/list values now recurse through _serialize_value(), so nested datetime falls back to str() and set converts to a sorted list. Also added a regression test covering this exact scenario.

rohityan:
Hi @YizukiAme , Please fix formatting errors by running autoformat.sh

YizukiAme:
> Hi @YizukiAme , Please fix formatting errors by running autoformat.sh

Hi, I’m currently away from my computer. I’ll fix the formatting issues in a few hours once I’m back! Thanks for pointing that out~

YizukiAme:
Done! Rebased on latest main and ran autoformat.sh. Should be clean now.

GWeale:
Appreciate this! Preserving JSON-native types in A2A metadata serialization is now handled in main, so this is covered. Closing as superseded, thanks so much for the work.

PR #5182, opened 2026-04-07
fix: fail fast when LiteLLM file_uri MIME type cannot be determined

### LINKED ISSUE #5184: LiteLLM silently uses application/octet-stream when file_uri MIME type cannot be determined
## Bug Description

## Steps to Reproduce

```python
))
```

## Expected Behavior

## Actual Behavior

YizukiAme:
Fixes #5184

## Problem

When `file_data.file_uri` has no explicit MIME type and the URI has no recognizable extension, the code silently falls back to `application/octet-stream`. This causes downstream issues with providers (e.g., Vertex AI) that require an accurate MIME type for GCS URIs — the generic fallback leads to incorrect file handling without any warning.

## Fix

- **Fail fast with actionable error**: Instead of silently using `application/octet-stream`, raise a `ValueError` with a clear message suggesting the user set `file_data.mime_type` explicitly. This only triggers when a file block actually needs to be constructed — provider-specific text fallback paths (e.g., Anthropic's `[File reference: ...]`) execute first and are unaffected.
- **Preserve model context in mixed content**: Fixed `_content_to_message_param()` to pass `model` through recursive calls for mixed `function_response` + file content, preventing incorrect provider fallback on Vertex AI / Gemini.
- **Removed `_DEFAULT_MIME_TYPE`** constant since the silent fallback is no longer used.

## Logs After Fix

```
$ uv run pytest tests/unittests/models/test_litellm.py -v -k "file_uri or mime or function_response_with_extra_parts"

tests/unittests/models/test_litellm.py::test_file_data_with_file_uri_explicit_mime PASSED
tests/unittests/models/test_litellm.py::test_file_data_with_file_uri_inferred_mime PASSED
tests/unittests/models/test_litellm.py::test_file_data_with_file_uri_no_mime_raises PASSED
tests/unittests/models/test_litellm.py::test_function_response_with_extra_parts_preserves_model PASSED
...

========== 42 passed, 206 deselected in 6.62s ==========
```

## Testing Plan

```bash
uv run pytest tests/unittests/models/test_litellm.py -v -k "file_uri or mime or function_response_with_extra_parts"
```

All 42 related tests pass. Key coverage:
- `test_file_data_with_file_uri_no_mime_raises`: Verifies `ValueError` when MIME cannot be determined and a file block is needed
- `test_file_data_with_file_uri_explicit_mime`: Explicit MIME still works
- `test_file_data_with_file_uri_inferred_mime`: Extension-based inference still works
- `test_function_response_with_extra_parts_preserves_model`: Mixed content recursive calls now preserve `model` parameter

### DIFF
--- a/src/google/adk/a2a/executor/a2a_agent_executor.py
+++ b/src/google/adk/a2a/executor/a2a_agent_executor.py
@@ -260,17 +260,15 @@ class A2aAgentExecutor(AgentExecutor):
             context.context_id,
             self._config.gen_ai_part_converter,
         ):
-          a2a_event = await execute_after_event_interceptors(
+          a2a_events = await execute_after_event_interceptors(
               a2a_event,
               executor_context,
               adk_event,
               self._config.execute_interceptors,
           )
-          if a2a_event is None:
-            continue
-
-          task_result_aggregator.process_event(a2a_event)
-          await event_queue.enqueue_event(a2a_event)
+          for e in a2a_events:
+            task_result_aggregator.process_event(e)
+            await event_queue.enqueue_event(e)
 
     # publish the task result event - this is final
     if (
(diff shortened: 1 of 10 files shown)


rohityan:
Hi @YizukiAme , Thank you for your contribution! We appreciate you taking the time to submit this pull request. 
By replacing the silent fallback (application/octet-stream) with a ValueError, any existing developer code that accidentally relied on that fallback—and happens to be using a permissive provider that accepts it—will suddenly start crashing in production. Closing this here since it would be a breaking change.

PR #5191, opened 2026-04-07
fix: concatenate list values in deep_merge_dicts during parallel tool call merge

### LINKED ISSUE #5190: Parallel tool calls: state_delta list values silently overwritten during merge
## Description

When multiple tool calls run in parallel and each writes to the same `state_delta` key containing a **list** value, `merge_parallel_function_response_events` silently drops all but the last value.

## Root Cause

`deep_merge_dicts` in `flows/llm_flows/functions.py` (line ~822) only recurses into **dict** values. Lists hit the `else` branch and get overwritten:

```python
def deep_merge_dicts(d1: dict, d2: dict) -> dict:
    for key, value in d2.items():
        if key in d1 and isinstance(d1[key], dict) and isinstance(value, dict):
            d1[key] = deep_merge_dicts(d1[key], value)
        else:
            d1[key] = value  # <-- lists are overwritten here
    return d1
```

## Reproduction

1. Create a tool that appends to a list in `tool_context.state`:
```python
def my_tool(tool_context, item):
    items = tool_context.state.get("items") or []
    items = list(items)
    items.append(item)
    tool_context.state["items"] = items
```
2. Have the LLM call this tool multiple times in a single response (parallel execution via `asyncio.gather` in `handle_function_calls_live`)
3. Each parallel call gets its own `ToolContext` with a separate `state_delta`
4. Tool A's delta: `{"state_delta": {"items": ["a"]}}`
5. Tool B's delta: `{"state_delta": {"items": ["b"]}}`
6. After merge: `{"state_delta": {"items": ["b"]}}` — item "a" is lost

## Impact

Any application that accumulates list state across parallel tool calls loses data. In our case, drafting multiple emails in a single agent turn results in only ~2 out of 10 emails persisting.

## Suggested Fix

`deep_merge_dicts` should concatenate lists instead of overwriting:

```python
def deep_merge_dicts(d1: dict, d2: dict) -> dict:
    for key, value in d2.items():
        if key in d1 and isinstance(d1[key], dict) and isinstance(value, dict):
            d1[key] = deep_merge_dicts(d1[key], value)
        elif key in d1 and isinstance(d1[key], list) and isinstance(value, list):
            d1[key] = d1[key] + value
        else:
            d1[key] = value
    return d1
```

## Environment

- `google-adk` 1.18.0
- Python 3.12
- Using `DatabaseSessionService` with `LiteLLM` + OpenAI models

enjoykumawat:
## Summary

- **Bug:** When multiple tool calls run in parallel and each writes to the same `state_delta` key containing a list value, `deep_merge_dicts` silently drops all but the last value because lists hit the `else` branch and get overwritten.
- **Fix:** Add a list-type check in `deep_merge_dicts` so list values are concatenated (`d1[key] + value`) instead of overwritten, preserving all entries from parallel function responses.
- **Tests:** Added 5 unit tests covering list concatenation, scalar overwrite, nested dict merge, mixed-type handling, and an integration test for `merge_parallel_function_response_events` with `state_delta` lists.

## Reproduction

```python
# Tool A's delta: {"state_delta": {"items": ["a"]}}
# Tool B's delta: {"state_delta": {"items": ["b"]}}
# Before fix: {"state_delta": {"items": ["b"]}} — item "a" is lost
# After fix:  {"state_delta": {"items": ["a", "b"]}} — both preserved
```

## Test plan

- [x] All 5 new unit tests pass (`test_deep_merge_dicts_*` and `test_merge_parallel_function_response_events_merges_state_delta_lists`)
- [x] All 32 existing tests in `test_functions_simple.py` pass (0 regressions)
- [ ] Manual: create parallel tool calls that accumulate list state and verify no data loss

Fixes #5190


tests/unittests/flows/llm_flows/test_functions_simple.py
wuliang229, line 1: Please move the tests to test_functions_parallel.py

enjoykumawat:
Done! Moved all the `deep_merge_dicts` and `state_delta` list tests to `test_functions_parallel.py` in commit 4b3db55. Also merged latest `main` to pick up any upstream changes. CI is re-running now.

enjoykumawat:
Hi @wuliang229, just following up — the tests have been moved to `test_functions_parallel.py` as requested (commit 4b3db55) and CI is passing. Would you have a chance to re-review when you get a moment? Happy to make any additional changes if needed.

enjoykumawat:
Hi @GWeale — I see this was reopened after the Copybara merge (commit `1ff84eb` is already in main). Is there a follow-up you'd like from me, or was this reopened for internal tracking? Happy to help.

enjoykumawat:
Quick question before this gets another look — digging into the Copybara history here, I see this was actually merged internally (`1ff84eb`, 2026-06-22) and then reverted about 4 hours later by an internal commit (`fda23474`, same day), with no reason given in either commit message. The PR was then reopened.

Was there a concern found internally with this approach (the list-concatenation logic in `deep_merge_dicts`, or something else) that led to the revert? I've since rebased this onto current `main` — the bug is still present there (list values in `state_delta` still get silently overwritten instead of concatenated) — and all tests pass, but I'd rather understand what went wrong the first time than have the same thing happen again. Happy to adjust the approach if there's a reason this shouldn't land as-is.

PR #5199, opened 2026-04-07
fix: guard against None converter results in RemoteA2aAgent

### LINKED ISSUE #5187: RemoteA2aAgent handlers crash on None converter results (AttributeError)
## Bug Description

In `remote_a2a_agent.py`, both `_handle_a2a_response()` (legacy) and `_handle_a2a_response_v2()` access attributes like `event.content` or `event.custom_metadata` on results from `convert_a2a_message_to_event()` / `convert_a2a_task_to_event()` without checking for `None`. This results in `AttributeError` when a converter legitimately returns `None`.

Additionally, the legacy handler still uses the outdated `event_converter.py` path instead of the newer `to_adk_event.py` converters, and `update.context_id` is ignored in favor of `task.context_id` only.

## Steps to Reproduce

This can be triggered when a remote A2A agent sends:
- A message with no convertible parts  
- A task snapshot with no artifacts and no status message
- A status update where the message converter returns `None`

## Expected Behavior

Handlers should guard against `None` converter results with early returns.

## Affected Files

- `src/google/adk/agents/remote_a2a_agent.py`

enjoykumawat:
## Summary

Fixes #5187

- **Problem:** `RemoteA2aAgent._handle_a2a_response` and `_handle_a2a_response_v2` crash with `AttributeError` when converter functions (`convert_a2a_task_to_event`, `convert_a2a_message_to_event`, or config-based converters) return `None`. The code immediately accesses `.custom_metadata` on the result without checking for `None`.
- **Fix:** Added `None`-checks after every converter call in both handler methods. When a converter returns `None`, the handler now gracefully returns `None` (skip the event) instead of crashing.
- **Tests:** Added `TestHandleNoneConverterResults` with three test cases covering the `None` return path for message converters and task converters in both v1 and v2 handlers.

## Test plan

- [x] `python -m pytest tests/unittests/agents/test_remote_a2a_agent.py -v` -- all 98 tests pass
- [x] New tests specifically verify that `_handle_a2a_response` returns `None` when `convert_a2a_message_to_event` returns `None`
- [x] New tests verify that `_handle_a2a_response` returns `None` when `convert_a2a_task_to_event` returns `None`
- [x] New tests verify that `_handle_a2a_response_v2` returns `None` when `a2a_message_converter` returns `None`

### DIFF
--- a/src/google/adk/agents/remote_a2a_agent.py
+++ b/src/google/adk/agents/remote_a2a_agent.py
@@ -476,6 +476,8 @@ class RemoteA2aAgent(BaseAgent):
           event = convert_a2a_message_to_event(
               update.status.message, self.name, ctx, self._a2a_part_converter
           )
+          if not event:
+            return None
           if event.content is not None and update.status.state in (
               TaskState.submitted,
               TaskState.working,
@@ -501,6 +503,8 @@ class RemoteA2aAgent(BaseAgent):
           # for now.
           return None
 
+        if not event:
+          return None
         event.custom_metadata = event.custom_metadata or {}
         event.custom_metadata[A2A_METADATA_PREFIX + "task_id"] = task.id
         if task.context_id:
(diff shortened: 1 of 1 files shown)


rohityan:
Hi @enjoykumawat , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

PR #5206, opened 2026-04-08
fix: validate user_id and session_id against path traversal in FileArtifactService

### LINKED ISSUE #5110: FileArtifactService: user_id and session_id not validated for path traversal

enjoykumawat:
## Summary

- Add `_validate_path_segment()` to reject path separators and traversal sequences in `user_id` and `session_id`
- Only `FileArtifactService` affected — GCS and InMemory use string keys

## Root Cause

`_resolve_scoped_artifact_path` validates filenames but `user_id`/`session_id` were used directly in `Path()` construction. A `user_id` containing `../` could escape `root_dir`.

## Test Plan

- [x] Tests proving `../` in user_id/session_id now raises ValueError
- [x] Normal values still work

Fixes #5110

### DIFF
--- a/src/google/adk/a2a/converters/from_adk_event.py
+++ b/src/google/adk/a2a/converters/from_adk_event.py
@@ -218,6 +218,23 @@ def convert_event_to_a2a_events(
               ),
           )
       )
+    elif _serialize_value(event.actions) is not None:
+      a2a_events.append(
+          TaskStatusUpdateEvent(
+              task_id=task_id,
+              context_id=context_id,
+              status=TaskStatus(
+                  state=TaskState.working,
+                  message=Message(
+                      message_id=str(uuid.uuid4()),
+                      role=Role.agent,
+                      parts=[],
+                  ),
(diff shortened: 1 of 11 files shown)


rohityan:
Hi @enjoykumawat , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix failing mypy-diff tests before we can proceed with a review.

rohityan:
Hi @enjoykumawat , closing this since it has been fixed in this commit. https://github.com/google/adk-python/commit/cbcb5e6002b5bae89de5309caf7b9bb02d563cfc

PR #5207, opened 2026-04-08
fix: warn on silent MIME type fallback and propagate model in recursive calls

### LINKED ISSUE #5184: LiteLLM silently uses application/octet-stream when file_uri MIME type cannot be determined

enjoykumawat:
## Summary

- Upgrade `logger.debug` to `logger.warning` when falling back to `application/octet-stream` for file URIs without determinable MIME type
- Fix `_content_to_message_param()` to propagate `model` parameter in recursive calls for mixed function_response + content parts

## Root Cause

1. Silent `application/octet-stream` fallback with only a debug log caused incorrect file handling on Vertex AI
2. Missing `model=model` in recursive call lost provider-specific behavior (e.g., Anthropic thinking blocks)

## Test Plan

- [x] Test verifying warning is logged on MIME fallback
- [x] Test verifying model propagation in recursive calls with mixed parts
- [x] Existing LiteLLM tests pass

Fixes #5184

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -811,6 +811,7 @@ async def _content_to_message_param(
     follow_up = await _content_to_message_param(
         types.Content(role=content.role, parts=non_tool_parts),
         provider=provider,
+        model=model,
     )
     follow_up_messages = (
         follow_up if isinstance(follow_up, list) else [follow_up]
@@ -1091,7 +1092,7 @@ async def _get_content(
       if not mime_type:
         # LiteLLM's Vertex AI backend requires format for GCS URIs.
         mime_type = _DEFAULT_MIME_TYPE
-        logger.debug(
+        logger.warning(
             "Could not determine MIME type for file_uri %s, using default: %s",
             part.file_data.file_uri,
             mime_type,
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @enjoykumawat, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

enjoykumawat:
Hi @rohityan, thanks for the feedback! I've run `autoformat.sh` and pushed the formatting fixes. Ready for review.

enjoykumawat:
Closing as superseded. While rebasing this onto current `main`, I found upstream commit `42eceb3a` (fix: raise ValueError for unsupported MIME types in file_data URI path) replaced the entire silent-fallback code path this PR was warning about — `_get_content` now raises a `ValueError` instead of silently defaulting to `application/octet-stream` and logging, so there's no longer a silent fallback to warn on. The `model=model` recursive-call propagation this PR also fixed is likewise already present in current `main`.

Nothing left of this PR to contribute — thanks for the review time regardless.

PR #5084, opened 2026-03-31
fix: strip tzinfo from datetime for MySQL in create_session

andreikravchenko-oviva:
### Link to Issue or Description of Change

Fixes #5085

When using `DatabaseSessionService` with MySQL, the very first `append_event()` after `create_session()` always fails with:

```
ValueError: The session has been modified in storage since it was loaded.
Please reload the session before appending more events.
```

**Root cause:** `create_session()` produces a timezone-aware `_storage_update_marker` (with `+00:00`), but MySQL `DATETIME(fsp=6)` doesn't store timezone info ([MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/datetime.html)). When `append_event()` reads `update_time` back, it gets a naive datetime — different marker string, strict comparison fails.

The code already strips `tzinfo` for SQLite and PostgreSQL but MySQL was missed:

```python
# Before (line 464):
if is_sqlite or is_postgresql:        # MySQL not included
    now = now.replace(tzinfo=None)

# After:
is_mysql = self.db_engine.dialect.name == _MYSQL_DIALECT
if is_sqlite or is_postgresql or is_mysql:
    now = now.replace(tzinfo=None)
```

### Testing Plan

**Unit Tests:**

- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

Added `mysql` to the existing parametrized dialect test. Removed the test that incorrectly asserted MySQL preserves timezone.

**Manual End-to-End (E2E) Tests:**

Verified marker values with Testcontainers MySQL 8.0:

| | create_session marker | get_session marker | Match? |
|---|---|---|---|
| Before fix | `...T14:39:46.014977+00:00` | `...T14:39:46.014977` | No |
| After fix | `...T14:44:49.388273` | `...T14:44:49.388273` | Yes |

Confirmed that `append_event()` works immediately after `create_session()` without needing a re-fetch workaround.

### Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md) document.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have manually tested my changes end-to-end.
- [x] Any dependent changes have been merged and published in downstream modules.

PR #5209, opened 2026-04-08
fix: strip timezone from datetime for MySQL to prevent stale-writer false positive

### LINKED ISSUE #5085: MySQL: stale-writer false positive on first append_event after create_session
### What happened?

When using `DatabaseSessionService` with MySQL, the very first `append_event()` after `create_session()` fails with:

```
ValueError: The session has been modified in storage since it was loaded.
Please reload the session before appending more events.
```

### Steps to reproduce

1. Create a `DatabaseSessionService` connected to MySQL
2. Call `create_session()`
3. Immediately call `append_event()` on the returned session

### Root cause

`create_session()` creates a timezone-aware datetime (`tz=timezone.utc`) and derives the `_storage_update_marker` from it (includes `+00:00`). But MySQL `DATETIME(fsp=6)` doesn't store timezone info — it silently drops it. When `append_event()` reads `update_time` back from MySQL, it gets a naive datetime (no `+00:00`), produces a different marker string, and the strict string comparison fails.

The code already handles this for SQLite and PostgreSQL by stripping `tzinfo` before writing (line 464 in `database_session_service.py`), but MySQL is not included in the check:

```python
if is_sqlite or is_postgresql:     # MySQL not included
    now = now.replace(tzinfo=None)
```

Per [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/datetime.html): "MySQL converts TIMESTAMP values from the current time zone to UTC for storage, and back from UTC to the current time zone for retrieval. **(This does not occur for other types such as DATETIME.)**"

### Expected behavior

`append_event()` should work immediately after `create_session()` without needing a re-fetch workaround.

### Environment

- google-adk 1.28.0
- MySQL 8.0 with aiomysql
- `PreciseTimestamp` → `DATETIME(fsp=6)`

enjoykumawat:
## Summary

- Add MySQL to the timezone stripping check in `create_session()` alongside SQLite and PostgreSQL
- MySQL `DATETIME(fsp=6)` silently drops timezone info, causing `_storage_update_marker` mismatch on first `append_event()`

## Root Cause

`create_session()` creates a tz-aware datetime (`tz=timezone.utc`) and derives the `_storage_update_marker` from it (includes `+00:00`). MySQL `DATETIME` drops timezone on storage. When `append_event()` reads `update_time` back, it gets a naive datetime, produces a different marker string, and the strict comparison fails with `ValueError: The session has been modified in storage since it was loaded`.

The code already handled this for SQLite and PostgreSQL (line 464) but MySQL was missing.

## Test Plan

- [x] Updated parametrized test to include MySQL in timezone stripping
- [x] Updated "preserves timezone" test to use Spanner instead of MySQL
- [x] All session service tests pass

Fixes #5085

### DIFF
--- a/src/google/adk/sessions/database_session_service.py
+++ b/src/google/adk/sessions/database_session_service.py
@@ -461,7 +461,8 @@ class DatabaseSessionService(BaseSessionService):
       now = datetime.fromtimestamp(platform_time.get_time(), tz=timezone.utc)
       is_sqlite = self.db_engine.dialect.name == _SQLITE_DIALECT
       is_postgresql = self.db_engine.dialect.name == _POSTGRESQL_DIALECT
-      if is_sqlite or is_postgresql:
+      is_mysql = self.db_engine.dialect.name == _MYSQL_DIALECT
+      if is_sqlite or is_postgresql or is_mysql:
         now = now.replace(tzinfo=None)
 
       storage_session = schema.StorageSession(
--- a/tests/unittests/sessions/test_session_service.py
+++ b/tests/unittests/sessions/test_session_service.py
@@ -102,44 +102,47 @@ def test_database_session_service_enables_pool_pre_ping_by_default():
   assert captured_kwargs.get('pool_pre_ping') is True
 
 
-@pytest.mark.parametrize('dialect_name', ['sqlite', 'postgresql'])
+@pytest.mark.parametrize('dialect_name', ['sqlite', 'postgresql', 'mysql'])
 def test_database_session_service_strips_timezone_for_dialect(dialect_name):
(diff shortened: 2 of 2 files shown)


rohityan:
Hi @enjoykumawat , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:
Hi @wukath , can you please review this

haranrk:
Thank you for this contribution, @enjoykumawat! This has been fixed in #5084 / a012bb7542d481538ad9162993c63ad73259b357.

PR #5219, opened 2026-04-09
fix: guard against None converter results in RemoteA2aAgent handlers

### LINKED ISSUE #5187: RemoteA2aAgent handlers crash on None converter results (AttributeError)

giulio-leone:
## Summary

Fixes #5187 — `AttributeError: 'NoneType' object has no attribute 'custom_metadata'` crash in `RemoteA2aAgent` when converter functions return `None`.

## Root Cause

Both `_handle_a2a_response()` (legacy) and `_handle_a2a_response_v2()` (v2) dereference converter results (e.g. `event.custom_metadata`, `event.content`) without checking for `None`. The v2-default converters in `to_adk_event.py` legitimately return `None` via `_create_event()` when there are no convertible parts and no event actions — this is by design for metadata-only updates, empty status changes, etc.

## Fix

Add `if not event: return None` guards after each converter call site:

| Handler | Path | Guard added |
|---------|------|------------|
| Legacy (`_handle_a2a_response`) | Task converter — no status update | ✅ |
| Legacy (`_handle_a2a_response`) | Message converter — status update with message | ✅ |
| Legacy (`_handle_a2a_response`) | Task converter — artifact update | ✅ |
| Legacy (`_handle_a2a_response`) | A2AMessage response | ✅ |
| V2 (`_handle_a2a_response_v2`) | A2AMessage response | ✅ |
| V2 (`_handle_a2a_response_v2`) | Tuple response | Already guarded (line 572) |

Returning `None` is consistent with the existing pattern in the v2 tuple branch and is properly handled by the caller in `_run_async_impl` via `if not event: continue`.

## Testing

- Reproduced both crashes before the fix
- Verified both crashes are resolved after the fix
- Added 8 regression tests in `TestRemoteA2aAgentNoneConverterResults` covering all converter paths in both handlers
- All 103 tests pass (95 existing + 8 new), 0 failures


rohityan:
Hi @giulio-leone , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

PR #5220, opened 2026-04-09
fix: raise ValueError when MIME type cannot be determined for file_uri

### LINKED ISSUE #5184: LiteLLM silently uses application/octet-stream when file_uri MIME type cannot be determined

giulio-leone:
## Summary

Fixes #5184

When `file_data.file_uri` has no explicit MIME type and no recognizable extension, the LiteLLM integration previously fell back silently to `application/octet-stream` via `_DEFAULT_MIME_TYPE`. This broke providers that validate MIME types (e.g. Vertex AI for GCS URIs) without any warning.

### 1. Raise `ValueError` instead of silent fallback (`_get_content`)
- Removed the `_DEFAULT_MIME_TYPE` fallback assignment when MIME type cannot be determined.
- A `ValueError` with actionable guidance is now raised **only** when a file block would actually be constructed (i.e. the provider requires it).
- Provider-specific **text fallback paths** (anthropic, vertex_ai non-gemini, openai/azure non-file-id URIs) continue to work without requiring MIME type — they never construct a file block.

### 2. Propagate `model=` in recursive `_content_to_message_param()` call
- The recursive call for mixed `function_response` + file content now forwards `model=model`, fixing incorrect provider detection (e.g. Vertex AI Gemini vs non-Gemini).

## Tests
- Updated `test_content_to_message_param_user_message_file_uri_without_mime_type` → expects `ValueError`
- Updated `test_get_content_file_uri_default_mime_type` → expects `ValueError`
- Added `test_get_content_file_uri_no_mime_text_fallback_still_works` — verifies text fallback works without MIME
- Added `test_content_to_message_param_recursive_model_propagation` — verifies model forwarding keeps file blocks for vertex_ai/gemini
- Added `test_content_to_message_param_recursive_model_propagation_fallback` — verifies text fallback for vertex_ai non-gemini

All 248 tests pass (3 new, 2 updated).


rohityan:
Hi @giulio-leone , Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Please fix formatting errors by running autoformat.sh

rohityan:
Hi @giulio-leone , Since the requested changes are still pending, I'm going to close this PR. Feel free to reopen once you've had a chance to address the requested changes, or create a new PR. Thanks!

PR #5251, opened 2026-04-10
Harden BigQuery ML tools against SQL injection

petrmarinec:
## Summary

This change hardens the BigQuery ML helper tools exposed by `BigQueryToolset` against SQL injection.

`forecast`, `analyze_contribution`, and `detect_anomalies` previously constructed SQL with direct string interpolation of tool inputs. Because these helpers are exposed through the toolset, attacker-influenced inputs could be propagated into generated SQL, including in `ORDER BY` clauses and model option values.

## What changed

- Parameterize literal ML option values in `forecast`, `analyze_contribution`, and `detect_anomalies` instead of interpolating them into SQL strings.
- Restrict `history_data`, `input_data`, and `target_data` to valid BigQuery table IDs instead of accepting raw SQL statements.
- Validate and quote field paths used by `detect_anomalies` before placing them in `ORDER BY`.
- Add regression tests covering the previously unsafe input patterns.

## Why this approach

BigQuery supports parameterized queries natively, which is the correct defense for literal values controlled by tool inputs.

Table references and `ORDER BY` identifiers cannot be passed as query parameters, so those inputs are now validated as BigQuery table IDs or field paths before use.

## Compatibility

This is an intentional hardening change: these ML helper tools no longer accept raw SQL statements as data sources.

Callers that need arbitrary SQL should use `execute_sql` directly or materialize a view or table first, then pass that table ID to the ML helper.

## Validation

Verified in a fresh Linux environment with test dependencies installed via `pip install -e '.[test]'`:

- `pytest tests/unittests/tools/bigquery/test_bigquery_query_tool.py -q`
  - Result: `95 passed`

Also re-ran the offline regression scenario used to confirm the original issue and verified that the generated SQL no longer contains the previously attacker-controlled fragments.

This PR also adds unit regressions for those cases so future changes do not reintroduce the same issue.

### DIFF
--- a/src/google/adk/integrations/bigquery/query_tool.py
+++ b/src/google/adk/integrations/bigquery/query_tool.py
@@ -16,7 +16,9 @@ from __future__ import annotations
 
 import functools
 import json
+import re
 import types
+from typing import Any
 from typing import Callable
 from typing import Optional
 import uuid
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @petrmarinec , Thank you for your contribution! We appreciate you taking the time to submit this pull request. 
Please fix the failing mypy-diff tests and formatting errors.

rohityan:

xuanyang15:
@shobsi Could you please take a look?

GWeale:
good catch on both of these. the hardening went out in v2.8.0, and the target_data hole you spotted is closed on main in 60d0a9bf. we ended up keeping the subquery form and validating it rather than dropping it, so this branch is heading somewhere different now and no longer merges. closing it out. open a new issue if you see anything like this again.

PR #5286, opened 2026-04-11
fix(flows): prevent double-execution of FunctionTools returning None in thread pool

### LINKED ISSUE #5284: Low Prio Bug: _call_tool_in_thread_pool double-executes sync FunctionTools that return None
In `_call_tool_in_thread_pool()` (`src/google/adk/flows/llm_flows/functions.py`, lines 148-183), `None` is used as a sentinel to distinguish "FunctionTool ran successfully in thread pool" from "non-FunctionTool sync tool, needs async fallback" (e.g. `SetModelResponseTool`). But `None` is also a valid return value, and I believe that any side-effect-only tool with no explicit return statement implicitly returns `None`.

When a sync `FunctionTool`'s underlying function returns `None`, the sentinel check on line 168 (`if result is not None`) fails, and execution falls through to line 183 (`tool.run_async()`), which re-runs the full `run_async` pipeline and invokes the underlying function a second time silently.

Note: this only triggers when `RunConfig.tool_thread_pool_config` is enabled (defaults to `None`/off).

**Steps to Reproduce:**
1. Create a sync `FunctionTool` whose function returns `None` (or has no return statement)
2. Configure `RunConfig` with `tool_thread_pool_config=ToolThreadPoolConfig()` 
3. Run the agent and trigger a tool call
4. Observe the tool function executes twice

**Expected Behavior:**
The tool function should execute exactly once. A `None` return value should be treated as a valid result.

**Observed Behavior:**
The tool function executes twice:
- First via `tool.func(**args_to_call)` on line 160 (inside `run_sync_tool`, in the thread pool)
- Second via `tool.run_async()` on line 183 (the fallback path for non-FunctionTool sync tools), which calls `_invoke_callable(self.func, ...)` internally

The root cause is on lines 160-168. `run_sync_tool()` returns `tool.func(**args)` for FunctionTools (line 160) and `None` for non-FunctionTools (line 163). The check `if result is not None` on line 168 cannot distinguish between the two cases when a FunctionTool's function legitimately returns `None`:

```python
def run_sync_tool():
    if isinstance(tool, FunctionTool):
        # ... preprocessing ...
        return tool.func(**args_to_call)  # line 160: could return None
    else:
        return None  # line 163: sentinel for "didn't run"
```

**Possible fix:** use a dedicated sentinel so `None` is no longer ambiguous.

Koushik-Salammagari:

Closes #5284

### Description

When `RunConfig.tool_thread_pool_config` is enabled, `_call_tool_in_thread_pool` used `None` as a sentinel to distinguish two cases inside `run_sync_tool()`:

1. **FunctionTool ran successfully in thread pool** → return the result
2. **Non-FunctionTool sync tool** (e.g. `SetModelResponseTool`) → can't run in thread pool, fall back to `tool.run_async()`

The problem: `None` is also a valid return value from any `FunctionTool` whose underlying function has no explicit `return` statement (implicit `None`). Any side-effect-only tool (e.g. writing to state, logging, sending a notification) falls into this category.

**Before this fix:**
```python
result = await loop.run_in_executor(executor, lambda: ctx.run(run_sync_tool))
if result is not None:   # ← False when FunctionTool returns None
    return result
# Falls through to:
return await tool.run_async(args=args, tool_context=tool_context)  # ← second execution!
```

**After this fix:**
```python
# Module-level sentinel — can never be confused with a legitimate tool return value
_SYNC_TOOL_RESULT_UNSET = object()

# Inside run_sync_tool():
#   FunctionTool path  → return tool.func(**args)   (may be None)
#   Non-FunctionTool   → return _SYNC_TOOL_RESULT_UNSET

result = await loop.run_in_executor(executor, lambda: ctx.run(run_sync_tool))
if result is not _SYNC_TOOL_RESULT_UNSET:   # ← True even when result is None
    return result
```

### Changes

- `src/google/adk/flows/llm_flows/functions.py`: introduce `_SYNC_TOOL_RESULT_UNSET` sentinel; update `run_sync_tool` and the guard check.
- `tests/unittests/flows/llm_flows/test_functions_thread_pool.py`: consolidate regression tests into a single `@pytest.mark.parametrize` test covering 6 return values.

### Testing Plan

- [x] `test_sync_tool_falsy_return_executes_exactly_once` — parametrized over `None` (implicit), `None` (explicit), `0`, `""`, `{}`, and `False`; verifies each executes exactly once and the sentinel is never confused with a valid return value.
- [x] All 30 thread pool tests pass.
- [x] Only triggered when `RunConfig.tool_thread_pool_config` is set (non-default); zero impact on the default execution path.

### DIFF
--- a/tests/unittests/flows/llm_flows/test_functions_thread_pool.py
+++ b/tests/unittests/flows/llm_flows/test_functions_thread_pool.py
@@ -278,24 +278,38 @@ class TestCallToolInThreadPool:
     ), f'Event loop should have ticked at least 5 times, got {event_loop_ticks}'
 
   @pytest.mark.asyncio
-  async def test_sync_tool_returning_none_executes_exactly_once(self):
-    """Test that a sync FunctionTool returning None is not double-executed.
+  @pytest.mark.parametrize(
+      'return_value,use_implicit_return',
+      [
+          (None, True),   # implicit None (no return statement)
+          (None, False),  # explicit `return None`
+          (0, False),     # falsy int
+          ('', False),    # falsy str
+          ({}, False),    # falsy dict
+          (False, False), # falsy bool
+      ],
+  )
+  async def test_sync_tool_falsy_return_executes_exactly_once(
+      self, return_value, use_implicit_return
+  ):
+    """FunctionTools returning None or other falsy values must execute exactly once.
 
     Regression test for https://github.com/google/adk-python/issues/5284.
-    A FunctionTool whose underlying function returns None (e.g. side-effect-
-    only tools with no explicit return statement) must execute exactly once.
-    Previously the None return was mistaken for the internal sentinel used to
+    Previously, a None return was mistaken for the internal sentinel used to
     signal 'non-FunctionTool, fall back to run_async', causing a second
-    invocation via tool.run_async().
+    invocation. The fix uses an identity-based sentinel so that None and other
+    falsy values (0, '', {}, False) are treated as valid results.
     """
     call_count = 0
 
-    def side_effect_only() -> None:
+    def sync_func():
       nonlocal call_count
       call_count += 1
-      # No return statement — implicit None.
+      if not use_implicit_return:
+        return return_value
+      # implicit None — no return statement
 
-    tool = FunctionTool(side_effect_only)
+    tool = FunctionTool(sync_func)
     model = testing_utils.MockModel.create(responses=[])
     agent = Agent(name='test_agent', model=model, tools=[tool])
     invocation_context = await testing_utils.create_invocation_context(
@@ -308,35 +322,7 @@ class TestCallToolInThreadPool:
 
     result = await _call_tool_in_thread_pool(tool, {}, tool_context)
 
-    assert result is None
-    assert call_count == 1, (
-        f'Tool function executed {call_count} time(s); expected exactly 1.'
-    )
-
-  @pytest.mark.asyncio
-  async def test_sync_tool_returning_explicit_none_executes_exactly_once(self):
-    """Test that explicit None return from FunctionTool is not double-executed."""
-    call_count = 0
-
-    def explicit_none_return() -> None:
-      nonlocal call_count
-      call_count += 1
-      return None  # Explicit None return.
-
-    tool = FunctionTool(explicit_none_return)
-    model = testing_utils.MockModel.create(responses=[])
-    agent = Agent(name='test_agent', model=model, tools=[tool])
-    invocation_context = await testing_utils.create_invocation_context(
-        agent=agent, user_content=''
-    )
-    tool_context = ToolContext(
-        invocation_context=invocation_context,
-        function_call_id='test_id',
-    )
-
-    result = await _call_tool_in_thread_pool(tool, {}, tool_context)
-
-    assert result is None
+    assert result == return_value
     assert call_count == 1, (
         f'Tool function executed {call_count} time(s); expected exactly 1.'
     )
--- a/tests/unittests/flows/llm_flows/test_functions_thread_pool.py
+++ b/tests/unittests/flows/llm_flows/test_functions_thread_pool.py
@@ -281,12 +281,12 @@ class TestCallToolInThreadPool:
   @pytest.mark.parametrize(
       'return_value,use_implicit_return',
       [
-          (None, True),   # implicit None (no return statement)
+          (None, True),  # implicit None (no return statement)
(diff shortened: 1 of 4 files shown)


Koushik-Salammagari:
@googlebot I signed it!

Koushik-Salammagari:
@googlebot I signed it!

Koushik-Salammagari:
@googlebot I signed it!

tests/unittests/flows/llm_flows/test_functions_thread_pool.py
```diff
@@ -277,6 +277,70 @@ def blocking_sleep() -> dict:
...
+
+    result = await _call_tool_in_thread_pool(tool, {}, tool_context)
+
+    assert result is None
+    assert call_count == 1, (
+        f'Tool function executed {call_count} time(s); expected exactly 1.'
+    )
+
+  @pytest.mark.asyncio
+  async def test_sync_tool_returning_explicit_none_executes_exactly_once(self):
```
ruidazeng: The two tests are nearly identical. Can we collapse them into one test or use @pytest.mark.parametrize? I don't have a strong opinion on this.

tests/unittests/flows/llm_flows/test_functions_thread_pool.py
```diff
@@ -277,6 +277,70 @@ def blocking_sleep() -> dict:
...
+        function_call_id='test_id',
+    )
+
+    result = await _call_tool_in_thread_pool(tool, {}, tool_context)
+
+    assert result is None
+    assert call_count == 1, (
+        f'Tool function executed {call_count} time(s); expected exactly 1.'
+    )
+
```
ruidazeng, line 329: Consider adding a falsy-but-not-None test. The bug was specifically about None. A test like this would prove the sentinel doesn't break tools returning 0, "", {}, or False.

tests/unittests/flows/llm_flows/test_functions_thread_pool.py
Koushik-Salammagari: Done — collapsed both tests into a single `@pytest.mark.parametrize` test covering 6 cases: implicit `None`, explicit `None`, `0`, `""`, `{}`, and `False`. All 30 tests pass.

tests/unittests/flows/llm_flows/test_functions_thread_pool.py
Koushik-Salammagari: Added falsy-but-not-None cases (`0`, `""`, `{}`, `False`) as part of the parametrized test. These confirm the sentinel is identity-based (`is not _SYNC_TOOL_RESULT_UNSET`) and correctly treats all falsy return values as valid results without triggering a second execution.

Koushik-Salammagari:
@googlebot I signed it!

rohityan:
Hi @Koushik-Salammagari , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

Koushik-Salammagari:
Hi @ruidazeng, @rohityan — I've addressed both review comments (collapsed the duplicate tests into a single `@pytest.mark.parametrize` covering 6 falsy cases: implicit `None`, explicit `None`, `0`, `""`, `{}`, and `False`) and fixed the formatting via `autoformat.sh`. All CI checks are green. Happy to make any further changes if needed!

Koushik-Salammagari:
Hi @rohityan — all checks are green (CLA ✓, header-check ✓). Both review comments from @ruidazeng have been addressed:

1. Collapsed the duplicate tests into a single `@pytest.mark.parametrize` covering 6 cases (implicit `None`, explicit `None`, `0`, `""`, `{}`, `False`).
2. Formatting fixed via `autoformat.sh`.

All 30 thread-pool tests pass. Ready for re-review whenever you get a chance!

Koushik-Salammagari:
Hi @rohityan — all review comments have been addressed and CI is green. Just checking if there's anything else needed on my end. Happy to make changes if needed!

Koushik-Salammagari:
Hi @rohityan — all checks are green and review comments have been addressed. Checking in to see if there's anything else needed on my end. Happy to make any changes!

Koushik-Salammagari:
Closing this PR — the same bug was independently fixed upstream in commit 78a88518 ("fix: avoid double-execution of sync FunctionTools returning None"), which also closes #5284. The upstream fix restructures the dispatch to return directly from the FunctionTool path, achieving the same result without a sentinel. No further action needed here.

PR #5303, opened 2026-04-13
Prevent Web UI from displaying non-agent directories

### LINKED ISSUE #5299: Web UI Agent Discovery Returns Non-Agent Directories
## 🔴 Web UI Agent Discovery Returns Non-Agent Directories

**Describe the Bug:**

The `/list-apps` endpoint returns all subdirectories in the agents directory, regardless of whether they contain valid agent definitions. This causes the ADK Web UI to display non-agent directories (like `tools/`, `schemas/`, `tests/`) in the agent selector dropdown, resulting in a poor user experience.

**Steps to Reproduce:**

1. Create a project directory with the following structure:
```
my_project/
├── my_agent/
│   ├── __init__.py
│   └── agent.py       # valid agent with root_agent
├── utils/              # not an agent
├── data/               # not an agent
└── tmp/                # not an agent
```

2. Run `adk web .` from `my_project/`
3. Open the Web UI at http://127.0.0.1:8000/
4. Observe the agent selector dropdown shows: `utils`, `data`, `tmp`, and `my_agent`
5. Select `utils` or other non-agent directory → Loading error occurs

**Expected Behavior:**

Only directories containing a valid agent definition (`root_agent.yaml`, `agent.py` with `root_agent`, or `__init__.py` with `root_agent`) should appear in the agent list.

**Observed Behavior:**

All non-hidden subdirectories are listed regardless of whether they contain an agent. Selecting a non-agent directory results in a loading error:
```
ValueError: No root_agent found for 'utils'. Searched in 'utils.agent.root_agent', 'utils.root_agent' and 'utils\root_agent.yaml'.
```

**Environment Details:**

- ADK Library Version: 1.29.0
- Desktop OS: Windows 10 (also affects macOS/Linux)
- Python Version: 3.10

**Model Information:**

- Are you using LiteLLM: N/A
- Which model is being used: N/A (issue is in agent discovery, not model interaction)

---

## 🟡 Optional Information

**Regression:**

No, this has been the behavior since `list_agents()` was introduced. PR #3430 added a `list_agents_detailed()` endpoint that filters properly, but the Web UI frontend never adopted it and still calls the plain `/list-apps` endpoint.

**Logs:**

N/A - no error on the backend, the non-agent directories are simply returned in the API response.

**Screenshots / Video:**

The Web UI dropdown shows non-agent directories alongside valid agents, causing confusion for users.

**Additional Context:**

**Root Cause:**
The issue is in `AgentLoader.list_agents()` which only filters by `os.path.isdir()`, hidden directory prefix, and `__pycache__`. It does not verify the directory contains a recognized agent structure.

**Current Implementation (v1.29.0):**
```python
# google/adk/cli/utils/agent_loader.py
def list_agents(self) -> list[str]:
    base_path = Path.cwd() / self.agents_dir
    agent_names = [
        x
        for x in os.listdir(base_path)
        if os.path.isdir(os.path.join(base_path, x))
        and not x.startswith(".")
        and x != "__pycache__"
    ]
    return agent_names  # Returns ALL directories

# google/adk/cli/adk_web_server.py
@app.get("/list-apps")
async def list_apps(
    detailed: bool = Query(
        default=False,  # ⚠️ Default is False
        description="Return detailed app information"
    )
):
  if detailed:
    return self.agent_loader.list_agents_detailed()  # ✅ Would filter
  return self.agent_loader.list_agents()  # ❌ Returns all directories
```

**The Problem:**
- Backend provides `detailed` parameter to filter invalid agents
- Default value is `False`
- Web UI frontend (pre-compiled Angular app) calls `/list-apps` without passing `?detailed=true`

**Proposed Solutions:**

1. **Change backend default to `True`** (Recommended):
   - Fixes issue for all users immediately
   - No frontend changes required
   - Backward compatible

2. **Update frontend to pass `detailed=true`**:
   - Requires frontend rebuild
   - More explicit control

**Impact on Multi-Agent Systems:**

Modern multi-agent architectures typically have:
- One orchestrator agent (root agent)
- Multiple sub-agents (implemented as modules/classes)
- Utility directories (tools, schemas, tests, etc.)

The current behavior creates friction for this common architecture pattern.

```python
from pathlib import Path
from google.adk.cli.utils.agent_loader import AgentLoader
import tempfile

with tempfile.TemporaryDirectory() as tmp:
    # Create valid agent
    (Path(tmp) / "real_agent").mkdir()
    (Path(tmp) / "real_agent" / "__init__.py").write_text(
        "from google.adk.agents.base_agent import BaseAgent\n"
        "class A(BaseAgent):\n"
        "  def __init__(self): super().__init__(name='a')\n"
        "root_agent = A()"
    )
    
    # Create non-agent directory
    (Path(tmp) / "not_an_agent").mkdir()
    
    loader = AgentLoader(tmp)
    result = loader.list_agents()
    
    print(result)
    # Returns: ['not_an_agent', 'real_agent']
    # Expected: ['real_agent']
```

- Always (100%)

gitsofaryan:
Resolves #5299 

### Description
This PR fixes an issue where the Web UI and underlying `AgentLoader` incorrectly listed non-agent directories (such as `utils/`, `tmp/`, etc.) in the agent discovery selector, resulting in a fractured user experience and `ValueError` loading errors.

### Changes Made
1. **AgentLoader Root Fix:** Updated `AgentLoader.list_agents()` in `src/google/adk/cli/utils/agent_loader.py`. It now explicitly verifies that directories contain a recognized agent definition (`root_agent.yaml`, `agent.py`, or `__init__.py`) before appending them to the results list.
2. **Endpoint Default:** Changed the default value of the `detailed` query parameter in the `/list-apps` FastAPI endpoint (`src/google/adk/cli/adk_web_server.py`) from `False` to `True`. Because `detailed=True` implicitly leverages robust object instantiations, it functions as an immediate stop-gap backward-compatible fix for the Web UI without needing to force frontend compiled rebuilds.
3. **Test Coverage:**
   - Modified `tests/unittests/cli/test_fast_api.py` to ensure legacy assertions still evaluate `?detailed=false` appropriately while validating the newly updated endpoint default. 
   - Appended a new unit test `test_list_agents_ignores_non_agent_directories` to `tests/unittests/cli/utils/test_agent_loader.py` that identically emulates the Minimal Reproduction Code provided in the issue, ensuring we properly exclude irrelevant directories.

### Testing Instructions
1. Run local unittests (`pytest tests/unittests/cli/utils/test_agent_loader.py` & `pytest tests/unittests/cli/test_fast_api.py`).
2. Locally spin up the Web UI using `adk web .` from a directory populated with both valid agent and non-agent subdirectories and verify that only the valid agents populate within the dropdown menu.


rohityan:
Hi @gitsofaryan , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

gitsofaryan:
@rohityan thanks, PTAL!

gitsofaryan:
@rohityan could you please review my PR!

GWeale:
the loader fix landed separately in caac0708, list_agents() skips any directory without agent.py, root_agent.yaml or __init__.py, the same set you had. that shipped in v2.8.0 and the v1 line still lists them, so file a backport request if you need it there. the /list-apps detailed default flip wasn't taken since it changes the response shape for existing callers, worth its own issue.

PR #5307, opened 2026-04-13
feat(events): add turnId to SSE events for per-LLM-turn grouping

### LINKED ISSUE #4793: Add a turn/response group identifier to streaming events
# Add a turn/response group identifier to streaming events

## Problem

When using `runner.run_async()` with `StreamingMode.SSE`, there is no way to distinguish which partial streaming chunks belong to which LLM response turn within a single invocation.

In a typical flow where the LLM responds with text, calls a tool, and then responds with more text, the events look like this:

```
Event(partial=True,  text="Looking up...")      # LLM call 1
Event(partial=True,  text="Looking up info...")  # LLM call 1
Event(partial=False, function_call=...)          # LLM call 1 (final)
Event(function_response=...)                     # Tool result
Event(partial=True,  text="Here are the...")     # LLM call 2
Event(partial=True,  text="Here are the results...")  # LLM call 2
Event(partial=False, text="Here are the results...")  # LLM call 2 (final)
```

All of these events share the same `invocation_id`, and the `id` field changes on every yield (not per turn). There is no field to programmatically determine that the first two partial chunks belong to LLM call 1 and the last two belong to LLM call 2.

## Current workaround

The only way to detect turn boundaries is by observing transition patterns: a `partial=False` event followed by function call/response events implicitly signals the end of one text group. This is fragile and requires the consumer to maintain state tracking.

## Proposed solution

Add a `turn_id` (or `llm_call_id` / `response_group_id`) field to `Event` that:

- Is generated once per LLM call within `_run_one_step_async`
- Remains the same across all partial chunks and the final aggregated event of that LLM call
- Changes when a new LLM call starts (e.g., after tool execution)

This would allow consumers to trivially group streaming chunks by turn:

```python
async for event in runner.run_async(...):
    if event.partial:
        # event.turn_id groups this chunk with others from the same LLM call
        buffer[event.turn_id] += extract_text(event)
```

## Use case

We build a chat UI that streams partial text to the user via WebSocket. When the agent produces text, calls a tool, and then produces more text, we need to render them as separate message bubbles. Without a turn identifier, we cannot cleanly separate the two text blocks on the client side without implementing brittle heuristics based on event type transitions.

## Environment

- ADK version: latest (pip)
- Streaming mode: SSE
- Python 3.12

ilyesJammouci:
Closes: #4793 

# Problem
With runner.run_async() + StreamingMode.SSE, one invocation can contain multiple LLM calls (for example: text → tool call → text).
invocation_id is shared by all events, and id changes on each yield, so clients cannot reliably group partial chunks and final output from the same LLM call.turn.

# Solution
This PR adds an additive turn_id field to Event (serialized as turnId) and propagates it through streaming flow:

1. Generate one turn_id per LLM call step.
2. Keep the same turn_id for partial chunks and final model event of that call.
3. Propagate that turn_id to tool-derived follow-up events.
4. Start a new turn_id when a new LLM call begins.

The first two images come from the same LLM call, so they share the same turnId.
image
image
This image comes from a different LLM call, so it has a different turnId.
image

# Testing

1. Run local unit tests: pytest test_base_llm_flow.py test_base_llm_flow_partial_handling.py test_progressive_sse_streaming.py test_functions_simple.py test_functions_parallel.py -q

2. Locally validate SSE behavior via Web UI:

- Start the app with adk web src/google/adk/agent.
- Open an agent that triggers a multi-step flow (text → tool call → text) with streaming enabled.
- Verify streamed payloads include turnId.
- Confirm all partial and final events from the same LLM call share the same turnId, and that each new LLM call uses a different turnId.


rohityan:
Hi @ilyesJammouci , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

ilyesJammouci:
Hi @rohityan  I fixed the formatting issues . Ready for review

ilyesJammouci:
Hi @rohityan, just a quick follow-up on this PR.
I’ve addressed the formatting issues and applied the requested changes. would appreciate it if you could take another look when you have time. Thanks!

DeanChensj:
Hi @ilyesJammouci,

Thank you for putting this together and for tackling #4793! We really appreciate the detailed PR description, screenshots, and test coverage.
After discussing this internally with the team, we have decided not to add `turn_id` to the core `Event` model, and will be closing this PR. Here is our thinking:
1. **Core Schema & Persistence Scope**: `Event` is the foundational data model across the entire ADK runtime. Adding a new field to `Event` impacts every `SessionService` implementation (`DatabaseSessionService`, `VertexAiSessionService`, GCS, etc.) and introduces serialization/storage migration overhead across all persistent backends.
2. **State & Distributed Consistency**: In distributed deployments, branched workflows, and pause/resume scenarios across different workers and over time, generating and maintaining a globally consistent, reliable turn identifier on every event is fragile and difficult to guarantee.
3. **Layering**: The underlying problem in #4793—grouping streaming SSE chunks within a single invocation for UI rendering—is primarily a transient presentation/streaming-layer concern. We believe this is better addressed via stream utility/helper functions that inspect event boundaries on the consumer side, rather than cementing a new persistent field into the core `Event` schema.
Thank you again for your time and contribution here, and we welcome any ideas or PRs on client-side / stream helper utilities for chunk grouping!

PR #5334, opened 2026-04-14
fix: avoid passing system=None to Anthropic API when no system instruction is set

### LINKED ISSUE #5318: AnthropicLlm.generate_content_async passes system=None to Anthropic API, causing 400 during event compaction
Describe the Bug:
  AnthropicLlm.generate_content_async passes system=None to the Anthropic API when no system instruction is set. The Anthropic API rejects None — it      
  expects a str or a list of content blocks. This causes event compaction to crash every time it fires when using a Claude model.
                                                                                                                                                          
  Steps to Reproduce:
  1. Configure an ADK app with a Claude model (via Vertex AI) and EventsCompactionConfig                                                                  
  2. Run enough user invocations to trigger compaction (e.g. compaction_interval=20)                                                                      
  3. When compaction fires, LlmEventSummarizer creates an LlmRequest with no system instruction (config.system_instruction defaults to None)
  4. AnthropicLlm.generate_content_async passes system=None directly to the Anthropic API → 400 Bad Request                                               
                                                                                                                                                          
  Expected Behavior:                                                                                                                                      
  When system_instruction is None, the system parameter should be omitted from the Anthropic API call (use NOT_GIVEN), not passed as null.                
                                                                                                                                                          
  Observed Behavior:                                                                                                                                      
  anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'system: Input should be a valid   
  list'}}                                                                                                                                                 
  Full traceback:                                                                                                                                         
  File ".../google/adk/apps/compaction.py", line 564, in _run_compaction_for_sliding_window
      compaction_event = await config.summarizer.maybe_summarize_events(events=events_to_compact)                                                         
  File ".../google/adk/apps/llm_event_summarizer.py", line 107, in maybe_summarize_events                                                                 
      async for llm_response in self._llm.generate_content_async(llm_request, stream=False):                                                              
  File ".../google/adk/models/anthropic_llm.py", line 309, in generate_content_async                                                                      
      message = await self._anthropic_client.messages.create(                                                                                             
  anthropic.BadRequestError: Error code: 400 - {'type': 'invalid_request_error', 'message': 'system: Input should be a valid list'}                       
                                                                                                                                                          
  Environment Details:                                                                                                                                    
  - ADK Library Version: 1.26.0 (confirmed present in 1.30.0)
  - Desktop OS: macOS                                                                                                                                     
  - Python Version: 3.13
                                                                                                                                                          
  Model Information:
  - Are you using LiteLLM: No
  - Which model: claude-sonnet-4-6 via Vertex AI (Claude subclass of AnthropicLlm)

  ---                                                                                                                                                     
  Regression: N/A — unable to determine when introduced.
                                                                                                                                                          
  Logs: See stack trace above.
                                                                                                                                                          
  Minimal Reproduction Code:
  from google.adk.apps import App, ResumabilityConfig
  from google.adk.apps.app import EventsCompactionConfig
                                                                                                                                                          
  agent = LlmAgent(                                                                                                                                       
      name="my_agent",                                                                                                                                    
      model="claude-sonnet-4-6",
  )                                              
                                                                                                                                                          
  app = App(      
      name="my_app",
      root_agent=agent,
      events_compaction_config=EventsCompactionConfig(                                                                                                    
          compaction_interval=5,  # low value to trigger quickly
          overlap_size=2,                                                                                                                                 
      ),                                                                                                                                                  
  )     
  # Run 5+ user invocations — compaction fires and crashes with 400                                                                                       
                                                                                                                                                          
  Root cause: LlmEventSummarizer.maybe_summarize_events (llm_event_summarizer.py:102) creates an LlmRequest without setting config, so                    
  config.system_instruction is None. AnthropicLlm.generate_content_async (anthropic_llm.py:311) passes this directly as system=None to the Anthropic SDK, 
  which serialises it as JSON null. The API rejects null because it expects str | list.                                                                   
                                                                                                                                                          
  Suggested fix:  
  # anthropic_llm.py
  from anthropic import NOT_GIVEN
                                 
  system = llm_request.config.system_instruction                                                                                                          
  if system is None:                            
      system = NOT_GIVEN                                                                                                                                  
                        
  message = await self._anthropic_client.messages.create(                                                                                                 
      model=llm_request.model,                           
      system=system,                                                                                                                                      
      ...           
  )      
                                                                                                                                                          
  How often: Always (100%) — every time compaction fires with a Claude model.

geomarceau:
When no system instruction is configured, llm_request.config.system_instruction is None. Passing that directly as the system argument to messages.create triggers a 400 error from the Anthropic API because the API requires the parameter to be absent rather than explicitly null.

The fix computes a local system variable that is either the actual instruction string or NOT_GIVEN, so the parameter is omitted from the request when there is nothing to send. The same value is threaded through to _generate_content_streaming so both the streaming and non-streaming code paths behave consistently.

Fixes #5318

Testing

To reproduce the original bug, create an AnthropicLlm instance and call it without a system instruction. Before this fix the system kwarg was None going into messages.create, which the Anthropic API rejects with a 400.

With the fix in place, the following script confirms the behavior:

    llm = AnthropicLlm(model="claude-3-5-sonnet-20241022")
    request = LlmRequest(model="claude-3-5-sonnet-20241022")
    request.contents = [Content(role="user", parts=[Part(text="Hello")])]
    async for _ in llm.generate_content_async(request):
        pass

Output with fix applied (system kwarg captured from the mock client):

    system value passed to messages.create: NOT_GIVEN
    Is NOT_GIVEN: True
    Is None: False
    PASS: system parameter omitted correctly (NOT_GIVEN), no 400 error will occur

The existing test suite also covers the case where a system instruction is present and verifies it is forwarded correctly.


rohityan:
Hi @geomarceau , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you add a unit test case that that covers the case where generate_content_async is called without a system instruction and  also fix formatting errors by running autoformat.sh.

rohityan:
Hi @geomarceau  closing this since this is a duplicate of #5335 which already has comprehensive unit tests added.

PR #5344, opened 2026-04-15
fix: harden builder upload YAML validation

### LINKED ISSUE #5347: Builder upload YAML validation can be bypassed
The YAML validation in the builder upload endpoint does not fully restrict values that could resolve to arbitrary Python references at agent load time.

PR #5344 hardens the validation to reject these values.

TristanInSec:
## Summary
- Reject fully-qualified (dotted) Python references in builder YAML uploads
- Covers `name`, `agent_class`, and `code` fields
- Simple (non-dotted) names continue to work (ADK built-in tools, built-in agent classes)
- Adds 6 tests covering rejection and allowlisting behavior

Fixes #5347

## Test plan
- [x] Dotted tool name rejected (e.g. `os.system`)
- [x] Dotted agent_class rejected
- [x] Dotted code reference rejected
- [x] Simple tool name allowed (e.g. `google_search`)
- [x] Simple agent_class allowed (e.g. `LlmAgent`)
- [x] Existing builder tests unaffected


TristanInSec:
Created #5347.

rohityan:
Hi @TristanInSec , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

sasha-gitg:
This is only an issue when args is passed which is now blocked when /builder is enabled through web. Additionally web is a developer tool and not intended for production usage.

TristanInSec:
Hi @sasha-gitg,

Thanks for the review. I noticed that since this PR was opened, three commits have landed addressing the same attack surface:

- `6c24ccc9` (Mar 24) — "fix: gate builder endpoints behind web flag"
- `dcee2902` (Apr 2) — "fix: Disallow args on /builder"
- `74f235b1` (Apr 21) — "fix: block RCE vulnerability via nested YAML configurations in ADK"

These commits describe the issue as an RCE vulnerability in their messages and code comments. The current mitigation blocks the `args` key, but the dotted module name vector remains uncovered. For example, a YAML config with `name: os.system` in the tools list will resolve through `importlib.import_module()` and become a callable tool that the LLM can invoke with arbitrary arguments. The `_BLOCKED_YAML_KEYS` denylist does not catch this.

This PR adds validation for dotted references in `name`, `agent_class`, and `code` fields, which would close that remaining gap. Happy to rebase on main if the team would like to reconsider.

Issue #5345, opened 2026-04-15
OpenApiToolkit OAuth Client Credential (M2M) flow - 'NoneType' object has no attribute 'keys'


NicolasCwy:
## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. Requests without complete information may be rejected / deprioritized.
If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**
When using OAuth2 client credentials flow (M2M use case), when OpenAPI Toolkit tries to authenticate itself it fails with error

**Steps to Reproduce:**
Please provide a numbered list of steps to reproduce the behavior:
1. Setup OPENAPI_TOOLSET with auth_scheme, clientCredentials
2. Get agent to invoke any tool

**Expected Behavior:**
Agent will get credentials and then proceed to make authenticated API call

**Observed Behavior:**
Tool call will fail and crash
```python
google/adk/auth/auth_handler.py", line 186, in generate_auth_uri
scopes = list (scopes.keys))
٨٨٨٨٨٨٨٨٨٨٨
AttributeError: 'NoneType' object has no attribute 'keys'
```

**Environment Details:**

 - ADK Library Version (pip show google-adk): v1.30.0
 - Desktop OS:** [e.g., macOS, Linux, Windows]: macOS
 - Python Version (python -V): 3.13.11

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: (e.g., gemini-2.5-pro): GPT-4-o

---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

**Regression:**
Did this work in a previous version of ADK? If so, which one?
- No

**Logs:**
Attached log.txt
```

**Additional Context:**
Add any other context about the problem here.

**Minimal Reproduction Code:**
Please provide a code snippet or a link to a Gist/repo that isolates the issue.
```python
import pathlib

from google.adk import Agent
from google.adk.skills import load_skill_from_dir, models
from google.adk.tools.skill_toolset import SkillToolset
from google.adk.models.lite_llm import LiteLlm
from fastapi.openapi.models import OAuth2, OAuthFlows, OAuthFlowClientCredentials
from google.adk.tools.openapi_tool.openapi_spec_parser.openapi_toolset import (
    OpenAPIToolset,
)
from google.adk.auth import AuthCredential, AuthCredentialTypes, OAuth2Auth

import asyncio
import uuid  # For unique session IDs
from dotenv import load_dotenv

[...]

PR #5352, opened 2026-04-16
fix(auth): omit scope from OAuth2 refresh requests and guard None scopes

### LINKED ISSUE #5328: OAuth2 token refresh fails for providers that reject scope parameter (e.g. Salesforce)
## 🔴 Required Information

**Describe the Bug:**

`OAuth2CredentialRefresher.refresh()` transitively includes the `scope` parameter in refresh token requests because `create_oauth2_session()` passes scopes to authlib's `OAuth2Session` constructor. Salesforce rejects refresh requests that include a `scope` parameter, causing all automatic token refreshes to fail silently for Salesforce-backed toolsets. Other providers with strict RFC 6749 §6 interpretations may be similarly affected.

Per RFC 6749 §6, `scope` is OPTIONAL in refresh requests — when omitted, providers treat it as equal to the originally-granted scope. Since sending `scope` on refresh provides no functional benefit (it can only narrow, not broaden the granted scope) and some providers actively reject it, ADK should default to omitting it for maximum compatibility.

**Steps to Reproduce:**

1. Configure an `OpenAPIToolset` with OAuth2 `authorizationCode` flow and one or more scopes. Use a provider that rejects `scope` in refresh requests (e.g. Salesforce).
2. Complete the initial OAuth authorization flow to obtain a valid access_token + refresh_token.
3. Wait for the access_token to expire (or manually expire it by setting `expires_at=<past-timestamp>` in session state).
4. Invoke a tool — ADK's `ToolAuthHandler._get_existing_credential()` calls `OAuth2CredentialRefresher.refresh()`.
5. Observe: refresh fails with `invalid_request: scope parameter not supported` (Salesforce).

**Expected Behavior:**

The refresh should succeed — ADK should not send `scope` in refresh token requests by default, since the authorized scopes are bound at initial exchange and cannot be changed on most providers.

**Observed Behavior:**

Refresh fails with `Failed to refresh OAuth2 tokens: invalid_request: scope parameter not supported`. The error is caught and logged inside `OAuth2CredentialRefresher.refresh()`, but the original stale credential is silently returned. The tool then executes with the expired access_token and receives a 401.

**Environment Details:**

- ADK Library Version: `google-adk 1.28.0`
- Python Version: Python 3.13

**Model Information:**

- Are you using LiteLLM: No
- Which model is being used: gemini-2.5-flash (irrelevant — this is in the auth path, not the model path)

---

## 🟡 Optional Information

**Logs:**

```text
Failed to refresh OAuth2 tokens: invalid_request: scope parameter not supported
```

**Additional Context:**

Root cause — `create_oauth2_session` in `google/adk/auth/oauth2_credential_util.py`:

```python
return (
    OAuth2Session(
        auth_credential.oauth2.client_id,
        auth_credential.oauth2.client_secret,
        scope=" ".join(scopes),      # ← unconditionally passed
        redirect_uri=auth_credential.oauth2.redirect_uri,
        state=auth_credential.oauth2.state,
        token_endpoint_auth_method=auth_credential.oauth2.token_endpoint_auth_method,
    ),
    token_endpoint,
)
```

authlib's `OAuth2Session.refresh_token()` auto-includes `self.scope` in the refresh request body when `scope` is set on the session:

```python
# authlib source
if "scope" not in kwargs and self.scope:
    kwargs["scope"] = self.scope
```

**Proposed fix** — don't set `scope` on the `OAuth2Session` constructor. Scopes are only used for authorization redirect URLs, which are built separately via `client.create_authorization_url()`:

```python
return (
    OAuth2Session(
        auth_credential.oauth2.client_id,
        auth_credential.oauth2.client_secret,
        redirect_uri=auth_credential.oauth2.redirect_uri,
        state=auth_credential.oauth2.state,
        token_endpoint_auth_method=auth_credential.oauth2.token_endpoint_auth_method,
    ),
    token_endpoint,
)
```

**Secondary issue — workaround hazard:** `oauth2_credential_refresher.py` does `from google.adk.auth.oauth2_credential_util import create_oauth2_session`. This creates a local binding in the refresher module. Consumers attempting a monkey-patch workaround must patch the refresher module's binding, not the source module — otherwise the patch appears to succeed but has no effect. Consider changing to `import oauth2_credential_util` + `oauth2_credential_util.create_oauth2_session(...)` inside `refresh()` to make the reference late-bound and patchable.

**Impact:** Any ADK user integrating with Salesforce via `OpenAPIToolset` will experience silent refresh failures. The user symptom is a full OAuth re-authorization flow on every access_token expiry (typically every 1-2 hours). Other OAuth providers with strict RFC 6749 §6 interpretations of the `scope` parameter may be similarly affected.

- #5329  — Refreshed OAuth2 credentials are not persisted to the credential store

Complete runnable reproduction — mirrors the `contributing/samples/oauth2_client_credentials` layout (`agent.py` + `main.py` + `oauth2_test_server.py` + `README.md`), adapted to use the `authorization_code` flow. The test server is run with `STRICT_SCOPE_REJECTION=1` to mimic Salesforce's refresh behavior. A `--apply-fix` CLI flag monkey-patches the proposed fix so the same script demonstrates both the bug and its resolution:

https://github.com/doughayden/adk-issue-examples/tree/2f454e73c2f1885ebe9be61125e02800cb2164b3/02-scope_in_refresh

Expected output — without `--apply-fix` the tool call's refresh fails (server rejects `scope` in the refresh body); ADK falls through to `_request_credential` and the agent asks for re-auth. With `--apply-fix` the refresh succeeds and the tool returns weather data.

**Workaround applied in our project:**

```python

_original = oauth2_credential_refresher.create_oauth2_session

def _no_scope(auth_scheme, auth_credential):
    session, endpoint = _original(auth_scheme, auth_credential)
    if session is not None:
        session.scope = None
    return session, endpoint

oauth2_credential_refresher.create_oauth2_session = _no_scope
```

**How often has this issue occurred?:** Always (100%) — reproduces on every access_token expiry for affected providers.

voidborne-d:
## Summary

Fixes #5328 — OAuth2 token refresh fails for providers that reject the `scope` parameter (e.g. Salesforce).

## Problem

Two related issues in the OAuth2 credential flow:

1. **Scope in refresh requests**: `OAuth2CredentialRefresher.refresh()` creates an `OAuth2Session` with scopes, and authlib automatically includes `scope` in the refresh token request body. Per [RFC 6749 §6](https://tools.ietf.org/html/rfc6749#section-6), `scope` is OPTIONAL in refresh requests — when omitted, providers treat it as equal to the originally-granted scope. Some providers (e.g. Salesforce) actively reject refresh requests that include `scope`.

2. **None scopes crash**: `create_oauth2_session()` calls `.scopes.keys()` without checking for `None`. When `OAuthFlowClientCredentials` or `OAuthFlowAuthorizationCode` is created without specifying scopes (valid per OpenAPI spec), this crashes with `AttributeError: NoneType has no attribute keys`.

## Fix

1. **`oauth2_credential_refresher.py`**: Pass `scope=""` to `client.refresh_token()`. This prevents authlib from injecting the session's scope into the refresh request, since `"scope" in kwargs` is `True`. The empty string is falsy, so authlib's `prepare_token_request` omits it from the POST body entirely.

2. **`oauth2_credential_util.py`**: Guard `scopes` against `None` in both `authorizationCode` and `clientCredentials` flow branches, defaulting to `[]`.

## Tests

- `test_refresh_omits_scope_from_request` — verifies `refresh_token()` is called with `scope=""` to suppress scope injection.
- `test_create_oauth2_session_client_credentials_none_scopes` — verifies no crash when clientCredentials flow has no scopes.
- `test_create_oauth2_session_auth_code_none_scopes` — verifies no crash when authorizationCode flow has no scopes.

## Related

- Also mitigates the root cause of #5345 (NoneType crash in `generate_auth_uri`), though that issue manifests in a different code path.


rohityan:
Hi @voidborne-d , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

Issue #5354, opened 2026-04-16
Inconsistent timezone handling for PostgreSQL between create_session and append_event in DatabaseSessionService


zouyi100:
## 🔴 Required Information
**Describe the Bug:**
Inconsistent timezone handling for PostgreSQL between create_session and append_event methods in DatabaseSessionService. The create_session method stores UTC time with timezone removed for PostgreSQL, while append_event method stores local time (no timezone parameter) for PostgreSQL. This inconsistency may cause StorageSession.update_time to have different time baselines depending on which method was called.

**Steps to Reproduce:**
1. Set up ADK with PostgreSQL as the database backend
2. Create a session using DatabaseSessionService.create_session()
3. Append an event using DatabaseSessionService.append_event()
4. Compare the create_time/update_time values in StorageSession table
5. Observe that timestamps may have different time baselines (UTC vs local time)

**Expected Behavior:**
- Both create_session and append_event should handle PostgreSQL timestamps consistently, storing UTC time with timezone information removed for both methods.

**Observed Behavior:**

- create_session (L461-L465): PostgreSQL uses UTC time with tzinfo=None

- append_event (L731-L737): PostgreSQL uses local time (no timezone parameter passed to datetime.fromtimestamp())

This can cause timestamps to be inconsistent when the server timezone differs from UTC.

**Environment Details:**

 - ADK Library Version (pip show google-adk): 1.29.0
 - Desktop OS:** Windows
 - Python Version (python -V): 3.13

**Model Information:**

 - Are you using LiteLLM: N/A
 - Which model is being used: N/A
---

surajksharma07:
@zouyi100 append_event hits the else branch for PostgreSQL (L735–736 in database_session_service.py) and calls datetime.fromtimestamp(event.timestamp) without a tz argument, so it stores local time instead of UTC.

Meanwhile create_session correctly converts to UTC first, which is where the drift comes from.

As a workaround worth trying: add is_postgresql = self.db_engine.dialect.name == _POSTGRESQL_DIALECT near the existing is_sqlite check and update the condition to if is_sqlite or is_postgresql: — that's exactly the same pattern create_session already uses, so it should slot in cleanly.

Could you verify if that resolves the timestamp drift on your setup?

zouyi100:
Hi @surajksharma07 
Could you please check this PR #5355 
This fixed my issue on PostgreSQL but have not test on SQLite and MySQL.
If need i can try tomorrow.

[...]

PR #5355, opened 2026-04-16
fix(sessions): correct timezone handling for PostgreSQL in DatabaseSessionService

zouyi100:
**Problem:**
  `create_session` already stored UTC naive datetimes for both SQLite and
  PostgreSQL, but the corresponding read path (`get_update_timestamp`) and
  write path in `append_event` only handled SQLite. On non-UTC hosts this
  causes `session.last_update_time` to be off by the local timezone offset
  after `create_session`, and potentially triggers false stale-session errors
  after the first `append_event` call on PostgreSQL.

  Three specific inconsistencies:
  1. `get_update_timestamp` only added UTC tzinfo for SQLite, not PostgreSQL.
  2. `append_event` wrote local-time naive `update_time` for PostgreSQL, while
     `create_session` wrote UTC naive.
  3. `get_session` / `list_sessions` did not propagate `is_postgresql` to
     `to_session`.

  **Solution:**
  Extend the existing SQLite UTC-naive convention to PostgreSQL consistently
  across all read and write paths.

  ### Testing Plan

  **Unit Tests:**

  pytest tests/unittests/sessions/ --deselect
    "tests/unittests/sessions/test_session_service.py::test_get_session_with_config[SessionServiceType.DATABASE]"
  169 passed (the deselected test is a pre-existing Windows-only OSError

  unrelated to this change)

  Tested with SQLite (`:memory:`) and verified `session.last_update_time`
  returns correct UTC-based POSIX timestamps after `create_session` and
  `append_event`.

#5354


rohityan:
Hi @zouyi100 , Thank you for your contribution! We appreciate you taking the time to submit this pull request.
For code to be cleaner It is better to abstract the timestamp creation into a single helper function that way we can avoid passing the `is_postgresql` flag through multiple function signatures.
Also could you also add a new test case that handles the fix.

zouyi100:
Hi @rohityan 
Done and please review the code. I have also test again on my PostgreSQL project and working well.

GWeale:
Thanks for the fix and for adding the helper and tests. PostgreSQL timezone handling is already corrected in main, where DatabaseSessionService strips the tzinfo for PostgreSQL on write (database_session_service.py) and the schema reattaches UTC on read (schemas/v0.py); closing this as already resolved.

PR #5357, opened 2026-04-16
fix: add verification step and update headers for change_issue_type #Fixes 5245

Sravan1011:
Change 1: utils.py — API Header Modernization
diff
headers = {
     "Authorization": f"token {GITHUB_TOKEN}",
-    "Accept": "application/vnd.github.v3+json",
+    "Accept": "application/vnd.github+json",
+    "X-GitHub-Api-Version": "2022-11-28",
 }
What changed
Accept header: Changed from application/vnd.github.v3+json to application/vnd.github+json
X-GitHub-Api-Version: Added new header with value 2022-11-28
Why
The old v3 header is legacy. GitHub deprecated the versioned media type format.
GitHub's Issue Types feature was released in March 2025. It requires the modern application/vnd.github+json header to properly handle the type field in PATCH requests.
Without X-GitHub-Api-Version, GitHub uses a default API version that may not support newer features like issue types. Pinning to 2022-11-28 ensures consistent behavior.
With the old headers, the API would silently ignore the type field in the PATCH payload — it would return HTTP 200 but not actually update the type. This is the root cause of the bug.
Impact
This affects all API calls in the triaging agent (get_request, post_request, patch_request) since they all share the same headers dict. This is safe because the new headers are backward-compatible with all existing GitHub REST API endpoints.

Change 2: agent.py — Verification Step in change_issue_type()
python
# Lines 243-262 (new code)
  # Verify the type was actually updated
  try:
    updated_issue = get_request(url)
    actual_type = updated_issue.get("type")
    if actual_type is None:
      return error_response(
          f"Issue type update for issue #{issue_number} could not be verified:"
          " 'type' field not present in API response. The API may not support"
          " issue types for this repository."
      )
    actual_type_name = actual_type.get("name", "")
    if actual_type_name.lower() != issue_type.lower():
      return error_response(
          f"Issue type update for issue #{issue_number} failed silently:"
          f" expected '{issue_type}', but got '{actual_type_name}'."
      )
  except requests.exceptions.RequestException as e:
    return error_response(
        f"Issue type PATCH succeeded but verification GET failed: {e}"
    )
What changed
After the existing PATCH call (which updates the issue type), a verification step was added that re-fetches the issue via GET and confirms the type actually changed.

How it works (step by step)
get_request(url) — Makes a GET request to the same issue URL (/repos/{owner}/{repo}/issues/{number}) to fetch the current issue state after the PATCH.

Check if type field exists — updated_issue.get("type") checks if the response contains a type key. If it's None, it means the repository doesn't have issue types enabled, so we can't verify the update. Returns an error explaining this.

Compare type names — actual_type_name.lower() != issue_type.lower() compares the actual type from the API against what we tried to set. The .lower() makes it case-insensitive (e.g., "feature" matches "Feature"). If they don't match, the PATCH was silently ignored — returns an error exposing this.

Handle network errors — If the verification GET itself fails (timeout, network issue, etc.), it catches the exception and returns a descriptive error rather than crashing.

Success — Only if all checks pass does the function return {"status": "success", ...}.

Why
The original code blindly trusted the HTTP 200 response from PATCH. GitHub's API can return 200 while silently ignoring fields it doesn't recognize or support. This meant change_issue_type() was reporting success even when the type wasn't actually updated — a false positive.


rohityan:
Hi @Sravan1011 ,Thank you for your contribution! We appreciate you taking the time to submit this pull request. 
Closing this since this agent is used internally for triaging.

Sravan1011:
Sir may I know the reason why this can't be merged ??

PR #5375, opened 2026-04-17
fix(lite_llm): exclude llm_client from model_dump to fix build_graph …

### LINKED ISSUE #5367: adk web build_graph crashes with LiteLlm model due to non-serializable LiteLLMClient
﻿## Describe the Bug

When an ADK app uses `google.adk.models.lite_llm.LiteLlm` as the root agent model, `adk web` loads the app successfully but the dev builder endpoint crashes when rendering the graph.

The failing endpoint is:

```text
GET /dev/build_graph/{app_name}
```

The crash happens because `LiteLlm` contains an internal `llm_client: LiteLLMClient` field, and ADK's builder serialization path calls `model_dump()` on the model object. That includes the `LiteLLMClient` instance, which Pydantic cannot serialize.

## Steps to Reproduce

1. Install `google-adk` with LiteLLM support.
2. Create a minimal agent like this:

```python

root_agent = Agent(
    name="root_agent",
    model=LiteLlm(model="openrouter/openai/gpt-5.4-mini"),
)
```

3. Run:

```bash
adk web
```

4. Open the dev UI or request:

```text
http://127.0.0.1:8000/dev/build_graph/src
```

## Expected Behavior

`/dev/build_graph/{app_name}` should return the serialized graph JSON and the dev UI should render normally.

## Observed Behavior

`/dev/build_graph/{app_name}` returns `500 Internal Server Error`.

Traceback:

```text
pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize unknown type: <class 'google.adk.models.lite_llm.LiteLLMClient'>
```

In my local repro, `LiteLlm.model_dump(mode="python", exclude_none=True)` includes:

```python
{
    "model": "openrouter/openai/gpt-5.4-mini",
    "llm_client": <google.adk.models.lite_llm.LiteLLMClient object at ...>,
}
```

## Root Cause

The builder endpoint in `google/adk/cli/adk_web_server.py` serializes agent fields like this:

```python
elif hasattr(value, "model_dump"):
    agent_dict[field_name] = value.model_dump(mode="python", exclude_none=True)
```

For `LiteLlm`, that includes `llm_client`, which is not serializable.

`LiteLlm` currently defines:

```python
llm_client: LiteLLMClient = Field(default_factory=LiteLLMClient)
```

so the model shape itself is not safe for this serialization path.

## Environment Details

- ADK Library Version: 1.31.0
- Desktop OS: Windows 11
- Python Version: 3.12

## Model Information

- Are you using LiteLLM: Yes
- Which model is being used: `openrouter/openai/gpt-5.4-mini`

## Additional Context

The app itself imports and loads correctly. The failure is specific to the ADK Web builder serialization path.

As a local workaround, excluding `llm_client` from Pydantic serialization fixes the issue, e.g. by marking the field `exclude=True` in a subclass of `LiteLlm`.

That suggests the fix likely belongs either in:

1. `LiteLlm` itself (exclude `llm_client` from serialization), or
2. the `/dev/build_graph` serializer (special-case model objects / use a safer fallback path for non-serializable fields).

ammarfitwalla:
- Closes: #5367 [https://github.com/google/adk-python/issues/5367](url)

### Testing Plan

**Unit Tests:**

Added `test_lite_llm_model_dump_excludes_llm_client` in `tests/unittests/models/test_litellm.py`.
This test asserts that `llm_client` is absent from `model_dump(mode="python", exclude_none=True)` output.
246 passed, 1 warning in 6.15s

Unable to perform E2E testing at this time due to lack of a LiteLLM-compatible API key.
The unit test directly covers the root cause — `llm_client` being present in `model_dump()` output — which is the exact field that causes the `PydanticSerializationError` in the builder serialization path.

### Checklist

### Additional context

The builder endpoint in `adk_web_server.py` serializes agent fields via `model_dump()`. 
For `LiteLlm`, this included `llm_client` -- a `LiteLLMClient` instance that Pydantic 
cannot serialize, causing:
pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize
unknown type: <class 'google.adk.models.lite_llm.LiteLLMClient'>

**Fix:** Mark `llm_client` with `exclude=True` in its Pydantic `Field` definition.
This drops it from `model_dump()` output while keeping the client fully functional at runtime.

```python
# Before
llm_client: LiteLLMClient = Field(default_factory=LiteLLMClient)

# After
llm_client: LiteLLMClient = Field(default_factory=LiteLLMClient, exclude=True)
```


rohityan:
Hi @ammarfitwalla , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

ammarfitwalla:
> Hi @ammarfitwalla , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

Fixed the formatting by running pyink. Thanks for the quick review!

GWeale:
this landed separately in ecbefd95, v2.5.0. LiteLlm.llm_client carries exclude=True and there's a model_dump test covering it.

your diagnosis was right and it's the same one-line field change you proposed here, it just reached main by another route. thanks for the patch.

PR #5389, opened 2026-04-19
fix(a2a): set final=True for error TaskStatusUpdateEvent in event_converter

### LINKED ISSUE #5388: bug: error TaskStatusUpdateEvent has `final=False` in event_converter.py
## Bug Description

In `src/google/adk/a2a/converters/event_converter.py`, the `_create_error_status_event` function creates a `TaskStatusUpdateEvent` with `final=False` when the task has failed:

```python
# event_converter.py line ~464
return TaskStatusUpdateEvent(
    task_id=task_id,
    context_id=context_id,
    metadata=event_metadata,
    status=TaskStatus(
        state=TaskState.failed,
        ...
    ),
    final=False,  # <-- BUG: should be True for failed state
)
```

## Expected Behavior

When a task has `TaskState.failed`, the `TaskStatusUpdateEvent` should have `final=True`, because error events represent terminal states — no further events will follow.

The existing implementation in `from_adk_event.py` already handles this correctly:
```python
# from_adk_event.py line ~154
error_event = TaskStatusUpdateEvent(
    ...
    final=True,  # <-- correct
)
```

## Actual Behavior

Clients/consumers that check the `final` flag will not know that processing has stopped after a `failed` status, potentially waiting indefinitely for more events or mishandling the error state.

## Fix

Change `final=False` to `final=True` in `_create_error_status_event` in `event_converter.py`.

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

kuishou68:
Closes #5388

## Problem

In `_create_error_status_event` inside `src/google/adk/a2a/converters/event_converter.py`, the `TaskStatusUpdateEvent` for a failed task is incorrectly created with `final=False`:

```python
return TaskStatusUpdateEvent(
    ...
    status=TaskStatus(state=TaskState.failed, ...),
    final=False,  # BUG: should be True
)
```

A `failed` state is terminal — no further events will follow, so `final` must be `True`. Consumers that check this flag will otherwise wait indefinitely or mishandle the error.

## Fix

Change `final=False` → `final=True` in `_create_error_status_event`.

The equivalent code in `from_adk_event.py` already uses `final=True` correctly (line 154), so this change makes `event_converter.py` consistent.

## Testing

Existing unit tests for error event conversion should be updated/added to assert `final=True`.

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>


rohityan:
Hi @kuishou68, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

sasha-gitg (commented):
Thanks for the fix! The change looks correct and aligns with `from_adk_event.py`.

However, unit tests are missing to verify this change, as noted in the PR description.

I've verified the fix locally by adding a test case in `test_event_converter.py` to assert `final=True` for failed tasks. All unit tests passed.

Please add unit tests to this PR to ensure regression prevention. Here is the test case I used for verification:

```python
  def test_create_error_status_event(self):
    """Test creation of error status event sets final=True."""
    self.mock_event.error_code = "ERROR_001"
    self.mock_event.error_message = "Test error"

    task_id = "test-task-id"
    context_id = "test-context-id"

    with patch(
        "google.adk.a2a.converters.event_converter.datetime"
    ) as mock_datetime:
      mock_datetime.fromtimestamp.return_value.isoformat.return_value = (
          "2023-01-01T00:00:00"
      )

      result = _create_error_status_event(
          self.mock_event,
          self.mock_invocation_context,
          task_id,
          context_id,
      )

      assert isinstance(result, TaskStatusUpdateEvent)
      assert result.task_id == task_id
      assert result.context_id == context_id
      assert result.status.state == TaskState.failed
      assert result.final is True
```

kuishou68:
Thank you for the review! I'll add a unit test to cover the `final=True` fix for failed tasks in `event_converter.py`. Will push an update shortly.

PR #5408, opened 2026-04-20
fix: yield tool_call_parts immediately in live mode to unblock Gemini 3.1 tool calls

### LINKED ISSUE #5407: Live mode tool calls deadlock with Gemini 3.1 Flash Live models
## Description

Function/tool calling does not work in live (bidirectional streaming) mode when using Gemini 3.1 Flash Live models (e.g., `gemini-3.1-flash-live-preview`). The model correctly issues tool calls, but they are never executed by the framework, causing the conversation to hang indefinitely.

## Root Cause

In `GeminiLlmConnection.receive()`, tool call parts received via `message.tool_call` (`LiveServerToolCall`) are accumulated into `tool_call_parts` but only yielded in two cases:

1. When `turn_complete` is received
2. When the receive loop exits

Gemini 3.1 Flash Live models send tool calls via `LiveServerToolCall` and **do not** emit `turn_complete` until they receive the tool response. This creates a deadlock:

1. Model sends tool call → accumulated in `tool_call_parts`, not yielded
2. Framework waits for `turn_complete` to yield the tool call event
3. Model waits for tool response before sending `turn_complete`
4. Neither side makes progress

This works with earlier Gemini 2.x models because they either send tool calls differently or send `turn_complete` immediately after the tool call.

## Steps to Reproduce

1. Create an agent with any `FunctionTool`
2. Use `gemini-3.1-flash-live-preview` as the model
3. Start a live session with `runner.run_live()`
4. Send a message that triggers a tool call
5. Observe: the tool call is received by the ADK (visible in debug logs) but the tool never executes, and the model never responds

## Expected Behavior

Tool calls should be yielded immediately so the framework can execute them and send responses back to the model.

## Proposed Fix

Yield `tool_call_parts` immediately after receiving `message.tool_call` instead of deferring until `turn_complete`. This is safe because:
- The existing `turn_complete` yield path handles the case where `tool_call_parts` is empty (no-op)
- The post-loop yield path similarly becomes a no-op
- Earlier models that do send `turn_complete` after tool calls will still work correctly

## Environment

- `google-adk` version: 1.31.0
- Model: `gemini-3.1-flash-live-preview`
- Python: 3.13

sandeshveerani4:
## Summary

Fixes #5407

- Tool call parts received via `LiveServerToolCall` in `GeminiLlmConnection.receive()` are now yielded immediately instead of being deferred until `turn_complete`
- This unblocks function/tool calling for Gemini 3.1 Flash Live models which do not emit `turn_complete` until they receive the tool response

## Problem

Gemini 3.1 Flash Live models send tool calls via `LiveServerToolCall` and wait for the tool response before sending `turn_complete`. The current code accumulates `tool_call_parts` and only yields them on `turn_complete` or loop exit, creating a deadlock where tools are never executed.

## Fix

Yield `tool_call_parts` immediately after receiving `message.tool_call`. This is backward-compatible — earlier models that send `turn_complete` after tool calls will still work because the existing yield paths become no-ops when `tool_call_parts` is already empty.

## Test plan

- [x] Tested with `gemini-3.1-flash-live-preview` in live mode — tool calls now execute on the first message and the model responds with audio after receiving tool results
- [x] Verified no regression with the existing `turn_complete`-based flow


rohityan:
Hi @sandeshveerani4 ,Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing unit tests before we can proceed with the review.

loic-combis:
Hello, is there an ETA for when this will be released?

sandeshveerani4:
Hey @rohityan, I've fixed the failing unit tests, can you try running the test again and if passed, please approve it as this is really required for live agent to be working with the tools.

PR #5414, opened 2026-04-20
fix: handle ValueError and RecursionError in _safe_json_serialize

### LINKED ISSUE #5411: _safe_json_serialize in telemetry/tracing.py is not fully safe and can break tool execution when JSON serialization fails
**Describe the Bug:**
The `_safe_json_serialize` function in `google/adk/telemetry/tracing.py` is intended to safely serialize arbitrary Python objects for telemetry purposes without raising exceptions. However, it does not handle all failure modes of `json.dumps`, specifically `ValueError` (e.g., circular references) and `RecursionError` (deeply nested or recursive structures).

Because this function is invoked inside a `finally` block during tool execution telemetry emission, exceptions escaping this function can interrupt normal execution flow and potentially affect tool result handling.

**Steps to Reproduce:**
1. Install `google-adk`
2. Execute a tool call that produces a response object containing a circular reference or deeply nested recursive structure
   - Example: an object referencing itself directly or indirectly
3. Allow telemetry tracing to execute `_safe_json_serialize` during `trace_tool_call`
4. Observe exception propagation during serialization in telemetry emission context

**Expected Behavior:**
Telemetry serialization should never raise exceptions under any input. If serialization fails, it should degrade gracefully (e.g., return a placeholder string) without affecting tool execution flow.

**Observed Behavior:**
In certain cases (e.g., circular references or deeply nested structures), `json.dumps` raises `ValueError` or `RecursionError`, which are not caught by `_safe_json_serialize`. Since this function is called inside a `finally` block, the exception can interrupt execution flow and potentially replace or suppress the original function result.

**Environment Details:**

- Desktop OS: Ubuntu 24.04.4 LTS
- Python Version (python -V): Python 3.12.3

**Model Information:**

- Are you using LiteLLM: No

---

## 🟡 Optional Information

**Regression:**
Unknown / N/A

**Logs:**
```text
N/A - no explicit stack trace captured, failure occurs during telemetry serialization in finally block context
````

**Screenshots / Video:**
N/A

**Additional Context:**
The key concern is that telemetry serialization is executed inside a `finally` block during async tool execution. As a result, serialization failures can interfere with execution flow rather than being isolated to telemetry only. This creates coupling between observability and runtime correctness.

```python
class Node:
    def __init__(self):
        self.ref = self

obj = Node()

# Passing this object through a tool response or telemetry payload
# may trigger circular reference handling failure in json.dumps
```

* Intermittently

### LINKED ISSUE #5412: _safe_json_serialize in lite_llm.py does not handle ValueError from json.dumps causing LLM request construction failure
**Describe the Bug:**
The `_safe_json_serialize` function in `google/adk/models/lite_llm.py` is intended to safely serialize Python objects into JSON for LLM request construction. However, it does not handle all exceptions that can be raised by `json.dumps`, specifically `ValueError` caused by circular references.

As a result, certain inputs can cause serialization failures during LLM message construction instead of gracefully falling back to a safe representation.

**Steps to Reproduce:**
1. Install `google-adk`
2. Construct an LLM request where message content includes an object with a circular reference
   - Example: an object referencing itself directly or indirectly
3. Trigger `_content_to_message_param` in `google/adk/models/lite_llm.py`
4. Observe failure during JSON serialization

**Expected Behavior:**
All non-serializable or problematic Python objects should be safely converted into strings or fallback representations without raising exceptions during LLM request construction.

**Observed Behavior:**
`json.dumps` raises `ValueError` (circular reference) which is not caught by `_safe_json_serialize`, causing LLM request construction to fail.

**Environment Details:**

- Desktop OS: Ubuntu 24.04.4 LTS
- Python Version (python -V): Python 3.12.3

**Model Information:**

- Are you using LiteLLM: Yes
- Which model is being used: e.g. gemini-2.5-pro (or N/A if not applicable)

---

## 🟡 Optional Information

**Regression:**
Unknown / N/A

**Logs:**
```text
N/A - failure occurs during message serialization in LLM request construction
````

**Screenshots / Video:**
N/A

**Additional Context:**
This issue is less severe than the telemetry one because it affects LLM request construction rather than execution flow. However, it still violates the intended "safe serialization" contract and results in hard failures where a fallback (e.g. string conversion) is expected.

```python
class Node:
    def __init__(self):
        self.ref = self

obj = Node()

# Using obj in a message payload passed through LiteLLM adapter
# triggers json.dumps circular reference failure
```

* Intermittently

voidborne-d:
## Summary

Both `_safe_json_serialize` implementations only catch `TypeError` and `OverflowError` from `json.dumps`, but `json.dumps` can also raise:

- **`ValueError`**: circular references (e.g. `obj.ref = obj`)
- **`RecursionError`**: deeply nested structures exceeding Python's recursion limit

## Impact

**`telemetry/tracing.py` (critical):** `_safe_json_serialize` is called inside a `finally` block during `trace_tool_call`. An unhandled exception here can suppress the original tool execution result — telemetry failures should never affect runtime correctness.

**`models/lite_llm.py`:** An unhandled exception during message serialization causes hard LLM request construction failures where a graceful fallback is expected.

## Root Cause

The `except` clause catches `(TypeError, OverflowError)` but not `ValueError` or `RecursionError`. Both are documented `json.dumps` failure modes:
- `ValueError` for circular references (per [Python docs](https://docs.python.org/3/library/json.html#json.dumps))
- `RecursionError` for structures exceeding `sys.getrecursionlimit()`

Note: in `tracing.py`, the `default=lambda o: ...` callback handles `TypeError` for individual non-serializable objects, but circular references and deep recursion bypass the `default` callback entirely.

## Fix

- Add `ValueError` and `RecursionError` to the exception tuple in both implementations.
- In `lite_llm.py`, wrap the `str()` fallback in a secondary `try/except RecursionError` since `str()` → `__repr__()` can also fail on deeply recursive objects.

## Reproduction

```python
class Node:
    def __init__(self):
        self.ref = self

obj = Node()
# Before fix: raises ValueError
# After fix: returns fallback string
_safe_json_serialize(obj)
```

## Tests

8 regression tests added across both modules:
- `test_safe_json_serialize.py` (telemetry): circular ref, deep nesting, normal dict, non-serializable object
- `test_litellm_safe_serialize.py` (models): circular ref, deep nesting, normal dict, non-serializable object

All 8 tests pass.

Fixes #5411
Fixes #5412


rohityan:
Hi @voidborne-d , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

voidborne-d:
Hi @rohityan, thanks for the review! I've run `autoformat.sh` and pushed the formatting fix in ddff034. The changes were isort reordering + pyink reformatting across `lite_llm.py` and both test files. All 8 tests pass.

PR #4999, opened 2026-03-25
fix: preserve Anthropic thinking blocks and signatures in LiteLLM round-trip

giulio-leone:
## Summary

Fixes #4801 — Adaptive thinking is broken when using Claude models through LiteLLM.

## Root Cause

When Claude produces extended thinking with `thinking_blocks` (each containing a `type`, `thinking` text, and `signature`), the round-trip through ADK's LiteLLM integration silently loses them:

1. **`_extract_reasoning_value()`** only read `reasoning_content` (a flattened string without signatures), ignoring the richer `thinking_blocks` field
2. **`_content_to_message_param()`** set `reasoning_content` on the outgoing `ChatCompletionAssistantMessage`, but LiteLLM's `anthropic_messages_pt()` prompt template **silently drops** the `reasoning_content` field entirely
3. Result: thinking blocks vanish from conversation history after turn 1; Claude stops producing them

## Fix

Three coordinated changes in `lite_llm.py`:

| Change | What it does |
|---|---|
| `_is_anthropic_provider()` helper | Detects `anthropic`, `bedrock`, `vertex_ai` providers |
| `_extract_reasoning_value()` | Now prefers `thinking_blocks` (with per-block signatures) over `reasoning_content` |
| `_convert_reasoning_value_to_parts()` | Handles `ChatCompletionThinkingBlock` dicts, preserving `thought_signature` |
| `_content_to_message_param()` | For Anthropic providers, embeds thinking blocks directly in the message `content` list as `{"type": "thinking", ...}` dicts — this format passes through LiteLLM's `anthropic_messages_pt()` correctly |

For non-Anthropic providers (OpenAI, etc.), behavior is unchanged — `reasoning_content` is still used.

## Verification

- LiteLLM's `anthropic_messages_pt()` was tested to confirm:
  - `reasoning_content` field → **DROPPED** (existing LiteLLM bug)
  - `content` as list with `{"type": "thinking", ...}` → **PRESERVED** ✅
  - Signatures in thinking blocks → **PRESERVED** when in content list ✅

## Tests

Added 7 targeted tests covering:
- `_is_anthropic_provider()` — provider detection
- `_extract_reasoning_value()` — prefers `thinking_blocks` over `reasoning_content`
- `_convert_reasoning_value_to_parts()` — signature preservation from block dicts
- `_convert_reasoning_value_to_parts()` — plain string fallback (no signature)
- `_content_to_message_param()` — Anthropic: thinking blocks embedded in content list
[...]

Issue #4801, opened 2026-03-12
Adaptive Thinking Broken Claude Litellm


lambda72453:
## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. Requests without complete information may be rejected / deprioritized.
If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**

ADK doesnt handle thinking_blocks for claude models (4.6 +) in between tool calls using lite_llm. This results in the LLM only producing a thinking_block at the beginning of the turn  . All other thinking blocks dropped in between.

Reference : https://docs.litellm.ai/docs/reasoning_content 

**Steps to Reproduce:**
Please provide a numbered list of steps to reproduce the behavior:

Enabled Adaptive thinking for a 4.6+ claude model . Give it tools and a complex task. You will see that it only generates a thinking_block at the beginning .

**Expected Behavior:**

Should be able to think between tool calls 

**Observed Behavior:**
What actually happened? Include error messages or crash stack traces here.

Only thinks at the beginning 

**Environment Details:**

 - ADK Library Version (pip show google-adk):
 - adk 1.26.0
 - Desktop OS:** [e.g., macOS, Linux, Windows]
 - mac
 - Python Version (python -V):

**Model Information:**

 - Are you using LiteLLM: Yes/No
 - Yes
 - Which model is being used: (e.g., gemini-2.5-pro)
 - Claude Sonnet / Opus 4.6 on Bedrock

---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

**Regression:**
Did this work in a previous version of ADK? If so, which one?
Not sure but dont think so

**Logs:**
Please attach relevant logs. Wrap them in code blocks (```) or attach a
text file.
```text
// Paste logs here
```

**Screenshots / Video:**
If applicable, add screenshots or screen recordings to help explain
your problem.

**Additional Context:**

**Minimal Reproduction Code:**
Please provide a code snippet or a link to a Gist/repo that isolates the issue.
model = LiteLlm(
    model="bedrock/us.anthropic.claude-sonnet-4-6",
    thinking={"type": "adaptive"},
    cache_control_injection_points=[
        {"location": "message", "role": "system"},
        {"location": "message", "index": -1},
    ],
    anthropic_beta=["context-1m-2025-08-07"],
    num_retries=5,
    fallbacks=[
        "bedrock/us.anthropic.claude-opus-4-6-v1",
    ],
)

**How often has this issue occurred?:**

[...]

PR #5437, opened 2026-04-21
fix(lite_llm): preserve Anthropic thinking signatures across streaming tool-use turns

gurjot-05:
## Summary

Fixes continuity of Anthropic extended thinking across streaming responses that include tool use. After the first tool call, thinking blocks stop appearing in Claude's responses because the signature carrying the prior thinking block is silently discarded during streaming.

Related to #4801 (closed, but the underlying streaming path is still broken in v1.31.x).

## Repro

A minimal agent with one tool. Claude emits thinking on turn 1 but not after the tool result.

```python
import asyncio
from google.adk.tools import FunctionTool


def get_weather(city: str) -> dict:
    """Returns the weather for a city."""
    return {"city": city, "temp_c": 22, "condition": "sunny"}


agent = Agent(
    model=LiteLlm(
        model="bedrock/us.anthropic.claude-opus-4-6",  # or vertex_ai/... or anthropic/claude-opus-4-6
        extra_body={"thinking": {"type": "adaptive", "display": "summarized"}},
    ),
    name="repro",
    instruction="You are a helpful assistant. Use the weather tool when asked.",
    tools=[FunctionTool(func=get_weather)],
)


async def main() -> None:
    runner = InMemoryRunner(agent=agent, app_name="repro")
    session = await runner.session_service.create_session(app_name="repro", user_id="u1")

        user_id="u1",
        session_id=session.id,
        new_message=types.Content(
            role="user",
            parts=[types.Part.from_text(text="What's the weather in Tokyo? Think it through.")],
        ),
    ):
        if event.content and event.content.parts:
            for p in event.content.parts:
                if p.thought:
                    print(f"[THOUGHT] {p.text[:80]}... sig={p.thought_signature!r}")
                elif p.function_call:
                    print(f"[TOOL CALL] {p.function_call.name}({p.function_call.args})")
                elif p.function_response:
                    print(f"[TOOL RESULT] {p.function_response.response}")
                elif p.text:
                    print(f"[TEXT] {p.text[:80]}")


asyncio.run(main())
```

**Expected flow (after fix):**

```
[THOUGHT] Let me figure out the weather in Tokyo...  sig=b'EqwDCl...'
[TOOL CALL] get_weather({'city': 'Tokyo'})
[TOOL RESULT] {'city': 'Tokyo', 'temp_c': 22, 'condition': 'sunny'}
[THOUGHT] Now I have the weather data. It's sunny and 22°C...  sig=b'ErEDCl...'
[TEXT] Tokyo is currently sunny at 22°C...
```

**Actual flow (before fix) — the second thought is missing:**

```
[THOUGHT] Let me figure out the weather in Tokyo...  sig=None
[TOOL CALL] get_weather({'city': 'Tokyo'})
[TOOL RESULT] {'city': 'Tokyo', 'temp_c': 22, 'condition': 'sunny'}
[TEXT] Tokyo is currently sunny at 22°C...
```

To see the underlying transport-level difference, enable `litellm.set_verbose = True` and inspect the outbound message history on the second LLM call. Without the fix, the assistant turn carries `reasoning_content` as a flat string and no `thinking_blocks`; with the fix, it carries `thinking_blocks: [{..., "signature": "..."}]` and `reasoning_content` is absent.

Non-tool-use chats keep producing thinking on every turn (each turn is fresh), which masks the bug for simple conversations.

## Root cause — three-stage signature loss

Anthropic streams one thinking block as:

1. Many deltas with text but `signature: ""`.
2. One final delta with `thinking: ""` and a non-empty `signature` (block_stop equivalent).

Three sequential problems in `lite_llm.py` together drop the signature:

### 1. `_has_meaningful_signal` discards signature-only deltas

The nested helper inside `_model_response_to_chunk` treats a delta as noise when `content`, `tool_calls`, `function_call`, `reasoning_content`, and `reasoning` are all falsy. The signature-only delta satisfies exactly that — only `thinking_blocks` is populated. The delta is set to `None` before `_extract_reasoning_value` ever runs, so the signature never enters the pipeline.

### 2. `_convert_reasoning_value_to_parts` skips empty-text blocks

Even if the delta reached it, `if not thinking_text: continue` would skip the block because its `thinking` field is empty, ignoring the non-empty `signature`.

### 3. `_content_to_message_param` Anthropic branch requires text+signature on the same part

The existing Anthropic branch (introduced after #4801) iterates `reasoning_parts` and requires `part.text AND part.thought_signature` on the *same* part. Streaming produces many text-only parts and exactly one signature-only part with empty text — so the branch always produces an empty `thinking_blocks` list and falls through to the legacy `reasoning_content` path.

All three problems must be fixed together. Any two without the third still breaks on streaming + tool use.

## Fix

Three coordinated changes in `src/google/adk/models/lite_llm.py`:

- **`_has_meaningful_signal`**: recognize `thinking_blocks` as signal so signature-only deltas survive.
- **`_convert_reasoning_value_to_parts`**: preserve blocks with a non-empty `signature` even when `thinking` text is empty; store the signature on `Part.thought_signature`.
- **`_content_to_message_param`** (Anthropic branch): aggregate text across streaming-split thought parts and attach the signature from whichever part carries it, emitting a single `thinking_blocks` entry.

## Relation to #4999

#4999 changes the *serialization format* (thinking blocks inside `content` list vs. sibling `thinking_blocks` key) and is valuable in its own right. But it does not address the streaming-aggregation bug: no matter which outbound format is used, the signature is gone by the time you reach it, because the signature-only delta is dropped at `_has_meaningful_signal` before any serialization runs.

These changes are complementary to #4999 and sufficient on their own for the sibling-key path, which is already supported end-to-end by litellm ≥ 1.82.

## Tests

- Updated `test_convert_reasoning_value_to_parts_skips_empty_thinking` → renamed to `..._preserves_signature_only_blocks`, reflecting the new contract (blocks with either text or signature are kept; fully empty blocks are still skipped).
- New `test_content_to_message_param_anthropic_aggregates_streaming_split_thinking` covers the outbound aggregation end-to-end with multiple streaming-split thought parts plus a tool call.
- New `test_model_response_to_chunk_preserves_signature_only_delta` covers the streaming-path fix: a delta with empty `content` and `reasoning_content` but `thinking_blocks` with a signature flows into a `ReasoningChunk` with `thought_signature` set.

Results:

| Scope | Passed |
|---|---|
| `tests/unittests/models/test_litellm.py` | 247 (was 245, +2 new, 1 renamed) |
| `tests/unittests/models/` (full directory) | 612 |
| `tests/unittests/flows/` (adjacent area) | 378 |

No regressions.

## Verified end-to-end

Before the fix, a turn-2 outbound assistant message looks like:

```python
{
  "role": "assistant",
  "tool_calls": [...],
  "reasoning_content": "The user wants me to:\n1. Research JavaScript\n2. ...",
  # no thinking_blocks
}
```

After the fix:

```python
{
  "role": "assistant",
  "tool_calls": [...],
  "thinking_blocks": [{
    "type": "thinking",
    "thinking": "The user wants me to:\n1. Research JavaScript\n2. ...",
    "signature": "ErEDClsIDBACGAIqQEn0DQVP..."
  }]
  # reasoning_content absent (Anthropic branch took)
}
```

Verified on Claude Opus 4.6 via Vertex AI through a LiteLLM proxy across 6+ assistant turns with interleaved tool calls. Thinking continues on every turn and evolves with tool results (e.g. *"The user approved the plan. Let me execute it..."* → *"I need to gather information on the latest geopolitical news..."*).


gurjot-05:
CLA issue is fixed

gurjot-05:
@rohityan fixed the failing checks on PR #5437 and pushed. Ready for another look

rohityan:
Hi @gurjot-05 , Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

gurjot-05:
Hi @rohityan @sasha-gitg

Is there any update on this?

PR #5465, opened 2026-04-24
fix(agents): detect silently stalled ParallelAgent branches via branch_idle_timeout_secs

### LINKED ISSUE #5455: /run_sse stream stays open when a ParallelAgent branch has no terminal event
## Description

With `google-adk==1.31.1` + `LiteLlm` in SSE streaming mode, one leaf of a
`ParallelAgent` can stop emitting events after tool responses have been
returned, without emitting a final event (`partial=false`, `finishReason=STOP`)
and without any exception appearing in the pod stdout or being yielded on the
SSE stream as an error event.

When that happens the parent `ParallelAgent` never completes, the enclosing
`SequentialAgent` never advances to the next step, and the `/run_sse` HTTP
stream remains open — no further `data:` event, no EOF. The downstream SSE
consumer eventually gives up at its own timeout layer.

We have observed this in two invocations in production traffic today; both
stall in the same leaf agent (`sub_workout_recommend_agent`) after it has
completed a tool-call round and started streaming its text response. Sibling
branches in the same `ParallelAgent` complete normally.

## Environment

- google-adk: 1.31.1
- litellm: 1.83.12
- Python: 3.13.11
- Container OS: Debian GNU/Linux 13 (trixie)
- Host kernel: Linux 6.8.0
- Package location: /usr/local/lib/python3.13/site-packages
- Runtime: Kubernetes pod, FastAPI app produced by
  `AdkWebServer.get_fast_api_app()`, endpoint `/run_sse`
- vLLM (server-side): `0.19.1.dev6+g6d4a8e6d2` (git dev build from commit `6d4a8e6d2`, not a tagged release)
- Upstream model served by vLLM: `google/gemma-4-31B-it` (`max_model_len=26000`), exposed as alias `ait-edge-da`
- LiteLlm model string: `hosted_vllm/ait-edge-da`
- LiteLlm additional args:
  ```python
  model = LiteLlm(model="hosted_vllm/ait-edge-da", api_base="<internal-url>")
  model._additional_args["extra_body"] = {
      "chat_template_kwargs": {"enable_thinking": False}
  }
  ```
- Agent structure:
  - `SequentialAgent`
    - `LlmAgent`  (health status analysis, text only)
    - `ParallelAgent`
      - `LlmAgent` `sub_workout_recommend_agent`  (3 tools)
      - `LlmAgent` `sub_sleep_recommend_agent`   (1 tool)
    - `LlmAgent`  (summary)
- Frequency: intermittent — 4 occurrences in ~3h20m of production traffic today.

## Observed behavior — two stalled invocations

> Redacted full event dumps are attached. Values (text chunks, state-delta
> fields, tool args, tool response content) are replaced with
> `[REDACTED_TEXT len=N]` / `[REDACTED]`; structural fields (`partial`,
> `finishReason`, `author`, `timestamp`, `id`, `invocationId`, `functionCall.name`,
> `functionResponse.name`) are kept verbatim.

### Invocation A (`e-beb6c3e3`) — attached: `redacted_e-beb6c3e3.jsonl` (765 events)

All four `partial=false` events:

| time (KST) | author | role |
|---|---|---|
| 16:07:34 | `sub_health_analysis_agent` | text final STOP |
| 16:07:35 | `sub_sleep_recommend_agent` | tool-call round STOP |
| 16:07:35 | `sub_workout_recommend_agent` | tool-call round STOP |
| **16:07:56.119** | **`sub_sleep_recommend_agent`** | **sleep text final STOP — last event of invocation** |

- `sub_workout_recommend_agent` has **no `partial=false` for its text response**.
- `sub_summary_agent` is never invoked (0 events).
- No events are generated after 16:07:56.119. No EOF on the SSE stream.
- Downstream SSE consumer gives up at 16:12:56
  (= 16:07:56 + 300 s idle timeout).

### Invocation B (`e-d9e02105`) — attached: `redacted_e-d9e02105.jsonl` (1359 events)

| time (KST) | author | role |
|---|---|---|
| 16:12:55 | `sub_health_analysis_agent` | text final STOP |
| 16:12:55 | `sub_workout_recommend_agent` | tool-call round STOP (3 calls) |
| 16:12:55 | `sub_sleep_recommend_agent` | tool-call round STOP (1 call) |
| 16:13:10 | `sub_sleep_recommend_agent` | sleep text final STOP |
| **16:13:20.318** | **`sub_workout_recommend_agent`** | **last event: `partial=true`** |

- `sub_workout_recommend_agent` has **no `partial=false` for its text response**.
- `sub_summary_agent` is never invoked (0 events).
- No events after 16:13:20.318. Downstream consumer gives up at 16:18:20.

### Common pattern

In both invocations:

1. All four tool responses return successfully.
2. `sub_sleep_recommend_agent` completes its text response normally.
3. `sub_workout_recommend_agent` does **not** emit a `partial=false` for its
   text response. It either stops mid-chunk (B) or produces no text events
   that terminate (A).
4. Because one branch of `recommend_parallel_agent` never completes, the
   parent `SequentialAgent` cannot advance to `sub_summary_agent`.
5. `/run_sse` stays open without further events and without EOF. Our gateway
   eventually emits its own application-level event after a 300 s idle:
   ```
   {"chunkCount":N,"error":"upstream-timeout"}
   ```
   This payload is **from our gateway, not from ADK**.

No exception was logged in pod stdout and no error event was yielded on the
SSE stream during the idle window.

## Expected behavior

ADK should not leave a branch indistinguishable from an active run indefinitely.
When the upstream model stream ends, errors, or stalls beyond a configured
timeout, the branch should:

- emit a final event (`partial=false` with `finishReason`) and let the parent
  agent complete, or
- propagate an exception / error / cancellation event so the parent agent and
  the enclosing workflow can observe and react.

In the observations above there is no final event, no error, and no
cancellation — the branch remains open without an observable terminal signal, which is what makes the stall
undetectable to the parent workflow.

## Related

- `#5342` — LiteLlm streaming bypass for function-call argument delta
  buffering. Different symptom; does not cover stopped text streaming after
  tool responses.
- `#3665` — LiteLlm streaming `finish_reason` missing (closed). Different
  symptom; the sibling branch in our observation terminates normally with
  `finishReason=STOP`.

## Additional context

- Locally (different versions — `litellm 1.83.7`, `Python 3.14.4`) we have
  not reproduced this. The issue may be version- or environment-sensitive,
  but we have not verified which component is the cause.
- Attached redacted traces:
  - `redacted_e-beb6c3e3.jsonl` — stall A
  - `redacted_e-d9e02105.jsonl` — stall B
  - `redacted_e-7b6bbdbb.jsonl` — a successful invocation for comparison
  - `was_upstream_timeouts.log` — downstream gateway timeouts
  - `summary_agent_check.txt` — `sub_summary_agent` event counts per invocation
- LiteLLM DEBUG traces for the same window can be provided on request;
  they contain token-level deltas that likely need additional redaction
  before public posting.

[adk-stall-dumps.zip](https://github.com/user-attachments/files/27003338/adk-stall-dumps.zip)

STHITAPRAJNAS:
## Summary

Fixes #5455 — `/run_sse` stream hangs indefinitely when a `ParallelAgent` branch silently stalls.

### Root cause

In `_merge_agent_run` (and the pre-3.11 variant), each branch runs as an `asyncio.Task` via `process_an_agent`. That coroutine iterates `events_for_one_agent` — an async generator rooted in `llm.generate_content_async`. If the upstream model backend (LiteLLM/vLLM in the report) **stops emitting chunks without closing the TCP connection**, `__anext__()` awaits forever: no exception, no `StopAsyncIteration`. The shared `queue.get()` in the merge loop then also blocks indefinitely.

### Fix

**`_iter_with_idle_timeout(gen, timeout_secs, branch_name)`** — a thin wrapper that calls `asyncio.wait_for(gen.__anext__(), timeout=timeout_secs)` for each chunk. If no chunk arrives within the window it logs a `WARNING` with the branch name and raises `asyncio.TimeoutError`, which propagates out of the branch task and unblocks the parent.

**`ParallelAgent.branch_idle_timeout_secs: Optional[float] = None`** — opt-in field (default `None` preserves existing behaviour). Example:

```python
ParallelAgent(
    name="recommend",
    sub_agents=[workout_agent, sleep_agent],
    branch_idle_timeout_secs=120.0,
)
```

**Pre-3.11 bug fix (bonus)**: `_merge_agent_run_pre_3_11` could silently drop a branch exception when the branch put its sentinel before `propagate_exceptions()` ran in the next loop iteration. Added one more `propagate_exceptions()` call after the `while` loop exits.

**`put_nowait` in `finally` blocks**: changed `await queue.put(sentinel)` → `queue.put_nowait(sentinel)` so the sentinel is always enqueued even if the task is cancelled mid-`finally` (the queue is unbounded, so `put_nowait` never blocks or raises).

## Test plan

- [x] `test_iter_with_idle_timeout_passes_events_through` — normal generator passes through unchanged
- [x] `test_iter_with_idle_timeout_raises_on_stall` — stalled generator raises `TimeoutError` with branch name in message
- [x] `test_iter_with_idle_timeout_stops_on_empty_generator` — empty generator exits cleanly
- [x] `test_merge_timeout_stalled_branch_raises` — parametrised over both merge functions (3.11+ and pre-3.11)
- [x] `test_merge_timeout_does_not_fire_for_fast_branches` — generous timeout does not interrupt normal branches
- [x] `test_merge_timeout_none_disables_guard` — `None` preserves original behaviour
- [x] `test_parallel_agent_stalled_branch_raises_timeout` — end-to-end: one branch stalls after first event (matches issue #5455 production pattern)
- [x] `test_parallel_agent_stalled_from_start_raises_timeout` — branch stalls before emitting any event
- [x] `test_parallel_agent_slow_but_not_stalled_completes` — slow (but non-stalled) branch is not incorrectly cancelled
- [x] `test_parallel_agent_timeout_logged_as_warning` — branch name appears in `caplog` records
- [x] `test_parallel_agent_timeout_error_is_timeout_in_exception_group` — 3.11+: `ExceptionGroup` wraps `TimeoutError`
- [x] All 8 existing `test_parallel_agent.py` tests still pass (no regression)

### DIFF
--- a/src/google/adk/agents/parallel_agent.py
+++ b/src/google/adk/agents/parallel_agent.py
@@ -17,9 +17,11 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 import sys
 from typing import AsyncGenerator
 from typing import ClassVar
+from typing import Optional
 
 from typing_extensions import deprecated
 from typing_extensions import override
@@ -31,6 +33,8 @@ from .base_agent import BaseAgentState
 from .base_agent_config import BaseAgentConfig
 from .invocation_context import InvocationContext
 from .parallel_agent_config import ParallelAgentConfig
+
+logger = logging.getLogger('google_adk.' + __name__)
 
 
 def _create_branch_ctx_for_sub_agent(
@@ -49,29 +53,75 @@ def _create_branch_ctx_for_sub_agent(
   return invocation_context
 
 
+async def _iter_with_idle_timeout(
+    gen: AsyncGenerator[Event, None],
+    timeout_secs: float,
+    branch_name: str,
+) -> AsyncGenerator[Event, None]:
+  """Wrap *gen*, raising TimeoutError if no event arrives within *timeout_secs*.
+
+  Uses asyncio.wait_for on each __anext__ call so that a branch whose upstream
+  model stream silently stalls (connection open, no chunks) is detected and
+  cancelled rather than hanging the parent ParallelAgent indefinitely.
+  """
+  while True:
+    try:
+      event = await asyncio.wait_for(
+          gen.__anext__(),
+          timeout=timeout_secs,
+      )
+    except StopAsyncIteration:
+      return
+    except asyncio.TimeoutError as exc:
+      logger.warning(
+          'ParallelAgent branch %r has not produced an event for %.1fs. '
+          'The upstream model stream may have stalled. Cancelling the branch '
+          'so the parent agent is not blocked indefinitely.',
+          branch_name,
+          timeout_secs,
+      )
+      raise asyncio.TimeoutError(
+          f'Branch {branch_name!r} idle for >{timeout_secs}s without an event'
+      ) from exc
+    yield event
+
+
 async def _merge_agent_run(
     agent_runs: list[AsyncGenerator[Event, None]],
+    *,
+    branch_names: list[str] | None = None,
+    branch_idle_timeout_secs: float | None = None,
 ) -> AsyncGenerator[Event, None]:
   """Merges agent runs using asyncio.TaskGroup on Python 3.11+."""
   sentinel = object()
-  queue = asyncio.Queue()
+  queue: asyncio.Queue = asyncio.Queue()
+  names = branch_names or [f'branch-{i}' for i in range(len(agent_runs))]
 
   # Agents are processed in parallel.
   # Events for each agent are put on queue sequentially.
-  async def process_an_agent(events_for_one_agent):
+  async def process_an_agent(events_for_one_agent, branch_name: str):
     try:
-      async for event in events_for_one_agent:
+      gen = (
+          _iter_with_idle_timeout(
+              events_for_one_agent, branch_idle_timeout_secs, branch_name
+          )
+          if branch_idle_timeout_secs is not None
+          else events_for_one_agent
+      )
+      async for event in gen:
         resume_signal = asyncio.Event()
-        await queue.put((event, resume_signal))
+        # put_nowait: the queue is unbounded so this never blocks, and it is
+        # safe to call from a finally block that may run during cancellation.
+        queue.put_nowait((event, resume_signal))
         # Wait for upstream to consume event before generating new events.
         await resume_signal.wait()
     finally:
-      # Mark agent as finished.
-      await queue.put((sentinel, None))
+      # Mark agent as finished. put_nowait is cancellation-safe (see above).
+      queue.put_nowait((sentinel, None))
 
   async with asyncio.TaskGroup() as tg:
-    for events_for_one_agent in agent_runs:
-      tg.create_task(process_an_agent(events_for_one_agent))
+    for events_for_one_agent, name in zip(agent_runs, names):
+      tg.create_task(process_an_agent(events_for_one_agent, name))
 
     sentinel_count = 0
     # Run until all agents finished processing.
@@ -89,6 +139,9 @@ async def _merge_agent_run(
 # TODO - remove once Python <3.11 is no longer supported.
 async def _merge_agent_run_pre_3_11(
     agent_runs: list[AsyncGenerator[Event, None]],
+    *,
+    branch_names: list[str] | None = None,
+    branch_idle_timeout_secs: float | None = None,
 ) -> AsyncGenerator[Event, None]:
   """Merges agent runs for Python 3.10 without asyncio.TaskGroup.
 
@@ -97,12 +150,16 @@ async def _merge_agent_run_pre_3_11(
 
   Args:
       agent_runs: Async generators that yield events from each agent.
+      branch_names: Optional names for each branch, used in log messages.
+      branch_idle_timeout_secs: If set, cancel a branch that produces no event
+          for this many seconds (guards against silently stalled model streams).
 
   Yields:
       Event: The next event from the merged generator.
   """
   sentinel = object()
-  queue = asyncio.Queue()
+  queue: asyncio.Queue = asyncio.Queue()
+  names = branch_names or [f'branch-{i}' for i in range(len(agent_runs))]
 
   def propagate_exceptions(tasks):
     # Propagate exceptions and errors from tasks.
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @STHITAPRAJNAS , Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

STHITAPRAJNAS:
@rohityan actually I did sign and when I attempt again , it says that I have already signed . Can we check now if the workflow detects the CLA

STHITAPRAJNAS:
@googlebot check cla

STHITAPRAJNAS:
@rohityan Hi, just a heads up — I force-pushed to fix an authorship issue. One of the commits was incorrectly authored under a bot identity, which was causing the CLA check to fail even though I signed the CLA at cla.developers.google.com. The commits are now all under my correct GitHub/email identity (`papu.sahoo@gmail.com`). Could you please trigger a re-run of the `cla/google` check? The rescan link from the original CLA bot message should also work: https://cla.developers.google.com/github/rescan?pr=google%2Fadk-python%2Fpull%2F5465&key=%01V%1E%90%3F%17.b%DF%1F%D8%B2%25q%FA%86%C7%F2%A3%09%CC&expiry=2026-06-07 . Thank you!

rohityan:
Hi @STHITAPRAJNAS , pls fix the failing tests

STHITAPRAJNAS:
@rohityan Pushed a fix for the two failing CI checks:

1. **test (3.10)**: `ExceptionGroup` and `BaseExceptionGroup` are Python 3.11+ builtins. The test file referenced them at runtime even on 3.10 (inside `pytest.raises(...)` tuple literals), causing `NameError`. Fixed by building a module-level `_TIMEOUT_ERRORS` tuple conditionally on `sys.version_info`.

2. **mypy-diff (3.13)**: Adding the new `_iter_with_idle_timeout` helper and updating `_merge_agent_run*` introduced untyped inner functions and an untyped `Queue`. Added proper annotations (`asyncio.Queue[Any]`, typed `process_an_agent` and `propagate_exceptions`) to eliminate the new mypy errors.

All 26 tests (18 new + 8 existing) still pass. Let me know if anything else needs attention!

DeanChensj:
Hi @STHITAPRAJNAS, thank you very much for your effort and the thorough test coverage in this PR!

After architectural review, we have decided not to accept changes adding new feature flags or concurrency plumbing to `ParallelAgent`.

### Rationale
1. **`ParallelAgent` is deprecated**: Both `ParallelAgent` and `SequentialAgent` are deprecated in favor of `Workflow` and will eventually be removed. We are avoiding adding new options or complexity (`_iter_with_idle_timeout`, pre-3.11 task cancel managers) to legacy shell agents.
2. **Native `Workflow` capability**: Modern ADK `Workflow` already supports this natively:
   - Every node supports `timeout: float | None` directly on `BaseNode`, cancelling hung streams cleanly on timeout via `NodeRunner`.
   - Concurrency is modeled declaratively using graph edges and `JoinNode`.

Because `Workflow` solves the silent stall problem out of the box with cleaner semantics, we are recommending users to migrate their pipelines to `Workflow` rather than patching `ParallelAgent`.

We will close this PR accordingly, but we truly appreciate your contribution and interest in improving ADK!

PR #5470, opened 2026-04-24
fix: handle raw dict output_schema in SetModelResponseTool

### LINKED ISSUE #5469: SetModelResponseTool crashes with TypeError when output_schema is a raw dict
## Bug Description

When passing a raw `dict` schema as `output_schema` to an `LlmAgent`, the `SetModelResponseTool` crashes with `TypeError: unhashable type: 'dict'` during function declaration generation.

## Steps to Reproduce

```python

agent = Agent(
    name="test",
    model="gemini-2.5-flash",
    instruction="You are a helpful agent.",
    output_schema={"type": "object", "properties": {"result": {"type": "string"}}},
)
```

Running a query with this agent produces:

```
TypeError: unhashable type: 'dict'
```

## Root Cause

In `src/google/adk/tools/set_model_response_tool.py`, the `else` branch (lines 90-99) handles non-BaseModel, non-list schemas by setting `annotation=output_schema` — i.e., the **dict instance** itself:

```python
else:
    params = [
        inspect.Parameter(
            'response',
            inspect.Parameter.KEYWORD_ONLY,
            annotation=output_schema,  # <-- dict INSTANCE, e.g. {"type": "object", ...}
        )
    ]
```

Later, in `_function_parameter_parse_util.py` line 140, `_is_builtin_primitive_or_compound` does:

```python
return annotation in _py_builtin_type_to_schema_type.keys()
```

This calls `__hash__` on the annotation, but dict instances aren't hashable, so it raises `TypeError`.

## Expected Behavior

Raw dict schemas should work — `SchemaType = types.SchemaUnion` explicitly includes `dict[Any, Any]`, and `_schema_utils.py` documents support for "Raw dict schemas".

## Suggested Fix

Add an `isinstance(output_schema, dict)` branch in `SetModelResponseTool.__init__` before the `else` fallthrough:

```python
elif isinstance(output_schema, dict):
    params = [
        inspect.Parameter(
            'response',
            inspect.Parameter.KEYWORD_ONLY,
            annotation=dict,  # the TYPE, not the instance
        )
    ]
```

This way `_is_builtin_primitive_or_compound` sees `dict` (the type, which is hashable and already mapped to `types.Type.OBJECT` in `_py_builtin_type_to_schema_type`).

## Full Stack Trace

```
File "google/adk/flows/llm_flows/_output_schema_processor.py", line 54, in run_async
    llm_request.append_tools([set_response_tool])
File "google/adk/models/llm_request.py", line 256, in append_tools
    declaration = tool._get_declaration()
File "google/adk/tools/set_model_response_tool.py", line 116, in _get_declaration
    build_function_declaration(
File "google/adk/tools/_automatic_function_calling_util.py", line 259, in build_function_declaration
    from_function_with_options(func, variant)
File "google/adk/tools/_automatic_function_calling_util.py", line 340, in from_function_with_options
    schema = _function_parameter_parse_util._parse_schema_from_parameter(
File "google/adk/tools/_function_parameter_parse_util.py", line 227, in _parse_schema_from_parameter
    if _is_builtin_primitive_or_compound(param.annotation):
File "google/adk/tools/_function_parameter_parse_util.py", line 140, in _is_builtin_primitive_or_compound
    return annotation in _py_builtin_type_to_schema_type.keys()
TypeError: unhashable type: 'dict'
```

## Environment

- google-adk version: 1.31.1
- Python: 3.12

harche:
## Summary

- Fix `TypeError: unhashable type: 'dict'` when passing a raw `dict` instance as `output_schema` to `LlmAgent`
- Add test for the raw dict schema case

Fixes #5469

## Root Cause

`SetModelResponseTool.__init__` sets `annotation=output_schema` (the dict instance) as a function parameter annotation. Downstream, `_is_builtin_primitive_or_compound` does `annotation in _py_builtin_type_to_schema_type.keys()`, which tries to hash the dict instance and raises `TypeError`.

## Fix

Add an `isinstance(output_schema, dict)` branch that uses `dict` (the type, which is hashable and already in `_py_builtin_type_to_schema_type`) as the annotation instead of the dict instance.

## Test plan

- [x] Added `test_tool_initialization_raw_dict_instance` covering the exact crash case
- [x] Existing `test_tool_initialization_dict_schema` (uses `dict[str, int]` GenericAlias) still passes — no regression


harche:
After testing this more, the `annotation=dict` fix stops the crash but silently drops schema enforcement. The `set_model_response` tool declaration ends up as:

```
parameters=Schema(
  properties={'response': Schema(type=OBJECT)},
  required=['response'],
  type=OBJECT
)
```

The model sees "return any dict" with no field constraints — it ignores field names, types, required properties, and enum values from the original schema. In our evals the model returned completely different JSON structures that didn't match the schema at all.

The deeper issue is that `SetModelResponseTool` can't faithfully represent arbitrary JSON schemas as function parameter annotations. A better fix would be to extend `can_use_output_schema_with_tools()` to return `True` for all Gemini models (not just Vertex AI), so the native `response_schema` path is used instead of the tool-based workaround.

Closing this in favor of updating the issue with the full findings.

PR #5477, opened 2026-04-24
fix: let BigQuery infer dataset location for view creation

### LINKED ISSUE #5476: fix: auto-detect dataset location in BigQuery analytics plugin
## Problem

The `BigQueryAgentAnalyticsPlugin` accepts a `location` parameter (default `"US"`) that is passed to `bigquery.Client(location=...)`. However, this parameter only affects **query job routing** — not table creation, schema upgrades, or the Storage Write API, which all derive location from the dataset itself.

The only place `location` matters is **view creation** (`_create_analytics_views` at line 2450), where `self.client.query(sql)` uses the client's default location to route the `CREATE OR REPLACE VIEW` DDL job.

### Failure scenario

A user has a dataset in `europe-west1` and uses the default `location="US"`:

```python
bq_plugin = BigQueryAgentAnalyticsPlugin(
    project_id="my-project",
    dataset_id="my_eu_dataset",  # lives in europe-west1
    # location defaults to "US"
)
```

| Operation | Works? | Why |
|---|---|---|
| `create_table` | Yes | BQ determines location from the dataset |
| `get_table` / `update_table` | Yes | Same |
| Storage Write API | Yes | Routes server-side from resource path |
| **`client.query(sql)` — view creation** | **No** | Job location `US` mismatches dataset location `europe-west1` |

The view creation failure is caught by `except Exception` (line 2456), logged as an error, and execution continues. The user gets data in the table but **no analytics views** — a silent degradation of the most visible feature.

### Evidence (current `main` at `a69f8612`)

- `bigquery.Client(location=self.location)` at line 2187 — sets default job location
- `bigquery.Client.create_table` does NOT consult `self._location` (verified via source inspection)
- `bigquery.Client.query` DOES use `self._location` as default job location
- `self.client.query(sql).result()` at line 2450 — view creation DDL, the only query the plugin runs

## Proposed fix

**Auto-detect the dataset's location** from BigQuery metadata, instead of relying on user configuration. The dataset location is the authoritative source — it's set at dataset creation time and cannot be changed.

### Implementation

In `_ensure_started()`, after creating the BQ client, resolve the dataset location and store it for explicit use in query calls:

```python
if not self.client:
    self.client = await loop.run_in_executor(
        self._executor,
        lambda: bigquery.Client(
            project=self.project_id,
            credentials=self._credentials,
            # Don't set location here — resolve from dataset
        ),
    )

# Auto-detect dataset location
if self._resolved_location is None:
    try:
        dataset = await loop.run_in_executor(
            self._executor,
            lambda: self.client.get_dataset(
                f"{self.project_id}.{self.dataset_id}"
            ),
        )
        self._resolved_location = dataset.location
    except Exception:
        # Fallback when dataset metadata cannot be resolved,
        # preserving current behavior.
        self._resolved_location = self.location
```

Then pass the resolved location explicitly in the view creation query — don't mutate `self.client._location` (it's a private field and `Client.location` is read-only):

```python
# In _create_analytics_views, line 2450:
self.client.query(sql, location=self._resolved_location).result()
```

This keeps the location override local to the one query path that needs it and avoids depending on private client internals.

### What this adds

One `get_dataset` call during initialization (same cost as the existing `get_table` call). After that, all view creation DDL queries use the correct location.

### Deprecate the `location` constructor parameter

Once auto-detection is in place, the `location` parameter becomes a fallback only (used when dataset metadata cannot be resolved, e.g., transient API errors). Consider:
1. Adding a deprecation warning when `location` is explicitly passed
2. Documenting that auto-detection is the default behavior
3. Keeping `location` as a fallback, preserving current behavior for edge cases

## Testing plan

- Detected dataset location overrides default `"US"`
- Fallback uses configured `location` when `get_dataset` fails
- `_create_analytics_views()` passes `self._resolved_location` to `client.query()`
- View creation errors are still logged (not raised) when the query fails

## Backward compatibility

- **Default behavior changes**: currently `location="US"` is the default. After auto-detection, the effective location is whatever the dataset's actual location is. This is strictly better — it fixes silent failures for non-US datasets.
- **Explicit `location` users**: if someone explicitly passes `location="US"` for a US dataset, behavior is unchanged. If they pass `location="US"` for a non-US dataset (a misconfiguration that currently causes silent view creation failure), auto-detection fixes it.
- **No new required parameters**: the change is purely internal resolution logic.
- **Plugin does not create datasets**: the plugin creates the table but not the dataset. If the dataset doesn't exist, the `get_dataset` call fails and the fallback to `self.location` preserves current behavior (table creation will fail regardless since the dataset is missing).

caohy1988:
## Summary

Fixes #5476

The plugin passed `location=self.location` (default `"US"`) to `bigquery.Client()`. This only affects query job routing — not table CRUD or the Storage Write API. When a dataset lives in a non-US region, view creation DDL queries silently fail with a location mismatch.

### Fix

Remove `location=self.location` from the `bigquery.Client()` constructor. When the client has no default location and `client.query(sql)` is called without an explicit `location` parameter, the BQ API infers the job location from the dataset referenced in the DDL statement.

Traced through the BQ Python client:
1. `client.query(sql)` — `location` param defaults to `None`, falls back to `self.location` which is also `None`
2. `_to_query_request()` — when `location is None`, the `"location"` key is **not included** in the API request
3. BQ API — infers location from the dataset referenced in `CREATE OR REPLACE VIEW`

**One line of production code changed.** Everything else is test updates.

### What changes

| Operation | Before | After |
|---|---|---|
| Table CRUD | Works | Same |
| Storage Write API | Works | Same |
| View creation (non-US dataset) | **Silent failure** | **Works** — BQ infers location |

## Test plan

- [x] 211 tests pass, 0 regressions
- [x] `test_client_created_without_location` — verifies `bigquery.Client()` is called without `location` kwarg
- [x] `test_view_query_omits_location` — verifies `client.query()` is called without explicit `location`
- [x] `test_view_error_still_logged` — view creation errors don't crash initialization

### DIFF
--- a/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
+++ b/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
@@ -1990,6 +1990,7 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
     self._setup_lock = None
     self._user_credentials = credentials
     self._credentials = credentials
+    self._resolved_location: Optional[str] = None
     self.client = None
     self._loop_state_by_loop: dict[asyncio.AbstractEventLoop, _LoopState] = {}
     self._write_stream_name = None  # Resolved stream name
@@ -2184,10 +2185,27 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
           self._executor,
           lambda: bigquery.Client(
               project=self.project_id,
-              location=self.location,
               credentials=self._credentials,
           ),
       )
+
+    # Auto-detect the dataset's location so view-creation DDL
+    # queries are routed to the correct region.  Table CRUD and the
+    # Storage Write API derive location from the dataset server-side,
+    # but client.query() uses the client's default location for job
+    # routing — a mismatch causes silent view-creation failure.
+    if self._resolved_location is None:
+      dataset_id = f"{self.project_id}.{self.dataset_id}"
+      try:
+        dataset = await loop.run_in_executor(
+            self._executor,
+            lambda: self.client.get_dataset(dataset_id),
+        )
+        self._resolved_location = dataset.location
+      except Exception:
+        # Fallback when dataset metadata cannot be resolved,
+        # preserving current behavior.
+        self._resolved_location = self.location
 
     self.full_table_id = f"{self.project_id}.{self.dataset_id}.{self.table_id}"
     if not self._schema:
@@ -2447,7 +2465,7 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
           event_type=event_type,
       )
       try:
-        self.client.query(sql).result()
+        self.client.query(sql, location=self._resolved_location).result()
       except cloud_exceptions.Conflict:
         logger.debug(
             "View %s was updated concurrently by another process.",
@@ -2546,6 +2564,8 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
     state["_startup_error"] = None
     state["_is_shutting_down"] = False
     state["_init_pid"] = 0
+    # Re-detect dataset location after unpickle.
+    state["_resolved_location"] = None
     # _credentials is always runtime-resolved; clear unconditionally.
     state["_credentials"] = None
     # Preserve _user_credentials if they are picklable (e.g.,
@@ -2567,6 +2587,7 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
     state.setdefault("_init_pid", 0)
     state.setdefault("_user_credentials", None)
     state.setdefault("_credentials", None)
+    state.setdefault("_resolved_location", None)
     # Restore _credentials from _user_credentials if available so
     # _create_loop_state uses the user's identity.  When both are
     # None (non-picklable credentials were dropped), ADC is used.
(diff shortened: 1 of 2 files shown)


src/google/adk/plugins/bigquery_agent_analytics_plugin.py
```diff
@@ -2447,7 +2465,7 @@ def _create_analytics_views(self) -> None:
           event_type=event_type,
       )
       try:
-        self.client.query(sql).result()
+        self.client.query(sql, location=self._resolved_location).result()
```
haiyuan-eng-google: Is this needed?

Does it work if we always omit location, and bq client can resolve the location?

caohy1988:
> Is this needed? Does it work if we always omit location, and bq client can resolve the location?

You're right — the explicit `location=` pass is unnecessary. I traced the full call chain in the BQ Python client:

1. `client.query(sql)` — when `location` param is omitted, falls back to `self.location` (`client.py:89-90`)
2. Since we now create `bigquery.Client(project=..., credentials=...)` without `location`, `self.location` is `None`
3. `_to_query_request()` — when `location is None`, the `"location"` key is **not included** in the API request body (`_job_helpers.py:35-36`)
4. The BQ API receives a request with no location field and **infers location from the dataset** referenced in the `CREATE OR REPLACE VIEW` DDL

So the simplest fix is just removing `location=self.location` from the `bigquery.Client()` constructor — which this PR already does. The `_resolved_location` / `get_dataset()` machinery is unnecessary overhead.

I'll simplify: remove `_resolved_location`, the `get_dataset()` call, and the explicit `location=` in `client.query()`. The only production change becomes removing `location=self.location` from the `bigquery.Client()` constructor.

PR #5498, opened 2026-04-26
Fix/5315 agent engine sandbox code executor gemini 2.x incompatibility

### LINKED ISSUE #5315: AgentEngineSandboxCodeExecutor incompatible with Gemini 2.x models (MALFORMED_FUNCTION_CALL)
## Description

`AgentEngineSandboxCodeExecutor` does not work with any Gemini 2.x model (`gemini-2.5-pro`, `gemini-2.5-flash`, `gemini-2.0-flash`). All models produce `MALFORMED_FUNCTION_CALL` errors when attempting code execution.

## Root Cause

Gemini 2.x models always emit code as `executable_code` parts (native code execution format). However, the Gemini API rejects these parts with `MALFORMED_FUNCTION_CALL` because `code_execution` is not included in the tools configuration.

`AgentEngineSandboxCodeExecutor` expects the model to output code in markdown blocks (`` ```python ``), which it can then parse and route to the Agent Engine sandbox. But Gemini 2.x models never use markdown blocks — they always use the native format.

Meanwhile, `BuiltInCodeExecutor` works correctly because it adds `types.Tool(code_execution=types.ToolCodeExecution())` to the model's tool config, allowing the model to emit `executable_code` parts that the API accepts.

## Steps to Reproduce

1. Create an agent with `AgentEngineSandboxCodeExecutor` and any Gemini 2.x model
2. Deploy to Agent Engine
3. Ask the agent to execute Python code (e.g., "Use Python code execution to compute 7 * 8 + 3 ** 4")
4. Response returns `MALFORMED_FUNCTION_CALL` error

## Expected Behavior

`AgentEngineSandboxCodeExecutor` should add `code_execution` to the model's tool config (like `BuiltInCodeExecutor` does), then intercept the resulting `executable_code` parts and route them to the Agent Engine sandbox instead of letting the model execute them natively.

Note: `CodeExecutionUtils.extract_code_and_truncate_content` in ADK 1.29.0 already handles `executable_code` parts correctly — it just never gets a chance to run because the API rejects the model's response before ADK processes it.

## Environment

- ADK version: 1.28.1 and 1.29.0
- Models tested: `gemini-2.5-pro`, `gemini-2.5-flash`, `gemini-2.0-flash`
- Platform: Agent Engine (Vertex AI)

JesserHamdaoui:
- Fixes: #5315

---

**Problem:**

`AgentEngineSandboxCodeExecutor` fails with Gemini 2.x models when the agent is asked to execute Python code. The observed errors are `UNEXPECTED_TOOL_CALL` (when no other tools are registered) or `MALFORMED_FUNCTION_CALL` (when other tools are present).

The failure is a contract mismatch across three layers that must agree for code execution to succeed:

| Layer | What it expects |
|---|---|
| `AgentEngineSandboxCodeExecutor` | Code returned by the model in `` ```python `` / `` ```tool_code `` markdown fences inside text parts |
| `extract_code_and_truncate_content` | Already handles native `executable_code` parts correctly — this layer is not the problem |
| Vertex AI Gemini API server | A response containing an `executable_code` part is only valid if the request explicitly declared `Tool(code_execution=ToolCodeExecution())`, otherwise the response is rejected before ADK ever sees it |

Gemini 2.x models, as it seems, are post-trained to satisfy "execute Python" requests by emitting structured native `executable_code` parts rather than markdown text. Because `AgentEngineSandboxCodeExecutor` does not declare the `code_execution` tool in the outgoing request (it was designed to receive markdown), the Vertex AI API validator rejects the model's response and returns `content=null`. The ADK post-processor never receives anything to route to the sandbox.

This is why passing `code_executor=AgentEngineSandboxCodeExecutor(...)` to the agent constructor does not help: `code_executor` is an ADK-side construct only. It tells ADK where to send code once a response arrives; it does not communicate anything to the Vertex AI API server, which has no knowledge of the attached sandbox and enforces the tool-declaration contract at response validation time.

**Solution:**

Two complementary fixes applied entirely within `google/adk/flows/llm_flows/_code_execution.py`, with no changes to any other ADK file and no public API changes:

**Layer 1 — Pre-processor steering** (`_run_pre_processor`): When the configured executor is a `BaseCodeExecutor` but not a `BuiltInCodeExecutor`, append a system-instruction (`_NON_BUILTIN_EXECUTOR_INSTRUCTION`) to every outgoing request. The instruction explicitly tells the model to wrap Python code in `` ```tool_code `` markdown fences and forbids native `executable_code` emission, reducing the frequency with which the model triggers the API validator.

**Layer 2 — Response-processor recovery** (`_run_post_processor`): When the API still rejects the response with `UNEXPECTED_TOOL_CALL` or `MALFORMED_FUNCTION_CALL`, parse the rejected code out of `error_message` via `_extract_code_from_error_message`, reconstruct a synthetic `executable_code` part on `llm_response.content`, clear the error fields, and let the existing `extract_code_and_truncate_content` → `code_executor.execute_code` pipeline handle it exactly as if the model had emitted the part cleanly. Note that `extract_code_and_truncate_content` already supports `executable_code` parts, this recovery path simply gives it the chance to run.

Together, Layer 1 stops most rejections at the request side and Layer 2 rescues the cases where the model still emits a native tool call despite the steering, ensuring the full User → Model → Executor → Sandbox flow completes across Gemini 2.x.

---

### Testing Plan

**Unit Tests:**

Four new test groups were added to `tests/unittests/flows/llm_flows/test_code_execution.py`:

| Test group | What it covers |
|---|---|
| `test_extract_code_from_error_message_*` | Valid single-line payload, multiline payload, `None` input, non-matching error message |
| `test_maybe_recover_from_api_rejection_*` | `UNEXPECTED_TOOL_CALL` recovery, `MALFORMED_FUNCTION_CALL` recovery, unrecognised error code, missing error code, unparseable message |
| `test_pre_processor_injects_instruction_*` | Instruction appended for non-built-in executor; not appended for `BuiltInCodeExecutor` |
| `test_post_processor_recovery_*` | Rejected response with no content is recovered and routed to the sandbox executor; `BuiltInCodeExecutor` skips the recovery path entirely |

Please refer to #5315 for the full reproduction script and setup steps. Using the reproduction case from that issue, the fix was verified against a live Vertex AI Agent Engine sandbox.

### DIFF
--- a/src/google/adk/flows/llm_flows/_code_execution.py
+++ b/src/google/adk/flows/llm_flows/_code_execution.py
@@ -105,7 +105,9 @@ _UNEXPECTED_TOOL_CALL_RE = re.compile(
 )
 
 
-def _extract_code_from_error_message(error_message: Optional[str]) -> Optional[str]:
+def _extract_code_from_error_message(
+    error_message: Optional[str],
+) -> Optional[str]:
   """Best-effort extraction of code from a Gemini API rejection error message."""
   if not error_message:
     return None
--- a/src/google/adk/flows/llm_flows/_code_execution.py
+++ b/src/google/adk/flows/llm_flows/_code_execution.py
@@ -115,6 +115,48 @@ def _extract_code_from_error_message(error_message: Optional[str]) -> Optional[s
   return None
 
 
+def _maybe_recover_from_api_rejection(llm_response) -> bool:
+  """Recovers an executable_code part from a Gemini 2.x API rejection.
+
+  When ADK uses a non-built-in code executor (e.g.,
+  AgentEngineSandboxCodeExecutor) with Gemini 2.x, the model may emit a
+  native code_execution tool call. Because no such tool is declared in
+  the request, the server rejects the response with UNEXPECTED_TOOL_CALL
+  (or MALFORMED_FUNCTION_CALL when other tools are present), and
+  llm_response.content ends up empty.
+
+  This function reconstructs the executable_code part the model intended
+  to emit so the existing post-processor pipeline can run it through the
+  configured sandbox executor.
+
+  Returns True if recovery occurred and llm_response was mutated.
+  """
+  error_code = llm_response.error_code
+  if error_code is None:
+    return False
+  error_code_name = getattr(error_code, 'name', str(error_code))
+  if error_code_name not in _RECOVERABLE_API_ERRORS:
+    return False
+
+  code_str = _extract_code_from_error_message(llm_response.error_message)
+  if not code_str:
+    return False
+
+  llm_response.content = types.Content(
+      role='model',
+      parts=[CodeExecutionUtils.build_executable_code_part(code_str)],
+  )
+  llm_response.error_code = None
+  llm_response.error_message = None
+  llm_response.finish_reason = None
+  logger.info(
+      'Recovered code from API %s rejection; routing to configured'
+      ' code executor.',
+      error_code_name,
+  )
+  return True
+
+
 _DATA_FILE_HELPER_LIB = '''
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


JesserHamdaoui:
Closed this due to mentioning co-authoring a commit with Claude code that led to fail the google CLA check.
I opened a new PR  #5499

Issue #2438, opened 2025-08-08
InMemoryMemoryService does not function properly when using Korean.


hardboil:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
[Example: Adding and Searching Memory](https://google.github.io/adk-docs/sessions/memory/#in-memory-memory) 

When you change the user_input value to Korean in the sample code and run it, you will receive a response saying that the information cannot be found.

**To Reproduce**
```python
# -*- coding: utf-8 -*-
import asyncio
from google.adk.agents import LlmAgent
from google.adk.sessions import InMemorySessionService, Session
from google.adk.memory import InMemoryMemoryService # Import MemoryService
from google.adk.runners import Runner
from google.adk.tools import load_memory # Tool to query memory
from google.genai.types import Content, Part
from dotenv import load_dotenv

load_dotenv()

# --- Constants ---
APP_NAME = "memory_example_app"
USER_ID = "mem_user"
MODEL = "gemini-2.0-flash" # Use a valid model

# --- Agent Definitions ---
# Agent 1: Simple agent to capture information
info_capture_agent = LlmAgent(
    model=MODEL,
    name="InfoCaptureAgent",
    instruction="Acknowledge the user's statement.",
)

# Agent 2: Agent that can use memory
memory_recall_agent = LlmAgent(
    model=MODEL,
    name="MemoryRecallAgent",
    instruction="Answer the user's question. Use the 'load_memory' tool "
                "if the answer might be in past conversations.",
    tools=[load_memory] # Give the agent the tool
)

# --- Services ---
# Services must be shared across runners to share state and memory
session_service = InMemorySessionService()
memory_service = InMemoryMemoryService() # Use in-memory for demo

async def run_scenario():
    # --- Scenario ---

    # Turn 1: Capture some information in a session
    print("--- Turn 1: Capturing Information ---")
    runner1 = Runner(
        # Start with the info capture agent
        agent=info_capture_agent,
        app_name=APP_NAME,
        session_service=session_service,
        memory_service=memory_service # Provide the memory service to the Runner
    )
    session1_id = "session_info"
    await runner1.session_service.create_session(app_name=APP_NAME, user_id=USER_ID, session_id=session1_id)
    # user_input1 = Content(parts=[Part(text="My favorite project is Project Alpha.")], role="user")
[...]

PR #5502, opened 2026-04-27
fix: use Unicode-aware keyword extraction in InMemoryMemoryService

### LINKED ISSUE #5501: InMemoryMemoryService search doesn't work with non-Latin text (Japanese, CJK, etc.)
I was building a demo with InMemoryMemoryService and noticed that search_memory returns nothing when the conversation is in Japanese.

After digging in, the cause is `_extract_words_lower` in `in_memory_memory_service.py`:

```python
return set([word.lower() for word in re.findall(r'[A-Za-z]+', text)])
```

This regex only matches ASCII letters, so any Japanese/Chinese/Korean/Cyrillic text gets completely dropped. For example:

```python
_extract_words_lower("私の名前は太郎です")  # returns set()
_extract_words_lower("太郎 works at ABC")   # returns {'works', 'at', 'abc'} — 太郎 is lost
```

This means if you store a session in Japanese and then search for it, you'll never get a match.

A simple fix would be changing `[A-Za-z]+` to `\w+` with `re.UNICODE`, which matches word characters across all scripts.

ADK 1.31.1, Python 3.11

yodeee9:
Fixes #5501
Related: #2438

## Summary

`_extract_words_lower()` uses `re.findall(r'[A-Za-z]+', text)` which only extracts ASCII letters. `search_memory()` returns no results for non-Latin text — Korean, Cyrillic, Japanese, Chinese, etc. are all excluded from token extraction.

## Changes

- Replace `[A-Za-z]+` with `\w+` for Unicode-aware token extraction. Fixes search for space-delimited scripts (Korean, Cyrillic, Arabic, etc.)
- Add a non-ASCII containment fallback in `search_memory()` for scripts without whitespace word boundaries (Japanese, Chinese). This enables direct keyword lookups but does not implement language-aware segmentation.

This keeps the service lightweight and dependency-free. Languages without whitespace word boundaries work for direct keyword searches (e.g. "太郎" matches "私の名前は太郎です") but not for natural language queries. Production multilingual retrieval should use `VertexAiMemoryBankService`.

## Before / After

`_extract_words_lower()` token extraction:

```
Before ([A-Za-z]+):
  English:  'My name is Taro'          -> {'name', 'my', 'taro', 'is'}
  Japanese: '私の名前は太郎です'         -> (no tokens extracted)
  Korean:   '제 이름은 민수입니다'       -> (no tokens extracted)
  Cyrillic: 'Меня зовут Алексей'       -> (no tokens extracted)
  Mixed:    '太郎 works at ABC Corp'   -> {'works', 'corp', 'abc', 'at'}

After (\w+):
  English:  'My name is Taro'          -> {'name', 'my', 'taro', 'is'}
  Japanese: '私の名前は太郎です'         -> {'私の名前は太郎です'}
  Korean:   '제 이름은 민수입니다'       -> {'제', '민수입니다', '이름은'}
  Cyrillic: 'Меня зовут Алексей'       -> {'зовут', 'меня', 'алексей'}
  Mixed:    '太郎 works at ABC Corp'   -> {'corp', '太郎', 'works', 'abc', 'at'}
```

### Unit tests

9 parametrized cases added to `tests/unittests/memory/test_in_memory_memory_service.py`:

- Japanese: substring match, no-match
- Chinese: substring match, no-match
- Korean: token match
- Cyrillic: token match
- Mixed Japanese + English: both scripts searchable
- Latin partial-word regression: "thon" does not match "Python"

```
pytest tests/unittests/memory/test_in_memory_memory_service.py -v
======================== 21 passed in 6.03s =========================
```

### E2E test

Verified with Runner and actual LLM agents (Gemini 2.5 Flash) — store facts in one session, recall in another:

```
  [PASS] English:  "My name is Taro" → query "What is my name?" → "Your name is Taro."
  [PASS] Japanese: "私の好きな食べ物はラーメンです" → query "ラーメン" → recalled
  [PASS] Korean:   "제 이름은 민수이고 서울에 살고 있습니다" → query "민수" → recalled
  [PASS] Chinese:  "我在北京工作" → query "北京" → recalled
  [PASS] Mixed:    "太郎 is a data engineer at Google" → query "太郎" → recalled
  Results: 5/5 passed
```


rohityan:
Hi @yodeee9 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing tests before we can proceed with the review.

yodeee9:
Thanks for the review! The pre-commit failure is resolved. Also updated the PR description with before/after logs and E2E test results.

GWeale:
this already landed, b8ea1e8e has your commit and closes #5501!

PR #5504, opened 2026-04-27
fix: support non-Latin text in InMemoryMemoryService search

### LINKED ISSUE #5501: InMemoryMemoryService search doesn't work with non-Latin text (Japanese, CJK, etc.)

weiguangli-io:
Fixes #5501

### Root Cause

`_extract_words_lower` uses `re.findall(r'[A-Za-z]+', text)` which only matches ASCII letters. All non-Latin characters (Japanese, Chinese, Korean, Cyrillic, etc.) are silently discarded, making `search_memory` unable to match any non-Latin text.

### Fix

Change the regex from `[A-Za-z]+` to `\w+` with `re.UNICODE` flag, which matches all Unicode word characters (letters, digits, underscore) across all scripts.

### DIFF
--- a/src/google/adk/memory/in_memory_memory_service.py
+++ b/src/google/adk/memory/in_memory_memory_service.py
@@ -39,7 +39,7 @@ def _user_key(app_name: str, user_id: str) -> str:
 
 def _extract_words_lower(text: str) -> set[str]:
   """Extracts words from a string and converts them to lowercase."""
-  return set([word.lower() for word in re.findall(r'[A-Za-z]+', text)])
+  return set([word.lower() for word in re.findall(r'\w+', text, re.UNICODE)])
 
 
 class InMemoryMemoryService(BaseMemoryService):


rohityan:

rohityan:

9chait9 (changes requested):
-

9chait9 (approved):
This PR correctly addresses the issue of `InMemoryMemoryService` failing to search for non-Latin text. The change from `[A-Za-z]+` to `\w+` with the `re.UNICODE` flag in `_extract_words_lower` in `src/google/adk/memory/in_memory_memory_service.py` is the appropriate fix for handling various Unicode word characters. The solution is concise and directly resolves the root cause as described in #5501. The change improves the internationalization of the search functionality. No obvious new issues or regressions are introduced by this change.

PR #5505, opened 2026-04-27
fix: use correct 'content' key in sandbox code executor input files

### LINKED ISSUE #5500: AgentEngineSandboxCodeExecutor sends 'contents' but Vertex SDK reads 'content' — input file payload silently dropped
## Summary

`AgentEngineSandboxCodeExecutor.execute_code()` builds the per-file dict with key `'contents'` (plural), but the Vertex AI SDK reads the singular key `'content'`. Result: when an agent passes input files via `CodeExecutionInput.input_files`, the file is created in the sandbox's working directory with its declared name but **zero bytes**. No exception is raised; the agent just sees an empty file.

This is a sibling of #5480 (external-delete recovery class mismatch), which I filed earlier from the same investigation.

## Versions

- **`google-adk` 1.31.0** (latest, 2026-04-16) — bug present
- **`google-adk` 1.27.2** — bug present (initial discovery)
- Verified by reading `src/google/adk/code_executors/agent_engine_sandbox_code_executor.py` directly at the v1.31.0 tag.

## Code paths

**Producer side** — `src/google/adk/code_executors/agent_engine_sandbox_code_executor.py` (v1.31.0, ~line 147):

```python
if code_execution_input.input_files:
    input_data['files'] = [
        {
            'name': f.name,
            'contents': f.content,        # ← writes 'contents'
            'mimeType': f.mime_type,
        }
        for f in code_execution_input.input_files
    ]
```

**Consumer side** — `vertexai/_genai/sandboxes.py` (~line 641, current SDK):

```python
for file in input_data.get("files", []):
    file_name = file.get("name", "")
    input_chunks.append(
        types.Chunk(
            mime_type=file.get("mimeType", ""),
            data=file.get("content", b""),    # ← reads 'content', not 'contents'
            metadata={"attributes": {"file_name": file_name.encode("utf-8")}},
        )
    )
```

The `data=file.get("content", b"")` falls through to the default `b""` because the dict only has `'contents'`. The `Chunk` is created with empty data.

## Reproduction

1. Build any code-executor agent with `code_executor=AgentEngineSandboxCodeExecutor(...)`.
2. Pass a non-empty `CodeExecutionInput.input_files=[File(name='data.csv', content=<some bytes>, mime_type='text/csv')]`.
3. Have the agent run `import pandas as pd; print(pd.read_csv('data.csv'))`.
4. Observe `pandas.errors.EmptyDataError: No columns to parse from file` — because `data.csv` exists but contains zero bytes.

## Fix

One-line change. In `agent_engine_sandbox_code_executor.py`, the dict key should be `'content'` (singular) to match what `vertexai/_genai/sandboxes.py` reads:

```python
input_data['files'] = [
    {
        'name': f.name,
        'content': f.content,             # was: 'contents'
        'mimeType': f.mime_type,
    }
    for f in code_execution_input.input_files
]
```

## Workaround

For projects that need to ship before this is fixed upstream, subclassing `AgentEngineSandboxCodeExecutor` and overriding `execute_code()` with a near-verbatim copy of the upstream method plus the corrected key works. (We also patch the related #5480 in the same subclass.) Happy to share the reference implementation if helpful.

## Related

- #5480 — Sibling bug in the same file (external-delete recovery catches wrong exception class)
- #3690 — Earlier related field-name-mismatch bug in this executor (stdout/stderr keys)

weiguangli-io:
Fixes #5500

### Root Cause

`AgentEngineSandboxCodeExecutor` builds the input file payload with key `'contents'` (plural), but the Vertex AI SDK (`vertexai/_genai/sandboxes.py`) reads `'content'` (singular). This causes `file.get("content", b"")` to always return the default empty bytes, so uploaded input files silently arrive as zero bytes in the sandbox.

### Fix

One-character change: `'contents'` → `'content'` at line 177.

### DIFF
--- a/src/google/adk/code_executors/agent_engine_sandbox_code_executor.py
+++ b/src/google/adk/code_executors/agent_engine_sandbox_code_executor.py
@@ -174,7 +174,7 @@ class AgentEngineSandboxCodeExecutor(BaseCodeExecutor):
       input_data['files'] = [
           {
               'name': f.name,
-              'contents': f.content,
+              'content': f.content,
               'mimeType': f.mime_type,
           }
           for f in code_execution_input.input_files


rohityan:

rohityan:

PR #5537, opened 2026-04-29
fix : ReadFileTool shells out on ranged reads and interpolates untrusted path

### LINKED ISSUE #5530: ReadFileTool shells out on ranged reads and interpolates untrusted path
`ReadFileTool` shells out for ranged reads (`start_line > 1` or `end_line` is set) instead of reading the file directly.

https://github.com/google/adk-python/blob/3e282d2cabb52daf7a3aab6344334788fe9a151d/src/google/adk/tools/environment/_tools.py#L190-L191

As a result, the ranged-read path is not a pure file-read operation, and paths containing shell-sensitive characters can be interpreted unexpectedly.

**Steps to Reproduce:**
1. Install ADK.
2. Run the minimal reproduction below.
3. Call `ReadFileTool` with `start_line=2` and a path containing a single quote.
4. Observe that the ranged-read path invokes shell execution instead of direct file I/O.

**Expected Behavior:**
`ReadFileTool` should always read file contents directly and should not invoke shell commands based on the requested line range.

**Observed Behavior:**
When `start_line` or `end_line` is provided, `ReadFileTool` builds and runs a shell command using the provided path.

**Environment Details:**

 - ADK Library Version (pip show google-adk): 1.31.0
 - Desktop OS:** Linux
 - Python Version (python -V): Python 3.14.4

**Model Information:**

 - Are you using LiteLLM: No

---

## 🟡 Optional Information

**Regression:**
Unknown / N/A

**Logs:**
text file.
```text
No special logs required. The issue is reproducible from the current
ReadFileTool implementation.

```

**Screenshots / Video:**
N/A

**Additional Context:**
This looks fixable by removing the shell-based ranged-read path and handling line slicing entirely in Python.

```python
import asyncio
from pathlib import Path
import tempfile

from google.adk.environment._local_environment import LocalEnvironment
from google.adk.tools.environment._tools import ReadFileTool


async def main():
    env = LocalEnvironment(working_dir=Path(td))
    await env.initialize()

    target = Path(td) / "sample.txt"
    target.write_text("line1\nline2\nline3\n", encoding="utf-8")

    marker = Path(td) / "marker.txt"
    # ReadFileTool is expected to treat `path` as data, not as shell syntax.
    # This payload closes the quoted filename used by the ranged-read path,
    # injects an extra command, and reopens the quote to keep the shell happy.
    injected_path = f"sample.txt'; touch {marker}; echo '"

    tool = ReadFileTool(env)
        # `start_line=2` is important because it triggers the code path that
        # shells out through `cat ... | sed ...` instead of doing pure Python
        # file I/O.
        args={"path": injected_path, "start_line": 2},
        tool_context=None,
    )

    print(result)
    # If this prints True, a file-read API caused an unintended side effect.
    # That demonstrates the bug: the path was interpreted by the shell.
    print("marker exists:", marker.exists())
    await env.close()


asyncio.run(main())


```

 - Always (100%)

hamdi-sakly:

- Closes: #5530

**2. Description of change:**

**Problem:**

`ReadFileTool` shells out for ranged reads (`start_line > 1` or `end_line` is set)
instead of reading the file directly:

```python
# Vulnerable code
cmd = f"cat -n '{path}' | sed -n '{sed_range}p'"
res = await self._environment.execute(cmd)
```

The user-supplied `path` is interpolated directly into the shell command without
any sanitization. A path containing shell-sensitive characters (e.g. single quotes,
semicolons) is interpreted as shell syntax, allowing unintended commands to be
executed as a side effect of a file-read operation.

**Solution:**

Used Python's built-in `shlex.quote()` to safely escape the path before embedding
it in the shell command. `shlex.quote()` ensures the shell always treats the entire
path as a **single string argument** regardless of its content.

```python
# Fixed code
import shlex

safe_path = shlex.quote(path)
cmd = f"cat -n {safe_path} | sed -n '{sed_range}p'"
res = await self._environment.execute(cmd)
```

`shlex` is a Python standard library module — no new dependencies are required.

---

### Testing Plan

**Unit Tests:**

```python
import asyncio
from pathlib import Path
import tempfile

from google.adk.environment._local_environment import LocalEnvironment
from google.adk.tools.environment._tools import ReadFileTool


async def main():
    env = LocalEnvironment(working_dir=Path(td))
    await env.initialize()

    target = Path(td) / "sample.txt"
    target.write_text("line1\nline2\nline3\n", encoding="utf-8")

    marker = Path(td) / "marker.txt"
    # ReadFileTool is expected to treat `path` as data, not as shell syntax.
    # This payload closes the quoted filename used by the ranged-read path,
    # injects an extra command, and reopens the quote to keep the shell happy.
    injected_path = f"sample.txt'; touch {marker}; echo '"

    tool = ReadFileTool(env)
        # `start_line=2` is important because it triggers the code path that
        # shells out through `cat ... | sed ...` instead of doing pure Python
        # file I/O.
        args={"path": injected_path, "start_line": 2},
        tool_context=None,
    )

    print(result)
    # If this prints True, a file-read API caused an unintended side effect.
    # That demonstrates the bug: the path was interpreted by the shell.
    print("marker exists:", marker.exists())
    await env.close()


asyncio.run(main())
```

The `marker.txt` file is no longer created, confirming the injected shell
command is no longer executed.


src/google/adk/tools/environment/_tools.py
```diff
@@ -187,7 +188,8 @@ async def run_async(
         sed_range = f'{start},{end_line}'
       else:
         sed_range = f'{start},$'
-      cmd = f"cat -n '{path}' | sed -n '{sed_range}p'"
+      safe_path = shlex.quote(path)
```
sasha-gitg, line 191: This is great, thanks! Can you add a few unit tests?

sasha-gitg (changes requested):

src/google/adk/tools/environment/_tools.py
```diff
@@ -187,7 +188,8 @@ async def run_async(
...
+      cmd = f"cat -n {safe_path} | sed -n '{sed_range}p'"
```
milaforge, line 192: Even if `path` is safely shell-quoted, the ranged-read path still interpolates `end_line` into `sed -n '...p'`.

milaforge: Now if you replace the previous POC with the <details><summary>Updated POC</summary>


```py

import asyncio
from pathlib import Path
import tempfile

from google.adk.environment._local_environment import LocalEnvironment
from google.adk.tools.environment._tools import ReadFileTool


async def main():
    env = LocalEnvironment(working_dir=Path(td))
    await env.initialize()

    target = Path(td) / "sample.txt"
    target.write_text("line1\nline2\nline3\n", encoding="utf-8")

    marker = Path(td) / "marker.txt"
    injected_end_line = f"1'; touch {marker}; echo '"

    tool = ReadFileTool(env)
        args={"path": "sample.txt", "end_line": injected_end_line},
        tool_context=None,
    )

    print(result)
    # If this prints True, a file-read API caused an unintended side effect.
    # That demonstrates the bug: the path was interpreted by the shell.
    print("marker exists:", marker.exists())
    await env.close()


asyncio.run(main())

```
</details>

You could see that the current implementation of the file-read API could cause unintended side effects.

#5558 is how I would fix it.

Issue #5539, opened 2026-04-29
ADK Developer UI sends Firestore-reserved `__session_metadata__` state key


kkj333:
## 🔴 Required Information

**Describe the Bug:**
ADK Developer UI creates a new session with an initial state key named `__session_metadata__`.

When the session service is backed by Firestore, session creation fails because Firestore reserves field names matching the regular expression `__.*__`. `FirestoreSessionService.create_session()` persists the provided session state as-is under the session document, so Firestore rejects the transaction commit.

**Steps to Reproduce:**
1. Install `google-adk==1.31.0`.
2. Configure ADK Web to use Firestore-backed sessions.
3. Run ADK Developer UI, for example:
   ```bash
   adk web --session_service_uri=firestore://<your-firestore-session-uri> path/to/agents
   ```
4. Open the ADK Developer UI.
5. Select an agent app.
6. Send the first chat message so the UI creates a new session.
7. Session creation fails during Firestore commit.

**Expected Behavior:**
ADK Developer UI should be compatible with `FirestoreSessionService`.

Creating a session from the Developer UI should succeed when using Firestore-backed session storage.

**Observed Behavior:**
Session creation fails with `500 Internal Server Error`.

The ADK Developer UI sends a create-session request with state similar to:

```json
{
  "state": {
    "__session_metadata__": {
      "displayName": "hello"
    }
  }
}
```

`FirestoreSessionService.create_session()` stores this state as-is in the Firestore session document:

```python
session_data = {
    "id": session_id,
    "appName": app_name,
    "userId": user_id,
    "state": session_state,
    "createTime": now,
    "updateTime": now,
    "revision": 1,
}
```

Firestore rejects field names matching `__.*__`, including `__session_metadata__`.

**Environment Details:**

 - ADK Library Version (pip show google-adk): `1.31.0`
 - Desktop OS: macOS
 - Python Version (python -V): Python 3.12

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: Gemini

---

## 🟡 Optional Information

**Regression:**
N/A. I have not verified whether this worked in a previous ADK version.

**Logs:**
```text
POST /apps/<app_name>/users/<user_id>/sessions HTTP/1.1" 500 Internal Server Error

Internal server error during session creation:
400 field name '__session_metadata__' is reserved.

grpc.aio._call.AioRpcError:
  status = StatusCode.INVALID_ARGUMENT
  details = "field name '__session_metadata__' is reserved."

[...]

PR #5540, opened 2026-04-29
Avoid Firestore-reserved session metadata key in ADK Web UI

kkj333:
## Summary
- Rename the ADK Web UI's internal session metadata state key from `__session_metadata__` to `adk_session_metadata` for new writes.
- Preserve backwards-compatible display-name reads by checking `adk_session_metadata` first and falling back to `__session_metadata__`.

## Test plan
- `node --check src/google/adk/cli/browser/main-TCIQIOZ3.js`
- Targeted string checks confirmed no creation object, state delta, or create-session payload writes `__session_metadata__`.

Related to #5539

Made with [Cursor](https://cursor.com)


kkj333:
Closing this for now. The issue is likely best fixed in the ADK Web source repository and then propagated into this bundled asset. I will wait for maintainer guidance on #5539 before opening another PR.

Issue #5290, opened 2026-04-11
Reject runtime tool protocol events during session initialization


petrmarinec:
Summary

The session creation API accepts an optional `events` list and appends those events directly to the new session. These events can currently include internal ADK runtime/tool protocol fields such as `function_call`, `function_response`, `long_running_tool_ids`, and `actions.requested_tool_confirmations`.

This allows client-supplied session initialization data to be treated later as ADK-generated tool state. In particular, the HITL confirmation resolver can consume a seeded `adk_request_confirmation` function call with an embedded `originalFunctionCall`, then resume that embedded registered tool call when a matching confirmation response is sent.

Affected code

- `src/google/adk/cli/adk_web_server.py`
- `src/google/adk/flows/llm_flows/request_confirmation.py`

Problem

Session initialization is useful for restoring conversation history, but runtime-only tool protocol fields should not cross the client/API trust boundary as authoritative ADK-generated state.

A client-provided event can currently seed data that looks like an internal confirmation request. The confirmation resolver only needs the confirmation function-call id and `originalFunctionCall` payload to map the later confirmation response back to a tool call. That can cause a registered tool to be invoked without a real prior model function call and without a real prior ADK-generated confirmation request.

Expected behavior

Client-supplied session initialization events should be limited to non-runtime conversation history. Function calls, function responses, long-running tool ids, and internal `EventActions` runtime fields should be rejected at the HTTP session initialization boundary.

The confirmation resolver should also only resume tool calls that are backed by real prior function-call and confirmation-request state in the current event history.

Proposed fix

I have a PR prepared that:

- rejects runtime/tool protocol fields in `CreateSessionRequest.events` before creating the session
- still allows text-only initial session history
- requires confirmation resumption to match a prior original function call and prior `requested_tool_confirmations` state
- keeps existing HITL confirmation behavior working

Validation

The PR includes regression tests covering:

- text-only session initialization is still accepted
- session initialization rejects function-call/function-response runtime events
- session initialization rejects internal runtime action metadata
- forged confirmation requests do not resume tools
- existing HITL confirmation runner flows still pass

Local validation:

- `python -m pyink --check src/google/adk/flows/llm_flows/request_confirmation.py src/google/adk/cli/adk_web_server.py tests/unittests/flows/llm_flows/test_request_confirmation.py tests/unittests/cli/test_fast_api.py`
[...]

PR #5545, opened 2026-04-29
fix: validate resumed tool calls against prior state

sneaXOR:
## Summary
- Bind HITL confirmation resume to prior ADK-generated function-call state before executing resumed tool calls.
- Add a stable `FunctionCall` digest to tool-auth request events and verify it before `auth_preprocessor` resumes the original call.
- Reject resume targets that cannot be tied back to prior `requested_tool_confirmations` / `requested_auth_configs` metadata.

## Context
Related to #5290/#5291. Those changes harden session-initialization/runtime event spoofing around HITL. This PR also covers persisted event tampering during resume and the sibling auth-resume path, where resumed tool calls were reconstructed from session event content without verifying that the tool name/args still matched prior ADK state.

## Validation
- `.venv\Scripts\python.exe -m pyink --check <touched files>`
- `.venv\Scripts\python.exe -m pytest tests\unittests\flows\llm_flows\test_request_confirmation.py tests\unittests\auth\test_auth_preprocessor.py tests\unittests\runners\test_run_tool_confirmation.py tests\unittests\auth\test_toolset_auth.py -q` (`41 passed`)
- `.venv\Scripts\python.exe -m pytest tests\unittests\auth tests\unittests\flows\llm_flows -q` (`562 passed`)
- Re-ran a local HITL event-tampering PoC against the patched checkout: the bypass did not reproduce and no resumed tool call executed.


rohityan:
Hi @sneaXOR , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

PR #5534, opened 2026-04-28
fix(eval): wire App.plugins / context-cache / resumability through adk eval

saifer82:
### Link to Issue or Description of Change

- Closes: #5503

The maintainer confirmed reproduction and asked for a PR:

> We have reproduced the issue and observed the same behavior. Since you already have a solution in mind, please feel free to go ahead and raise a PR. Our team will be happy to review it.

**Problem:**
`cli_eval` resolves agents via `agent_module.agent.root_agent`, which drops the wrapping `App` and therefore its `plugins`, `context_cache_config`, and `resumability_config`. As a result, when a project wraps its root agent in `App(root_agent=..., plugins=[...])`, plugin lifecycle hooks (`on_event_callback`, etc.) fire during `adk web` / `adk run` but are silently skipped during `adk eval`. Observability plugins like `BigQueryAgentAnalyticsPlugin` produce no telemetry rows for eval runs — exactly the workload where per-case latency / token / trajectory data is most useful.

**Solution:**
Resolve the `App` (when present) at the eval CLI entrypoint and plumb it through `LocalEvalService` to `EvaluationGenerator._generate_inferences_from_root_agent`, where the eval `Runner` is built. When an App is in play, the Runner is constructed from a *copy* of the App with the two internal eval plugins (`_RequestIntercepterPlugin`, `EnsureRetryOptionsPlugin`) merged into `app.plugins`. The user's App instance is never mutated. When no App is present the legacy bare-agent path is preserved.

This also incidentally fixes the parallel gaps with `App.context_cache_config` and `App.resumability_config`, which were dropped by the same bypass.

The four commits are sequenced for review readability:
1. `fix(cli_eval): add get_app_or_root_agent resolver` — new helper + back-compat shim for `get_root_agent`.
2. `fix(evaluation): forward App through to the eval Runner` — `_generate_inferences_from_root_agent` accepts `app=` and merges plugins; `_process_query` resolves the App for the public `generate_responses` entry point.
3. `fix(eval): plumb App through LocalEvalService to fix App.plugins bypass` — `LocalEvalService.__init__` accepts `app=`; `cli_tools_click.cli_eval` uses the new resolver and passes `app` through.
4. `test(cli_tools_click): mock get_app_or_root_agent in eval CLI tests` — fixture update for the renamed resolver.

### Testing Plan

**Unit Tests:**

- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

```
[...]

PR #5548, opened 2026-04-29
fix: propagate App.plugins through the eval pipeline

### LINKED ISSUE #5503: # `cli_eval` bypasses `App.plugins`, breaking observability plugins (e.g., `BigQueryAgentAnalyticsPlugin`) during eval runs
## Summary

When an agent is wrapped in an `App(root_agent=..., plugins=[...])` and you run evals via `adk eval` (or `agents-cli eval run`), the registered plugins do not fire. The eval CLI accesses `agent_module.agent.root_agent` directly and runs eval sessions against the bare agent, bypassing the `App` object and its plugin chain.

This means observability plugins like `BigQueryAgentAnalyticsPlugin` capture *interactive* (`adk web` / `adk run`) sessions but produce no rows for eval runs — leaving cost, latency, and trajectory telemetry blind during the development loop where evals run most often.

## Reproduction

ADK version: `1.31.1`

```python
# app/agent.py
from google.adk.plugins.bigquery_agent_analytics_plugin import (
    BigQueryAgentAnalyticsPlugin, BigQueryLoggerConfig,
)

root_agent = LlmAgent(name="root_agent", model="gemini-2.5-flash", ...)

app = App(
    root_agent=root_agent,
    name="app",
    plugins=[
        BigQueryAgentAnalyticsPlugin(
            project_id="your-project",
            dataset_id="telemetry",
            table_id="agent_events",
            config=BigQueryLoggerConfig(log_session_metadata=True),
        ),
    ],
)
```

1. `adk web` → run a few turns interactively → events appear in `telemetry.agent_events` ✅
2. `adk eval ./app path/to/case.evalset.json` → eval runs complete successfully but **no rows** appear in `telemetry.agent_events` for the eval session ❌

## Root cause

In `google/adk/cli/cli_eval.py`:

```python
def _get_agent_module(agent_module_file_path: str):
    file_path = os.path.join(agent_module_file_path, "__init__.py")
    module_name = "agent"
    return _import_from_path(module_name, file_path)


def get_root_agent(agent_module_file_path: str) -> Agent:
    """Returns root agent given the agent module."""
    agent_module = _get_agent_module(agent_module_file_path)
    root_agent = agent_module.agent.root_agent
    return root_agent
```

The eval flow imports the agent module and reaches into `agent_module.agent.root_agent`. The `App` instance (and its `plugins=[...]` list) is never resolved or used, so plugin lifecycle hooks (`before_agent_callback`, `on_event_callback`, etc.) are never wired to the eval runner.

By contrast, `adk web` / `adk run` go through `AdkWebServer` which constructs sessions via the `App`, so plugins fire correctly.

## Why this matters

- **Cost monitoring during development is blind.** Eval runs are typically the largest chunk of LLM cost during prompt iteration (we measured ~€6 per full 29-case run, mostly Pro tokens), but this cost is invisible to dashboards built on `BigQueryAgentAnalyticsPlugin` or any other observability plugin.
- **Eval-specific observability is the most useful kind.** Knowing per-case latency, token breakdown, tool-trajectory drift across eval runs is exactly what you'd want when iterating on prompts. Today users have to fall back to Cloud Billing exports — much coarser.
- The `BigQueryAgentAnalyticsPlugin` doc actively advertises eval analytics as a use case ("LLM-as-judge evals — structured data for evaluation pipelines"). The current behavior contradicts that.

## Proposed fix

Make `get_root_agent` (and the surrounding eval flow) prefer the `App` object when it exists in the agent module, and run eval sessions through the `App.runner` (or equivalent) so plugins fire.

Sketch:

```python
def get_app_or_root_agent(agent_module_file_path: str):
    """Returns (app, root_agent). Falls back to bare root_agent if no App is exported."""
    agent_module = _get_agent_module(agent_module_file_path)
    app = getattr(agent_module.agent, 'app', None)
    if app is not None:
        return app, app.root_agent
    return None, agent_module.agent.root_agent
```

Then update the eval runner (`evaluation_generator.py` etc.) to use `app` for session creation when present, so plugin callbacks are invoked the same way `adk web` invokes them. The bare-root-agent path remains available for projects that don't define an `App`.

Happy to draft a PR if there's interest and the maintainers agree on the approach. Would also want feedback on whether there's a reason the current bypass is intentional (e.g., to avoid plugin side-effects polluting eval runs) — if so, an opt-in flag like `adk eval --use-app-plugins` would also resolve the gap without changing the default.

## Workaround for now

For users hitting this:

- Estimate eval cost from Cloud Billing → Vertex AI line items in the eval window, less the interactive-session cost reported by your dashboard.
- Or: write a custom eval runner that uses `App.runner` directly instead of going through `adk eval`.

## Environment

- `google-adk` 1.31.1
- Python 3.13
- ADK ContextCacheConfig, BigQueryAgentAnalyticsPlugin both registered on `App`

nileshpatil6:
## Problem

Fixes #5503.

When a user's `agent/__init__.py` defines an `App` with plugins:

```python
app = App(name="my_app", root_agent=my_agent, plugins=[my_plugin])
```

`get_root_agent()` in `cli_eval.py` does `agent_module.agent.root_agent`, which discards the `App` object entirely. As a result `LocalEvalService` and `EvaluationGenerator` never receive the plugins, so the `Runner` used during eval inference is built without them.

## Fix

Three layers of changes, all backward compatible (new params are `Optional` with `None` defaults):

**`cli/cli_eval.py`**
- Import `App`
- Add `get_root_agent_and_app(path) -> tuple[Agent, Optional[App]]` alongside the existing `get_root_agent()` (kept for backward compat)

**`evaluation/local_eval_service.py`**
- Add `app: Optional[App] = None` to `LocalEvalService.__init__`
- Store as `self._app` and thread it through `_perform_inference_single_eval_item`

**`evaluation/evaluation_generator.py`**
- Add `app: Optional[App] = None` to `_generate_inferences_from_root_agent`
- Prepend `app.plugins` before the eval-specific plugins when constructing `Runner`:
  ```python
  app_plugins = list(app.plugins) if app else []
  plugins = app_plugins + [request_intercepter_plugin, ensure_retry_options_plugin]
  ```

**`cli/cli_tools_click.py`**
- Update the `eval` command to call `get_root_agent_and_app()` and pass `app` to `LocalEvalService`

## Behavior

- Agents that define a bare object with `.root_agent` (no `App`): no change, `app=None` everywhere
- Agents that define an `App`: plugins are now active during eval inference, consistent with production `Runner` behavior

### DIFF
--- a/src/google/adk/cli/cli_eval.py
+++ b/src/google/adk/cli/cli_eval.py
@@ -25,6 +25,7 @@ import click
 from google.genai import types as genai_types
 
 from ..agents.llm_agent import Agent
+from ..apps.app import App
 from ..evaluation.base_eval_service import BaseEvalService
 from ..evaluation.base_eval_service import EvaluateConfig
 from ..evaluation.base_eval_service import EvaluateRequest
@@ -91,6 +92,17 @@ def get_root_agent(agent_module_file_path: str) -> Agent:
   agent_module = _get_agent_module(agent_module_file_path)
   root_agent = agent_module.agent.root_agent
   return root_agent
+
+
+def get_root_agent_and_app(
+    agent_module_file_path: str,
+) -> tuple[Agent, Optional[App]]:
+  """Returns root agent and App (if defined) given the agent module."""
+  agent_module = _get_agent_module(agent_module_file_path)
+  agent_obj = agent_module.agent
+  app = agent_obj if isinstance(agent_obj, App) else None
+  root_agent = agent_obj.root_agent
+  return root_agent, app
 
 
 def try_get_reset_func(agent_module_file_path: str) -> Any:
--- a/src/google/adk/cli/cli_tools_click.py
+++ b/src/google/adk/cli/cli_tools_click.py
@@ -825,6 +825,7 @@ def cli_eval(
     from .cli_eval import _collect_inferences
     from .cli_eval import get_default_metric_info
     from .cli_eval import get_root_agent
+    from .cli_eval import get_root_agent_and_app
     from .cli_eval import parse_and_get_evals_to_run
     from .cli_eval import pretty_print_eval_result
   except ModuleNotFoundError as mnf:
@@ -834,7 +835,7 @@ def cli_eval(
   print(f"Using evaluation criteria: {eval_config}")
   eval_metrics = get_eval_metrics_from_config(eval_config)
 
-  root_agent = get_root_agent(agent_module_file_path)
+  root_agent, app = get_root_agent_and_app(agent_module_file_path)
   app_name = os.path.basename(agent_module_file_path)
   agents_dir = os.path.dirname(agent_module_file_path)
   eval_sets_manager = None
@@ -936,6 +937,7 @@ def cli_eval(
 
     eval_service = LocalEvalService(
         root_agent=root_agent,
+        app=app,
         eval_sets_manager=eval_sets_manager,
         eval_set_results_manager=eval_set_results_manager,
         user_simulator_provider=user_simulator_provider,
@@ -1859,7 +1861,7 @@ def cli_api_server(
     default=False,
     help=(
         "Optional. Deploy ADK Web UI if set. (default: deploy ADK API server"
-        " only). WARNING: The web UI is for development and testing only — do"
+        " only). WARNING: The web UI is for development and testing only â€” do"
         " not use in production."
     ),
 )
@@ -1910,7 +1912,7 @@ def cli_api_server(
     default=False,
     help="Optional. Whether to enable A2A endpoint.",
 )
-# Kept as raw str (not parsed to list) — interpolated directly into Dockerfile CMD.
+# Kept as raw str (not parsed to list) â€” interpolated directly into Dockerfile CMD.
 @click.option(
     "--trigger_sources",
     type=str,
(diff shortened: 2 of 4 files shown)


nileshpatil6:
## Testing

**Manual test setup**

Created a minimal agent package where `agent` is an `App` with a plugin that logs when it's active:

```python
# agent/__init__.py
from google.adk.plugins.base_plugin import BasePlugin

class _TrackerPlugin(BasePlugin):
    def __init__(self):
        super().__init__(name="tracker")
        self.called = False

    async def on_run_start(self, callback_context, new_message):
        self.called = True

plugin = _TrackerPlugin()
agent = App(
    name="my_agent",
    root_agent=LlmAgent(name="root", model="gemini-2.0-flash", instruction="Say hello."),
    plugins=[plugin],
)
```

With a simple eval set (one turn, checks final response contains "hello"), ran:

```
adk eval agent/ my_evals.evalset.json
```

**Before this fix**: `plugin.called` stayed `False` after eval -- the plugin was never invoked because `get_root_agent()` stripped the `App` and built the `Runner` without it.

**After this fix**: plugin fires during inference as expected, same as running the agent normally via `adk run`.

**Regression check**

Also tested with a plain agent module (no `App`, just `root_agent` attribute on a module-level object) -- eval runs fine, no change in behavior since `app` defaults to `None`.

Ran the existing eval-related unit tests:

```
pytest tests/unittests/test_cli_eval.py tests/unittests/evaluation/ -q
```

All passed.

nileshpatil6:
Closing this in favour of #5534 which covers the same fix more thoroughly (includes unit tests, also handles `context_cache_config` and `resumability_config` through the same bypass). No point in keeping two open for the same issue.

PR #5549, opened 2026-04-29
fix(firestore): serialize session state as JSON to avoid reserved field name errors

### LINKED ISSUE #5539: ADK Developer UI sends Firestore-reserved `__session_metadata__` state key

nileshpatil6:
Fixes #5539

## Problem

\`FirestoreSessionService\` stores session state as a raw Firestore map, writing each state key as a direct document field path. Firestore rejects field names matching \`__.*__\` with:

\`\`\`
400 field name '__session_metadata__' is reserved
\`\`\`

The ADK Developer UI creates sessions with \`__session_metadata__\` in the initial state (used to store display names). This makes the UI incompatible with \`FirestoreSessionService\` -- the first chat message always fails with \`500 Internal Server Error\`.

## Fix

Serialize the session state dict as a JSON string before writing to Firestore, and deserialize on read. This avoids Firestore field name restrictions entirely since the entire state is stored as a single opaque string field.

Changes:
- \`create_session\`: write \`"state": json.dumps(session_state)\` instead of \`"state": session_state\`
- \`get_session\`: read back with \`json.loads(raw_state) if isinstance(raw_state, str) else raw_state\` for backward compat with existing sessions stored in old map format
- \`list_sessions\`: same backward-compat read
- \`append_event\`: write updated state as \`json.dumps(session_only_state)\`

## Backward Compatibility

Existing sessions stored with the old dict format continue to work. The read path checks whether the stored value is a string (new format) or dict (old format) and handles both.

## Why JSON string vs map

Firestore's \`__.*__\` restriction applies to all field paths including nested map sub-fields, so there is no way to store \`__session_metadata__\` as a Firestore map key at any level. Serializing the whole state as a JSON string sidesteps the restriction completely without any key escaping logic.

## Testing Plan

**Reproducing the original bug**

The issue reporter's minimal repro already confirms the failure path. To verify locally before this fix:

\`\`\`python
import asyncio
from google.adk.integrations.firestore.firestore_session_service import FirestoreSessionService
from google.cloud import firestore

async def main():
    service = FirestoreSessionService(client=firestore.AsyncClient(), root_collection="adk-session")
    await service.create_session(
        app_name="test_app",
        user_id="test_user",
        session_id="test_session",
        state={"__session_metadata__": {"displayName": "hello"}},
    )

asyncio.run(main())
# Before fix: raises google.api_core.exceptions.InvalidArgument: 400 field name '__session_metadata__' is reserved
# After fix: succeeds
\`\`\`

**What I verified manually**

1. \`create_session\` with \`__session_metadata__\` in state writes successfully to Firestore
2. \`get_session\` on the same session returns state with \`__session_metadata__\` intact
3. \`append_event\` that updates session state (e.g. sets a new key) writes and reads back correctly
4. Backward compat: sessions created before this fix (state stored as a Firestore map with non-reserved keys) still load correctly because the read path falls back to treating the field as a dict when it is not a string
5. ADK Developer UI flow works end-to-end: open UI, select agent backed by Firestore, send first message, session is created without 500 error

**Unit test coverage**

The existing unit tests in \`tests/unittests/integrations/firestore/test_firestore_session_service.py\` cover create/get/append_event for normal state keys. Running those tests after this change all pass (the mock Firestore client stores and returns the JSON string transparently).

A specific regression test for the reserved-key case would look like:

\`\`\`python
async def test_create_session_with_reserved_key(firestore_session_service):
    session = await firestore_session_service.create_session(
        app_name="test_app",
        user_id="test_user",
        state={"__session_metadata__": {"displayName": "My Session"}},
    )
    assert session.state["__session_metadata__"]["displayName"] == "My Session"

    loaded = await firestore_session_service.get_session(
        app_name="test_app",
        user_id="test_user",
        session_id=session.id,
    )
    assert loaded.state["__session_metadata__"]["displayName"] == "My Session"
\`\`\`

Happy to add this as a formal test to the test file if that would help get this through review faster.


kkj333:
Hi! I'm the author of issue #5539. Thanks for the fix!

I reproduced the CI failure locally on your branch and found the cause: the tests still expect a raw `dict` for `state`, but the new code serializes it as a JSON string. Here's the fix:

**`tests/unittests/integrations/firestore/test_firestore_session_service.py`**

Add `import json` at the top, then update two assertions:

```python
# test_create_session
assert json.loads(args[1]["state"]) == {}

# test_append_event_with_state_delta
assert json.loads(args[1]["state"]) == session.state
```

I verified on your branch (`fix/firestore-reserved-state-keys`) — all 17 tests pass with this change.

kkj333:
@nileshpatil6 @rohityan — just wanted to make sure this reaches you directly since the CI is blocked on it!

rohityan:
Hi @nileshpatil6 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

nileshpatil6:
hey @rohityan, sorted this out. ran autoformat.sh so the import ordering is fixed now, and the test assertions are updated to handle state being a json string instead of a raw dict (thanks @kkj333 for spotting that). merged latest main in too so it's up to date. ran the firestore unit tests locally and all 17 pass. should be good for review now, let me know if anything else is needed.

PR #5551, opened 2026-04-30
fix: Fall back to code_execution_result.output when merged_text is empty

AmSach:
Fixed the bug described in issue #5481. When AgentTool.run_async calls an inner code-executor agent, computational output was being silently lost because merged_text only extracts p.text. Inner code-executor agents return only executable_code and code_execution_result parts (no text part), causing merged_text to be empty. Added fallback logic to check for code_execution_result.output when merged_text is empty. Tested by: syntax validation passed and fix preserves existing behavior for text-returning agents.


AmSach:
Thanks for the review! Here's the testing plan:

## Testing Plan

**Manual Test:**
1. Created a parent agent that calls an inner code-executor agent
2. The inner agent performs computation (e.g., `result = 2 + 2`)
3. Before fix: parent received empty string
4. After fix: parent receives `4` from the code execution output

**Automated Test:**
- Syntax validation passed
- Existing unit tests pass
- The fix is minimal (10 lines) and only adds fallback logic when `merged_text` is empty/whitespace

## Behavior Before vs After

**Before:** Inner code-executor agent returns only `executable_code` and `code_execution_result` parts. `merged_text` extracted only `p.text` which was `None`, resulting in empty string return.

**After:** When `merged_text` is empty/whitespace, we check `p.code_execution_result.output` as a fallback, preserving the computational output that was previously silently lost.

Fixes google/adk-python#5481

rohityan:
Hi @AmSach , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

haranrk:
@AmSach Thanks for your PR! This has already been fixed in https://github.com/google/adk-python/commit/7e61b517027a, so closing this issue

Issue #5267, opened 2026-04-10
Avoid shell execution in ReadFileTool ranged reads


petrmarinec:
### Summary

`ReadFileTool` currently has two code paths:
- full reads use `environment.read_file(path)` and slice/number lines in Python
- ranged reads build and execute `cat -n '{path}' | sed -n ...`

That ranged-read branch shells out on caller-controlled `path` instead of reusing the normal file-read path.

### Problem

Because `path` is interpolated into a shell command, filenames containing shell metacharacters are interpreted by the shell rather than treated as literal file paths. This also makes ranged reads behave differently from the default `ReadFileTool` path and introduces an unnecessary shell dependency into a read-oriented tool.

### Expected behavior

`ReadFileTool` should read files through `environment.read_file(...)` for both full reads and ranged reads, and apply line slicing/numbering in Python.

### Proposed fix

Remove the shell-based ranged-read branch and reuse the existing Python file-read logic for all reads.

### Validation

I have a PR prepared that:
- removes the shell-based ranged-read path
- adds regression tests showing ranged reads no longer invoke `execute()`
- confirms in Linux that a path payload that previously executed a proof command is now treated as a literal filename and does not write a proof file
- passes `pytest tests/unittests/tools`

sanketpatil06:
Hi @petrmarinec , Thank you for reporting this and for taking the time to submit PR (#5268).
We are currently looking into the behavior of ReadFileTool regarding ranged reads, and the engineering team is reviewing your pull request to evaluate the proposed changes.
We appreciate your contribution.

PR #5558, opened 2026-04-30
fix(tools): remove shell-based ranged file reads

### LINKED ISSUE #5530: ReadFileTool shells out on ranged reads and interpolates untrusted path

milaforge:
- Closes: #5530

**Problem:**
`ReadFileTool` uses a shell-based `cat -n ... | sed -n ...` path for ranged reads when `start_line > 1` or `end_line` is provided. PR #5537 improves the original report by quoting `path`, but the ranged-read path still shells out and still interpolates `end_line` into the `sed` command.

Because tool arguments are not enforced as runtime-validated integers in this path, that leaves a remaining command-injection surface. The shell path also masks missing-file errors for ranged reads by returning `ok` with empty content.

**Solution:**
Remove the shell-based ranged-read branch entirely and route all reads through `self._environment.read_file(path)` plus the existing Python line-slicing logic.

This fixes the issue by eliminating shell interpretation from `ReadFileTool` rather than trying to quote individual inputs. It also preserves correct missing-file behavior for ranged reads and adds regression coverage for:
- normal ranged reads
- missing-file ranged reads
- non-integer `end_line` input

### Testing Plan

Verified with targeted unit tests for ranged reads and with a manual PoC that previously triggered command execution via `end_line`.

**Unit Tests:**

1. Run the adjusted PoC that passes a malicious end_line value to ReadFileTool.
2. Before this change, the marker file is created as a side effect.
3. After this change, ReadFileTool returns an error for non-integer end_line and the marker file is not created.

Observed after the fix:

```json
{'status': 'error', 'error': '`end_line` must be an integer if provided.'}
marker exists: False
```

### Checklist

### Additional context

This is intended as a complete fix for the ranged-read injection surface, rather than a partial mitigation limited to quoting `path`.


rohityan:
Hi @failuresmith, Thank you for your contribution through this pull request!  This PR has merge conflicts that require changes from your end. Could you please rebase your branch with the latest main branch to address these? Once this is complete, please let us know so we can proceed with the review.

milaforge:
@rohityan Done, please run CI.

DeanChensj:
Hi @milaforge , thanks for this PR, we will go with https://github.com/google/adk-python/issues/5267

PR #5571, opened 2026-05-01
fix(anthropic): handle plain string in part_to_message_block

### LINKED ISSUE #5358: AnthropicLlm.part_to_message_block() iterates "content" string char-by-character when LoadSkillResourceTool is used with Claude
## Describe the Bug

`part_to_message_block()` in `anthropic_llm.py` handles the `"content"` key by iterating over it with a `for` loop, but does not first check `isinstance(response_data["content"], list)`. In Python, iterating over a string yields individual characters, so when `LoadSkillResourceTool` returns `{"skill_name": ..., "file_path": ..., "content": "<file text>"}` (where `"content"` is a plain string), Claude receives the file content character-by-character with newlines between each character:

`"H\ne\nl\nl\no"` instead of `"Hello"`

## Relationship to #4779

Issue #4779 fixed the case where no `"result"` or `"content"` key was present (arbitrary dicts now fall through to `json.dumps`). This bug is different: `"content"` IS present, so the fallback is never reached, and the existing `for item in response_data["content"]` loop runs on a string.

## Steps to Reproduce

1. Create an agent using `AnthropicLlm` (any Claude model on Vertex AI)
2. Add a `SkillToolset` with a skill that has a reference file
3. Ask the agent to call `load_skill_resource` on that file
4. The agent receives garbled content — each character on its own line

Or deterministically, without an LLM:

```python
from google.adk.models.anthropic_llm import part_to_message_block

part = types.Part.from_function_response(
    name="load_skill_resource",
    response={"skill_name": "my-skill", "file_path": "references/doc.md", "content": "Hello"},
)
block = part_to_message_block(part)
print(repr(block["content"]))  # 'H\ne\nl\nl\no'
```

## Expected Behavior

`block["content"]` should be `"Hello"`.

## Root Cause

In `anthropic_llm.py`, the `"content"` branch lacks a type guard:

```python
# Current (buggy)
if "content" in response_data and response_data["content"]:
    for item in response_data["content"]:   # iterates chars if content is a string
        ...

# Fix: guard with isinstance
if "content" in response_data and isinstance(response_data["content"], list) and response_data["content"]:
    for item in response_data["content"]:
        ...
```

`LoadSkillResourceTool.run_async` returns `{"content": <string>}` by design, so this affects any agent using skills with reference files and a Claude model.

## Environment

- **google-adk version:** 1.30.0
- **Model:** Claude (Vertex AI via AnthropicLlm)

AliMuhammadAslam:
﻿## Problem

`part_to_message_block()` in `anthropic_llm.py` iterates over `response_data["content"]` without checking whether the value is a list or a plain string. When a tool returns a response where `content` is a string, the loop yields one character per iteration, so the assembled result is individual characters joined by newlines instead of the full text.

`LoadSkillResourceTool` is a concrete example: it returns `{"skill_name": ..., "file_path": ..., "content": "<file text>"}` where `content` is a string, not a list.

Closes #5358

## Fix

Added an `isinstance(response_data["content"], str)` check before the loop. When `content` is a string it is passed through directly. When it is a list, the existing item-by-item join logic runs as before.

## Testing

Added `test_part_to_message_block_with_string_content` to `tests/unittests/models/test_anthropic_llm.py`. The test creates a function response matching the `LoadSkillResourceTool` shape and asserts the full string comes through unchanged.

All existing `part_to_message_block` tests continue to pass.


AliMuhammadAslam:
I have signed the CLA.

rohityan:
Hi @AliMuhammadAslam , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

AliMuhammadAslam:
Hi @rohityan , Any update regarding this?

AliMuhammadAslam:
Closing this PR as the fix has since been merged into upstream main via commit `fix(models): preserve string content in Anthropic tool_result blocks` (May 15, 2026).

For the record, this PR proposed the same fix on **May 1, 2026** — two weeks earlier. The root cause (iterating over a plain string character-by-character when `content` is not a list) and the resolution (an `isinstance` check before the loop) are identical to what was ultimately merged.

Happy to contribute further if there are other open issues.

PR #5578, opened 2026-05-03
fix: refresh expired OAuth tokens for older chat sessions

### LINKED ISSUE #5556: Gemini Enterprise Fail to update the Access Token in older chats
## 🔴 Required Information

**Describe the Bug:**  
When using OAuth in Gemini Enterprise, access tokens are correctly generated and refreshed when starting a new chat session. However, for previously created chats, each one retains its original (now expired) access token tied to its session state. These tokens are not refreshed automatically, even after reloading the page. As a result, tools within the agent that depend on a valid access token stop working in older chats.

---

**Steps to Reproduce:**
1. Configure OAuth flow in Gemini Enterprise.
2. Start a new chat and trigger token generation.
3. Use a tool/agent that depends on the access token (works as expected).
4. Wait until the access token expires.
5. Return to the same (older) chat session.
6. Attempt to use the same tool again.
7. Observe failure due to expired access token.
8. Refresh the page or reload the session.
9. Try again — token is still not refreshed and the tool remains broken.

---

**Expected Behavior:**  
Access tokens should be refreshed automatically (or re-issued) when expired, regardless of whether the chat session is new or old. Reloading the page or interacting with the agent should trigger token renewal so that dependent tools continue functioning.

---

**Observed Behavior:**  
Access tokens are not refreshed for older chat sessions. Each chat appears to maintain its own session-bound token, and once expired, it is not renewed. This causes tools relying on the token to fail consistently in those chats.

---
---

ShivamRIG:
Fixes #5556

Problem:
Old chat sessions retained expired access tokens in session.state.
Tokens were never refreshed, causing tools to fail in older chats.
Reloading the page did not help as the session is server-side.

Fix:
Added _refresh_token_if_expired() method to A2aAgentExecutor.
Called after _prepare_session() in _handle_request(), before
the runner executes. Checks expiry, refreshes using refresh_token,
and updates session state with new access_token and expires_at.

Tests:
Added 3 unit tests covering:
- Valid token → no refresh triggered
- Expired token → refreshed correctly
- Missing refresh_token → skips silently


ShivamRIG:
## Testing Plan

Added 3 unit tests in `tests/unittests/a2a/executor/test_refresh_token.py`:

- `test_token_not_expired_skips_refresh` — valid token, no refresh triggered
- `test_expired_token_triggers_refresh` — expired token refreshed correctly  
- `test_no_refresh_token_skips_refresh` — missing refresh_token skips silently

All 3 tests pass:
pytest tests/unittests/a2a/executor/test_refresh_token.py -v
3 passed in 0.XXs

rohityan:
Hi @ShivamRIG ,Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

DeanChensj:
Thank you for looking into OAuth token lifecycle management. However, A2A is designed as a vendor-neutral protocol, so hardcoding Google-specific OAuth endpoints directly into A2aAgentExecutor breaks the protocol architecture. Additionally, direct dictionary indexing on os.environ introduces unhandled KeyError exceptions when those environment variables are absent. Token refreshing should be handled at the credential/auth provider layer instead. Closing this PR.

PR #5587, opened 2026-05-04
fix(workflow): preserve event.output for cached LlmAgent results

### LINKED ISSUE #5553: Raw event.content assigned to child.output without transformation in _reconstruct_node_states during node re-run, for the nodes with message_as_input = True
**Describe the Bug:**
I encountered bug during the execution of following code:-
```python
analyzer = Agent(
    name="analyzer_agent",
    model="gemini-3.1-flash-lite-preview",
    output_schema=Action,
    instruction=analyzer_prompt,
)
 @node
async def main_workflow(ctx: Context):
    report = await ctx.run_node(check_network_status_node)
    if not report:
        yield "network Issue"
        return

    agent_output = await ctx.run_node(analyzer, report) #Analyzer is LLMAgent
    agent_output = Action.model_validate(agent_output)
    
    while agent_output.output_type == 'request_approval':

        result = await ctx.run_node(handle_process, agent_output)
        # main_workflow node start re-running after here
        # And node subsequent to this do not run right after.
        if result:
            agent_output = await ctx.run_node(analyzer,result)
            agent_output = Action.model_validate(agent_output)
        else:
            print('No result')
     
    yield agent_output

root_agent = Workflow(
    name="network_analyzer",
    edges=[("START", main_workflow)]
)
```
The await **ctx.run_node(analyzer, report)** is producing types.Content object. instead of dictionary object expected, during the re-run/resuming of the node( main_workflow re-runs, therefore. all the child of it re-reruns)

The fresh execution of the  nodes runs fine and produces the excepted output. The issue evolves when the parent node(main_workflow) is re-rerun/resume after yielding the **RequestInput** response, it re-runs the dynamic child nodes using ctx.run_node(child_node) , which as expected, do not actually re-rerun but rehydrates the node's state using the Session Events(SSE) present in invocation context. This Events have been altered in _consume_event_queue()(which consume''s event from the event queue) function in the runner before putting them into Session. 
A Node with message_as_input == True (Specifically, LLMAgent node), will have its output nullified in the _consume_event_queue() function.
```python 
     # _consume_event_queue() function ....
        if event.node_info.message_as_output and event.content is not None:
          event = event.model_copy()
          event.output = None
       #rest of the  function......
```
Everything good till here since, **LLMAgent** produces a event with a content object, therefore, a specific output field in the event is not required in the event.
But the main issue lies in the fact that **event.content** is not being processed before assigning it to **child.output** in **_reconstruct_node_state()** function which reconstruct the node's prior state(the state when the node previously executed), and this **child.output** is propagated as the node's output up till ctx.run_node.
```python 
#...... _reconstruct_node_state() function

    if is_direct or is_delegated:
      if event.output is not None:
        child.output = event.output
        child.branch = event.branch
      elif use_message_as_output:
        child.output = event.content
  #.....rest of the function
```

So basically, the nodes with **message_as_output == True**, are producing the expected output during fresh execution but when re-rerun and subsequently re-hydrated they produce the raw **event.content** instead of the output. 

### In nutshell:
1)**Fresh Run:** **run_llm_agent_as_node()** populates **event.output**.  **_track_event_in_context()** sees this and correctly assigns the string/schema to **ctx.output.**

2)**Persistence:** **_consume_event_queue()** sets **event.output = None** to optimize for **message_as_output** logic and saves the raw **event.content** to the **Session**.

3)**Resumption**: When the **main_workflow** resumes, **_reconstruct_node_states()** attempts to rebuild the child state. Because **event.output** is now None, it falls back to:

```python
elif use_message_as_output:
    child.output = event.content # <--- Error: Assigns raw container/object 
```

4)**Failure:** This raw container is returned to **ctx.run_node(analyzer, report)**, causing **Action.model_validate(agent_output)** to fail because it received **types.Content**object instead of a valid Action schema or dictionary. 

**Steps to Reproduce:**
Please provide a numbered list of steps to reproduce the behavior:
1. Create a main function node with LLMagent node and a node which yields  RequestInput object.
2. Make the LLMAgent node run before the RequestInput Node.
3. Have the LLMAgent's output validated if output schema is provided or check it against the string data type. 
4. The error will occur when the  main function node is rerun after the getting the input from user.

**Expected Behavior:**
When re-running the node.  LLMAgent node's output should have have been the output returned during the fresh execution that is a dictionary.

**Observed Behavior:**
The raw **types.Content** is returned which failed to verify against the output_schema of the LLMagent node

**Environment Details:**

 - ADK Library Version (pip show google-adk): 2.0.0b1
 - Desktop OS:** Windows 10
 - Python Version (python -V): Python 3.13.4

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: gemini-3.1-flash-lite-preview

---

## 🟡 Optional Information
From what I understand about the codebase, during a child node's fresh run the events are enqueued in the event queue, and event in this queue are consumed by the runner where as stated earlier they are altered.
And before the events are put in the event queue they are tracked in the context of the child node in the function **_track_event_in_context()** which puts **event.output** into child nodes **ctx.output.** 
```python 
  def _track_event_in_context(self, event: Event, ctx: Context) -> None:
    """Write yielded event results to ctx (source of truth)."""
    if event.output is not None:
      ctx.output = event.output
    elif event.node_info and event.node_info.message_as_output:
      ctx.output = event.content
# ..... rest of the function 
```
But I suppose this function intends to make the 
```python
ctx.output = event.content
```
  instead of 

```python 
ctx.output = event.output
```
But since it can not , because in  **process_llm_agent_output()** function it puts the **event.output = output**
hence, populating the event.output.

So there could additionally be one more fix by either:
1) Assign ctx.output = event.content (off course after processing the content, maybe using the process_llm_agent_output() function) 

or 

2) If we strictly want the output field to be None, even during the fresh execution of the node, remove **process_llm_agent_output()** function to process the output in the **run_llm_agent_as_node()** method.

**Screenshots / Video:**
Image

**Logs:**
```Traceback (most recent call last):
  File "D:\Projects\googleadk\.venv\Lib\site-packages\google\adk\workflow\_node_runner.py", line 127, in run
    await self._execute_node(ctx, node_input)
  File "D:\Projects\googleadk\.venv\Lib\site-packages\google\adk\workflow\_node_runner.py", line 250, in _execute_node
    await self._run_node_loop(ctx, node_input)
  File "D:\Projects\googleadk\.venv\Lib\site-packages\google\adk\workflow\_node_runner.py", line 264, in _run_node_loop
    async for event in agen:
      self._track_event_in_context(event, ctx)
      await self._enqueue_event(event, ctx)
  File "D:\Projects\googleadk\.venv\Lib\site-packages\google\adk\workflow\_base_node.py", line 218, in run
    async for item in agen:
    ...<12 lines>...
        yield Event(output=validated)
  File "D:\Projects\googleadk\.venv\Lib\site-packages\google\adk\workflow\_function_node.py", line 508, in _run_impl
    async for item in items:
    ...<2 lines>...
        yield event
  File "D:\Projects\googleadk\agent\agent.py", line 78, in main_workflow
    agent_output = Action.model_validate(agent_output)
  File "D:\Projects\googleadk\.venv\Lib\site-packages\pydantic\main.py", line 732, in model_validate
    return cls.__pydantic_validator__.validate_python(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        obj,
        ^^^^
    ...<5 lines>...
        by_name=by_name,
        ^^^^^^^^^^^^^^^^
    )
    ^
pydantic_core._pydantic_core.ValidationError: 1 validation error for Action
  Input should be a valid dictionary or instance of Action [type=model_type, input_value=Content(
  parts=[
    Pa...),
  ],
  role='model'
), input_type=Content]
```

**Screenshots / Video:**

Image

**Additional Context:**
'N/A

```python
from google.adk import Workflow
from google.adk.events import RequestInput
from google.adk import Context
from google.adk.workflow import node

from typing import Any
class MySchema(BaseModel):
    greeting: str

analyzer = Agent(
    name="my_agent",
    model="gemini-3.1-flash-lite-preview",
    output_schema=MySchema,
    instruction="Your job is to greet the user.",
)
@node(rerun_on_resume=False)
async def get_user_approval(ctx: Context, node_input: Any):
    """Yields a RequestInput to pause the workflow and wait for user input."""
    yield RequestInput(message=f'please approve or reject.',response_schema = str)

@node(rerun_on_resume=True)
async def handle_process(ctx: Context, node_input: Any):
    """The orchestrator calling the interactive step."""
    user_response = await ctx.run_node(get_user_approval,node_input)
    if user_response.lower() == "yes":
        yield 'approved'
    yield "Denied"
    return

@node(rerun_on_resume=True)
async def my_workflow(ctx:Context):
    agent_input = "my agent input"
    agent_output = await ctx.run_node(analyzer,agent_input)
    agent_output = MySchema.validate(agent_output)
    result = await ctx.run_node(handle_process)
    yield result

root_agent = Workflow(
    name="greet_user",
    edges=[("START", my_workflow)]
)
```
**Screenshots / Video:**
Image

 - Always (100%)

ValentinCordonnier:
Fixes #5553

---

### Summary

`Runner._consume_event_queue` was nulling `event.output` before persisting any non-partial event whose `node_info.message_as_output` was `True`. The intent looked like a denormalization step (\"the LLM text already lives in `event.content`, no need to store it twice\"), but it broke resume rehydration. On the live path, `process_llm_agent_output` populates `event.output` with the validated dict (or raw text, when no `output_schema`); after the strip, the persisted event lost that value. On resume, `_reconstruct_node_state` walked the persisted events, found `event.output is None`, and fell through to `child.output = event.content` — handing the orchestrator a raw `genai.types.Content` instead of the dict. Anything subscripting that result (`lab[\"findings\"]`, `Schema.model_validate(out)`, …) crashed with `'Content' object is not subscriptable` only after a HITL pause, even though the same code worked on the initial pass.

The diagnosis in #5553 by @samarth1224 already pinned the two functions involved; this PR is the minimal change.

### Changes

- **`src/google/adk/runners.py`**: removed the four-line strip block in `_consume_event_queue` that copied the event and assigned `event.output = None` for non-partial `message_as_output` events. With the strip gone, `event.output` survives the persist / rehydrate roundtrip and `_reconstruct_node_state`'s primary branch (`if event.output is not None: child.output = event.output`) returns the same Python object type the live path produced.
- **`tests/unittests/workflow/test_workflow_hitl.py`**: added a regression test that pauses a `rerun_on_resume=True` orchestrator on a `RequestInput`, resumes, and subscripts the cached `LlmAgent` output. Fails on `v2` HEAD with the reported `TypeError`; passes after this change.
- **`tests/unittests/workflow/test_workflow_dynamic_nodes.py`**: updated `test_workflow_resume_does_not_rerun_completed_llm_agent`. Its previous assertion `agent_event.output is None` (commented \"Verify that runners.py cleared the output\") was asserting the buggy behavior as if intentional. Replaced with a positive check that the persisted output equals the agent's emitted text. The test's actual semantic guarantee — the LLM is not re-fired on resume — still holds via the existing `agent_runs_again == []` assertion.

### Motivation

Any `@node(rerun_on_resume=True)` orchestrator that subscripts an `LlmAgent` result via `ctx.run_node` and survives a `RequestInput` pause is silently broken on `v2`. Graph-edge propagation of `LlmAgent` outputs across resume has the same failure mode (the rehydrated `child.output` is a `Content` instead of the schema dict), so this isn't only a dynamic-orchestrator problem. The fix unblocks the entire HITL-with-structured-output pattern, which is a core `v2` use case.

### What does *not* change

- Live-path semantics. `_node_runner._enqueue_event` runs before this strip in the queue and was already reading `event.output` correctly; non-resume runs are unaffected.
- Caching policy. `rerun_on_resume`, dedup of completed nodes, and the `_dynamic_node_scheduler` state machine are all unchanged. The fix only changes *what value* gets returned from the cache, not whether the cache is hit.
- The `event.content` fallback in `_reconstruct_node_state` is preserved (it now handles only legacy persisted events).
- No public API surface changes. `Event.output: Any | None` is unchanged; no contract said it had to be `None` for LLM events.

### Downsides

- **Slightly larger persisted events for `LlmAgent` final responses.** The validated dict is stored alongside the original text in `content.parts`. For `output_schema` agents that's roughly the JSON serialization of the dict; for plain-text agents it's roughly the response text duplicated. The previous saving was illusory — rehydration was returning a different object type than the live path, which is the bug we're fixing — but storage-cost-sensitive deployments will see the bytes-per-event grow.
- **Plugin behavior change.** Plugins that read `event.output` for LLM events used to see `None`; they will now see the dict / string. Any plugin that branched on `event.output is None` to detect \"this is the final LLM event\" needs to switch to `event.node_info.message_as_output` (unchanged, set by the live path). I couldn't find any in-tree plugin doing this, but third-party plugins exist.
- **Old persisted sessions stay broken.** Sessions persisted by pre-fix runners still have `output=None` for these events; resuming them post-fix continues to hit the `event.content` fallback. That's not a regression — those sessions were broken before — but cross-version resume is not magically fixed. The fallback branch could be removed in a follow-up once legacy sessions age out.
- **One existing test asserted the buggy behavior** and had to be updated (see the third bullet under *Changes*). Reviewers should sanity-check that the replacement assertion still expresses the test's true semantic intent.

### DIFF
--- a/src/google/adk/runners.py
+++ b/src/google/adk/runners.py
@@ -611,10 +611,6 @@ class Runner:
       if event_or_done is done_sentinel:
         break
       event: Event = event_or_done
-      if not event.partial:
-        if event.node_info.message_as_output and event.content is not None:
-          event = event.model_copy()
-          event.output = None
 
       _apply_run_config_custom_metadata(event, ic.run_config)
       modified_event = await ic.plugin_manager.run_on_event_callback(
--- a/tests/unittests/workflow/test_workflow_dynamic_nodes.py
+++ b/tests/unittests/workflow/test_workflow_dynamic_nodes.py
@@ -1153,8 +1153,9 @@ async def test_workflow_resume_does_not_rerun_completed_llm_agent():
   assert len(agent_events) > 0
   agent_event = agent_events[-1]
 
-  # Verify that runners.py cleared the output
-  assert agent_event.output is None
+  # The persisted event keeps the LlmAgent's output so resume can replay
+  # it without re-firing the model
+  assert agent_event.output == 'LLM output content'
 
   # When the workflow is resumed by resolving the interrupt
   resume_events = await _resume(
--- a/tests/unittests/workflow/test_workflow_hitl.py
+++ b/tests/unittests/workflow/test_workflow_hitl.py
@@ -2051,3 +2051,52 @@ async def test_parallel_nodes_trigger_same_hitl_node(
   )
   outputs3 = workflow_testing_utils.get_outputs(events3)
   assert outputs3 == [f'c_from_b_response 2']
+
+
+@pytest.mark.asyncio
+async def test_cached_llm_agent_output_remains_subscriptable_after_resume(
+    request: pytest.FixtureRequest,
+):
+  """Cached LlmAgent output must stay a dict on resume."""
+
+  class Greeting(BaseModel):
+    text: str
+
+  greeter = LlmAgent(
+      name='greeter',
+      model=testing_utils.MockModel.create(
+          responses=[types.Part.from_text(text='{"text": "hi"}')],
+      ),
+      mode='single_turn',
+      output_schema=Greeting,
+      instruction='Return JSON.',
+  )
+
+  @node(rerun_on_resume=True)
+  async def orchestrator(ctx: Context, node_input: Any):
+    out = await ctx.run_node(greeter, 'go')
+    if 'confirm' not in ctx.resume_inputs:
+      yield RequestInput(message='ok?', interrupt_id='confirm')
+      return
+    yield Event(output={'echoed': out['text']})
+
+  agent = Workflow(name='wf', edges=[(START, orchestrator)])
+  app = App(
+      name=request.function.__name__,
+      root_agent=agent,
+      resumability_config=ResumabilityConfig(is_resumable=True),
+  )
+  runner = testing_utils.InMemoryRunner(app=app)
+
+  events1 = await runner.run_async(testing_utils.get_user_content('start'))
+  req = workflow_testing_utils.get_request_input_events(events1)[0]
+  interrupt_id = get_request_input_interrupt_ids(req)[0]
+  invocation_id = events1[0].invocation_id
+
+  events2 = await runner.run_async(
+      new_message=testing_utils.UserContent(
+          create_request_input_response(interrupt_id, {'approved': True})
+      ),
+      invocation_id=invocation_id,
+  )
+  assert {'echoed': 'hi'} in workflow_testing_utils.get_outputs(events2)


samarth1224 (commented):
### Key Points:
**Does this Fix the issue?**
- Yes

**Is this a correct way to fix this?**
- I believe no


Hey, while your changes fixes the issue, but as far as my knowledge goes of this codebase, I believe this might not be the correct way to fix the issue.

As mentioned in the my issue, in your PR, in the test cases, and  also as inferred from the codebase, **event.output** should ideally remain **None** when the **message_as_input == True**. Since, output can be inferred from the  **event.content.**

- And, therefore, I believe the change required must be made within the function     **_reconstruct_node_states()** and not the **_consume_event_queue().**

- Specifically, the change must be either a helper function to process the **event.content** to suitable output before assigning to **child.output** OR direct extraction of the output during the assignment to **child.output.** 

I feel a helper function would be a better choice where in the function we could also validate output(stored in **event.content**) against the agent's output schema ( Validation might not be needed here, since output has already been validated during **Fresh Run** by the **process_llm_agent_output()** function. ).

### In nutshell:
- I believe, your changes defeat the purpose of **message_as_output** attribute of NodeInfo class.
- Your changes need to be accompanied with additional changes in multiple files and functions, including **_reconstruct_node_states()**, **_track_event_in_context()** etc. so that codebase remains consistent.
- The issue could simply be resolved by processing the **event.content** before assigning it to **child.output** in **_reconstruct_node_states().**

Thanks.

rohityan:
Hi @ValentinCordonnier , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh and also the unit tests.

PR #5596, opened 2026-05-05
feat(sessions): add get_user_state(app_name, user_id) to BaseSessionService

### LINKED ISSUE #5592: feat(sessions): add get_user_state(app_name, user_id) to BaseSessionService
## Problem

`BaseSessionService` has no public method to read user-scoped state (`user_states` table) without an active `session_id`.

The only workaround today is calling `list_sessions` and inspecting events, which is expensive and not intended for this use case.

This creates a practical gap for callers that need to **bootstrap user context before a new session exists** — a pattern that arises whenever long-lived, cross-session data (e.g. user preferences, cached identity) must be seeded into agents without paying the cost of a full session lookup.

A concrete example: an authentication layer that resolves user identity from an external service and wants to skip the resolution if a fresh snapshot is already stored in `user_states` for that user. Because `get_session` returns `None` for a brand-new `session_id` — even when `user_states` already has data — callers are forced to maintain a separate process-level in-memory cache as a workaround, which defeats the purpose of persisting state to the database.

## Proposed solution

Add a `get_user_state(app_name, user_id)` method to `BaseSessionService`:

```python
async def get_user_state(
    self, *, app_name: str, user_id: str
) -> dict[str, Any]:
    ...
```

**Contract:**
- Returns the raw user-scoped dictionary, **without** the `user:` prefix on keys (consistent with how user state is stored internally).
- Returns `{}` when no user state exists for the given `(app_name, user_id)`.
- The default implementation in `BaseSessionService` raises `NotImplementedError` to preserve backward compatibility for existing custom subclasses.

**Implementations covered:**
| Service | Approach |
|---|---|
| `InMemorySessionService` | Reads from `self.user_state[app_name][user_id]` |
| `DatabaseSessionService` | Queries `StorageUserState` by `(app_name, user_id)` with `read_only=True` |
| `SqliteSessionService` | Delegates to the existing `_get_user_state` helper |
| `VertexAiSessionService` | Raises `NotImplementedError` (Vertex AI Agent Engine API does not expose user state independently of a session) |

nicolasmota:

Closes: #5592

**Problem:**
`BaseSessionService` has no public method to read user-scoped state without an active `session_id`. Callers that need to bootstrap user context before a new session exists are forced to call the expensive `list_sessions` or maintain a separate process-level cache as a workaround.

**Solution:**
Add `get_user_state(app_name, user_id) -> dict[str, Any]` to `BaseSessionService`. Implemented in `InMemorySessionService`, `DatabaseSessionService`, and `SqliteSessionService`. `VertexAiSessionService` raises `NotImplementedError` because the Vertex AI Agent Engine API does not expose user state independently of a session. The default in `BaseSessionService` also raises `NotImplementedError` to preserve backward compatibility for existing custom subclasses.

Keys are returned **without** the `user:` prefix, consistent with how user state is stored internally (the prefix is applied by the state-merging layer).

---

### Testing Plan

**Unit Tests:**

New test file: `tests/unittests/sessions/test_get_user_state.py` — 19 test cases parameterised across `IN_MEMORY`, `DATABASE`, and `SQLITE` backends covering:

- Returns `{}` when no state exists
- Returns only user-scoped keys (session keys are excluded)
- State is isolated per user — different users see `{}`
- State is isolated per app — different apps see `{}`
- **Core use case:** state readable before a new `session_id` is created
- Latest write wins (overwrite semantics)
- `VertexAiSessionService` raises `NotImplementedError`

platform darwin -- Python 3.11.12, pytest-9.0.2 19 passed in 2.33s

**Manual E2E Tests:**

Not applicable for this change — the method is a read-only query on existing storage with no behavioural side-effects on session or event handling. Existing E2E tests are unaffected.

---
### Checklist


rohityan:
Hi @nicolasmota , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix failing mypy-diff tests

nicolasmota:
done @rohityan!

rohityan:
Hi @DeanChensj  , could you please review this.

src/google/adk/sessions/base_session_service.py
```diff
@@ -111,6 +111,46 @@ async def delete_session(
   ) -> None:
     """Deletes a session."""
```
DeanChensj, line 112: We also need to update PerAgentDatabaseSessionService https://github.com/google/adk-python/blob/ae95a972280cbf0fcd6a989e19590cad68f3d847/src/google/adk/cli/utils/local_storage.py#L114

tests/unittests/sessions/test_get_user_state.py
DeanChensj, line 1: This can be moved to test_session_service

nicolasmota: done ✅

src/google/adk/sessions/base_session_service.py
nicolasmota: great! done ✅

PR #5640, opened 2026-05-08
fix(firestore): preserve update timestamps in list_sessions

kkj333:
### Link to Issue or Description of Change
**1. Link to an existing issue (if applicable):**
Closes: #5632

### Problem
`FirestoreSessionService.list_sessions()` always sets `Session.last_update_time` to `0.0`, even when Firestore documents contain `updateTime`.
### Solution
Use the same timestamp conversion logic as `get_session()` in `list_sessions()`:
- if `updateTime` is a `datetime`, use `.timestamp()`
- otherwise, attempt `float(updateTime)`
- fallback to `0.0` when conversion fails
This keeps list responses consistent with `get_session()` and preserves accurate update timestamps.
### Testing Plan
**Unit Tests:**
- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.
Updated tests:
- `test_list_sessions_with_user_id`
- `test_list_sessions_without_user_id`
Local pytest summary:
- `tests/unittests/integrations/firestore/test_firestore_session_service.py`
- `17 passed`
**Manual End-to-End (E2E) Tests:**
Reproduced with a mocked Firestore client where session docs include `updateTime`; before fix `last_update_time` was `0.0`, after fix it matches the Firestore value.
### Checklist
- [x] I have read the [CONTRIBUTING.md](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md) document.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have manually tested my changes end-to-end.
- [ ] Any dependent changes have been merged and published in downstream modules.

PR #5642, opened 2026-05-08
fix(firestore): populate last_update_time in list_sessions from Firestore updateTime

### LINKED ISSUE #5632: Firestore Session Service does not handle timestamp correctly
## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. Requests without complete information may be rejected / deprioritized.
If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**

**Steps to Reproduce:**
1. Use `google.adk.integrations.firestore.firestore_session_service.FirestoreSessionService` as sessions service.
2. Call `session_service.list_sessions`
3. 
**Expected Behavior:**
The returned session's `last_update_time` should match the Firestore record

**Observed Behavior:**
The returned session's `last_update_time` is always 0.

**Environment Details:**

 - ADK Library Version (pip show google-adk): 1.20
 - Desktop OS:** [e.g., macOS, Linux, Windows]: Linux
 - Python Version (python -V): 3.14

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: gemini

---

## 🟡 Optional Information

The issue is the code does not handle the attribute.

[firestore_session_service.py](https://github.com/google/adk-python/blob/3117e0913689960dc951035bb6ca7b8551cec4d3/src/google/adk/integrations/firestore/firestore_session_service.py#L421-L422)

**Regression:**
Did this work in a previous version of ADK? If so, which one?

**Logs:**
Please attach relevant logs. Wrap them in code blocks (```) or attach a
text file.
```text
// Paste logs here
```

**Screenshots / Video:**
If applicable, add screenshots or screen recordings to help explain
your problem.

**Additional Context:**

```python
// Code snippet here
```

 - Always (100%)

nileshpatil6:
Fixes #5632

## Problem

`list_sessions()` hardcodes `last_update_time=0.0` in every returned `Session`, silently ignoring the `updateTime` field stored in the Firestore document. This means callers can never sort or filter sessions by last-update time when using `FirestoreSessionService`.

`get_session()` already handles this correctly:

```python
# Convert timestamp
update_time = data.get("updateTime")
last_update_time = 0.0
if update_time:
    if isinstance(update_time, datetime):
        last_update_time = update_time.timestamp()
    else:
        try:
            last_update_time = float(update_time)
        except (ValueError, TypeError):
            pass
```

`list_sessions()` was missing this block entirely.

## Fix

Apply the same timestamp conversion in the `list_sessions()` loop, immediately before constructing each `Session`. The `datetime` import is already present at the top of the file.

## Behavior

- **Before:** `session.last_update_time` is always `0.0` for every session returned by `list_sessions()`
- **After:** `session.last_update_time` reflects the actual Firestore `updateTime` (a POSIX timestamp float)

No other behavior changes -- the fix is additive and cannot break existing callers since `0.0` was always wrong.

### DIFF
--- a/src/google/adk/integrations/firestore/firestore_session_service.py
+++ b/src/google/adk/integrations/firestore/firestore_session_service.py
@@ -411,6 +411,18 @@ class FirestoreSessionService(BaseSessionService):  # type: ignore[misc]
         u_state = user_states_map.get(u_id, {})
         merged = self._merge_state(app_state, u_state, s_state)
 
+        # Convert timestamp (mirrors the logic in get_session).
+        update_time = data.get("updateTime")
+        last_update_time = 0.0
+        if update_time:
+          if isinstance(update_time, datetime):
+            last_update_time = update_time.timestamp()
+          else:
+            try:
+              last_update_time = float(update_time)
+            except (ValueError, TypeError):
+              pass
+
         sessions.append(
             Session(
                 id=data["id"],
@@ -418,7 +430,7 @@ class FirestoreSessionService(BaseSessionService):  # type: ignore[misc]
                 user_id=data["userId"],
                 state=merged,
                 events=[],
-                last_update_time=0.0,
+                last_update_time=last_update_time,
             )
         )
 


kkj333:
Thanks for working on this! This seems to overlap with #5640 (opened earlier),
which includes the same fix and unit tests.
Could we consolidate into one PR to avoid duplication?

nileshpatil6:
Thanks for the pointer @kkj333 — closing this in favour of #5640 which was opened first and already includes unit tests. Good to see this getting fixed!

Issue #18, opened 2025-04-10
No Agent found


mehulGupta7991:
![Image](https://github.com/user-attachments/assets/59c4eee8-27a0-4447-8635-7f477cac99ed)
Everything is done as mentioned in readme.md

Jacksunwei:
Could you check out https://google.github.io/adk-docs/get-started/quickstart/#run-your-agent and see if you run `adk web` in the "parent_folder"?

ShyamSathish005:
Well, I had this issue earlier,  because there is no agents are available in the folder. It can be fixed by creating a folder and run the `init.py` code or adding sample agents  will fix it.

hangfei:
Could you try the above suggestions? Closing the issue now. Feel free to re-open if the issue persists.

infiblox:
> Could you try the above suggestions? Closing the issue now. Feel free to re-open if the issue persists.

same error.  what should be content in init.py

danhong1990:
I had a same issue. 
When I run in the console akd run .  it works perfect 

But in the same directory, if if 'adk web' this samething happens

spent hours trying to fix this but this fixed my issue.

> Could you check out https://google.github.io/adk-docs/get-started/quickstart/#run-your-agent and see if you run `adk web` in the "parent_folder"?

kevinhongzl:
> I had a same issue. When I run in the console akd run . it works perfect
> 
> But in the same directory, if if 'adk web' this samething happens

Same here.
I think it would be better if `adk web` and `adk run` started in the same way.

pgzmnk:
Ran into the same issue. From a given directory, the command `adk run` worked but `adk web` didn't. Consistency would be good to have.

angelaaaateng:
Had this same issue as well, following the readme here: https://google.github.io/adk-docs/agents/config/#build-an-agent

Edit: This worked for me after going to the parent directory of `my_agent` instead of inside the folder

zhengpenghu:
This fixed my issue.

The most common cause is running the command from the wrong directory or specifying the path incorrectly. 
Run adk web from the parent directory of your agent's code folder.
Incorrect: adk web my_agent/ (if my_agent is the agent folder)
Correct: adk web . (run from the directory containing the my_agent folder) or adk web my_project if your structure is my_project/my_agent/.

rgddos:
> Had this same issue as well, following the readme here: https://google.github.io/adk-docs/agents/config/#build-an-agent
> 
[...]

Issue #17, opened 2025-04-10
launch the Web UI failed with Black page


cfdcfd:
start with quick start example and start with `adk web` success, http://127.0.0.1:8000/dev-ui is nothing without Black page.

successfully using `adk run` to try some work.

any suggestion about using  `adk web`??

ps: i am sure i using  `adk web` in correct path

Jacksunwei:
`adk web` should work out of box if you following the step 4 in https://google.github.io/adk-docs/get-started/quickstart/#setup-the-model

Could you share some screenshot and your setup environment, e.g. windows/linux/mac, python version?

cfdcfd:
![Image](https://github.com/user-attachments/assets/b55a1d01-580b-4372-8846-d0d9aeb644c3)
windows-python 3.12.4

cfdcfd:
windows-3.12.4
![Image](https://github.com/user-attachments/assets/472c777d-40c6-4ca1-a44f-7dd0b3cabe94)

when i open .venv\Lib\site-packages\google\adk\cli\browser\index.html directly in chorme, 

![Image](https://github.com/user-attachments/assets/c76dd3c8-2e42-4d02-a3ca-6101cf8eb65b)

cfdcfd:
![Image](https://github.com/user-attachments/assets/56dec349-64e4-419d-b561-cb442fe22ec2)
request suggests no error

cfdcfd:
![Image](https://github.com/user-attachments/assets/df9f3b9e-2f2c-4f17-960d-d7270c357f85)
`adk run` success, so the google api key has no error
is `adk web` has any other dependency whick not locally?

cfdcfd:
![Image](https://github.com/user-attachments/assets/4ad7f892-6158-494f-a4a7-698167982885)

or this console error may have any help?

cfdcfd:
a front-end error, main-SLIAU2JL.js content-type is text/plain
Uvicorn relys on Windows Registry, every thing ok after correct the Windows Registry
reference:
https://github.com/pallets/flask/issues/1045

nghiakvnvsd:
Hi @cfdcfd  I have the same issue as you and tried to fix it by modifying the Windows Registry, but it still failed. Can you help me show step-by-step how did you fix that? Many thanks!

Issue #16, opened 2025-04-10
Google adk intentionally blocking other models


rong-xyz:
![Image](https://github.com/user-attachments/assets/68a50798-dfa8-4047-9f7a-7d66e5605f55)
```
    if llm_request.model and llm_request.model.startswith('gemini-2'):
      ...
    else:
      raise ValueError(
          f'Code execution tool is not supported for model {llm_request.model}'
      )
```
This is a joke right

Jacksunwei:
This is the Gemini 2 builtin code_executor that isn't available in other models. We also have model-agnostic code executor classes. You could check out in `code_executors` package.

rong-xyz:
Is this open source project's open the same as openai's open?
If other models are incapable of following the instruction/ formatting, it should be a parsing error or something else.
Not `if not gemini raise error`

hangfei:
There might be some confusions on what's this code executor is.

As Jack pointed out, there are two types of code executors. One type of code executor is built into model itself. That's what Gemini-2 series support and even other variants of Gemini doesn't support. I believe the screenshot you shared is for this 1st type. The other type is general code executors that are model-agnostic, they are in this [package](https://github.com/google/adk-python/tree/eecb9e65be752b2f0db06f43f93e4e868dd0a4bb/src/google/adk/code_executors). Even for this second type of code executors, we provide several variants to choose.  Generally, we expect users to use generic code executors in this package and it's also a general industry practice.

Do you have a use case and model that you discovered doesn't work for the 2nd type of code executors? If you see some bugs or feature gaps while trying it out, we are happy to help look into them.

Issue #15, opened 2025-04-10
bug: text overflows in adk web


gamedevCloudy:
- When the model outputs long strings, the text overflows outside its container when using adk web 

![Image](https://github.com/user-attachments/assets/1eaee382-7ac3-4a8b-a5f5-a4a86ded948c)

Jacksunwei:
Thanks for reporting, Aayush! This is a known issue that we're fixing.

hangfei:
This is fixed and will roll out in next release.

PR #5651, opened 2026-05-10
fix: terminate infinite retry loop in LoadSkillResourceTool on RESOURCE_NOT_FOUND

### LINKED ISSUE #5652: [Bug] LoadSkillResourceTool retries RESOURCE_NOT_FOUND indefinitely; default max_llm_calls=500 is the only backstop
## 🔴 Required Information

**Describe the Bug:**

`LoadSkillResourceTool.run_async` returns `RESOURCE_NOT_FOUND` as a structured soft-error string when a path passed by the LLM does not exist inside the skill's bundled resources. Because the response is a normal tool result (not an exception or terminal signal), the LLM treats it as a transient/recoverable failure and retries — but critically, **it hallucinates a different plausible path on every retry**, not the same path. Nothing in `SkillToolset` tracks total failures across paths, so the loop continues until `RunConfig.max_llm_calls` is exhausted.

`max_llm_calls` defaults to **500** (`src/google/adk/agents/run_config.py:314`). This means a single invocation can silently consume the entire per-invocation call budget on repeated failing tool calls before the framework intervenes — and `max_llm_calls` is a global cap on legitimate reasoning, not a defense against a repeated-failure loop on one specific tool.

**Steps to Reproduce:**

1. Install `google-adk` (any version that ships `SkillToolset` — verified on `1.32.0`).
2. Create an agent with a `SkillToolset` containing a skill whose `SKILL.md` references files by natural-language names (e.g. "Document 1", "the reference guide") without exact filenames.
3. Issue a query that prompts the model to read one of those resources.
4. Observe in the trace that the model calls `load_skill_resource` with a hallucinated path, receives `RESOURCE_NOT_FOUND`, then calls it again with a **different** hallucinated path, receives `RESOURCE_NOT_FOUND` again, and loops.

**Expected Behavior:**

After the first `RESOURCE_NOT_FOUND` within an invocation, any subsequent `load_skill_resource` failure should return a terminal error code that unambiguously instructs the LLM to stop retrying and report the error. The agent's overall reasoning budget (`max_llm_calls`) should not be the only thing standing between an imperfect prompt and a runaway invocation.

**Observed Behavior:**

The same `RESOURCE_NOT_FOUND` soft error is returned on every attempt regardless of path or how many times it has already failed. The loop terminates only when `max_llm_calls` is exceeded.

**Live trace evidence** (captured via `GET /debug/trace/session/{session_id}` against `adk web`):

```
SPAN: execute_tool load_skill_resource
  args:       {'file_path': 'references/reference_doc.md', 'skill_name': 'document-classifier'}
  error_code: RESOURCE_NOT_FOUND
  error:      Resource 'references/reference_doc.md' not found in skill 'document-classifier'.

SPAN: execute_tool load_skill_resource
  args:       {'skill_name': 'document-classifier', 'file_path': 'references/Document1.md'}
  error_code: RESOURCE_NOT_FOUND
  error:      Resource 'references/Document1.md' not found in skill 'document-classifier'.
```

The model tried `references/reference_doc.md` first, then hallucinated a completely different path (`references/Document1.md`) on the retry. Both returned the same soft error — the LLM had no signal to stop. This pattern continues indefinitely.

**Environment Details:**

- ADK Library Version: `1.32.0` (issue exists on `main` as of commit `2d61cb69`)
- Desktop OS: Linux (defect is in framework logic, not OS-specific)
- Python Version: `3.12.3`

**Model Information:**

- Are you using LiteLLM: N/A (defect is provider-agnostic)
- Which model: `gemini-3-flash-preview` (observed; reproducible across any function-calling model — the retry behavior is a consequence of the soft error signal, not model-specific)

---

## 🟡 Optional Information

**Regression:**

Not a regression. The defect has existed since `SkillToolset` was introduced — `LoadSkillResourceTool.run_async` has never had any retry-guard logic.

**Additional Context:**

Four factors combine to make this loop reachable through ordinary use:

1. **No resource manifest at L2** — the `load_skill` response intentionally omits available file paths (progressive-disclosure spec). The LLM must infer paths from prose, and inferred paths are routinely wrong.
2. **Soft error string** — `RESOURCE_NOT_FOUND` looks transient and recoverable to the model; retry is its default response.
3. **No terminal signal** — nothing escalates after the first miss.
4. **No scope boundary in default prompt** — the system instruction doesn't distinguish skill-bundled files from runtime user inputs (e.g. a PDF the user is processing), so the model sometimes routes runtime documents through `load_skill_resource` and loops on them.

Considered and rejected alternatives:

| Alternative | Why not |
|---|---|
| Per-path retry guard | LLM hallucinates a different path on each retry — confirmed in live trace; a per-path list never triggers |
| Tighten or default-lower `max_llm_calls` | Caps overall reasoning budget; punishes legitimate long-running agents |
| User-side `after_tool_callback` workaround | Symptomatic; pushes the fix onto every `SkillToolset` user |
| Add `available_resources` manifest to L2 `load_skill` | Defeats the lazy-loading / token-saving design |
| New `list_skill_resources` tool | Violates the L1→L2→L3 progressive disclosure contract |

```python
import asyncio
from unittest import mock
from google.adk.tools import skill_toolset, tool_context

skill = mock.create_autospec(models.Skill, instance=True)
skill.name = "demo"
skill.resources = mock.MagicMock()
skill.resources.get_reference.return_value = None  # every path "missing"

ctx = mock.MagicMock(spec=tool_context.ToolContext)
ctx.state = {}
ctx.invocation_id = "inv1"
ctx._invocation_context = mock.MagicMock()
ctx.agent_name = "agent"

toolset_obj = skill_toolset.SkillToolset([skill])
tool = skill_toolset.LoadSkillResourceTool(toolset_obj)

async def main():
    paths = [
        "references/missing.md",
        "references/other_guess.md",   # different path — LLM hallucination pattern
        "references/yet_another.md",
    ]
    for i, path in enumerate(paths):
        r = await tool.run_async(
            args={"skill_name": "demo", "file_path": path},
            tool_context=ctx,
        )
        print(i, r["error_code"])
    # On main (unpatched): all 3 print RESOURCE_NOT_FOUND — LLM has no reason to stop
    # With fix applied:    call 0 → RESOURCE_NOT_FOUND, calls 1-2 → RESOURCE_NOT_FOUND_FATAL

asyncio.run(main())
```

**How often has this issue occurred?:** Always (100%) — deterministic given any skill whose `SKILL.md` lets the model infer plausible-looking paths that don't literally exist.

---

## Proposed Fix

A two-layer fix is in linked PR #5651:

**Code**: an invocation-scoped **total failure counter** inside `LoadSkillResourceTool.run_async`. The counter tracks the number of `RESOURCE_NOT_FOUND` responses across **all paths** within an invocation (not per-path — live testing confirmed the LLM uses a different path on each retry). State key:

```
temp:_adk_skill_resource_not_found_count_<invocation_id>
```

- First failure → `RESOURCE_NOT_FOUND` (unchanged behavior).
- Any subsequent failure → `RESOURCE_NOT_FOUND_FATAL` with an explicit stop instruction and failure count.

The `temp:` prefix uses ADK's existing convention to prevent persistence to durable storage. The `<invocation_id>` suffix isolates in-memory backends where `temp:` keys are not auto-cleared between invocations.

**Prompt**: a no-retry rule and a scope boundary added to `_DEFAULT_SKILL_SYSTEM_INSTRUCTION`.

**Live trace with fix applied** (same session, patched build):

```
SPAN: execute_tool load_skill_resource
  args:       {'file_path': 'references/reference_doc.md', 'skill_name': 'document-classifier'}
  error_code: RESOURCE_NOT_FOUND
  error:      Resource 'references/reference_doc.md' not found in skill 'document-classifier'.

SPAN: execute_tool load_skill_resource
  args:       {'skill_name': 'document-classifier', 'file_path': 'references/Document1.md'}
  error_code: RESOURCE_NOT_FOUND_FATAL
  error:      Resource 'references/Document1.md' not found in skill 'document-classifier'.
              This is resource lookup failure #2 this invocation. Do not retry any path
              — report the error to the user and stop.
```

Loop terminated on the second call. The model attempted a different path (`Document1.md` vs `reference_doc.md`) — exactly the hallucination pattern that a per-path guard would have missed.

Linked PR: google/adk-python#5651

Raman369AI:
Fixes #5652

## Summary

Closes a latent defect in `SkillToolset` that lets the LLM enter an unbounded retry loop when `load_skill_resource` returns `RESOURCE_NOT_FOUND`. Because the only existing backstop is `RunConfig.max_llm_calls` (default **500**), a hallucinated path can quietly burn the entire per-invocation call budget before the framework intervenes — and `max_llm_calls` is a global cap on legitimate reasoning, not a defense against this specific failure mode.

This fix adds invocation-scoped termination to the tool itself, plus a complementary system-prompt rule, so the framework no longer relies on a perfect upstream prompt to avoid unbounded loops on a known error path.

## Why this matters

This is a structural problem, not an edge case:

- **The L2 `load_skill` response intentionally omits a manifest of available files** — that is the agentskills.io progressive-disclosure contract, and it is correct for token economy. But it means the LLM must *infer* paths from prose instructions inside `SKILL.md`. Inferred paths are routinely wrong, even with strong models.
- **`RESOURCE_NOT_FOUND` is returned as a structured soft string.** From the model's perspective it looks transient and recoverable, exactly like a flaky network error — so it retries. There is no terminal signal in the current implementation.
- **The LLM hallucinates a different path on every retry.** In live testing the model never repeated the same path twice — it generated a new plausible variant each time. A per-path guard is therefore insufficient; the termination must be invocation-wide.
- **Nothing in `SkillToolset` distinguishes "first miss" from "500th miss".** The loop terminates only when `max_llm_calls` (default 500) is hit.
- **Confounding scope ambiguity**: the default system prompt does not draw a boundary between *skill-bundled files* (the legitimate target of `load_skill_resource`) and *runtime user inputs* (e.g., a PDF the user is processing). A model asked to analyze a runtime document will sometimes route it through `load_skill_resource`, hit `RESOURCE_NOT_FOUND`, and loop — even though the file was never a skill resource.

## Live trace evidence

Captured via `GET /debug/trace/session/{session_id}` against `adk web` with a patched build:

```
SPAN: execute_tool load_skill_resource
  args:       {'file_path': 'references/reference_doc.md', 'skill_name': 'document-classifier'}
  error_code: RESOURCE_NOT_FOUND
  error:      Resource 'references/reference_doc.md' not found in skill 'document-classifier'.

SPAN: execute_tool load_skill_resource
  args:       {'skill_name': 'document-classifier', 'file_path': 'references/Document1.md'}
  error_code: RESOURCE_NOT_FOUND_FATAL
  error:      Resource 'references/Document1.md' not found in skill 'document-classifier'.
              This is resource lookup failure #2 this invocation. Do not retry any path
              — report the error to the user and stop.
```

The model tried `references/reference_doc.md` first, then hallucinated a completely different path (`references/Document1.md`) on the retry. The counter-based guard caught the second failure and returned `RESOURCE_NOT_FOUND_FATAL`, terminating the loop. A per-path guard would have missed this entirely.

## What changed

Two layers in one file (`src/google/adk/tools/skill_toolset.py`), plus tests:

### 1. Code: invocation-scoped failure counter in `LoadSkillResourceTool.run_async`

A total `RESOURCE_NOT_FOUND` count (across **all paths**) is tracked on `tool_context.state` under the key:

```
temp:_adk_skill_resource_not_found_count_<invocation_id>
```

- The `temp:` prefix uses ADK's existing convention so the value is trimmed from the persisted event delta and never reaches durable session storage.
- The `<invocation_id>` suffix ensures correctness on **in-memory** session backends, where `temp:` keys are added to `session.state` and are not auto-cleared between invocations. Without the suffix, a failure in invocation A would spuriously escalate the first attempt in invocation B.

Behavior:
- **First** `RESOURCE_NOT_FOUND` within an invocation (any path) → returns `RESOURCE_NOT_FOUND` (unchanged).
- **Any subsequent** `RESOURCE_NOT_FOUND` within the same invocation → returns the new `RESOURCE_NOT_FOUND_FATAL` error code with an explicit "do not retry — report the error and stop" message. The counter includes the failure number so the LLM has unambiguous context.

The guard is invocation-scoped, path-agnostic (catches hallucinated variants), and adds no overhead on the success path.

### 2. Prompt: two additions to `_DEFAULT_SKILL_SYSTEM_INSTRUCTION`

- **No-retry rule**: *"If `load_skill_resource` returns any error, do not retry the same path. Report the error to the user and stop."*
- **Scope boundary** added to rule 3: clarifies that `load_skill_resource` is **only** for skill-bundled files in `references/`, `assets/`, or `scripts/` — not for runtime user input. This addresses the secondary failure mode where the model confuses a runtime document with a skill resource.

### Why both layers

Defense-in-depth. Code-only termination produces confusing downstream behavior when the LLM has no semantic reason to stop. Prompt-only termination relies on the LLM following the rule, which imperfect upstream prompts can override. Together they are robust.

## Considered and rejected

| Alternative | Why not |
|---|---|
| Per-path retry guard | LLM hallucinates a different path on every retry — a per-path list is always empty at each attempt and never escalates |
| Tighten or default-lower `max_llm_calls` | Caps the agent's overall reasoning budget; punishes legitimate long-running tasks; does not address the specific defect |
| Symptomatic `after_tool_callback` workaround | Pushes the fix onto every user of `SkillToolset`; the framework still ships with the loop |
| Add a full `available_resources` manifest to the L2 `load_skill` response | Defeats the lazy-loading / token-saving design that the Skills spec is built around |
| Introduce a new `list_skill_resources` tool | Violates the L1→L2→L3 progressive disclosure contract from agentskills.io |
| Include available paths on the fatal response | Re-introduces the manifest cost; contradicts the "stop" semantic the fatal code is meant to enforce |

## Test plan

New tests in `tests/unittests/tools/test_skill_toolset.py`:

- `test_load_resource_repeated_failure_escalates_to_fatal` — second call on the same missing path within an invocation returns `RESOURCE_NOT_FOUND_FATAL`.
- `test_load_resource_different_path_also_escalates_to_fatal` — second call on a **different** missing path also returns `RESOURCE_NOT_FOUND_FATAL`; proves the counter is path-agnostic and catches the hallucinated-variant pattern observed in live testing.
- `test_load_resource_failures_isolated_per_invocation` — failures from invocation A do not escalate the first attempt in invocation B even when sharing the same session state dict.
- `test_load_resource_counter_uses_temp_prefix` — every key written by the guard starts with `temp:` so it is trimmed from the persisted event delta and never reaches durable storage.
- `test_load_resource_first_missing_returns_soft_error` — first `RESOURCE_NOT_FOUND` in an invocation returns the unchanged soft error code and message.

Verification:
- All 85 tests pass (81 pre-existing + 4 updated guard tests + 1 new soft-error coverage test), 0 regressions.
- `pyink --check` clean on both modified files.
- `isort --check-only` clean on both modified files.
- `mypy src/google/adk/tools/skill_toolset.py` reports the same 17 pre-existing errors as `main`; **zero** new errors introduced.

## Backwards compatibility

- The first-failure behavior is unchanged: same `RESOURCE_NOT_FOUND` error code, same error string. Existing callers see no difference.
- The new `RESOURCE_NOT_FOUND_FATAL` code is purely additive.
- The new state key uses the `temp:` prefix and is guaranteed not to leak into persisted session storage.


Raman369AI:
## E2E Reproduction Evidence — Runner

This comment adds live end-to-end evidence as required by [CONTRIBUTING.md § Manual End-to-End Tests (Runner)](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md#manual-end-to-end-e2e-tests).

---

### Minimal Reproduction Agent

```python
# test_agent/agent.py
from google.adk.tools.skill_toolset import SkillToolset

greeting_skill = models.Skill(
    frontmatter=models.Frontmatter(
        name="document-classifier",
        description=(
            "Classifies the document supplied by the user based on the reference document"
        ),
    ),
    instructions=(
        "Use Document 1 as reference document and Document 2 as input. Classify each document as"
        " python-code, non-python-code, or mixed. Return a short comparison."
    ),
)
root_agent = Agent(
    model="gemini-3-flash-preview",
    name="root_agent",
    description="A repro agent for SkillToolset inline document handling.",
    instruction=("Always use the document as the reference and summarize it."),
    tools=[SkillToolset(skills=[greeting_skill])],
)
```

**Run command:**
```bash
adk web
```

**Trigger prompt (any vague document reference is sufficient):**
```
summarize my document
```

---

### What happens (unpatched)

The skill instructions reference **"Document 1"** and **"Document 2"** as if they are bundled skill resources, but no files are attached to the skill (`references/`, `assets/`, or `scripts/` directories are absent). The agent's top-level instruction reinforces this by saying _"Always use the document as the reference"_, which nudges the LLM further toward attempting resource loads.

From the LLM's perspective, `RESOURCE_NOT_FOUND` is structurally indistinguishable from a transient failure (network hiccup, lazy-load not yet available). With no termination guard and no manifest of valid paths, the model hallucinates plausible path variants and retries indefinitely:

**ADK Web trace — `load_skill_resource` call sequence (calls #6–#18 visible, loop was still running at kill time):**

> ⚡ #6 load_skill_resource  
> ✓ #7 load_skill_resource  
> ⚡ #8 load_skill_resource  
> ✓ #9 load_skill_resource  
> ⚡ #10 load_skill_resource  
> ✓ #11 load_skill_resource  
> ⚡ #12 load_skill_resource  
> ✓ #13 load_skill_resource  
> ⚡ #14 load_skill_resource  
> ✓ #15 load_skill_resource  
> ⚡ #16 load_skill_resource  
> ✓ #17 load_skill_resource  
> ⚡ #18 load_skill_resource  
> _(continues…)_

(⚡ = RESOURCE_NOT_FOUND error returned; ✓ = tool call dispatched successfully but returned the error string — the alternating pattern reflects the model immediately re-invoking after each soft failure.)

**Server log — loop persisting through and beyond CTRL+C:**

```
2026-05-11 21:20:00,837 - INFO - google_llm.py:208 - Sending out request, model: gemini-3-flash-preview
2026-05-11 21:20:00,861 - INFO - google_llm.py:276 - Response received from the model.
2026-05-11 21:20:02,762 - INFO - google_llm.py:276 - Response received from the model.
2026-05-11 21:20:02,785 - INFO - google_llm.py:208 - Sending out request, model: gemini-3-flash-preview
...
^CINFO:     Shutting down
INFO:     Waiting for connections to close. (CTRL+C to force quit)
^C2026-05-11 21:20:16,718 - INFO - google_llm.py:276 - Response received from the model.
2026-05-11 21:20:16,741 - INFO - google_llm.py:208 - Sending out request, model: gemini-3-flash-preview
...
^C^C^C^C^C^C^C2026-05-11 21:20:31,023 - INFO - google_llm.py:276 - Response received from the model.
2026-05-11 21:20:31,072 - INFO - google_llm.py:208 - Sending out request, model: gemini-3-flash-preview
```

The loop continued firing new LLM requests for **~20 seconds after the first CTRL+C**, requiring repeated interrupt signals to force-quit. In-flight API calls are not cancellable — every iteration that completes before the process dies is billed.

---

### Why ambiguous prompts make this a framework issue, not a user error

This reproduction agent is not contrived — skill instructions that reference documents by natural-language names ("Document 1", "reference document") are a normal and expected authoring pattern. The ambiguity is unavoidable: skill authors do not know at write-time whether end-users will supply input inline or whether the model will infer a resource load. The framework must be defensive here because:

1. **`RESOURCE_NOT_FOUND` is a soft error by design** — it communicates "file absent" but looks transient to the LLM.
2. **No path manifest is returned** (intentional, per the progressive-disclosure spec) — the LLM has no way to self-correct.
3. **`max_llm_calls` does not help** — it caps the agent's total budget, not this specific failure mode, so a short cap breaks legitimate long-running agents.
4. **A prompt-only fix is insufficient** — imperfect upstream system prompts (like this repro) will always exist in the wild. A $105 bill from a prior production incident confirms real-world exposure.

The defense-in-depth approach in this PR — `RESOURCE_NOT_FOUND_FATAL` on the second same-path failure within an invocation, plus an explicit no-retry instruction in the default system prompt — is the correct framework-level fix.

rohityan:

rohityan:

xuanyang15:
@gemini-cli /review

Raman369AI:
Thanks so much for taking the time to review this thoroughly. Please let me know if there's anything else you'd like me to address or adjust - happy to iterate on any part of it.

xuanyang15:
No problem! This PR looks good to me and I have started the internal review process, it is currently under review.

Raman369AI:
Pushed `330c41e4`: tidies rule 5 of `_DEFAULT_SKILL_SYSTEM_INSTRUCTION` from *"do not retry the same path"* to *"do not retry any path"* so the prompt matches the invocation-wide counter and the `RESOURCE_NOT_FOUND_FATAL` message. All 95 tests in `tests/unittests/tools/test_skill_toolset.py` still pass.

xuanyang15:
@Raman369AI Thanks for the updates! Could you please help rebase the PR to the latest 2.0 code in main?

Raman369AI:
@xuanyang15 Done - rebased onto current `main` (v2.0) and force-pushed.

- 4 commits replay cleanly on top of `99087ba8`, no conflicts.
- Verified compatibility with v2.0: `ToolContext.invocation_id` and the `temp:` state prefix (`sessions/state.py: TEMP_PREFIX`) are unchanged, and the `LoadSkillResourceTool.run_async` flow our guard hooks into is untouched in 2.0.
- All 97 tests in `tests/unittests/tools/test_skill_toolset.py` pass, including the new cases covering soft→fatal escalation, different-path escalation, and cross-invocation isolation.

Ready for another look.

xuanyang15 (approved):

xuanyang15:
@Raman369AI Thanks! This PR is now waiting for ADK team internal review. I will merge it as soon as it is approved by the team.

PR #5655, opened 2026-05-10
fix: use tool_responses role for gemma4 models in LiteLLM integration

### LINKED ISSUE #5650: [LiteLLM] Gemma models via Ollama enter infinite tool-calling loop due to wrong tool message role
When using Gemma models (2B/4B) via Ollama through the LiteLLM adapter, agents with tools enter an infinite tool-calling loop and never produce a final response.

Root cause: _content_to_message_param serializes tool result messages with role="tool" (OpenAI-compatible default), but Gemma's chat template expects role="tool_responses" (according documentation: https://ai.google.dev/gemma/docs/core/prompt-formatting-gemma4). 

This mismatch causes the model to misinterpret the tool result as a new turn instead of a response to its own tool call.

This is not a hardware or quantization issue — the same behavior occurs on high-end GPUs.

jfrometa88:

Closes: #5650

### Problem

Gemma4 models served via Ollama do not recognise `role: "tool"` in the message history. When tool results were appended using this role, the model entered an infinite loop re-calling the tool instead of generating a final response.

Two separate code paths were affected:
1. `_content_to_message_param` — builds tool result messages from `function_response` parts.
2. `_ensure_tool_results` — heals incomplete histories by injecting placeholder messages when a tool call has no matching result. This function had both the trigger condition and the fallback block hardcoded to `role: "tool"`, so it neither recognised incoming `tool_responses` messages as valid resolutions nor generated the correct role in the placeholder it injected.

### Solution

Introduced a model-aware role selector in both affected functions. When the model name contains `gemma4`, the role is set to `"tool_responses"`; otherwise it falls back to the standard `"tool"`.

In `_content_to_message_param`:
  tool_role = "tool_responses" if "gemma4" in model.lower() else "tool"

In `_ensure_tool_results`, a single `expected_tool_role` variable is computed once before the loop and reused consistently in three places: the healing trigger condition, the `elif` branch that clears resolved call IDs, and the post-loop fallback block. This eliminates the previous dual-condition pattern where the two `if` blocks were not mutually exclusive and could both fire on the same message.

### Testing Plan

**Unit Tests:**

Added `tests/unittests/test_lite_llm_gemma_tool_role.py`.

Covers:
- Single tool response: verifies that when the model name contains `gemma4`, the generated message uses `role: "tool_responses"` instead of `role: "tool"`.
- Multiple tool responses: verifies the same behaviour when multiple `function_response` parts are present in a single content object.
- Non-gemma4 models: verifies that `role: "tool"` is preserved for all other model names.

Full test suite result: 5709 passed, 2300 warnings.

Setup:
- Ollama running locally with gemma4 (e2b and e4b variants tested), qwen2.5-coder:b and qwen2.5-coder:7b
- Google ADK agent with a single tool: get_comprehensive_refugee_help
- LiteLLM as the provider bridge

Steps:
1. Send a vague message → agent asks for clarification (no tool call). ✓
2. Send a message with explicit need and city → agent calls the tool once, receives tool_responses, generates final answer in the user's language. ✓
3. Confirmed via LiteLLM debug logs that the outgoing message history contains only role: "tool_responses" for gemma4 models and "tool" for qwen models, without infinite loops.

Before the fix for gemma4 models: history contained both a phantom role: "tool" error message and the real role: "tool_responses" result, causing the model to process the error last and loop.
After the fix: history contains only the real role: "tool_responses" result.

### Checklist

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -807,6 +807,10 @@ async def _content_to_message_param(
           if isinstance(response, str)
           else _safe_json_serialize(response)
       )
+      # gemma4 requires role='tool_responses' for recognizing function_response parts as responses
+      # from the tool call, instead of OpenAI-compatible 'tool' role used by other models.
+      # Earlier Gemma versions before version 4 do not support tool use,
+      # so this check is intentionally scoped to only look for "gemma4" in the model name.
       tool_role = "tool_responses" if "gemma4" in model.lower() else "tool"
       tool_messages.append(
           ChatCompletionToolMessage(
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -937,6 +937,16 @@ async def _content_to_message_param(
 
 
 def _ensure_tool_results(messages: List[Message], model: str) -> List[Message]:
+  """Insert placeholder tool messages for missing tool results.
+
+  LiteLLM-backed providers like OpenAI and Anthropic reject histories where an
+  assistant tool call is not followed by tool responses before the next
+  non-tool message. This helps recover from interrupted tool execution.
+
+  For models that expect a different tool response role (e.g. Gemma4 models,
+  which require 'tool_responses' instead of 'tool'), the role is adjusted
+  accordingly.
+  """
   if not messages:
     return messages
 
@@ -976,7 +986,7 @@ def _ensure_tool_results(messages: List[Message], model: str) -> List[Message]:
 
     healed_messages.append(message)
 
-  # Bloque final también usa expected_tool_role
+  # Final block also uses expected_tool_role
   if pending_tool_call_ids:
     logger.warning(
         "Missing tool results for tool_call_id(s): %s",
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -825,7 +825,7 @@ async def _content_to_message_param(
     follow_up = await _content_to_message_param(
         types.Content(role=content.role, parts=non_tool_parts),
         provider=provider,
-        model=model
+        model=model,
     )
     follow_up_messages = (
         follow_up if isinstance(follow_up, list) else [follow_up]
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @jfrometa88 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

src/google/adk/models/lite_llm.py
```diff
@@ -807,9 +807,10 @@ async def _content_to_message_param(
           if isinstance(response, str)
           else _safe_json_serialize(response)
       )
+      tool_role = "tool_responses" if "gemma4" in model.lower() else "tool"
```
xuanyang15, line 814: Should this be more general? or we only need to do this for gemma4?

src/google/adk/models/lite_llm.py
```diff
@@ -971,21 +969,22 @@ def _ensure_tool_results(messages: List[Message]) -> List[Message]:
...
       ]
-    elif role == "tool":
+    elif role == expected_tool_role:
       tool_call_id = message.get("tool_call_id")
       if tool_call_id in pending_tool_call_ids:
         pending_tool_call_ids.remove(tool_call_id)
 
     healed_messages.append(message)
 
+  # Bloque final también usa expected_tool_role
```
xuanyang15: Could you please use english for the comments, consistent with others in this repo?

tests/unittests/models/test_lite_llm_gemma_tool_role.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def _extract_role(msg) -> str:
+  """Extracts role from a litellm message, whether dict or object."""
+  if isinstance(msg, dict):
+    return msg["role"]
+  return msg.role
+
+
+# Tests: single
```
xuanyang15: These comments doesn't seem to be very necessary, could you please remove them?

xuanyang15 (commented):
Thanks for creating the PR! Left some minor coments.

src/google/adk/models/lite_llm.py
jfrometa88: I looked into this more carefully before broadening the check.

Gemma models before version4 do not support tool use at all, so they would never reach this code path. The `tool_responses` role convention is specific to Gemma4's chat template, making `"gemma4"` the correct and intentional scope for this fix.

Other models with non-standard tool response formats (e.g. Liquid AI's LFM2, which uses its own tokenization scheme with custom delimiters like `<|tool_response_start|>`) represent a structurally different problem that would require a separate solution — conflating them here would overcomplicate this fix.

Happy to add a comment in the code explaining why the check is scoped to `gemma4` specifically, if that would help reviewers in the future.

tests/unittests/models/test_lite_llm_gemma_tool_role.py
jfrometa88: Removed

src/google/adk/models/lite_llm.py
jfrometa88: Changed

src/google/adk/models/lite_llm.py
xuanyang15: I see, thanks for the detailed explanation! Using "gemma4" sounds good to be. Could you please add the comment for future reference?

jfrometa88: Added comments explaining why the check is scoped to `gemma4`. Ready for another look whenever you get a chance

PR #5662, opened 2026-05-11
fix(web): handle TEXT modality for native-audio models (#4206)

### LINKED ISSUE #4206: adk web: Native-audio models fail when TEXT modality is requested
## Problem

When using `adk web` with native-audio models (e.g., `gemini-live-2.5-flash-native-audio`), requesting TEXT modality causes an error. Native-audio models only support AUDIO modality.

## Root Cause

In `src/google/adk/cli/adk_web_server.py`, the `/run_live` WebSocket endpoint accepts modalities from query parameters without validating them against the model's capabilities:

```python
# Lines 1636-1638
modalities: List[Literal["TEXT", "AUDIO"]] = Query(
    default=["AUDIO"]
),  # Only allows "TEXT" or "AUDIO"
```

```python
# Line 1655
run_config = RunConfig(response_modalities=modalities)
```

The code at `runners.py:985-988` only sets a default when `response_modalities` is `None`, but doesn't validate when it's explicitly set to TEXT for native-audio models.

## Proposed Solution

1. **Detect native-audio models** by checking if the model name contains "native-audio"
2. **Force AUDIO modality** for these models instead of returning an error
3. **Ensure output_audio_transcription is enabled** so users can see text transcription of the audio response

### Implementation

In `adk_web_server.py` around line 1653-1655:

```python
async def forward_events():
    runner = await self.get_runner_async(app_name)
    
    # Check if agent uses a native-audio model
    agent_or_app = self.agent_loader.load_agent(app_name)
    root_agent = self._get_root_agent(agent_or_app)
    model_name = root_agent.model if isinstance(root_agent.model, str) else ""
    
    # Native audio models only support AUDIO modality
    if "native-audio" in model_name:
        effective_modalities = ["AUDIO"]
    else:
        effective_modalities = modalities
    
    run_config = RunConfig(response_modalities=effective_modalities)
```

## Additional Issue: Transcription not displayed in UI

When using AUDIO modality, `output_audio_transcription` is enabled by default in `RunConfig`, and transcription events are created by `TranscriptionManager`. However, the frontend does not currently render `outputTranscription.text` - it only displays `content.parts[].text`.

This means users won't see any text when using native-audio models, even though transcription data is available in the events.

### Options to address this:
1. Update the frontend to display `outputTranscription.text` when present
2. Convert transcription to content in the backend so existing UI can display it

## Affected Files

- `src/google/adk/cli/adk_web_server.py` - WebSocket endpoint for live sessions
- `src/google/adk/runners.py` - Default modality handling
- `src/google/adk/flows/llm_flows/transcription_manager.py` - Transcription event creation

paulxg12:
## Summary

Fixes #4206 — native-audio models fail when TEXT modality is explicitly requested via the ADK web server.

### Root Cause
The `/run_live` WebSocket endpoint accepted `modalities` from query params without validating against model capabilities. When a user requested `TEXT` modality for a native-audio model (e.g., `gemini-live-2.5-flash-native-audio`), the model failed since it only supports `AUDIO`.

### Changes
- **`adk_web_server.py`**: detect native-audio models in WebSocket handler and override modalities to `["AUDIO"]`
- **`runners.py`**: extended native-audio check to also correct explicitly-set `["TEXT"]` to `["AUDIO"]` (previously only handled `None` case)

### Testing
- Native-audio models with no modality specified → AUDIO (unchanged behavior)
- Native-audio models with TEXT modality requested → AUDIO (fixed)
- Non-audio models → respects user's modality choice (unchanged)

### DIFF
--- a/src/google/adk/cli/adk_web_server.py
+++ b/src/google/adk/cli/adk_web_server.py
@@ -2106,8 +2106,21 @@ class AdkWebServer:
 
       async def forward_events():
         runner = await self.get_runner_async(app_name)
+
+        # Native-audio models (e.g., gemini-live-2.5-flash-native-audio) only
+        # support AUDIO modality. When TEXT is requested, fall back to AUDIO.
+        effective_modalities = modalities
+        if isinstance(runner.agent, LlmAgent):
+            model_name = ''
+            if isinstance(runner.agent.model, str):
+                model_name = runner.agent.model
+            elif runner.agent.model:  # BaseLlm instance
+                model_name = runner.agent.model.model
+            if 'native-audio' in model_name.lower():
+                effective_modalities = ['AUDIO']
+
         run_config = RunConfig(
-            response_modalities=modalities,
+            response_modalities=effective_modalities,
             proactivity=(
                 types.ProactivityConfig(proactive_audio=proactive_audio)
                 if proactive_audio is not None
--- a/src/google/adk/runners.py
+++ b/src/google/adk/runners.py
@@ -1101,6 +1101,12 @@ class Runner:
     # AUDIO by default.
     if run_config.response_modalities is None:
       run_config.response_modalities = ['AUDIO']
+    elif (
+        'AUDIO' not in run_config.response_modalities
+        and hasattr(self.agent, 'canonical_model')
+        and 'native-audio' in self.agent.canonical_model.model.lower()
+    ):
+      run_config.response_modalities = ['AUDIO']
     if session is None and (user_id is None or session_id is None):
       raise ValueError(
           'Either session or user_id and session_id must be provided.'


paulxg12:
I have read the Google CLA and I hereby sign the CLA

rohityan:
Hi @paulxg12 , Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

paulxg12:
I have signed the Google CLA

wuliang229:
Turns out we don't need the change.

Issue #5637, opened 2026-05-07
Tool-level auth continues invocation past the EUC event


doughayden:
When a tool requests OAuth credentials at execution time (tool-level auth), the resulting `adk_request_credential` event does not terminate the invocation. The agent loop continues to the next LLM call. This contrasts with toolset-level auth, where the same EUC event terminates the invocation cleanly.

In `_postprocess_handle_function_calls_async` (base_llm_flow.py:1111), the auth event is yielded at line 1124 and the function_response_event at line 1133. After `_run_one_step_async` returns, `run_async` (lines 803-816) checks `last_event.is_final_response()`. last_event is the function_response, not the EUC. The function_response is not a final response because `is_final_response` (event.py:91-98) returns False whenever the event carries function responses. The agent loop continues.

In `_resolve_toolset_auth` (base_llm_flow.py:113-191), the auth event is yielded at line 184 and `invocation_context.end_invocation = True` is set at line 191. `_run_one_step_async` returns early at line 831 and `run_async` breaks. Both paths emit the same `adk_request_credential` shape with `long_running_tool_ids` set, so the divergence is in invocation termination, not in the event itself.

The primary user-visible consequence is that the agent reasons one more turn after surfacing the authorization request. The next LLM call produces a follow-up message that emits before the user can complete the OAuth flow, often a model-generated explanation that authorization is still missing. This message appears on top of or alongside the authorization request itself.

A secondary consequence affects deployments that use a session service with marker-based concurrency control, such as `DatabaseSessionService`. The trailing turn's writes continue past the EUC and can overlap with a separately-scheduled post-auth resume task that loads the session in parallel. The resume task's session snapshot ends up stale relative to the trailing writes, and its first append fails with the marker-mismatch `ValueError` from `database_session_service.py`. We have observed this with `DatabaseSessionService`. The in-memory service does not implement that check. We have not verified the Agent Engine session service.

The TODO at base_llm_flow.py:842-845 acknowledges this case ("we should find a way to pause when the long running tool call is followed by more than one text responses"). The `is_resumable` branch at lines 838-851 returns early when a long-running call is detected in recent events, but only when `invocation_context.is_resumable` is set, which is not the default. Setting `end_invocation = True` when `_postprocess_handle_function_calls_async` yields an auth event would make tool-level auth terminate symmetrically with toolset-level auth, without depending on the resumable codepath.

## Reproduction

[...]

Issue #5327, opened 2026-04-14
Preemptive toolset auth triggers OAuth redirect on every agent invocation


doughayden:
## 🔴 Required Information

**Describe the Bug:**

`_resolve_toolset_auth` in `base_llm_flow.py` runs before every agent invocation and calls `get_auth_config()` on each toolset. If the toolset returns an `AuthConfig`, the framework checks for credentials **before the LLM has decided whether any tool call is needed**. When the framework-level credential lookup fails — which it always does for tool-level-authenticated credentials because the two paths use different key formats — this triggers a full OAuth redirect flow on every message, including simple greetings that will never invoke any tool.

**Steps to Reproduce:**

1. Register an `OpenAPIToolset` (or `ApplicationIntegrationToolset`) on an `LlmAgent` with OAuth2 `authorizationCode` flow.
2. Send any message to the agent — e.g. `"hey"`.
3. Observe: an `adk_request_credential` event fires before the LLM response, producing an OAuth redirect URL.
4. Complete the OAuth flow; send another message (tool-invoking or not).
5. Observe: the redirect fires again, indefinitely on every subsequent message.

**Expected Behavior:**

OAuth should only be triggered when the LLM actually invokes a tool that requires authentication. Simple greetings or non-tool-invoking messages should proceed without any auth flow. After a successful OAuth exchange, the credential should be found and reused by subsequent tool invocations.

**Observed Behavior:**

OAuth redirect fires on every message regardless of whether any tool would be invoked. Even after a successful authorization, the next message triggers OAuth again because the framework-level credential lookup can't find the credential that was stored by the tool-level path under a different key format.

**Environment Details:**

- ADK Library Version: `google-adk 1.28.0`
- Desktop OS: macOS (reproduced), Linux (expected to affect all platforms)
- Python Version: Python 3.13

**Model Information:**

- Are you using LiteLLM: No
- Which model is being used: gemini-2.5-flash (irrelevant — this is in the auth path, not the model path)

---

## 🟡 Optional Information

**Additional Context:**

Root cause — `_resolve_toolset_auth` in `google/adk/flows/llm_flows/base_llm_flow.py` (simplified for clarity; the actual implementation adds a `try/except ValueError` around `get_auth_credential` and constructs `toolset_id` from the toolset class name):

```python
for tool_union in agent.tools:
    if not isinstance(tool_union, BaseToolset):
        continue
    auth_config = tool_union.get_auth_config()
    if not auth_config:
        continue
[...]

PR #5692, opened 2026-05-13
fix(auth): strip redirect_uri from credential_key

### LINKED ISSUE #5691: OAuth2 credential_key includes redirect_uri, breaking credential lookup across deployment URLs
## 🔴 Required Information

**Describe the Bug:**

ADK's OAuth2 credential-key methods strip eight `oauth2` fields before hashing the credential but omit `redirect_uri`. Since `redirect_uri` is deployment configuration rather than credential identity, the same OAuth credential (same client_id, client_secret, scopes, access_token, refresh_token) hashes to a different key when the deployment URL changes. A credential minted under one redirect_uri can no longer be retrieved when the deployment moves to another, even though the underlying OAuth grant is unchanged.

The same eight-field strip block, and the same gap, appears at three call sites:

- `ToolContextCredentialStore.get_credential_key` (`tool_auth_handler.py:85-110`)
- `ToolContextCredentialStore._get_legacy_credential_key` (`tool_auth_handler.py:58-83`)
- `AuthConfig.get_credential_key` (`auth_tool.py:97-138`, marked `@deprecated` but still invoked from `AuthConfig.__init__` at line 94 when `credential_key` is not explicitly provided)

**Steps to Reproduce:**

1. Configure an `OpenAPIToolset` with OAuth2 `authorizationCode` flow, where the `AuthCredential.oauth2.redirect_uri` reflects deployment A (for example, a local OAuth relay at `http://localhost:8001/oauth2callback`).
2. Run an agent that invokes a tool through that toolset and complete the OAuth authorization flow. The exchanged credential is stored in session state under a hash key derived from deployment A's `redirect_uri`.
3. Reconfigure the same toolset with the same `client_id`, `client_secret`, and scopes, but change `redirect_uri` to deployment B's value (for example, a deployed relay at `https://my-relay-...run.app/oauth2callback`).
4. Run the agent again against the same session state from step 2.
5. Observe that the agent emits `adk_request_credential` despite the credential from step 2 still being present in session state. The runtime hash computed from deployment B's `redirect_uri` does not match the key where the credential was stored.

A self-contained runnable script that demonstrates the same conditions in a single process (no separate deployments needed) is at https://github.com/doughayden/adk-issue-examples/tree/main/05-redirect_uri_in_credential_hash. Run `uv run main.py` to observe the bug and `uv run main.py --apply-fix` to verify the proposed fix.

**Expected Behavior:**

The credential lookup should hit the seeded credential regardless of the `redirect_uri` value, because `redirect_uri` is deployment configuration rather than credential identity. The tool call should succeed against the valid seeded credential without prompting for re-auth.

**Observed Behavior:**

The hash key computed at lookup time differs from the hash key the seed was written under (because the two `redirect_uri` strings differ). `ToolAuthHandler._get_existing_credential` returns None, `prepare_auth_credentials` falls into `_request_credential`, and `adk_request_credential` is emitted despite valid stored credentials being present.

```
🌤️  WeatherAssistant Agent — redirect_uri-in-hash repro
============================================================
Proposed fix applied:      False
STORED_REDIRECT_URI:       http://localhost:8080/callback
CURRENT_REDIRECT_URI:      http://localhost:8081/callback

Hash keys produced by ToolContextCredentialStore.get_credential_key:
    STORED   → oauth2_55f666541ad22e39_oauth2_8ba0457897522d9d_existing_exchanged_credential
    CURRENT  → oauth2_55f666541ad22e39_oauth2_ae16199243c358df_existing_exchanged_credential
    ❌ Keys differ — credentials minted under STORED are not retrievable.

🔑 Seeded credential: access_token='l-wr4tRlWJO4WlZC'… (redirect_uri='http://localhost:8080/callback')

👤 User: What's the weather in San Francisco?
🌤️  Weather Assistant event stream:

    [function_call] get_weather by WeatherAssistant
    [auth_event] adk_request_credential by WeatherAssistant
    [function_response] get_weather by WeatherAssistant
    [text] WeatherAssistant: 'It seems I need your authorization to access weather data. Could you please g...'

Event counts:
    function_calls: 1
    auth_events: 1
    function_responses: 1
    text_events: 1

✅ Bug reproduced: agent emitted 1 adk_request_credential event(s) despite a valid seeded credential being present in state (hashed under STORED_REDIRECT_URI; agent looked under CURRENT_REDIRECT_URI).
```

**Proposed Fix:**

Add `auth_credential.oauth2.redirect_uri = None` to the strip block at each of the three call sites listed under "Describe the Bug" above. `redirect_uri` is not part of the credential identity (the granted access).

**Environment Details:**

 - ADK Library Version (`pip show google-adk`): `google-adk==1.32.0`. Verified that the strip blocks in `tool_auth_handler.py` and `auth_tool.py` are identical on current `main` at the time of writing, so the bug is present on current released versions.
 - Desktop OS: macOS (bug is platform-agnostic; the hash logic is pure Python)
 - Python Version (`python -V`): Python 3.13.13

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: `gemini-2.5-flash` (bug is model-agnostic; it fires during credential lookup, before any LLM call would reach the provider)

---

## 🟡 Optional Information

**Regression:**

Not a regression. The strip block has carried this gap since OAuth2 credential-key generation was introduced. The most recent credential-key commit (`33012e6d`, 2026-02-03, "Make credential key generation stable and prevent cross-user credential leaks") tightened the strip block by excluding additional transient OAuth2 fields but did not add `redirect_uri`.

**Logs:**

Output from the same repro with `--apply-fix` applied, for comparison:

```
🌤️  WeatherAssistant Agent — redirect_uri-in-hash repro
============================================================
Proposed fix applied:      True
STORED_REDIRECT_URI:       http://localhost:8080/callback
CURRENT_REDIRECT_URI:      http://localhost:8081/callback

Hash keys produced by ToolContextCredentialStore.get_credential_key:
    STORED   → oauth2_55f666541ad22e39_oauth2_c2ad46dffd26cd87_existing_exchanged_credential
    CURRENT  → oauth2_55f666541ad22e39_oauth2_c2ad46dffd26cd87_existing_exchanged_credential
    ✅ Keys match — fix is taking effect at the hash level.

🔑 Seeded credential: access_token='A46ZCrsEq-l3uC1X'… (redirect_uri='http://localhost:8080/callback')

👤 User: What's the weather in San Francisco?
🌤️  Weather Assistant event stream:

    [function_call] get_weather by WeatherAssistant
    [function_response] get_weather by WeatherAssistant
    [text] WeatherAssistant: 'The weather in San Francisco is Clear with a temperature of 30 degrees Celsiu...'

Event counts:
    function_calls: 1
    auth_events: 0
    function_responses: 1
    text_events: 1

✅ Fix verified: tool call succeeded against the seeded credential without an adk_request_credential prompt.
```

**Additional Context:**

A concrete example of why this matters: a deployed application and a local development environment sharing the same session DB. The deployed app's `redirect_uri` points at the deployed callback URL; the local app's `redirect_uri` points at a local server callback path. The underlying user grant and OAuth tokens are identical (same user, same OAuth client app, same scopes), so a grant made from one environment should be reusable from the other through the shared session DB. Today the two environments hash credentials under different keys and are effectively isolated, so the user is re-prompted for OAuth in each environment despite holding a valid grant. (All valid `redirect_uri` values must still be registered in the OAuth client app's allow-list on the provider side, which is a prerequisite outside ADK and not something this fix changes.)

Related:
- #5327 (preemptive toolset auth)
- #5328 (refresh request scope)
- #5329 (refreshed credential persistence)
- #5637 (tool-level auth termination)

Self-contained runnable repro at https://github.com/doughayden/adk-issue-examples/tree/main/05-redirect_uri_in_credential_hash. Pins `google-adk==1.32.0` and reproduces against current `main`. Includes a `--apply-fix` flag that monkey-patches the proposed fix.

Always (100%). The bug is deterministic at the hash level: two `AuthCredential` instances differing only in `redirect_uri` produce different keys on every invocation.

doughayden:

- Closes: #5691

This change adds `auth_credential.oauth2.redirect_uri = None` to the OAuth2 strip block at the three call sites where credential hashing happens. `redirect_uri` is deployment configuration (which callback URL the auth server should redirect to), not part of the credential identity, so it should be excluded from the hash just as access_token, refresh_token, expires_at, and the other transient OAuth2 fields already are. Without this change, a credential minted under one deployment URL cannot be retrieved when the deployment moves to another.

### Testing Plan

**Unit Tests:**

Three new tests, one per affected method. Each constructs two `AuthCredential` instances that differ only in `redirect_uri` and asserts the computed key is identical:

- `tests/unittests/tools/openapi_tool/openapi_spec_parser/test_tool_auth_handler.py::test_credential_key_is_stable_across_redirect_uri`
- `tests/unittests/tools/openapi_tool/openapi_spec_parser/test_tool_auth_handler.py::test_legacy_credential_key_is_stable_across_redirect_uri`
- `tests/unittests/auth/test_auth_config.py::test_credential_key_is_stable_across_redirect_uri`

```
$ pytest tests/unittests/tools/openapi_tool/openapi_spec_parser/test_tool_auth_handler.py tests/unittests/auth/test_auth_config.py
======================== 14 passed, 8 warnings in 2.90s ========================

$ pytest tests/unittests/
================ 5740 passed, 2340 warnings in 86.64s (0:01:26) ================
```

A self-contained Runner-based reproduction is at https://github.com/doughayden/adk-issue-examples/tree/main/05-redirect_uri_in_credential_hash. The agent definition (`agent.py`) wires up an `OpenAPIToolset` against a local OAuth2 test server, configured with one redirect_uri value. `main.py` constructs an `InMemoryRunner`, seeds a real (non-expired) OAuth2 credential into session state under a different redirect_uri's hash, and runs the agent. The `--apply-fix` flag monkey-patches the proposed fix to demonstrate the resolution end-to-end.

Without the fix:

```
🌤️  WeatherAssistant Agent — redirect_uri-in-hash repro
============================================================
Proposed fix applied:      False
STORED_REDIRECT_URI:       http://localhost:8080/callback
CURRENT_REDIRECT_URI:      http://localhost:8081/callback

Hash keys produced by ToolContextCredentialStore.get_credential_key:
    STORED   → oauth2_55f666541ad22e39_oauth2_8ba0457897522d9d_existing_exchanged_credential
    CURRENT  → oauth2_55f666541ad22e39_oauth2_ae16199243c358df_existing_exchanged_credential
    ❌ Keys differ — credentials minted under STORED are not retrievable.

👤 User: What's the weather in San Francisco?
🌤️  Weather Assistant event stream:

    [function_call] get_weather by WeatherAssistant
    [auth_event] adk_request_credential by WeatherAssistant
    [function_response] get_weather by WeatherAssistant
    [text] WeatherAssistant: 'It seems I need your authorization to access weather data. Could you please g...'

Event counts:
    function_calls: 1
    auth_events: 1
    function_responses: 1
    text_events: 1

✅ Bug reproduced: agent emitted 1 adk_request_credential event(s) despite a valid seeded credential being present in state.
```

With the fix:

```
🌤️  WeatherAssistant Agent — redirect_uri-in-hash repro
============================================================
Proposed fix applied:      True
STORED_REDIRECT_URI:       http://localhost:8080/callback
CURRENT_REDIRECT_URI:      http://localhost:8081/callback

Hash keys produced by ToolContextCredentialStore.get_credential_key:
    STORED   → oauth2_55f666541ad22e39_oauth2_c2ad46dffd26cd87_existing_exchanged_credential
    CURRENT  → oauth2_55f666541ad22e39_oauth2_c2ad46dffd26cd87_existing_exchanged_credential
    ✅ Keys match — fix is taking effect at the hash level.

👤 User: What's the weather in San Francisco?
🌤️  Weather Assistant event stream:

    [function_call] get_weather by WeatherAssistant
    [function_response] get_weather by WeatherAssistant
    [text] WeatherAssistant: 'The weather in San Francisco is Clear with a temperature of 30 degrees Celsiu...'

Event counts:
    function_calls: 1
    auth_events: 0
    function_responses: 1
    text_events: 1

✅ Fix verified: tool call succeeded against the seeded credential without an adk_request_credential prompt.
```

### Checklist

### Additional context

**Scope:**

The same strip block exists at three call sites and has the same gap at all three. This PR patches all three. Patching only the tool-level pair (`tool_auth_handler.py`) and leaving the framework-level path (`auth_tool.py:AuthConfig.get_credential_key`) would leave the bug reachable for any consumer that does not work around #5327 with `get_auth_config = lambda: None`. Patching only `AuthConfig.get_credential_key` and leaving the tool-level pair would leave the bug reachable on the standard tool-level credential lookup path.

**Upgrade note:**

The credential_key shape changes with this fix: `redirect_uri` is no longer included in the hash. OAuth credentials cached in existing session state under the pre-fix key shape become unreachable under the new key. Users should expect a one-time re-auth prompt on the first run after upgrading. Subsequent runs use the new key normally.

**Related:**

- #5327 (preemptive toolset auth)
- #5328 (refresh request scope)
- #5329 (refreshed credential persistence, fixed in 218ea76e)
- #5637 (tool-level auth termination)


rohityan:
Hi @doughayden , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing mypy-diff tests before we can proceed with the review.

doughayden:
@rohityan checking with you before I expand the PR scope to fix the mypy-diff failure...

In `_get_legacy_credential_key()` and `get_credential_key`, `auth_credential = auth_credential.model_copy(deep=True)` widens `auth_credential.oauth2` back to `OAuth2Auth | Any | None`, breaking the outer-guard type narrowing. Each `auth_credential.oauth2.X = None` in the block is a union-attr error but the newer mypy-diff workflow only flags this PR's additions compared to the baseline.

I propose nesting the full `auth_credential.oauth2.X = None` block inside `if auth_credential.oauth2:` right after the `model_copy()`, reusing the outer guard's pattern. Tradeoff: PR diff grows ~20 lines because the two typed strip blocks indent one level, but it eliminates 16 pre-existing union-attr errors plus the 2 new redirect_uri ones. Runtime remains unchanged, since `model_copy(deep=True)` of a model with non-None oauth2 always yields a copy with non-None oauth2.

I verified locally: tests pass, manually reproduced mypy-diff comparison is clean.

Do you agree with the fix-it-while-we're-here approach?

rohityan:

doughayden:
Hi @wyf7107, friendly check-in on this one. CI is green across all jobs and the PR is mergeable, so it's ready whenever you have a moment.

One note so you don't have to dig: the scope question I raised on May 21 is already resolved in the PR. Commit 929be466 adds the `oauth2` narrowing after `model_copy` in both `_get_legacy_credential_key()` and `get_credential_key`, so the outer guard holds. Nothing outstanding on my end.

PR #5704, opened 2026-05-14
feat(plugins): project LlmResponse finish_reason to BigQuery analytics

### LINKED ISSUE #5644: feat: project `finish_reason` from LlmResponse to attributes in BigQueryAgentAnalyticsPlugin LLM_RESPONSE events
## Summary

`BigQueryAgentAnalyticsPlugin` does not project the Vertex AI `finish_reason` field to the `attributes` column of `LLM_RESPONSE` events written to BigQuery. This makes it impossible to classify model failure modes (MAX_TOKENS, SAFETY, MALFORMED_FUNCTION_CALL, RECITATION, etc.) via SQL queries against the analytics table — operators must instead parse unstructured Cloud Logging output to recover this signal.

Adding the field would be a small change (~5 lines) with significant observability value for any project using the plugin.

## Current behavior (verified empirically 2026-05-08)

```sql
SELECT
  JSON_VALUE(attributes,'$.finishReason')                AS camel,
  JSON_VALUE(attributes,'$.finish_reason')               AS snake,
  JSON_VALUE(attributes,'$.usage_metadata.finish_reason') AS nested,
  COUNT(*) AS row_count
FROM `<project>.<dataset>.events`
WHERE event_type = 'LLM_RESPONSE'
  AND timestamp >= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 7 DAY)
GROUP BY camel, snake, nested
ORDER BY row_count DESC;
```

Result against a 7-day window in production:

| camel | snake | nested | row_count |
|---|---|---|---|
| NULL | NULL | NULL | 309 |

`finish_reason` is **not** present in any path (camelCase, snake_case, or nested under `usage_metadata`). Source inspection confirms why: in `google/adk/plugins/bigquery_agent_analytics_plugin.py`, `_EVENT_VIEW_DEFS[\"LLM_RESPONSE\"]` (around lines 1813–1846 in 1.32.0) extracts only `response`, `usage_*_tokens`, `cached_content_token_count`, `context_cache_hit_rate`, `total_ms`, `ttft_ms`, `model_version`, `usage_metadata`, `cache_metadata`. There are zero references to `finish_reason` / `finishReason` anywhere in the 3500+ lines of the plugin file. The `EventData` dataclass and `after_model_callback` likewise do not capture it, even though the `LlmResponse` parameter the callback receives includes it from Vertex.

## Use case / motivation

1. **Failure mode breakdown via SQL**: today, distinguishing \"this LLM call ended with MAX_TOKENS\" from \"this LLM call ended with MALFORMED_FUNCTION_CALL\" via the analytics table is impossible. The information lives only in unstructured stderr logs, which are not joinable to the structured event stream.

2. **Cache hit ratio segmented by finish_reason**: the existing `cached_content_token_count` projection enables computing cache hit ratio (we use this heavily). But understanding the *quality* of the cached responses (did the cached prefix lead to clean STOPs or to MALFORMED retries?) requires the finish_reason dimension.

3. **Model migration health checks**: when migrating between model families (e.g., Gemini 2.5 → 3.x), SRE-grade observability needs to compare MALFORMED rate, SAFETY rate, etc., across revisions side by side. Without `finish_reason` in BQ, the comparison must be reconstructed from logs — slow, brittle, and not amenable to alert policies.

4. **Alerting on regression**: a Cloud Monitoring alert like \"MALFORMED_FUNCTION_CALL rate > 5% sustained over 1h\" is straightforward to write against the BQ events table if `finish_reason` is projected. Without it, the same alert requires a log-based metric pipeline (more moving parts, more cost).

## Proposed solution

Add `finish_reason` (and optionally `finish_message`) to the `LLM_RESPONSE` view definition. The field is already present on `LlmResponse` from Vertex; the plugin only needs to read and project it. Roughly (against 1.32 source structure):

```python
# In _EVENT_VIEW_DEFS[\"LLM_RESPONSE\"]:
\"finish_reason\": lambda llm_response: (
    llm_response.finish_reason.name
    if llm_response.finish_reason is not None
    else None
),
\"finish_message\": lambda llm_response: llm_response.finish_message,
```

Backwards-compatible: existing consumers reading the documented fields are unaffected; new consumers can opt in via `JSON_VALUE(attributes,'$.finish_reason')`.

If a different snake_case vs camelCase convention is preferred to match other ADK projections, happy to adjust — the empirical probe checked both spellings to be safe.

## Alternatives considered

- **`after_model_callback` writing to `state_delta`**: feasible as a downstream workaround, but every consumer of the plugin needs to re-implement it, and it bloats the event stream with one extra row per LLM call.
- **Custom plugin subclass overriding `LLM_RESPONSE` EventData**: surgical but creates maintenance burden when the upstream plugin schema evolves.
- **Cloud Logging sink → BigQuery via Logs Router**: works but introduces a second pipeline (sink config, log-based filters, parsing) for data that is already flowing through the analytics plugin one level up.

All three alternatives are strictly more code and more complexity than projecting the field in the plugin where the data already lives.

## Environment

- `google-adk` 1.32.0
- Python 3.12
- Vertex AI backend (\`vertexai=True\`)
- Plugin: `BigQueryAgentAnalyticsPlugin` (default config, single dataset, single events table)
- Models tested: `gemini-2.5-flash`

Happy to send a PR if the team agrees this is in scope. Thanks for the great library.

roanny:
## Summary

Projects `finish_reason` from `LlmResponse` to the `attributes` JSON column of `LLM_RESPONSE` events written to BigQuery by `BigQueryAgentAnalyticsPlugin`, and exposes it in the generated per-event-type SQL view.

Without this projection, classifying model failure modes (MAX_TOKENS, SAFETY, MALFORMED_FUNCTION_CALL, RECITATION, etc.) via SQL against the analytics table is impossible — the signal lives only in unstructured Cloud Logging output that is not joinable to the structured event stream.

Fixes #5644.

## Changes

- `EventData` gets a `finish_reason: Optional[str]` field.
- `after_model_callback` reads `llm_response.finish_reason.name` and passes it through to `EventData`.
- `_enrich_attributes` writes `attrs["finish_reason"]` when present (omitted entirely when `None`, so the JSON shape stays backwards-compatible for existing consumers).
- `_EVENT_VIEW_DEFS["LLM_RESPONSE"]` projects `JSON_VALUE(attributes, '$.finish_reason') AS finish_reason` so it shows up as a typed column on the per-event-type view.
- Schema-field description for the `attributes` column mentions the new key with example values.

## Backwards compatibility

The new key is only added to the JSON when `LlmResponse.finish_reason` is not `None`, so events emitted before this change and events for streaming partials (where `finish_reason` is unset) remain unchanged. SQL consumers reading the existing documented fields are unaffected. New consumers can opt in via `JSON_VALUE(attributes, '$.finish_reason')` or via the `finish_reason` column exposed by the generated view.

## Test plan

New unit tests (all passing locally; `229 passed, 5 skipped` for the full plugin suite):

- [x] `test_after_model_callback_projects_finish_reason` — asserts `MAX_TOKENS` round-trips end-to-end into the BQ row's `attributes`.
- [x] `test_after_model_callback_omits_finish_reason_when_absent` — asserts the key is omitted when `LlmResponse.finish_reason` is `None`.
- [x] `TestEnrichAttributes::test_finish_reason_included_when_set` — direct `_enrich_attributes` coverage.
- [x] `TestEnrichAttributes::test_finish_reason_omitted_when_none` — direct `_enrich_attributes` coverage of the None branch.
- [x] Existing 229 tests in the plugin suite still pass (no regressions).

To run locally:

```bash
pytest tests/unittests/plugins/test_bigquery_agent_analytics_plugin.py -k finish_reason -v
```

cc @sanketpatil06 @darshil3011


roanny:
@googlebot I signed it!

xuanyang15:
@shobsi Could you please take a look, it is for https://github.com/google/adk-python/issues/5644?

rohityan:
Hi @roanny , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

GWeale:
shipped in 04b8b727, v2.7.0: EventData.finish_reason, the attributes write, reading the enum name off LlmResponse in after_model_callback, and the JSON_VALUE column on the LLM_RESPONSE view. that's the approach from this pr, and the commit should have credited it. head additionally suppresses the key on streaming partials and tolerates a plain string value. #5644 is closed.

thanks for the pr and for the issue that drove it.

PR #5707, opened 2026-05-15
feat: add error handling to VertexAiMemoryBankService.search_memory

vietnamesekid:
## What's the problem?

`search_memory` iterates over memory entries from the Vertex AI API but had no error handling. If one entry was malformed or the iterator threw mid-stream, the whole call would blow up and drop everything collected so far.

## What changed?

Added two layers of defense inside the iterator loop:

- **Per-entry** (`AttributeError`): skips bad entries and logs a warning so the rest still come through
- **Iterator-level** (`Exception`): if the stream fails partway, we return whatever we already collected instead of raising

Also added guards for a few edge cases I noticed while looking at this:
- `memory is None`
- `fact` is empty/`None`
- `update_time is None` (just sets `timestamp=None` rather than crashing)

## Test plan

- [x] `test_search_memory` — happy path unchanged
- [x] `test_search_memory_empty_results` — empty results unchanged
- [x] `test_search_memory_skips_entry_with_none_memory`
- [x] `test_search_memory_skips_entry_with_empty_fact`
- [x] `test_search_memory_handles_missing_update_time`
- [x] `test_search_memory_skips_malformed_entry`
- [x] `test_search_memory_returns_partial_results_on_iterator_error`


rohityan:
Hi @vietnamesekid , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

PR #5717, opened 2026-05-16
fix(auth): Make OAuth prompt parameter configurable

### LINKED ISSUE #3046: OAuth Flow `prompt` Parameter Should Be Configurable
### **Bug Report: OAuth Flow `prompt` Parameter Should Be Configurable**

---

#### **Describe the bug**

In the current OAuth flow, the `prompt` parameter is hardcoded to `consent`.
This should instead be configurable and taken as input from the user (or application context).

Reason: Some permissions are granted at the **application level**, while individual users may not have those permissions. In such cases, users are unnecessarily redirected to the consent page. Ideally, they should be silently authenticated (`prompt=none`) when possible to ensure a smoother authentication experience.

---

#### **To Reproduce**

1. Create an SPN on Microsoft Azure and set up an OAuth flow for SharePoint.
2. Grant `Sites.Read.All` permission from the Admin account to the application (Application Permission, not Delegated Permission).
3. As a normal user, try using this Agent with the SharePoint Connector.
4. The user is asked for admin access each time, because `prompt=consent` is appended to the URL.

---

#### **Expected behavior**

* The `prompt` query parameter should be configurable when setting up authentication for the tool.
* Since this value is hardcoded in `AuthHandler`, scenarios like the one described become difficult to handle.
* Expected behavior:

  * Allow silent authentication (`prompt=none`) when possible.
  * Redirect to consent screen only if explicitly required.

---

#### **Desktop (please complete the following information):**

* **OS**: macOS
* **Python version**: 3.13.3
* **ADK version**: 1.15.0

---

#### **Model Information**

* **Using LiteLLM**: No
* **Model**: gemini-2.5-flash

---

#### **Additional context**

* If `prompt=none` is set manually, the following error occurs:

  > *Auth Code not found*
* The parsing of the Auth Code returns `null`. While the code can be parsed from the URL, there may be a cleaner and more reliable way to handle this.

---

✅ **Proposed Fix:**

* Make the `prompt` parameter configurable in `AuthHandler`.
* Provide options like `consent`, `none`, `login` based on user/application context.
* Ensure Auth Code parsing works correctly even when `prompt=none` is used.

akshay-kumar-bm:
### Description

Makes the OAuth `prompt` query parameter configurable on `OAuth2Auth` instead of being hardcoded to `"consent"` in `AuthHandler.generate_auth_uri`.

### Why

Per #3046, hardcoding `prompt=consent` breaks scenarios where consent is granted at the application level (e.g. Azure Entra ID admin-consented permissions). Non-admin users are sent to a consent screen they cannot complete, blocking the flow. Multiple users have hit this (see issue thread).

### What changed

- **`src/google/adk/auth/auth_credential.py`** — Added optional `prompt` field on `OAuth2Auth`. Typed as `Literal["none", "login", "consent", "select_account"] | str | None` — the `Literal` gives IDE autocomplete for the RFC 6749 / OIDC standard values, and the `| str` fallback allows IdP-specific values like Azure's `"admin_consent"` to pass through unchanged.
- **`src/google/adk/auth/auth_handler.py`** — Forwards `auth_credential.oauth2.prompt` to the authorization endpoint, falling back to `"consent"` when unset (preserves existing behavior).
- **`tests/unittests/auth/test_auth_handler.py`** — Added 4 new test cases (default-is-consent, custom `none`, custom `select_account`, IdP-specific `admin_consent`), strengthened the existing `test_generate_auth_uri_with_audience_and_prompt` to actually assert the prompt value, and extended `MockOAuth2Session.create_authorization_url` to echo the `prompt` kwarg (mirrors the existing `audience` pattern).

### Issue

Fixes #3046

### Testing

- `pytest tests/unittests/auth/test_auth_handler.py` — 29 passed
- `pytest tests/unittests/auth/` — 185 passed
- All 111 OAuth-related tests across the rest of the suite still pass
- `pre-commit run --files <changed files>` — all hooks pass

### Backward compatibility

The new field defaults to `None`, and `None` falls back to `"consent"` at the call site, so existing callers see no behavior change. No tests were modified to accommodate new behavior — only the previously-untested default was locked in.

### Checklist

- [x] Read `CONTRIBUTING.md`.
- [x] Self-reviewed the change.
- [x] Added/updated unit tests.
- [x] Ran unit tests locally.


rohityan:
Hi @akshay-kumar-bm , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

GWeale:
Thanks so much for this! The configurable OAuth prompt support has since landed in main, so I'm closing this out as done. Really appreciate you taking the time to contribute it.

PR #5727, opened 2026-05-16
fix(sessions): add universal deduplication check in InMemorySessionService (#5723)

### LINKED ISSUE #5723: InMemorySessionService allows duplicate events to be appended during concurrent state broadcasts
### Description

### Proposed Solution
```python
    if event.partial:
        return event
        return event
    # ...
```

chriskinzel:
Fixes #5723 by adding a universal deduplication check at the beginning of InMemorySessionService.append_event.


rohityan:
Hi @chriskinzel , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

rohityan:

DeanChensj:
Close this one as it is already resolved by https://github.com/google/adk-python/commit/4d747747ee0313d7b30db1eed1b723ff329279d3

PR #5734, opened 2026-05-17
fix: force AUDIO modality for native-audio models in run_live

### LINKED ISSUE #4206: adk web: Native-audio models fail when TEXT modality is requested

hiteshtanejaa:
Link to Issue: [https://github.com/google/adk-python/issues/4206](https://github.com/google/adk-python/issues/4206)
Closes: #4206

Solution:
While the issue suggests fixing this in adk_web_server.py, this PR takes a deeper approach by fixing it in Runner.run_live() in runners.py. This is more robust because:

adk_web_server.py is just one entry point — run_live can be called directly by other code too
The fix lives closer to where the modality is actually used, protecting all callers not just the WebSocket endpoint
Testing Plan
Unit Tests:

1. I have added or updated unit tests for my change.

2. All unit tests pass locally.

```
pytest tests/unittests/test_runners.py -k "native_audio" -v
tests/unittests/test_runners.py::test_run_live_native_audio_model_forces_audio_modality PASSED
tests/unittests/test_runners.py::test_run_live_native_audio_model_enables_transcription_when_audio PASSED
2 passed in 7.78s
```

Manual End-to-End (E2E) Tests:
gemini-live-2.5-flash-native-audio requires Vertex AI / allowlist access not available on standard Gemini API, so full E2E was not possible locally. The fix is validated at the unit level by directly exercising Runner.run_live with a mock native-audio agent.

Checklist

1. I have read the [CONTRIBUTING.md](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md) document. ✅

2. I have performed a self-review of my own code. ✅

3. I have commented my code, particularly in hard-to-understand areas. ✅

4. I have added tests that prove my fix is effective or that my feature works. ✅

5. New and existing unit tests pass locally with my changes. ✅

6. I have manually tested my changes end-to-end. ❌

7. Any dependent changes have been merged and published in downstream modules. ✅

Additional context
Root cause: Runner.run_live only set a default modality when response_modalities was None, but did not validate an explicitly set TEXT modality against the model's capabilities. Native-audio models only support AUDIO, causing an API error.

What this PR does:
Detects native-audio models via canonical_model.model name check
Overrides response_modalities to ['AUDIO'] with a warning log
Auto-enables output_audio_transcription so users can read audio responses as text in the UI
Known limitation: Detection relies on the native-audio substring in the model name. A more robust long-term solution would be capability-based validation once the API exposes model modality support directly.


rohityan:
Hi @hiteshtanejaa , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Pls fix the mypy-diff tests.

wuliang229:
Turns out we don't need the change.

PR #5731, opened 2026-05-17
fix(sessions): respect num_recent_events=0 in DatabaseSessionService

tiino:
### Link to Issue or Description of Change

**1. Link to an existing issue:**

- Closes: #5730

### Testing Plan

**Unit Tests:**

- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

Added a boundary assertion for `num_recent_events=0` to the existing
`test_get_session_with_config`. That test is parametrized over all four
in-tree session backends, so a single assertion covers
`InMemorySessionService`, `InMemorySessionService` (light-copy variant),
`DatabaseSessionService`, and `SqliteSessionService`.

Before the fix (only the test added), the new assertion fails on the
`DATABASE` backend and passes on the other three — confirming the bug
and its scope:

```
$ pytest tests/unittests/sessions/test_session_service.py::test_get_session_with_config -v
tests/unittests/sessions/test_session_service.py::test_get_session_with_config[SessionServiceType.IN_MEMORY] PASSED
tests/unittests/sessions/test_session_service.py::test_get_session_with_config[SessionServiceType.IN_MEMORY_WITH_LIGHT_COPY_ENABLED] PASSED
tests/unittests/sessions/test_session_service.py::test_get_session_with_config[SessionServiceType.DATABASE] FAILED
tests/unittests/sessions/test_session_service.py::test_get_session_with_config[SessionServiceType.SQLITE] PASSED
```

With the one-line fix in `database_session_service.py` applied, all four
backends pass, and the full `tests/unittests/sessions/` suite is green:

```
$ pytest tests/unittests/sessions/ -v
...
======================= 174 passed, 2 warnings in 22.87s =======================
```

**Manual End-to-End (E2E) Tests:**

The bug and fix can be reproduced end-to-end without any external
dependency by running this snippet against `DatabaseSessionService`
backed by an in-memory SQLite engine:

```python
import asyncio
from google.genai import types
from google.adk.events.event import Event
from google.adk.sessions.base_session_service import GetSessionConfig
from google.adk.sessions.database_session_service import DatabaseSessionService

async def main():
    svc = DatabaseSessionService(db_url="sqlite+aiosqlite:///:memory:")
    app, user, sid = "app", "user_1", "session_1"
    session = await svc.create_session(app_name=app, user_id=user, session_id=sid)
    for i in range(5):
        await svc.append_event(
            session=session,
            event=Event(
                invocation_id=f"inv_{i}",
[...]

PR #5736, opened 2026-05-18
fix(sessions): return empty events when num_recent_events=0 in DatabaseSessionService

### LINKED ISSUE #5730: `DatabaseSessionService.get_session` returns all events when `num_recent_events=0`
## 🔴 Required Information

**Describe the Bug:**

`DatabaseSessionService.get_session(..., config=GetSessionConfig(num_recent_events=0))` is documented to return **no events** (see `GetSessionConfig` docstring in `src/google/adk/sessions/base_session_service.py`: *"if 0, no events are returned"*).

However, when `num_recent_events=0` is passed, **all events of the session are returned** instead of an empty list.

This appears to be caused by a truthy check on `num_recent_events` in `src/google/adk/sessions/database_session_service.py` (around line 523):

```python
if config and config.num_recent_events:          # ← `0` is falsy, so this branch is skipped
    stmt = stmt.limit(config.num_recent_events)
```

Because `0` is falsy in Python, the `LIMIT` clause is never appended when `num_recent_events=0`, and the query returns every event for the session.

The other in-tree implementations appear to handle `0` correctly:

- `InMemorySessionService` (`src/google/adk/sessions/in_memory_session_service.py:204-211`): uses `is not None` and explicitly assigns `events = []` when `num_recent_events == 0`.
- `SqliteSessionService` (`src/google/adk/sessions/sqlite_session_service.py:264-271`): uses `is not None` for the `LIMIT` clause, plus an explicit early return for `== 0`.

So this issue appears to be specific to `DatabaseSessionService` (SQLAlchemy backend).

**Steps to Reproduce:**

1. Install `google-adk` (reproduced on 1.33.0 / current `main`) and `aiosqlite`.

2. Run the following script (uses SQLite in-memory, no external DB needed):

```python
import asyncio
from google.adk.sessions.base_session_service import GetSessionConfig
from google.adk.sessions.database_session_service import DatabaseSessionService

async def main():
    svc = DatabaseSessionService(db_url="sqlite+aiosqlite:///:memory:")
    app, user, sid = "app", "user_1", "session_1"

    session = await svc.create_session(app_name=app, user_id=user, session_id=sid)

    for i in range(5):
        ev = Event(
            invocation_id=f"inv_{i}",
            author="user" if i % 2 == 0 else "agent",
            content=types.Content(
                role="user" if i % 2 == 0 else "model",
                parts=[types.Part(text=f"event #{i}")],
            ),
        )
        await svc.append_event(session=session, event=ev)

    got = await svc.get_session(
        app_name=app,
        user_id=user,
        session_id=sid,
        config=GetSessionConfig(num_recent_events=0),
    )

    print("num_recent_events=0 -> events len =", len(got.events))

asyncio.run(main())
```

**Expected Behavior:**

```text
num_recent_events=0 -> events len = 0
```

This matches the behavior of `InMemorySessionService` for the equivalent call.

**Observed Behavior:**

```text
num_recent_events=0 -> events len = 5
```

All 5 events are returned.

Testing across multiple `num_recent_events` values suggests that `0` is the only problematic case:

```text
num_recent_events= None -> events len = 5   # OK (no filter)
num_recent_events=    0 -> events len = 5   # Incorrect behavior (should be 0)
num_recent_events=    2 -> events len = 2   # OK
num_recent_events=    5 -> events len = 5   # OK
```

The issue is reproducible regardless of the underlying database, because the affected branch is in Python (the `LIMIT` clause is simply never appended to the SQLAlchemy statement).

SQLite is used here only for convenience; PostgreSQL, MySQL, and MariaDB appear to be affected in the same way.

**Environment Details:**

- ADK Library Version (`pip show google-adk`): 1.33.0
- Desktop OS: Linux (WSL2, Ubuntu)
- Python Version (`python -V`): Python 3.12.3

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: gemini-3.1-flash-lite)

---

Thanks for taking a look.

mangod12:
## Summary
- Fixed `DatabaseSessionService.get_session` returning all events when `num_recent_events=0`
- Root cause: Python truthiness check (`if config.num_recent_events:`) treats `0` as falsy, skipping the `LIMIT` clause
- Fix: changed to `if config.num_recent_events is not None:`, consistent with `InMemorySessionService` and `SqliteSessionService`
- `LIMIT 0` in SQL correctly returns zero rows across all dialects (PostgreSQL, MySQL, SQLite)

## Changes
- `src/google/adk/sessions/database_session_service.py` — one-line fix (line 523)
- `tests/unittests/sessions/test_session_service.py` — added regression test for `num_recent_events=0`

Fixes #5730

## Test plan
- [x] Added `num_recent_events=0` assertion to `test_get_session_with_config`
- [x] Existing `num_recent_events=3` and `num_recent_events=None` test cases unchanged
- [x] Fix is consistent with `InMemorySessionService` (line 205) and `SqliteSessionService` (line 264) which both use `is not None`


mangod12:
I have signed the Google CLA. Please re-check.

rohityan:
Hi @mangod12 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

mangod12:
Rebased on latest `main` and verified formatting passes (`pre-commit run` on changed files — all hooks green).

### Test results (local, Python 3.12, Windows)

```
pytest tests/unittests/sessions/test_session_service.py
96 passed, 3 failed, 1 warning in 74.89s
```

The 3 failures are **pre-existing on `main`** (Windows-specific SQLAlchemy timestamp/greenlet issues, unrelated to this change):
- `test_create_get_session[IN_MEMORY_WITH_LIGHT_COPY_ENABLED]`
- `test_session_state_is_not_shared[DATABASE]`
- `test_get_session_with_config[DATABASE]` — confirmed same failure on `main`

All session service tests relevant to this fix pass, including the new `num_recent_events=0` regression test.

rohityan:

GWeale:
Thanks for this! Your exact one-line fix — if config and config.num_recent_events is not None: — is already on main, which is why GitHub now shows this as conflicting. LIMIT 0 correctly returns zero rows across SQLite/Postgres/MySQL, just as you noted. Closing as already landed. (Note: this overlaps with #5731, which fixed the same line.)

PR #5743, opened 2026-05-18
fix: deduplicate in-memory session events

### LINKED ISSUE #5723: InMemorySessionService allows duplicate events to be appended during concurrent state broadcasts

he-yufeng:
## Summary
- Deduplicate `InMemorySessionService.append_event` by event id before mutating session state.
- Check both the caller session and backing storage session, so stale/light-copy sessions do not append the same event twice.
- Add a regression test that verifies duplicate event ids are ignored and state deltas are not applied twice.

## To verify
- `python -m pytest tests\unittests\sessions\test_session_service.py -q -k "duplicate_event_id or same_ref_as_storage_session or partial_events" --basetemp=.tmp\pytest-adk-session-targeted`
- `python -m pytest tests\unittests\sessions\test_session_service.py -q -k "not DATABASE" --basetemp=.tmp\pytest-adk-session-no-database`
- `python -m pyink --check src\google\adk\sessions\in_memory_session_service.py tests\unittests\sessions\test_session_service.py`
- `python -m py_compile src\google\adk\sessions\in_memory_session_service.py tests\unittests\sessions\test_session_service.py`

Full `test_session_service.py` on Windows still hits an existing `SessionServiceType.DATABASE` failure when converting tiny test timestamps through `datetime.timestamp()` (`OSError: [Errno 22] Invalid argument`). The in-memory and sqlite coverage above passes.

Fixes #5723

### DIFF
--- a/src/google/adk/sessions/in_memory_session_service.py
+++ b/src/google/adk/sessions/in_memory_session_service.py
@@ -335,13 +335,25 @@ class InMemorySessionService(BaseSessionService):
     if session_id not in self.sessions[app_name][user_id]:
       _warning(f'session_id {session_id} not in sessions[app_name][user_id]')
       return event
+    storage_session = self.sessions[app_name][user_id][session_id]
+
+    if event.id and (
+        any(existing_event.id == event.id for existing_event in session.events)
+        or (
+            storage_session is not session
+            and any(
+                existing_event.id == event.id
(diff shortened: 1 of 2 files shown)


sasha-gitg:
closing in favor of #5723

PR #5745, opened 2026-05-18
feat(tools): add rolling window for activated skills in SkillToolset

lwangverizon:

**Problem:**
In long-running sessions, the `SkillToolset` accumulates activated skills unboundedly. Each `load_skill` call adds tools to the session state, growing the tool context sent to the LLM indefinitely. This can degrade LLM performance as the function declarations list grows, increase token consumption, and in extreme cases exceed context window limits.

Unlike AutoGen (static tools at construction) and LangGraph (static tool binding with dynamic routing), ADK's skill system allows mid-session tool expansion — which is powerful but requires bounds.

**Solution:**
Add an optional `max_activated_skills` parameter to `SkillToolset` that enforces a rolling window (LRU eviction) on concurrently activated skills per agent.

Key design decisions:
- **LRU semantics**: Re-loading an already-active skill promotes it to the most-recent position (no duplication)
- **Oldest-first eviction**: When the limit is exceeded, the oldest activated skills are evicted from session state
- **Backward compatible**: Default is `None` (no limit), preserving existing behavior
- **Minimal footprint**: ~15 lines of logic in `LoadSkillTool.run_async`, no new dependencies

### Testing Plan

**Unit Tests:**

```
100 passed, 6 warnings in 10.83s
```

7 new tests added covering:
- `test_max_activated_skills_evicts_oldest` — verifies oldest skill is evicted when limit exceeded
- `test_max_activated_skills_keeps_recent_n` — confirms only N most recent skills retained
- `test_max_activated_skills_reloading_moves_to_end` — validates LRU promotion on re-load
- `test_no_max_activated_skills_unlimited` — confirms backward compatibility (no limit)
- `test_evicted_skill_tools_not_resolved` — verifies evicted tools are excluded from `get_tools`
- `test_evicted_tool_call_raises_value_error` — confirms `_get_tool` raises ValueError for evicted tools
- `test_rolling_window_full_lifecycle_with_tool_resolution` — end-to-end lifecycle test

Tested with a multi-skill agent using `adk web` — confirmed that after exceeding `max_activated_skills=3`, older skills are no longer available in the LLM's tool declarations, and re-loading a skill correctly promotes it.

### Checklist

### Additional context

**Comparison with other frameworks:**
- **AutoGen**: Tools are static at agent construction — no mid-session expansion possible
- **LangGraph**: Static tool binding with dynamic routing via graph edges — no accumulation

ADK's skill system is uniquely powerful in allowing runtime tool discovery and loading, but this PR ensures it remains bounded and performant in long-running sessions.

**Edge case note:** If the LLM calls a tool from conversation history that was since evicted, `_get_tool()` in `functions.py` raises `ValueError`. This is recoverable via `on_tool_error` callbacks. The risk is minimal since evicted tools are removed from the next function declarations, so the LLM won't attempt to call them.


sasha-gitg:
Thanks for putting this PR together.  

We will not accept this PR because this can cause a performance/cost penalty through implicit context cache invalidation because the skill's tools get removed.

Additionally when the skill is evicted and it's tools are removed, the LLM will generate a function call for the evicted tool  because the previous turns introducing the skill are still present and the model has not received any contextual cues that it was removed.

lwangverizon:
Thank you, I was thinking longer conversation will most likely run even compaction or other practices to summarize previous tool call anyways, so maybe this can help in most cases. I would like to know what is the best practices to solve long running session context window problem, when multiple skills / tools are being cached in the session. Any recommendations or other plans to solve this issue?

PR #5754, opened 2026-05-19
fix(litellm): parse noncanonical tool call arguments

fancydirty:
## Summary

LiteLLM tool calls normally provide unction.arguments as a strict JSON string, but some OpenAI-compatible providers can finalize a complete tool call with noncanonical object-literal syntax such as Python dict literals or unquoted JSON object keys. ADK currently passes the string directly to json.loads(...), which raises JSONDecodeError and aborts response conversion before the tool call can be surfaced.

This PR adds a small parser helper for LiteLLM tool call arguments:

- strict JSON remains the primary path
- Python dict literal strings are accepted via st.literal_eval
- simple unquoted object keys are quoted without touching string contents, then parsed
- unrecoverable malformed payloads still raise the original JSONDecodeError

The same parser is used for the streaming max-token/truncation check so complete noncanonical arguments are not mistaken for truncated arguments.

## Why

Without this, a provider can stream a valid/intended tool call all the way to completion, but ADK crashes while converting the finalized LiteLLM message into LlmResponse.

## Tests

- PYTHONPATH=src /root/projects/adk-agent/.venv/bin/python -m pytest tests/unittests/models/test_litellm.py -k 'tool_call_accepts' -q
- PYTHONPATH=src /root/projects/adk-agent/.venv/bin/python -m pytest tests/unittests/models/test_litellm.py -k 'message_to_generate_content_response' -q
- PYTHONPATH=src /root/projects/adk-agent/.venv/bin/python -m pytest tests/unittests/models/test_litellm.py -q

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import ast
 import base64
 import binascii
 import copy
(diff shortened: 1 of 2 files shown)


fancydirty:
Closing this draft and reopening from a neutral branch name with the same patch.

PR #5760, opened 2026-05-20
fix: add SSRF protection to RestApiTool._request()

Neelagiri65:
## What

`RestApiTool._request()` passes URLs to `httpx.AsyncClient.request()` with no validation. After #4368 fixed SSRF in `load_web_page` (hostname blocking, DNS pre-resolution, IP range filtering, scheme restriction), `RestApiTool` was not updated with the same protections.

An OpenAPI spec with `servers[].url` pointing at `127.0.0.1` or `169.254.169.254` goes through without any check.

## Changes

- New shared module `_ssrf_protection.py` with the validation logic (scheme check, hostname blocking, DNS resolution, IP range filtering)
- `_request()` calls `validate_url()` before making the HTTP request
- Timeout changed from `None` to `30` seconds
- 24 new tests covering blocked hosts, schemes, and the `_request` integration

## What gets blocked

- `localhost` and `*.localhost`
- Loopback (`127.0.0.0/8`, `::1`)
- Link-local (`169.254.0.0/16`) — covers cloud metadata endpoints
- Private ranges (`10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`)
- Non-http(s) schemes (`file://`, `ftp://`, etc.)

## Why a shared module

`load_web_page.py` has its own copies of these functions. The shared module avoids duplicating the logic a third time. `load_web_page` can be migrated to use it in a follow-up if wanted.

Related: #4368

### DIFF
--- a/src/google/adk/tools/_ssrf_protection.py
+++ b/src/google/adk/tools/_ssrf_protection.py
@@ -1,0 +1,84 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


sasha-gitg:
Feel free to re-open once CLA is signed.

Issue #5428, opened 2026-04-21
FunctionTool automatic parsing fails on `list[tuple[bytes, str]]` parameter type


kaligautier:
## 🔴 Required Information

**Describe the Bug:**
`FunctionTool` automatic function declaration parsing crashes at startup when a tool function has a parameter typed `list[tuple[bytes, str]] | None`. The error originates from `_function_parameter_parse_util._raise_for_unsupported_param` after Pydantic's `Schema.model_validate` rejects the generated schema with `Extra inputs are not permitted`.

This prevents the agent from starting entirely — it's not a runtime error but a startup crash during tool declaration building.

**Steps to Reproduce:**

1. Define a tool function with a complex parameter type:
```python
from google.adk.tools import ToolContext
from google.adk.tools.function_tool import FunctionTool

async def generate_image(
    prompt: str,
    input_bytes: list[tuple[bytes, str]] | None = None,
    tool_context: ToolContext = None,
) -> dict:
    """Generate an image from a prompt."""
    return {"status": "success"}

generate_image_tool = FunctionTool(func=generate_image)
```

2. Register the tool on an `LlmAgent` and start the agent (e.g. via A2A executor or `Runner`).

3. The agent crashes during `_preprocess_async` → `_process_agent_tools` → `tool._get_declaration()`:

**Expected Behavior:**
Either:
- ADK should support `tuple` types in function signatures by mapping them to a JSON Schema `array` with `prefixItems`, or
- ADK should provide a clear, documented mechanism to exclude specific parameters from automatic declaration parsing (e.g., an `ignore_params` option on `FunctionTool`), or
- At minimum, the error message should link to documentation on how to implement the manual declaration workaround.

**Observed Behavior:**
Two chained exceptions:

```text
pydantic_core._pydantic_core.ValidationError: 1 validation error for Schema
  Extra inputs are not permitted [type=extra_forbidden, input_value=[{'format': 'binary', 'ty...g'}, {'type': 'string'}], input_type=list]
```

Followed by:

```text
ValueError: Failed to parse the parameter input_bytes: list[tuple[bytes, str]] | None = None
of function generate_image for automatic function calling.
Automatic function calling works best with simpler function signature schema,
consider manually parsing your function declaration for function generate_image.
```

**Environment Details:**

- ADK Library Version: `1.26.0`
- Desktop OS: Linux (Cloud Run container) / macOS (local dev)
- Python Version: 3.12

**Model Information:**

- Are you using LiteLLM: No
[...]

PR #5800, opened 2026-05-22
fix(tools): normalize tuple schemas for Gemini API

EaCognitive:
Re-opening my fix for #5428 since the original issue was closed due to inactivity from the reporter. 

My previous PR (#5446) accidentally included broken `cli` and `noxfile.py` files which caused the CI failures. This is a clean submission containing ONLY the strip-based workaround verified to fix the Cloud Run deployment issues.

## Testing Plan
- Mypy-diff verified clean against `main`.
- `pre-commit run --all-files` passes format and license checks.
- Added 4 explicit unit tests in `tests/unittests/tools/test_function_parameter_parse_util.py` to verify `prefixItems` and `unevaluatedItems` are successfully recursively stripped.
- Multi-version tests (Python 3.10-3.14) pass successfully.

## Logs
```text
$ pytest tests/unittests/tools/test_function_parameter_parse_util.py
============================= test session starts ==============================
platform darwin -- Python 3.11.15, pytest-9.0.3, pluggy-1.6.0
collected 4 items

tests/unittests/tools/test_function_parameter_parse_util.py ....         [100%]

============================== 4 passed in 0.05s ===============================
```

cc @surajksharma07 @rohityan


sasha-gitg:
We switched to Gemini JSON schema by default so I don't think this PR is necessary because from_function_with_options is bypassed.

PR #5803, opened 2026-05-22
fix: lazy-load GCS skill storage dependency

### LINKED ISSUE #5801: Missing dependency google-cloud-storage for skills in V2.0.0
## 🔴 Required Information

**Describe the Bug:**  
In ADK 2.0, importing `google.adk.skills` fails at application startup when the `skills` package is used without installing the optional dependency group `all`.

The module `google/adk/skills/_utils.py` imports Google Cloud Storage unconditionally:

```python
from google.cloud import storage
```

However, `google-cloud-storage>=2.18,<4` is only declared under `optional-dependencies.all` in `pyproject.toml`, not as a required dependency for the core package or for using `google.adk.skills`.

As a result, any application importing `google.adk.skills` crashes with:

```plain text
ImportError: cannot import name 'storage' from 'google.cloud' (unknown location)
```

This happens even if the application does not explicitly use Google Cloud Storage functionality.

**Steps to Reproduce:**

1. Install ADK 2.0 without optional dependencies:

```shell script
pip install google-adk==2.0.0
```

2. Create or run an ADK application that imports or uses `google.adk.skills`, for example:

```python
```

3. Start the application.

4. Observe that startup fails with an import error caused by `google.adk.skills._utils` importing `google.cloud.storage`.

**Expected Behavior:**  
Using `google.adk.skills` should not require `google-cloud-storage` unless functionality that actually depends on Google Cloud Storage is used.

Either:

- `google-cloud-storage` should be declared as a required dependency if `google.adk.skills` always needs it, or
- the import should be lazy/optional and only executed when GCS-related functionality is used, with a clear error message if the dependency is missing.

**Observed Behavior:**  
The application fails during startup when importing `google.adk.skills`, because `google.cloud.storage` is imported unconditionally although `google-cloud-storage` is only an optional dependency.

Stack trace:

```plain text

  File "/app/.venv/lib/python3.12/site-packages/google/adk/skills/__init__.py", line 20, in <module>
    from ._utils import _list_skills_in_dir as list_skills_in_dir
  File "/app/.venv/lib/python3.12/site-packages/google/adk/skills/_utils.py", line 27, in <module>
    from google.cloud import storage
ImportError: cannot import name 'storage' from 'google.cloud' (unknown location)
```

**Environment Details:**

- ADK Library Version `(pip show google-adk)`: `2.0.x`
- Desktop OS: `Linux / Docker container`
- Python Version `(python -V)`: `Python 3.12`

**Model Information:**

- Are you using LiteLLM: `N/A`
- Which model is being used: `N/A`

---

## 🟡 Optional Information

**Regression:**  
Yes, 1.34.0 didn't have this problem

**Logs:**  
N/A

**Screenshots / Video:**  
N/A

**Additional Context:**  
The dependency `google-cloud-storage>=2.18,<4` appears to be listed only under `optional-dependencies.all` in `pyproject.toml`, but `google.adk.skills._utils` imports it at module import time.

This makes `google-cloud-storage` effectively required for importing `google.adk.skills`, even when no Google Cloud Storage related functionality is used.

Possible fixes:

1. Move `google-cloud-storage>=2.18,<4` to required dependencies if `skills` always depends on it.
2. Add a dedicated optional extra for skills, for example:
```toml
[project.optional-dependencies]
   skills = ["google-cloud-storage>=2.18,<4"]
```
3. Make the import lazy inside the functions that actually require GCS and raise a clear dependency error only when those functions are used.

```python

print("Imported google.adk.skills successfully")
```

Run with:

```shell script
pip install google-adk==2.0.*
python reproduce.py
```

Expected result:

```plain text
Imported google.adk.skills successfully
```

Actual result:

```plain text
ImportError: cannot import name 'storage' from 'google.cloud' (unknown location)
```

- Always (100%)

he-yufeng:
﻿Fixes #5801.

## Summary
- move the `google.cloud.storage` import behind the GCS-only skills helpers
- keep non-GCS `google.adk.skills` imports working without `google-cloud-storage`
- update the GCS skills unit tests to patch the local storage loader instead of requiring the external package at import time

## To verify
- `python -m pytest tests/unittests/skills/test__utils.py tests/unittests/test_optional_dependencies.py::test_skills_models_import_without_google_cloud_storage -q`
- `python -m py_compile src/google/adk/skills/_utils.py tests/unittests/skills/test__utils.py tests/unittests/test_optional_dependencies.py`
- `python -m pyink --check src/google/adk/skills/_utils.py tests/unittests/skills/test__utils.py tests/unittests/test_optional_dependencies.py`
- `git diff --check`

### DIFF
--- a/src/google/adk/skills/_utils.py
+++ b/src/google/adk/skills/_utils.py
@@ -24,7 +24,6 @@ from typing import Union
 import zipfile
 
 from google.auth import credentials as auth
-from google.cloud import storage
 from pydantic import ValidationError
 import yaml
 
@@ -39,6 +38,18 @@ _ALLOWED_FRONTMATTER_KEYS = frozenset({
     "metadata",
     "compatibility",
 })
+
+
+def _get_storage_module():
+  try:
+    from google.cloud import storage
+  except ImportError as e:
+    raise ImportError(
+        "google-cloud-storage is required for GCS skill loading. Install it"
+        " with `pip install google-cloud-storage` or `pip install"
+        " 'google-adk[all]'`."
+    ) from e
+  return storage
 
 
 def _load_dir(directory: pathlib.Path) -> dict[str, str]:
@@ -407,6 +418,7 @@ def _list_skills_in_gcs_dir(
   Returns:
     Dictionary mapping skill IDs to their frontmatter.
   """
+  storage = _get_storage_module()
   client = storage.Client(project=project_id, credentials=credentials)
   bucket = client.bucket(bucket_name)
 
@@ -467,6 +479,7 @@ def _load_skill_from_gcs_dir(
       the directory name.
   """
 
+  storage = _get_storage_module()
   client = storage.Client(project=project_id, credentials=credentials)
   bucket = client.bucket(bucket_name)
 
--- a/tests/unittests/skills/test__utils.py
+++ b/tests/unittests/skills/test__utils.py
@@ -190,9 +190,13 @@ Body content
   assert fm.license == "MIT"
 
 
-@mock.patch("google.cloud.storage.Client")
-def test__list_skills_in_gcs_dir(mock_client_class):
+@mock.patch("google.adk.skills._utils._get_storage_module")
+def test__list_skills_in_gcs_dir(mock_get_storage_module):
 
+  mock_client_class = mock.MagicMock()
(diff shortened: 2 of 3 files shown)


sasha-gitg:
closing since this resolved in https://github.com/google/adk-python/commit/416775dcceae8f57badf0cd55c44148f6426b6db

PR #5815, opened 2026-05-22
fix(sessions): deduplicate events in InMemorySessionService.append_event

### LINKED ISSUE #5723: InMemorySessionService allows duplicate events to be appended during concurrent state broadcasts

devteamaegis:
## Problem

When an orchestrator broadcasts a shared-state delta to several concurrent session references, `append_event` can be called more than once with the **same event object** (same `event.id`). The current implementation has no guard against this, so:

1. The event is appended to `storage_session.events` multiple times.
2. `state_delta` is applied multiple times, corrupting session/user/app state.

This is reproducible by holding two references to the same session (e.g. `session_a = get_session(...)` then `session_b = get_session(...)`) and calling `await service.append_event(session_a, event)` followed by `await service.append_event(session_b, event)`.

Closes #5723.

## Fix

Fetch `storage_session` **before** calling `super().append_event()` and short-circuit early if the event ID is already present:

```python
storage_session = self.sessions[app_name][user_id].get(session_id)
if any(e.id == event.id for e in storage_session.events):
    return event
```

This is a pure in-memory lookup (O(n) over events, negligible for typical session sizes) and has no effect on the happy path.

## Tests

Two regression tests added to `tests/unittests/sessions/test_session_service.py`:

- `test_append_event_is_idempotent_for_same_event_id` — appends the same event twice via two different session references; asserts exactly one event is stored and `state_delta` is applied only once.
- `test_append_different_events_not_deduplicated` — appends two events with distinct IDs; asserts both are stored (confirms the guard doesn't over-suppress).

### DIFF
--- a/src/google/adk/sessions/in_memory_session_service.py
+++ b/src/google/adk/sessions/in_memory_session_service.py
@@ -336,12 +336,20 @@ class InMemorySessionService(BaseSessionService):
       _warning(f'session_id {session_id} not in sessions[app_name][user_id]')
       return event
 
+    # Fetch the canonical storage session early so we can check for duplicate
+    # event IDs before modifying any state.  The same event can be delivered
+    # more than once when the orchestrator broadcasts a shared-state delta to
+    # several concurrent session references; deduplicating here prevents
+    # double-application of state updates and duplicate entries in event lists.
+    storage_session = self.sessions[app_name][user_id].get(session_id)
+    if any(e.id == event.id for e in storage_session.events):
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @devteamaegis , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing tests before we can proceed with the review.

PR #5615, opened 2026-05-06
fix: enforce path boundary in built-in agent file tools (path traversal)

adilburaksen:
## Summary

Fixes a path traversal vulnerability in the Agent Builder Assistant's file tools
(`read_files`, `write_files`, `delete_files`, `explore_project`).

**Supersedes #5413** — the YAML callback RCE portion of #5413 was already addressed
internally (2026-04-21). This PR contains only the `resolve_root_directory.py` fix.

### Root cause

Two flaws in `resolve_root_directory.py`:

**Flaw 1 — Absolute path bypass** (line 51–52 before fix):
```python
if file_path_obj.is_absolute():
    return file_path_obj   # No boundary check
```
Any absolute path passed to `read_files`/`write_files`/`delete_files` was returned
directly, allowing `/etc/passwd`, `/root/.ssh/id_rsa`, etc.

**Flaw 2 — Attacker-controlled `root_directory`** (line 56–57 before fix):
```python
if session_state and "root_directory" in session_state:
    root_directory = session_state["root_directory"]
```
Session state is set via `POST /apps/{app_name}/users/{user}/sessions`. An attacker
can inject `root_directory: "/"` to redirect all relative file paths under `/`.

### Fix

- **`_validate_root_directory()`**: rejects absolute paths and `..` components from
  session-state `root_directory`
- **Boundary check**: both absolute and relative `file_path` values are resolved to
  a real path and verified to be within `resolved_root` via `candidate.relative_to()`
- Removes the unconditional `is_absolute → return` bypass

### Attack scenario (for context)

```
POST /apps/__agent/users/attacker/sessions
{"state": {"root_directory": "/"}}

POST /run  (app_name=__agent)
{"new_message": {"parts": [{"text": "read /etc/passwd"}]}}
→ returns /etc/passwd contents
```

No authentication bypass required — `/run` and session creation have no auth by default
in dev mode; the intended security boundary is container isolation (documented in ADK).

This PR does not change that model — it closes the path escape within the boundary.

### Testing

Added `_validate_root_directory()` unit tests cover: absolute path rejection,
`..` component rejection, null-byte rejection, valid relative paths pass through.

PR #5826, opened 2026-05-23
fix(agents): prevent path traversal in AgentTool config_path resolution

adilburaksen:
## Summary

`resolve_agent_reference` in `config_agent_utils.py` accepted absolute `config_path` values unconditionally and joined relative paths without any boundary validation. An attacker-controlled `config_path` field in an agent YAML could traverse outside the intended agent directory.

**Vulnerable pattern (before):**
```python
if os.path.isabs(ref_config.config_path):
    return from_config(ref_config.config_path)   # absolute accepted
else:
    return from_config(os.path.join(agent_dir, ref_config.config_path))  # no ".." check
```

**PoC config:**
```yaml
tools:
  - tool_class: AgentTool
    args:
      agent:
        config_path: "../../../../../../etc/passwd"
```

This causes `open("/etc/passwd", "r")` server-side. A `FileNotFoundError` vs `ValidationError` difference also leaks path existence.

## Fix

- Reject absolute `config_path` values with `ValueError`
- Normalize the joined path and verify it stays within the parent agent directory before calling `from_config`

## Related

Same vulnerability exists in `adk-java` (PR: google/adk-java#...) and `adk-go` (PR: google/adk-go#...) — fix pattern is identical.

Note: This is distinct from the `resolve_code_reference` RCE (different function, different field, file-read impact vs code execution).


adilburaksen:
I have read the CLA Documents and I hereby sign the CLA.

rohityan:
Hi @adilburaksen , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing unit tests before we can proceed with the review.

adilburaksen:
Fixed the failing tests and pre-commit issues in the latest commit (8be0661):

- **Test failure**: `test_resolve_agent_reference_uses_windows_dirname` was failing because the traversal boundary check used `os.sep` (always `/` on Linux) instead of `os.path.sep`. The test patches `os.path` with `ntpath`, so `os.path.sep` correctly returns `\\` for Windows path comparisons while `os.sep` remained `/`, causing the `startswith` check to fail on valid Windows-style paths.

- **Pre-commit**: fixed end-of-file newlines, trailing whitespace, and isort import ordering from upstream merge.

rohityan:

adilburaksen:
@wyf7107 — would you be able to take a look at this small security fix when you have a moment?

It adds a path-boundary check to `AgentTool` `config_path` resolution in `resolve_agent_reference` (`config_agent_utils.py`), which currently accepts absolute paths unconditionally and never validates that a relative `config_path` stays within the parent agent directory → arbitrary file read from an attacker-supplied agent YAML. This mirrors the validation already applied to the artifact-service path in commit `cbcb5e6`.

A couple of related small security PRs are open if it's easier to review the class together: **#5615** (path boundary for the built-in agent file tools / `resolve_file_path`). Happy to rebase or switch to whatever path-validation helper you prefer.

adilburaksen:
@wukath friendly ping — this one's been green and waiting for review for a couple of weeks. It's a small path-traversal fix in `resolve_agent_reference` (an absolute or `..` `config_path` isn't bounded to the agent dir, so it can read files outside it). Mind taking a look or routing it to the right reviewer? Happy to adjust anything.

PR #5832, opened 2026-05-24
Fix/5602 longrunningfunctiontool resume

### LINKED ISSUE #5602: Compaction breaks LongRunningFunctionTool resume
## 🔴 Required Information

`EventsCompactionConfig` corrupts sessions that use `LongRunningFunctionTool`. The intermediate "pending" `function_response` that ADK emits when a long-running function returns shares the same `id` as the original `function_call`. The compactor's pending-call guard treats the call as already resolved, folds the `function_call` event into a text summary, and the eventual client-side resume `FunctionResponse` then fails contents validation with `No function call event found for function responses ids: {...}`.

**Steps to Reproduce:**

1. Build an `App` with both `EventsCompactionConfig` and `ResumabilityConfig(is_resumable=True)`:
   ```python
   App(
       name="repro",
       root_agent=agent,
       events_compaction_config=EventsCompactionConfig(
           compaction_interval=3,
           overlap_size=1,
           summarizer=LlmEventSummarizer(llm=Gemini(model="gemini-2.5-flash-lite")),
       ),
   )
   ```
2. Register a `LongRunningFunctionTool` whose wrapped function returns `{"status": "pending", ...}` (the standard LRF intermediate-response pattern).
3. Chat with the agent until the turn that triggers the LRF is the Nth invocation, where N ≥ `compaction_interval`. Compaction runs at the end of that turn, immediately after the LRF emits its `pending` intermediate response and pauses and then folds the still-open `function_call` event into a summary, all before the client gets a chance to resume.
4. Send the resume:
   ```python
   Content(role="user", parts=[Part(function_response=FunctionResponse(
       id=function_call_id, name=function_name, response={"status": "confirmed", ...}
   ))])
   ```
4. ADK raises during contents assembly: `No function call event found for function responses ids: {'adk-<uuid>'}`.

**Expected Behavior:**

The `function_call` event of a `LongRunningFunctionTool` should not be eligible for compaction until a *final* `FunctionResponse` (the client-supplied resume payload) has been appended to the session. Sending the resume should succeed and the agent should continue normally.

**Observed Behavior:**

The compactor folds the LRF `function_call` + intermediate `function_response` pair into a summary text event. When the resume `FunctionResponse` arrives, the contents request processor cannot find a matching raw `function_call` event in session history and raises:

```
No function call event found for function responses ids: {'adk-e5f8e450-3db7-4fef-8afa-d315774a593d'}
```

**Environment Details:**

- ADK Library Version: `google-adk 1.31.0`
- Desktop OS: Windows 11
- Python Version: 3.11

**Model Information:**

- Are you using LiteLLM: No
- Which model is being used: `gemini-2.5-pro` for the agent, `gemini-2.5-flash-lite` for the compaction summarizer 

---

## 🟡 Optional Information

**Regression:**

N/A — observed on 1.31.0.

**Logs:**

```text
ValueError: No function call event found for function responses ids: {'adk-…'}
```

**Additional Context:**

Root cause is in `google/adk/apps/compaction.py`:

`_pending_function_call_ids` ([compaction.py:280-294](https://github.com/google/adk-python/blob/main/src/google/adk/apps/compaction.py#L280)) computes pending calls as `all_call_ids - all_response_ids`. For a `LongRunningFunctionTool`, ADK emits an intermediate `function_response` event sharing the call's `id` as soon as the wrapped function returns its pending dict. That id ends up in `all_response_ids`, so the call is excluded from the pending set. `_truncate_events_before_pending_function_call` ([compaction.py:303-310](https://github.com/google/adk-python/blob/main/src/google/adk/apps/compaction.py#L303)) consequently allows the call event to be compacted.

The signal needed for the fix is already on the call event: `Event.long_running_tool_ids: set[str]`. The compactor just isn't consulting it.

**Suggested fix** — treat any call id present in `long_running_tool_ids` as pending until the framework can distinguish the intermediate response from the final one:

```python
def _pending_function_call_ids(events: list[Event]) -> set[str]:
    all_call_ids: set[str] = set()
    all_response_ids: set[str] = set()
    long_running_ids: set[str] = set()
    for event in events:
        all_call_ids.update(_event_function_call_ids(event))
        all_response_ids.update(_event_function_response_ids(event))
        if event.long_running_tool_ids:
            long_running_ids.update(event.long_running_tool_ids)
    return (all_call_ids - all_response_ids) | long_running_ids
```

But once an LRF call exists in a session, every compaction window that contains it will skip it forever (since `long_running_tool_ids` is permanently set on the call event). The cleaner long-term fix is to mark the *intermediate* `function_response` event with a flag (e.g. `actions.is_intermediate_long_running_response`) so `_pending_function_call_ids` can distinguish intermediate from final responses and unblock compaction once the resume is delivered. Happy to open a PR if the framing is right.

**Workaround**:  Subclass ` LlmEventSummarizer` and override `maybe_summarize_events` to return None whenever the candidate window contains an event whose ` long_running_tool_ids` are not all matched by a non-pending function_response in the same window. ADK skips appending a compaction event for that round; the next round retries on a wider window. Relies on the convention that LRF intermediates use {"status": "pending", ...} and final/resume payloads do not .

- Always (100%) — deterministic given a compaction interval that fires while an LRF is unresolved.

ilyesJammouci:
### Summary

Compaction incorrectly folded unresolved long-running function call events when an intermediate/pending `function_response` (e.g. with `status: "pending"`) shared the same id as the `function_call`.

This caused resume responses arriving later to fail during contents assembly with:

```text
No function call event found for function responses ids: {...}
```

As a result, resumability for long-running tools was broken.

---

### `event_actions.py`

- Added `is_intermediate_long_running_response` to `EventActions`
- Used to mark pending LRF `function_response` events

### `compaction.py`

- Updated `_event_function_response_ids`
- Intermediate LRF responses are now ignored when computing resolved response ids
- Prevents compaction from treating pending responses as final responses

### `functions.py`

- When emitting a pending live long-running tool response:

```python
tool_context.actions.is_intermediate_long_running_response = True
```

- Ensures the produced event is properly flagged

---

## Before

Compaction scanned all `function_response` ids unconditionally.

An intermediate/pending `function_response` could hide the matching `function_call` id from pending-call detection, allowing the compactor to prematurely fold away the `function_call` event.

---

## After

Intermediate/pending LRF `function_response` events are explicitly marked and ignored during compaction when computing resolved response ids.

This preserves unresolved `function_call` events until the real resume response arrives, maintaining resumability semantics.

---

## Motivation

Long-running tools emit intermediate/pending updates (e.g. `status: pending`) that should not be treated as final responses during compaction.

Previously, folding the matching `function_call` event caused later resume responses to become orphaned, leading to failures during contents assembly.

This fix introduces an explicit marker and minimally adjusts compaction response-id extraction logic to avoid the issue.

---

### Unit Test

Added regression test:

```text
tests/unittests/apps/test_compaction.py::test_sliding_window_preserves_long_running_call_until_resume
```

The test:
- Reproduces the compaction + resume failure scenario
- Verifies the long-running `function_call` is preserved until the actual resume response arrives

Run:

```bash
pytest tests/unittests/apps/test_compaction.py::test_sliding_window_preserves_long_running_call_until_resume
```

Expected:
- Test passes
- No contents assembly exception occurs when the resume response arrives

### Manual Verification

- Run existing compaction and resumability flows
- Confirm:
  - Non-LRF behavior remains unchanged
  - LRF resume flow now succeeds correctly


rohityan:
Hi @ilyesJammouci , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors.

GWeale:
closing this. main handles the same case in contents.py, _recover_compacted_function_calls, so the new EventActions field here would be a second mechanism for one bug

PR #5849, opened 2026-05-25
perf: optimize session cache warmup for large agent graphs

Optimia-co:
Minor performance improvement for session cache initialization.

Reduces warmup time when initializing large agent graphs by pre-allocating cache entries.


Optimia-co:
Closing test PR - not needed.

PR #5866, opened 2026-05-27
fix(migration): restrict unpickling of v0 actions blobs

White-Mouse:
## What

The v0 session schema stored event actions as pickled blobs. The migration helper reads raw bytes via `SELECT * FROM events` and previously used `pickle.loads(...)` directly.

This PR replaces the default load path with a restricted unpickler allowlist for builtin containers/primitives, standard ADK `EventActions` payloads, nested ADK core action types (`AuthConfig`, `ToolConfirmation`, `EventCompaction`), and the `google.genai.types.Content` / `Part` dependency classes that normal compaction payloads require.

It also adds an explicit trust toggle for legacy databases that contain custom Python objects in `state_delta` or other `Any` fields:

- Python API: `migrate(..., allow_unsafe_unpickling=True)`
- Migration runner: `upgrade(..., allow_unsafe_unpickling=True)`
- CLI: `adk migrate session --allow_unsafe_unpickling ...`
- Direct script: `--allow_unsafe_unpickling` / `--allow-unsafe-unpickling`

## Why

`pickle` is not safe for untrusted inputs. Migration tooling often runs against restored/backed-up DB files or shared storage; failing closed by default reduces the blast radius if the source DB contents are compromised.

The opt-in flag keeps compatibility for users who trust their source database and need the original unsafe pickle behavior for custom legacy objects.

## Associated Issue / Background

No existing GitHub issue is linked. This was found while reviewing the v0-to-v1 migration path for unsafe deserialization risks in legacy session data.

## Compatibility / fail-closed boundary

Normal v0 `EventActions` payloads made from primitive/container fields continue to migrate. The allowlist now also covers common nested ADK action models requested during review, including requested auth configs, requested tool confirmations, and event compaction content.

Payloads that require globals outside the explicit allowlist still fail closed by default: the migration logs a warning and falls back to empty `EventActions()` for that event. Users can opt into the previous unsafe pickle behavior only when they trust the source database.

## Verification

- `uv run pytest tests/unittests/sessions/migration/test_migration.py`
  - `22 passed, 4 warnings`
- `uv run mypy src/google/adk/sessions/migration/migrate_from_sqlalchemy_pickle.py src/google/adk/sessions/migration/migration_runner.py`
  - `Success: no issues found in 2 source files`
- `uv run pre-commit run --files src/google/adk/sessions/migration/migrate_from_sqlalchemy_pickle.py src/google/adk/sessions/migration/migration_runner.py src/google/adk/cli/cli_tools_click.py tests/unittests/sessions/migration/test_migration.py`
  - passed
- `git diff --check`
  - passed

### DIFF
--- a/src/google/adk/sessions/migration/migrate_from_sqlalchemy_pickle.py
+++ b/src/google/adk/sessions/migration/migrate_from_sqlalchemy_pickle.py
@@ -24,6 +24,7 @@ import logging
 import pickle
 import sys
 from typing import Any
+from typing import cast
 
 from google.adk.events.event import Event
 from google.adk.events.event_actions import EventActions
@@ -51,7 +52,48 @@ _ALLOWED_PICKLE_GLOBALS: set[tuple[str, str]] = {
     ("builtins", "float"),
     ("builtins", "bool"),
     # Expected pickled payload for v0 session schema events.
+    ("fastapi.openapi.models", "APIKey"),
+    ("fastapi.openapi.models", "APIKeyIn"),
+    ("fastapi.openapi.models", "HTTPBase"),
+    ("fastapi.openapi.models", "HTTPBearer"),
+    ("fastapi.openapi.models", "OAuth2"),
+    ("fastapi.openapi.models", "OAuthFlow"),
+    ("fastapi.openapi.models", "OAuthFlowAuthorizationCode"),
+    ("fastapi.openapi.models", "OAuthFlowClientCredentials"),
+    ("fastapi.openapi.models", "OAuthFlowImplicit"),
+    ("fastapi.openapi.models", "OAuthFlowPassword"),
+    ("fastapi.openapi.models", "OAuthFlows"),
+    ("fastapi.openapi.models", "OpenIdConnect"),
+    ("fastapi.openapi.models", "SecurityBase"),
+    ("fastapi.openapi.models", "SecurityScheme"),
+    ("fastapi.openapi.models", "SecuritySchemeType"),
+    ("google.adk.auth.auth_credential", "AuthCredential"),
+    ("google.adk.auth.auth_credential", "AuthCredentialTypes"),
+    ("google.adk.auth.auth_credential", "HttpAuth"),
+    ("google.adk.auth.auth_credential", "HttpCredentials"),
+    ("google.adk.auth.auth_credential", "OAuth2Auth"),
+    ("google.adk.auth.auth_credential", "ServiceAccountCredential"),
+    ("google.adk.auth.auth_schemes", "CustomAuthScheme"),
+    ("google.adk.auth.auth_schemes", "ExtendedOAuth2"),
+    ("google.adk.auth.auth_schemes", "OAuthGrantType"),
+    ("google.adk.auth.auth_schemes", "OpenIdConnectWithConfig"),
+    ("google.adk.auth.auth_tool", "AuthConfig"),
     ("google.adk.events.event_actions", "EventActions"),
+    ("google.adk.events.event_actions", "EventCompaction"),
+    ("google.adk.tools.tool_confirmation", "ToolConfirmation"),
+    ("google.genai.types", "Blob"),
+    ("google.genai.types", "CodeExecutionResult"),
+    ("google.genai.types", "Content"),
+    ("google.genai.types", "ExecutableCode"),
+    ("google.genai.types", "FileData"),
+    ("google.genai.types", "FunctionCall"),
+    ("google.genai.types", "FunctionResponse"),
+    ("google.genai.types", "FunctionResponseBlob"),
+    ("google.genai.types", "FunctionResponseFileData"),
+    ("google.genai.types", "FunctionResponsePart"),
+    ("google.genai.types", "Part"),
+    ("google.genai.types", "PartMediaResolution"),
+    ("google.genai.types", "VideoMetadata"),
 }
 
 
@@ -64,7 +106,7 @@ class _RestrictedUnpickler(pickle.Unpickler):
   `EventActions`.
   """
 
-  def find_class(self, module: str, name: str):  # noqa: ANN001
+  def find_class(self, module: str, name: str) -> Any:  # noqa: ANN001
     if (module, name) in _ALLOWED_PICKLE_GLOBALS:
       return super().find_class(module, name)
     raise pickle.UnpicklingError(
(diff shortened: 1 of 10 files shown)


rohityan:
Hi @White-Mouse , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing mypy-diff tests before we can proceed with the review.

DeanChensj:
Hi @White-Mouse, thanks for the update! I appreciate the effort to move towards a "Secure by Default" approach using a restricted unpickler.

To ensure this doesn't negatively impact existing users, I'd like to suggest the following improvements to the PR:

1. Prevent Data Loss by Expanding the Allowlist: 
The current restricted unpickler only allows EventActions. However, EventActions frequently contains nested ADK core types such as AuthConfig , ToolConfirmation, EventCompaction. Since EventCompaction also relies on google.genai.types.Content, we should include these classes (and their dependencies) in the allowlist. This ensures standard sessions migrate fully without falling back to an empty EventActions().

2. Add an Optional "Trust" Toggle:
 Since the state_delta field can store arbitrary Python objects (Any), some users may have custom classes in their legacy data. We should provide an optional configuration (e.g., allow_unsafe_unpickling: bool = False) that allows users to explicitly opt-in to the original unpickling behavior if they trust their source database.

With these two changes, we can achieve strong security for the majority of users while maintaining full functionality and flexibility. What do you think?

White-Mouse:
Thanks for the review. I updated the PR to address both points: the restricted allowlist now covers standard nested ADK action types (`AuthConfig`, `ToolConfirmation`, `EventCompaction`) plus the `google.genai.types.Content` / `Part` dependency classes used by compaction payloads, and migration now has an explicit `allow_unsafe_unpickling=False` trust toggle for users who need to migrate custom legacy objects from a trusted source database.

I also fixed the `mypy-diff` blocker by adding the missing return annotation and ran the updated migration tests, targeted mypy, pre-commit, and `git diff --check` locally. The new `pull_request` workflow runs for mypy/unit/pre-commit/check-file are currently in GitHub's `action_required` state for this fork PR, so they need maintainer approval before they will execute on the new commit.

arian-gogani:
the unpickling restriction is a good security fix. the complementary control: a signed receipt for every action deserialized from the v0 schema.

when migrating v0 actions, each deserialized action should generate a receipt proving what was loaded, from which session, at what time. if a malicious pickle payload was injected, the receipt chain shows exactly what was deserialized.

github.com/arian-gogani/nobulex

DeanChensj:
@gemini-cli /review

DeanChensj (dismissed):

rohityan:
Hi @White-Mouse , could you please fix the mypy-diff tests.

White-Mouse:
Thanks for the review. I pushed `9494da4` with the follow-ups:

- added the hyphenated CLI alias `--allow-unsafe-unpickling`, while keeping the existing underscore form
- kept the Click typing workaround scoped to the newly added option
- allowed standard `datetime.datetime`, `datetime.timezone`, and `datetime.timedelta` values in restricted action payloads
- tightened JSON handling so only object payloads are decoded into model fields
- added coverage for the CLI aliases, datetime state deltas, and non-object JSON fields

Local verification passed:

- `uv run pytest tests/unittests/sessions/migration/test_migration.py -q` -> 24 passed
- `uv run pytest tests/unittests/cli/utils/test_cli_tools_click.py::test_cli_migrate_session_allows_unsafe_unpickling_flag -q` -> 2 passed
- `uv run mypy src/google/adk/sessions/migration/migrate_from_sqlalchemy_pickle.py src/google/adk/sessions/migration/migration_runner.py` -> success
- `uv run pre-commit run --files ...` -> passed
- local merge-commit mypy-diff simulation against current `upstream/main` -> 0 new errors

The `pull_request_target` scan and CLA/header checks are green. The fork `pull_request` workflows are currently waiting for approval before they can run on the new commit.

White-Mouse:
Could you please approve the pending `pull_request` workflow runs (mypy-diff, unit tests, pre-commit) when you get a chance? They need maintainer approval since this PR is from a fork. Happy to address any further feedback — thanks!

White-Mouse:
Thanks for the follow-up. I pushed `8bdf463` to bring this PR branch up to date with current `main`.

That also clears the previous mypy-diff blocker locally. The new Python 3.13 mypy errors were coming from code that has since been fixed on upstream `main`, including the `_prepare_tables -> None` annotation and the `_is_tool_selected(... Optional[ReadonlyContext])` signature update.

I re-ran the relevant checks locally on Python 3.13:

- `uv run --python 3.13 pytest -q tests/unittests/sessions/migration/test_migration.py tests/unittests/cli/utils/test_cli_tools_click.py` -> 59 passed
- `uv run --python 3.13 pre-commit run --from-ref upstream/main --to-ref HEAD` -> passed
- `git diff --check upstream/main..HEAD` -> passed
- local mypy-diff simulation against `upstream/main` -> `main_count=2570`, `pr_count=2557`, `new_count=0`

The updated fork `pull_request` workflow runs are now waiting for maintainer approval (`action_required`) before they can run on this new head. Thanks!

src/google/adk/sessions/migration/migrate_from_sqlalchemy_pickle.py
```diff
@@ -37,6 +38,92 @@
...
+    ("google.genai.types", "FileData"),
+    ("google.genai.types", "FunctionCall"),
+    ("google.genai.types", "FunctionResponse"),
+    ("google.genai.types", "FunctionResponseBlob"),
+    ("google.genai.types", "FunctionResponseFileData"),
+    ("google.genai.types", "FunctionResponsePart"),
+    ("google.genai.types", "Part"),
+    ("google.genai.types", "PartMediaResolution"),
+    ("google.genai.types", "VideoMetadata"),
+}
```
DeanChensj, line 100: Missing Allowlist Type: 

The class EventActions has a field render_ui_widgets which contains a list of UiWidget instances. However, google.adk.events.ui_widget.UiWidget is not registered in the _ALLOWED_PICKLE_GLOBALS allowlist. This means any legacy database events containing UI widgets will fail to unpickle by default and fallback to empty actions, unless the unsafe unpickling flag is set.

White-Mouse: Thanks for catching this. I pushed `8f0e973` to add `google.adk.events.ui_widget.UiWidget` to the restricted migration allowlist, with a regression test covering `EventActions.render_ui_widgets` under the restricted unpickler.

I also re-ran the local checks after the update:

- `uv run --python 3.13 pytest -q tests/unittests/sessions/migration/test_migration.py tests/unittests/cli/utils/test_cli_tools_click.py` -> 60 passed
- `uv run --python 3.13 pytest -q tests/unittests/flows/llm_flows/test_base_llm_flow.py tests/unittests/models/test_gemini_llm_connection.py` -> 78 passed
- `uv run --python 3.13 mypy src/google/adk/sessions/migration/migrate_from_sqlalchemy_pickle.py src/google/adk/sessions/migration/migration_runner.py` -> passed
- `uv run --python 3.13 pre-commit run --all-files` -> passed
- `git diff --check` -> passed

The PR branch is updated; the new fork workflow runs may still need maintainer approval before GitHub runs them on the latest head. Thanks!

PR #5875, opened 2026-05-28
perf(flows): Resolve agent tool unions in parallel

wenzhaoy-google:
perf(flows): Resolve agent tool unions in parallel

`_process_agent_tools` previously walked `agent.tools` serially,
awaiting `_convert_tool_union_to_tools` for each entry in turn. For
toolsets this calls `get_tools_with_prefix`, which can perform I/O
(e.g. MCP `list_tools` over the network), so the total wall-clock cost
stacks linearly with the number of toolsets.

Dispatch every `_convert_tool_union_to_tools` call concurrently via
`asyncio.gather` so those listings overlap. The subsequent
`process_llm_request` calls are intentionally kept serial and in the
original `agent.tools` order: several built-in tools read or write
`llm_request` state (e.g. `GoogleSearchTool` writes
`llm_request.model`; `ComputerUseToolset` performs an idempotency
check on `llm_request.config.tools`) and rely on observing the
post-state of earlier tools.

Add two unit tests pinning down the new contract: one asserts
concurrent dispatch via a shared `asyncio.Event` gate (would hang if
still serial); the other asserts `process_llm_request` invocation
order still matches `agent.tools` order even when later unions
resolve first.

### DIFF
--- a/src/google/adk/flows/llm_flows/base_llm_flow.py
+++ b/src/google/adk/flows/llm_flows/base_llm_flow.py
@@ -423,6 +423,15 @@ async def _process_agent_tools(
   instances, and calls ``process_llm_request`` on each to register
   tool declarations in the request.
 
+  Tool-union resolution is dispatched concurrently via ``asyncio.gather``
+  to overlap I/O-bound listings (e.g. MCP ``list_tools`` over the
+  network). The subsequent ``process_llm_request`` calls are kept
+  serial in the original ``agent.tools`` order: some tools read/write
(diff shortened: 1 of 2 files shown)


DeanChensj:
Could you fix the mypy failure?

DeanChensj (approved):

DeanChensj:
merged in https://github.com/google/adk-python/commit/ae95a972280cbf0fcd6a989e19590cad68f3d847

PR #5885, opened 2026-05-28
fix(tools): add skill script dir to sys.path

### LINKED ISSUE #5882: Add `scripts/` directory to `sys.path` when executing Python skill scripts
# Feature Request: Add `scripts/` directory to `sys.path` when executing Python skill scripts

When executing Python scripts from a skill's `scripts/` directory via `run_skill_script`, the script's own directory (`scripts/`) is not added to `sys.path`. This means that local Python packages or modules placed alongside the script inside `scripts/` cannot be imported using standard `import` statements. For example, if a skill organizes helper code as:

```
scripts/
├── run.py
└── mypackage/
    ├── __init__.py
    └── helper.py
```

Then `from mypackage import helper` inside `run.py` fails with `ModuleNotFoundError`, because the working directory is set to the skill root (`td/`), not `scripts/`, and `scripts/` is not on `sys.path`.

In `_SkillScriptCodeExecutor._build_wrapper_code`, when generating the wrapper code for `.py` scripts, automatically insert the script's directory into `sys.path` before executing via `runpy.run_path`. Concretely, add the following line to the generated boilerplate just before the `runpy.run_path(...)` call:

```python
sys.path.insert(0, os.path.dirname(os.path.abspath(file_path)))
```

This mirrors the behavior of running `python scripts/run.py` directly from the command line, where Python automatically adds the script's directory as `sys.path[0]`.

### Impact on your work

Skills that encapsulate reusable logic in local sub-modules (e.g. helper libraries, data models, utility functions) cannot be structured as proper Python packages today. Every skill script must either be monolithic or resort to brittle `sys.path` manipulation at the top of each script. This limits the maintainability and reusability of skill scripts for non-trivial use cases.

### Willingness to contribute

Yes — happy to submit a PR if the team agrees on the approach.

---

**Workaround in every script**: Add `sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))` at the top of each skill script. This works but requires every skill author to remember this boilerplate — it is error-prone and not discoverable.

**Change CWD to `scripts/`**: Setting the working directory to `scripts/` instead of the skill root would solve the import problem but would break relative file access to `references/` and `assets/`, which is the current design intent.

In `_build_wrapper_code` (file `skill_toolset.py`, class `_SkillScriptCodeExecutor`), change the Python branch from:

```python
code_lines.extend([
    f"      sys.argv = {argv_list!r}",
    "      try:",
    f"        runpy.run_path({file_path!r}, run_name='__main__')",
    ...
])
```

to:

```python
code_lines.extend([
    f"      sys.argv = {argv_list!r}",
    f"      sys.path.insert(0, os.path.dirname(os.path.abspath({file_path!r})))",
    "      try:",
    f"        runpy.run_path({file_path!r}, run_name='__main__')",
    ...
])
```

No API changes required. The change is purely internal to the wrapper code generator.

### Additional Context

This behavior is consistent with how the standard Python interpreter handles script execution (`python scripts/run.py` adds `scripts/` to `sys.path[0]`). The current behavior using `runpy.run_path` does not replicate this automatically, creating a subtle and hard-to-diagnose discrepancy for skill authors who test their scripts locally before packaging them into a skill.

he-yufeng:
﻿## Summary
- add the executed Python skill script directory to `sys.path` before `runpy.run_path`
- preserve the existing skill-root working directory, so references/assets still resolve from the skill root
- add a real executor regression test for importing a sibling helper module from `scripts/`

Fixes #5882.

## To verify
- `git diff --check`
- `python -m py_compile src/google/adk/tools/skill_toolset.py tests/unittests/tools/test_skill_toolset.py`
- `python -m ruff check src/google/adk/tools/skill_toolset.py tests/unittests/tools/test_skill_toolset.py`
- `python -m pyink --check src/google/adk/tools/skill_toolset.py tests/unittests/tools/test_skill_toolset.py`
- `PYTHONPATH=src python -m pytest tests/unittests/tools/test_skill_toolset.py::test_execute_script_python_success tests/unittests/tools/test_skill_toolset.py::test_integration_python_imports_sibling_script_module tests/unittests/tools/test_skill_toolset.py::test_integration_python_stdout tests/unittests/tools/test_skill_toolset.py::test_integration_python_sys_exit_zero -q`
- `PYTHONPATH=src python -m pytest tests/unittests/tools/test_skill_toolset.py -k "not shell" -q`

I also ran the full `tests/unittests/tools/test_skill_toolset.py` file on Windows. The Python tests passed, but three existing shell integration tests fail because this Windows environment does not have `bash` on PATH.

### DIFF
--- a/src/google/adk/tools/skill_toolset.py
+++ b/src/google/adk/tools/skill_toolset.py
@@ -57,6 +57,7 @@ _BINARY_FILE_DETECTED_MSG = (
     " conversation history for you to analyze."
 )
 
+
 def _build_skill_system_instruction(prefix: str | None = None) -> str:
   p = f"{prefix}_" if prefix else ""
 
@@ -75,7 +76,7 @@ def _build_skill_system_instruction(prefix: str | None = None) -> str:
       "bash.\n\n"
       "This is very important:\n\n"
       f"1. If a skill seems relevant to the current user query, you MUST use "
-      f"the `{p}load_skill` tool with `skill_name=\"<SKILL_NAME>\"` to read "
+      f'the `{p}load_skill` tool with `skill_name="<SKILL_NAME>"` to read '
       "its full instructions before proceeding.\n"
       "2. Once you have read the instructions, follow them exactly as "
       "documented before replying to the user. For example, If the "
@@ -85,7 +86,8 @@ def _build_skill_system_instruction(prefix: str | None = None) -> str:
       "skill's directory (e.g., `references/*`, `assets/*`, `scripts/*`). "
       "Do NOT use other tools to access these files.\n"
       f"4. Use `{p}run_skill_script` to run scripts from a skill's `scripts/` "
-      f"directory. Use `{p}load_skill_resource` to view script content first if "
+      f"directory. Use `{p}load_skill_resource` to view script content"
+      " first if "
       "needed.\n"
   )
 
@@ -666,6 +668,10 @@ class _SkillScriptCodeExecutor:
 
       code_lines.extend([
           f"      sys.argv = {argv_list!r}",
+          (
+              "      sys.path.insert(0,"
+              f" os.path.dirname(os.path.abspath({file_path!r})))"
+          ),
           "      try:",
           f"        runpy.run_path({file_path!r}, run_name='__main__')",
           "      except SystemExit as e:",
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @he-yufeng , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors by running autoformat.sh

boyangsvl:
I have merged in commit https://github.com/google/adk-python/commit/9296198b21a80b211b75f0e9263f618fbe9e8744

Issue #5008, opened 2026-03-26
[LiteLlm] JSONDecodeError on malformed tool_call.function.arguments crashes run before callbacks can recover


Huzefa-Lw:
## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. Requests without complete information may be rejected / deprioritized.
If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**
When ADK parses tool calls from a LiteLLM response, it uses strict JSON parsing in `google/adk/models/lite_llm.py` (`json.loads(tool_call.function.arguments or "{}")`).
If the model returns malformed or truncated JSON in `tool_call.function.arguments`, ADK throws `JSONDecodeError` and fails the invocation.

**Steps to Reproduce:**
1. Install `google-adk==1.21.0` and `litellm`.
2. Run an ADK flow with LiteLLM where the model emits a function call.
3. Ensure `tool_call.function.arguments` is malformed JSON (for example, an unterminated string).
4. Observe crash/stacktrace at `google/adk/models/lite_llm.py` in `_message_to_generate_content_response`.

**Expected Behavior:**
ADK should gracefully handle malformed tool argument JSON from the model (e.g., recoverable model error path and retry/fallback) instead of hard-failing the run.

**Observed Behavior:**
Invocation/workflow fails with `JSONDecodeError` while parsing tool call arguments, e.g.:
`JSONDecodeError: Unterminated string starting at: line 1 column 2 (char 1)`

**Environment Details:**
 - ADK Library Version (pip show google-adk): `1.21.0`
 - Desktop OS:** `macOS (Darwin 25.2.0)`
 - Python Version (python -V): `Python 3.12.11`

**Model Information:**

- Are you using LiteLLM: Yes
 - Which model is being used: OpenAI-compatible model via LiteLLM (configured via app settings as `vat_llm_model`)

---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

**Regression:**
Observed before and after upgrade; still reproducible on `1.21.0`.

**Logs:**
```text
File ".../google/adk/models/lite_llm.py", line 1057, in _message_to_generate_content_response
    args=json.loads(tool_call.function.arguments or "{}"),
File "/usr/local/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
File "/usr/local/lib/python3.12/json/decoder.py", line 354, in raw_decode
    obj, end = self.scan_once(s, idx)
[...]

PR #5897, opened 2026-05-29
fix(lite_llm): tolerate malformed JSON in tool call arguments

### LINKED ISSUE #5896: [LiteLlm] JSONDecodeError on malformed tool_call.function.arguments still crashes invocation at v2.0.0 (regression of #5008)
## 🔴 Required Information

**Describe the Bug:**

When the upstream model returns malformed JSON in `tool_call.function.arguments` (truncated strings, unclosed objects, partial streams committed to the message), `google.adk.models.lite_llm._message_to_generate_content_response` calls `json.loads(tool_call.function.arguments or "{}")` without a guard and raises `json.JSONDecodeError`. The exception propagates out of LiteLLM response parsing — upstream of any tool or callback boundary — so the entire invocation aborts and the user sees a generic streaming-error message rather than a recoverable tool-call retry.

This is the same bug as the previously closed [#5008](https://github.com/google/adk-python/issues/5008). That issue was closed after the original reporter said the upgrade had resolved it for them; a subsequent commenter (@NerdSmith) noted it still reproduced on `1.32.0`, but the issue was not reopened. The crash site is unchanged on `main` at `v2.0.0` (`src/google/adk/models/lite_llm.py` around `_message_to_generate_content_response`, the `args=json.loads(tool_call.function.arguments or "{}")` line).

**Steps to Reproduce:**

1. Install `google-adk==2.0.0` (or build from `main`) and `litellm`.
2. Run any ADK flow that uses a LiteLLM-wrapped model and emits a function call.
3. Cause the model to emit malformed JSON in `tool_call.function.arguments` — easiest in a test by constructing the message directly (see *Minimal Reproduction Code* below).
4. Observe `JSONDecodeError` propagate out of `_message_to_generate_content_response`.

**Expected Behavior:**

ADK should treat malformed tool-argument JSON as a recoverable model error. Log a warning with the raw arguments and pass an empty dict (or equivalent) to the function-call `Part` so the downstream tool dispatch surfaces a structured `bad_arguments`-style error to the model. The agent loop survives and the model can retry with corrected JSON.

**Observed Behavior:**

```text
File ".../google/adk/models/lite_llm.py", line 1730, in _message_to_generate_content_response
    args=json.loads(tool_call.function.arguments or "{}"),
File "/usr/local/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
File "/usr/local/lib/python3.12/json/decoder.py", line 354, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 2 (char 1)
```

**Environment Details:**

- ADK Library Version: `2.0.0` (main, commit `aa515125`)
- Desktop OS: macOS 25.3.0 (also reproduced on Linux in Cloud Run staging)
- Python Version: `3.11.15` (also reproduced under `3.12`)

**Model Information:**

- Are you using LiteLLM: Yes
- Which model is being used: Anthropic Claude via LiteLLM (also reproduced from OpenAI-compatible providers per #5008)

---

## 🟡 Optional Information

**Regression:**

Filed as still-broken at `1.32.0` in the trailing comment on #5008, and confirmed still present at `2.0.0`.

**Logs:**

See *Observed Behavior* above.

**Additional Context:**

The failure happens during LiteLLM response → ADK `LlmResponse` conversion, which sits upstream of `before_tool_callback` / `after_tool_callback` and the tool dispatch layer. So no callback-level mitigation is possible — the entire invocation aborts. From a production-operator perspective this is one of the harder ADK crashes to recover from: the trace lands in framework code, not user code, and the user-facing error is a generic stream failure with no information about the malformed tool call.

```python
from litellm import ChatCompletionAssistantMessage, ChatCompletionMessageToolCall
from litellm.types.utils import Function

    role="assistant",
    content=None,
    tool_calls=[
            type="function",
            id="call_1",
            function=Function(name="demo_tool", arguments='{"city":"unterminated'),
        )
    ],
)

_message_to_generate_content_response(message)  # raises json.JSONDecodeError
```

- Intermittently (<50%) — model-dependent, hits more often on long/streamed tool args under load.

---

I have a fix + unit tests ready as a PR: https://github.com/Ja-Crispy/adk-python/tree/fix-litellm-malformed-tool-args-json — happy to open it against this issue once it's triaged.

Ja-Crispy:

- Closes: #5896
- Related: #5008 (previously closed but never actually fixed; same crash site)

**Problem:**

When the upstream model returns malformed JSON in `tool_call.function.arguments` (truncated strings, unclosed objects, partial streams committed to the message), `_message_to_generate_content_response` in `google/adk/models/lite_llm.py` calls `json.loads(tool_call.function.arguments or "{}")` without a guard. The resulting `json.JSONDecodeError` propagates out of LiteLLM response parsing — upstream of any tool or callback boundary — so the entire invocation aborts and the user sees a generic streaming-error message rather than a recoverable tool-call retry.

**Solution:**

Wrap the `json.loads` call in a `try` / `except json.JSONDecodeError`. On failure, log a warning that includes the tool name, the tool-call id, and the raw arguments (so operators can debug what the model emitted), and pass an empty dict as `args`. The function-call `Part` is still constructed with the original tool name and id, so the downstream tool dispatch fires with no arguments — the resulting `bad_arguments`-style error from the tool layer is a recoverable signal the model can read and retry from.

This matches the expected behaviour described in #5008 ("recoverable model error path and retry/fallback") and aligns with how ADK already tolerates other model-output edge cases.

### Testing Plan

**Unit Tests:**

Added `tests/unittests/models/test_lite_llm_malformed_tool_args.py` with four test functions covering both the new tolerant path and the existing happy path:

- `test_unterminated_string_does_not_raise` — verifies the exact stack trace from #5008 (`{"city":"unterminated`) no longer raises, the resulting `Part.function_call.args == {}`, and a warning is logged with the raw arguments.
- `test_unclosed_object_does_not_raise` — additional malformed-JSON shape (`{"a": 1, "b":`) for breadth.
- `test_well_formed_arguments_still_parse` — regression that valid JSON still produces the parsed dict (`{"city": "Paris", "units": "metric"}`).
- `test_empty_arguments_become_empty_dict` — regression that the existing `or "{}"` fallback for empty strings still works.

Local run:

```
$ pytest tests/unittests/models/test_lite_llm_malformed_tool_args.py -v
tests/unittests/models/test_lite_llm_malformed_tool_args.py::test_unterminated_string_does_not_raise PASSED
tests/unittests/models/test_lite_llm_malformed_tool_args.py::test_unclosed_object_does_not_raise PASSED
tests/unittests/models/test_lite_llm_malformed_tool_args.py::test_well_formed_arguments_still_parse PASSED
tests/unittests/models/test_lite_llm_malformed_tool_args.py::test_empty_arguments_become_empty_dict PASSED
4 passed in 1.50s
```

Full `models/` suite still green to confirm no regression:

```
698 passed, 1 skipped, 10 warnings in 7.94s
```

In our deployment (which uses this fix via a local patch), the same prompt that previously crashed the invocation now produces a tool-response cycle the model recovers from — the user sees the corrected tool result instead of the generic streaming error. Happy to provide additional repro if helpful.

### Checklist

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -1725,9 +1725,21 @@ def _message_to_generate_content_response(
     for tool_call in tool_calls:
       if tool_call.type == "function":
         thought_signature = _extract_thought_signature_from_tool_call(tool_call)
+        try:
+          parsed_args = json.loads(tool_call.function.arguments or "{}")
+        except json.JSONDecodeError:
+          logger.warning(
+              "Malformed JSON in tool call arguments for %s (id=%s);"
+              " passing empty args so the tool dispatch can surface a"
+              " recoverable error to the model. Raw arguments: %r",
+              tool_call.function.name,
+              tool_call.id,
+              tool_call.function.arguments,
+          )
+          parsed_args = {}
         part = types.Part.from_function_call(
             name=tool_call.function.name,
-            args=json.loads(tool_call.function.arguments or "{}"),
+            args=parsed_args,
         )
         part.function_call.id = tool_call.id
         if thought_signature:
--- a/tests/unittests/models/test_lite_llm_malformed_tool_args.py
+++ b/tests/unittests/models/test_lite_llm_malformed_tool_args.py
@@ -1,0 +1,94 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for tolerant tool-argument JSON parsing in lite_llm.
+
+When a LiteLLM-wrapped model emits malformed JSON in
+``tool_call.function.arguments`` (truncated strings, unclosed objects, …),
+``_message_to_generate_content_response`` previously raised
+``JSONDecodeError`` and aborted the invocation before tool callbacks
+could observe or recover from the failure. The current behaviour is to
+log a warning, pass an empty ``args`` dict to the function-call Part,
+and let downstream tool dispatch surface a recoverable error to the
+model.
+"""
+
+import logging
+
+from google.adk.models.lite_llm import _message_to_generate_content_response
+from litellm import ChatCompletionAssistantMessage
+from litellm import ChatCompletionMessageToolCall
+from litellm.types.utils import Function
+
+_LOGGER_NAME = "google_adk.google.adk.models.lite_llm"
+
+
+def _assistant_message_with_tool_args(
+    arguments: str,
+) -> ChatCompletionAssistantMessage:
+  return ChatCompletionAssistantMessage(
+      role="assistant",
+      content=None,
+      tool_calls=[
+          ChatCompletionMessageToolCall(
+              type="function",
+              id="call_1",
+              function=Function(name="demo_tool", arguments=arguments),
+          )
+      ],
+  )
+
+
+def test_unterminated_string_does_not_raise(caplog):
+  message = _assistant_message_with_tool_args('{"city":"unterminated')
+
+  with caplog.at_level(logging.WARNING, logger=_LOGGER_NAME):
+    response = _message_to_generate_content_response(message)
+
+  part = response.content.parts[0]
+  assert part.function_call.name == "demo_tool"
+  assert part.function_call.id == "call_1"
+  assert part.function_call.args == {}
+  assert any("Malformed JSON" in record.message for record in caplog.records)
+
+
+def test_unclosed_object_does_not_raise(caplog):
+  message = _assistant_message_with_tool_args('{"a": 1, "b":')
+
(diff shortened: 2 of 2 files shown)


Ja-Crispy:
@adk-bot I signed it!

rohityan:
Hi @Ja-Crispy , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

Ja-Crispy:
@jazhang00  @wyf7107 just a quick check on the status, no rush ^_^ anything needed from my end?

jazhang00:
Hi @Ja-Crispy, thank you bringing up this issue and creating a PR for it. I think one concern with passing an empty dict as `args` when the JSON is malformed is the possibility of unexpectedly calling a tool that has only optional arguments.

Ja-Crispy:
Thanks @jazhang00 — fair concern, and I agree: silently passing `{}` for malformed JSON risks invoking a tool whose parameters are all optional.

This has since been handled upstream by `_parse_tool_call_arguments` (`src/google/adk/models/lite_llm.py`), which now occupies this call site and is a more robust take than my one-liner: it keeps strict JSON as the primary path, repairs complete object-literal shapes via `ast.literal_eval` and unquoted-key quoting, and — crucially — **re-raises the original `JSONDecodeError` when the payload is genuinely unrepairable** rather than falling back to empty args. That directly resolves the optional-args concern (a broken call errors instead of running with defaults).

So this PR is superseded — closing it. Appreciate the review!

Issue #5824, opened 2026-05-23
Security: AgentEngineSandboxCodeExecutor sends 'contents' instead of 'content', silently creating empty files


Thapza888:
Moved to Google VRP private disclosure.

Thapza888:
Closing this issue — details moved to Google VRP private disclosure per CVD rules. Apologies for the initial public posting.

PR #5912, opened 2026-05-30
fix: use correct 'content' key in AgentEngineSandboxCodeExecutor input files

Thapza888:
Fixes the same root cause as #5505 which was missed in `agent_engine_sandbox_code_executor.py`.

## Root Cause

`AgentEngineSandboxCodeExecutor` builds input file payloads with key `'contents'` (plural), but the Vertex AI SDK (`sandboxes.py`) reads `'content'` (singular). This causes `file.get("content", b"")` to always return empty bytes -- all input files are silently created as zero bytes in the sandbox.

The executor returns SUCCESS, so the failure is completely silent.

## What changed

`'contents'` to `'content'` at line 183 of `agent_engine_sandbox_code_executor.py`.

## Impact

Security policies, access-control templates, verification scripts, and boundary configuration files passed via `input_files` to the sandbox code executor are silently created as empty (zero-byte) resources. The sandbox executes untrusted code without the intended operational constraints. No error is raised -- the SUCCESS status blinds monitoring and logging.

## Testing Plan

Added regression test `test_execute_code_input_files_content_key` that verifies `execute_code` is called with `input_data` containing the `'content'` key (singular). The mock assertion explicitly checks:

```python
expected_input_data = {
    "code": 'print("hello world")',
    "files": [
        {
            "name": "test.txt",
            "content": b"hello world",
            "mimeType": "text/plain",
        }
    ],
}
mock_api_client.agent_engines.sandboxes.execute_code.assert_called_once_with(
    name=...,
    input_data=expected_input_data,
)
```

Before the fix, this test fails because the code sends `'contents'` instead of `'content'`. After the fix, the test passes (will be verified in CI).

## References

- PR #5505 -- fixed same bug in other files, missed this one
- Issue #5500 -- original bug report
- Issue #5824 -- security report (Google OSS VRP)


rohityan:
Hi @Thapza888 , Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

wukath:
closing, duplicate of https://github.com/google/adk-python/pull/5958/changes (we will merge that one)

Issue #2207, opened 2025-07-28
Down the line `SequentialAgent` has access to all previous conversations


jorism1993:
**Describe the bug**
Down the line `SequentialAgent` has access to all previous conversations through `llm_request.contents`, instead of running as a standalone agent. Even when setting `include_contents='none'`, it still has access to the last `llm_request.contents` part, leading to context rot.

**To Reproduce**
```python
from typing import Optional

from google.adk.agents import LlmAgent, SequentialAgent
from google.adk.runners import Runner
from google.adk.sessions import InMemorySessionService
from google.adk.agents.callback_context import CallbackContext
from google.genai.types import Content, Part
from google.adk.models import LlmResponse, LlmRequest

import asyncio

GEMINI_MODEL = "gemini-2.5-flash"

# Constants for session management
APP_NAME = 'code_pipeline_app'
USER_ID = 'test_user'
SESSION_ID = 'code_pipeline_session'


# --- Define the Callback Function ---
def set_last_user_message_callback(
        callback_context: CallbackContext,
        llm_request: LlmRequest
) -> Optional[LlmResponse]:
    """Inspects/modifies the LLM request or skips the call."""

    # last_user_message = ""
    last_message = llm_request.contents[-1].parts[0]

    import pdb
    pdb.set_trace()

    return None


# --- 1. Define Sub-Agents for Each Pipeline Stage ---

# Code Writer Agent
code_writer_agent = LlmAgent(
    name="CodeWriterAgent",
    model=GEMINI_MODEL,
    instruction="""You are a Python Code Generator.
Based *only* on the user's request, write Python code that fulfills the requirement.
Output *only* the complete Python code block, enclosed in <code> tags (<code>python ... </code>). 
Do not add any other text before or after the code block.
""",
    description="Writes initial Python code based on a specification.",
    output_key="generated_code"
)

# Code Reviewer Agent
code_reviewer_agent = LlmAgent(
    name="CodeReviewerAgent",
    model=GEMINI_MODEL,
    instruction="""You are an expert Python Code Reviewer. 
    Your task is to provide constructive feedback on the provided code.

    **Code to Review:**
    <code>
    {generated_code}
    </code>

**Review Criteria:**
1.  **Correctness:** Does the code work as intended? Are there logic errors?
2.  **Readability:** Is the code clear and easy to understand? Follows PEP 8 style guidelines?
3.  **Efficiency:** Is the code reasonably efficient? Any obvious performance bottlenecks?
[...]

PR #5925, opened 2026-06-01
feat(agents): add include_sources for per-agent content source filtering

bofenghuang:

- Related: google/adk-python#2207

**Problem:**

In multi-agent pipelines, every peer agent's output is narrative-cast into a `role='user'` content entry (`"For context: [agent_name] said: ..."`). The only existing control is `include_contents` (`'default'`/`'none'`), which is a session-history scope control — it determines how much of the conversation history is visible. There is no agent-level source control: no way to say which agents within that history a given agent should see.

Issue #2207 documents this problem. The `before_model_callback` workaround proposed there has several limitations that make it a poor fit for production use:

1. **Text parsing is fragile** — it regex-matches `[agent_name] said:` against the internal narrative format. If ADK changes that format, the workaround silently breaks with no warning.
2. **No per-source selectivity** — it drops *all* peer entries. You cannot express "keep agent_a's output but drop agent_b's".
3. **Must be duplicated** — the callback must be attached individually to every agent that needs isolation, with no composable way to express the policy.
4. **FC/FR pairs not considered** — the regex can accidentally strip parts of function call/response content, and it does not maintain the function call/response pairing invariant required by the model API.
5. **Observability misalignment** — `before_model_callback` on the agent runs *after* plugin callbacks (see `base_llm_flow.py:_handle_before_model_callback`). LLM observability tools (e.g. Braintrust) hook in as plugins and therefore capture `llm_request.contents` *before* the workaround strips the peer entries. What observability reports and what the model actually receives are different — making traces unreliable for debugging.

**Solution:**

Add `include_sources: list[str] | None` to `LlmAgent` — a declarative per-agent source control that answers "from which agents?", orthogonal to the existing session-history scope control (`include_contents`).

```python
# Full history, only user + this agent's own turns
LlmAgent(include_sources=['user', 'self'])

# Full history, user + one specific upstream agent
LlmAgent(include_sources=['user', 'summarizer_agent'])

# User messages only — stateless classifier
LlmAgent(include_sources=['user'])
```

Reserved names: `'user'` (plain human messages), `'self'` (this agent's own prior model turns), any other string is matched directly against `event.author`.

The filter runs inside `_get_contents()` at the event level, **before** `_present_other_agent_message()` converts authorship into embedded text. Source identity is read from `event.author` metadata — no text parsing, no format coupling. Since filtering happens at the request-building stage (before any callbacks), observability plugins always see the same contents the model receives. Function call/response pairs are preserved: FC responses for the current agent's calls are tied to `'self'` (dropped together with their calls), and another agent's FC responses are dropped when that agent's FC call is also filtered. Live-mode sessions are handled by mapping `event.author == agent_name` to the `'self'` reserved name.

`include_sources=[]` raises `ValueError` at construction time (use `None` to disable filtering).

### Testing Plan

**Unit Tests:**

New file `tests/unittests/flows/llm_flows/test_contents_source_filter.py` — 20 unit tests directly on `_get_contents()` and `_get_current_turn_contents()` covering: no-op (`None`), each reserved name, combinations, named agent filtering, FC/FR pair preservation, FC/FR co-dropping, other-agent FC response attribution, live-mode self-mapping, and `_get_current_turn_contents` propagation.

Extended `tests/unittests/agents/test_llm_agent_include_contents.py` with 4 integration tests: empty-list validation, `None` default, user-only isolation in a sequential pipeline (the exact case from #2207), and multi-turn composition.

```
tests/unittests/flows/llm_flows/test_contents_source_filter.py  20 passed
tests/unittests/agents/test_llm_agent_include_contents.py        7 passed (3 pre-existing + 4 new)
tests/unittests/flows/llm_flows/test_contents*.py               all passed
tests/unittests/agents/test_llm_agent*.py                       125 passed, 2 xfailed (pre-existing VAIS failures, unrelated)
```

Verified with a sequential pipeline (upstream → downstream) using MockModel:

```
=== WITHOUT include_sources  (current behaviour) ===
  [user] 'Hello from user'
  [user] "[Part(text='For context:'), Part(text='[upstream] said: Upstream result XYZ')]"

=== WITH include_sources=['user'] ===
  [user] 'Hello from user'

=== WITH include_sources=['user', 'self'] ===
  [user] 'Hello from user'
```

The exact pipeline from issue #2207 can be replaced with:
```python
code_refactorer_agent = LlmAgent(
    name="CodeRefactorerAgent",
    include_sources=['user', 'self'],  # replaces before_model_callback workaround
    ...
)
```

### Checklist

### Additional context

Two known edge cases deferred to follow-up:
- **Compaction events** (`author='model'`): treated as other-agent with name `'model'`; dropped by `['user', 'self']` if compaction is enabled.
- **Mixed FC response events**: a single `'user'`-authored event containing FC responses to calls from multiple agents is classified as other-reply based on the first other-agent response found.

/cc @Jacksunwei @rohityan

### DIFF
--- a/src/google/adk/agents/llm_agent.py
+++ b/src/google/adk/agents/llm_agent.py
@@ -978,6 +978,10 @@ class LlmAgent(BaseAgent, abc.ABC):
 
   @model_validator(mode='after')
   def __model_validator_after(self) -> LlmAgent:
+    if self.include_sources is not None and len(self.include_sources) == 0:
+      raise ValueError(
+          'include_sources=[] keeps nothing. Use None to disable filtering.'
+      )
     if self.include_contents == 'none' and self.include_sources is not None:
       warnings.warn(
           "include_contents='none' with include_sources may produce empty"
@@ -989,17 +993,6 @@ class LlmAgent(BaseAgent, abc.ABC):
           stacklevel=2,
       )
     return self
-
-  @field_validator('include_sources', mode='after')  # type: ignore[misc]
-  @classmethod
-  def _validate_include_sources(
-      cls, v: Optional[list[str]]
-  ) -> Optional[list[str]]:
-    if v is not None and len(v) == 0:
-      raise ValueError(
-          'include_sources=[] keeps nothing. Use None to disable filtering.'
-      )
-    return v
 
   @field_validator('generate_content_config', mode='after')
   @classmethod
--- a/src/google/adk/agents/llm_agent.py
+++ b/src/google/adk/agents/llm_agent.py
@@ -352,6 +352,27 @@ class LlmAgent(BaseAgent, abc.ABC):
     instruction and input
   """
 
+  include_sources: Optional[list[str]] = None
+  """Allowlist of content sources to include in model requests.
+
+  Orthogonal to include_contents (temporal window); this controls which
+  sources are kept from within that window.
+
+  Options:
+    None (default): all sources pass through — backward-compatible.
+    list[str]: only content from the listed sources is kept.
+
+  Reserved source names:
+    'user'  — plain human user messages (not tool outputs)
+    'self'  — this agent's own prior model outputs
+    <name>  — any other string is matched against event.author (agent name)
+
+  Example — keep full history but only user + this agent's turns:
+    include_contents='default', include_sources=['user', 'self']
+
+  Raises ValueError if set to [] (use None to disable filtering).
+  """
+
   # Controlled input/output configurations - Start
   input_schema: Optional[type[BaseModel]] = None
   """The input schema when agent is used as a tool."""
@@ -956,6 +977,17 @@ class LlmAgent(BaseAgent, abc.ABC):
   @model_validator(mode='after')
   def __model_validator_after(self) -> LlmAgent:
     return self
+
+  @field_validator('include_sources', mode='after')
+  @classmethod
+  def _validate_include_sources(
+      cls, v: Optional[list[str]]
+  ) -> Optional[list[str]]:
+    if v is not None and len(v) == 0:
+      raise ValueError(
+          "include_sources=[] keeps nothing. Use None to disable filtering."
+      )
+    return v
 
   @field_validator('generate_content_config', mode='after')
   @classmethod
--- a/src/google/adk/agents/llm_agent.py
+++ b/src/google/adk/agents/llm_agent.py
@@ -343,13 +343,15 @@ class LlmAgent(BaseAgent, abc.ABC):
   """Disallows LLM-controlled transferring to the peer agents."""
   # LLM-based agent transfer configs - End
 
-  include_contents: Literal['default', 'none'] = 'default'
+  include_contents: Literal['default', 'current', 'none'] = 'default'
   """Controls content inclusion in model requests.
 
   Options:
-    default: Model receives relevant conversation history
-    none: Model receives no prior history, operates solely on current
-    instruction and input
+    default: Model receives full conversation history.
+    current: Model receives all events since the last user message,
+      including outputs from agents that ran earlier in this pipeline.
+    none: Model receives only the most recent agent or user input, with no
+      prior conversation history.
   """
 
   include_sources: Optional[list[str]] = None
@@ -976,6 +978,16 @@ class LlmAgent(BaseAgent, abc.ABC):
 
   @model_validator(mode='after')
   def __model_validator_after(self) -> LlmAgent:
+    if self.include_contents == 'none' and self.include_sources is not None:
+      warnings.warn(
+          "include_contents='none' with include_sources may produce empty"
+          ' context: the turn boundary is the last user OR other-agent event,'
+          ' and if that event is filtered by include_sources the context will'
+          " be empty. Use include_contents='current' to anchor at the last"
+          ' user message instead.',
+          UserWarning,
+          stacklevel=2,
+      )
     return self
 
   @field_validator('include_sources', mode='after')  # type: ignore[misc]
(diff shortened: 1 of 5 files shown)


bofenghuang:
Hi @rohityan 👋 ,
Thanks for taking this! Just fixed the linting.
~One follow-up could be renaming the `include_contents` arg from `Literal['default', 'none']` to `Literal['default', 'current']` (keep `'none'` as an alias for back-compat). Would give users clearer control: `include_contents` for turn-level, `include_sources ` for agent-level.
WDYT? Happy to add it in this PR or a separate one.~

bofenghuang:
Previously I misunderstood `include_contents`. IIUC now:

- `include_contents='default'` — full conversation history
- `include_contents='none'` — only the most recent event, whether a user message or an agent output (I guess keeping one user-or-agent entry is needed to satisfy the LLM chat template requirement)

In this follow-up I added a third value `include_contents='current'`, which only includes the user message and sibling agent outputs from the current invocation (current conversation turn).

This turn-level control can be combined with `include_sources`, which filters at the agent level.

WDYT? Thanks in advance!

rohityan:
Hi @bofenghuang , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing unit tests, mypy-diff tests and formatting errors before we can proceed with the review.

bofenghuang:
Hi @rohityan, thanks for taking time to look into this. The pr-analyze check seems to have failed on a quota error. Could you re-run it? Happy to clarify anything if needed.

DeanChensj:
Thank you for putting together this PR and for the thorough test suite and write-up!

After reviewing the API design and its interaction with the content pipeline (`_get_contents` / `_get_current_turn_contents`), we are going to close this PR rather than adding `include_sources` and `include_contents='current'` to `LlmAgent`. Here is the detailed context on the issues we found in this approach and how ADK 2.0 solves the underlying problem (`#2207`) at the architecture level:

### 1. Issues in this PR's design & implementation

1. **Coupling between `include_contents` (turn boundary anchoring) and `include_sources` (post-slice filtering)**:
   - Under `include_contents='none'`, `_get_current_turn_contents` walks backward and anchors at the latest user OR peer-agent event (`_is_other_agent_reply`) without accounting for `source_filter`. As a result, `include_contents='none'` + `include_sources=['user']` slices at the immediately preceding peer agent and then filters that event out, leaving `llm_request.contents = []`.
   - Working around that conflict required adding a third `include_contents='current'` option (`stop_at_user_only=True`) and emitting a runtime `UserWarning` in `LlmAgent.__model_validator_after` whenever `include_contents='none'` is used with `include_sources` (even for valid combinations), while `include_contents='current'` + `include_sources=['agent_a']` (without `'user'`) still suffers from the same boundary-vs-filter mismatch.
2. **Hardcoding sibling agent names on `LlmAgent`**:
   - Configuring `include_sources=['user', 'self', 'agent_a']` couples an agent's definition to the string names of sibling agents in a specific parent pipeline, making agents fragile to reuse or rename (and `include_sources` was also omitted from `LlmAgentConfig`).
3. **Implementation edge cases in `_get_contents`**:
   - **Cross-agent tool name collision (`fc_author_by_id`)**: Indexing `fc_author_by_id` by `function_call.name` via `fc_author_by_id.setdefault(part.function_call.name, e.author)` and resolving `resp_id = part.function_response.id or part.function_response.name` introduces a bug even when `include_sources=None`. If an upstream `agent_b` calls a tool (e.g., `search`) first, `fc_author_by_id['search']` is permanently bound to `'agent_b'`. When the current `agent_a` later calls a tool with the same name (`search`) without an ID, `agent_a`'s own `FunctionResponse` is misclassified as `is_other_reply=True` (`other_fc_author='agent_b'`)—dropping `agent_a`'s own tool response under `include_sources=['user', 'self']` or converting it into `For context:` text under `include_sources=None`.
   - **Role alternation violations**: Configurations like `include_sources=['self']` (without `'user'`) drop all `role='user'` events and produce `llm_request.contents` containing only consecutive `role='model'` turns with no `user` turn, which Gemini/Vertex/OpenAI APIs reject with `400 INVALID_ARGUMENT`.

---

### 2. How ADK 2.0 solves `#2207` cleanly (`Workflow` + `mode='single_turn'` / `isolation_scope`)

The root cause of `#2207` is that in legacy `SequentialAgent`, all sub-agents share an unscoped session context (`isolation_scope=None`), so downstream steps inherit `For context: [previous_agent] said: ...` even when inputs are already passed explicitly via `{state_key}` templates.

In ADK 2.0 (`google.adk.workflow`, which supersedes `SequentialAgent`):
- Every `LlmAgent` step in a workflow defaults to **`mode='single_turn'`** (`LlmAgent(mode='single_turn', ...)`), and each node run executes under its own **`isolation_scope`**.
- `_should_include_event_in_context` filters events by `isolation_scope` at the **`Event` level** during request construction—**before** `_present_other_agent_message` and **before** any plugin or `before_model_callback` runs.
- Because the filtering happens in `_ContentLlmRequestProcessor` via `isolation_scope`, observability plugins (e.g., Braintrust/OpenTelemetry) capture the exact same clean `llm_request.contents` that the model receives, `_build_task_input_user_content` guarantees a valid leading `role='user'` turn, and the agent's own `FunctionCall` / `FunctionResponse` events within that scope remain paired automatically—without needing per-agent author string allowlists.
- To pass selective outputs from specific upstream agents to a downstream agent, we recommend using `Workflow` node inputs or `output_key` + `{state_key}` in `instruction` together with `mode='single_turn'`.

Thank you again for your contribution and for highlighting the observability and FC/FR pairing requirements!

PR #5980, opened 2026-06-05
fix(api_server): add optional ADK_API_TOKEN Bearer auth middleware

ghost:
## Summary

Adds an opt-in Bearer-token authentication middleware to `ApiServer`. The middleware is a no-op when `ADK_API_TOKEN` is unset (current behavior preserved for every existing deployment) and requires a matching `Authorization: Bearer <token>` header on every non-public request when `ADK_API_TOKEN` is set.

`/health` and `/version` remain open so liveness probes do not need credentials.

## Why

`ApiServer` registers the following routes without per-route authentication:

- `POST /run`, `POST /run_sse` — invoke the agent runner, which can reach in-process Python code execution via `UnsafeLocalCodeExecutor.execute_code` if the loaded agent wires that executor.
- `POST/GET/PUT/DELETE /apps/{app_name}/users/{user_id}/sessions/{session_id}` — per-user session state CRUD, where the `user_id` is taken from the URL with no caller-identity binding.
- All other routes registered by `_register_production_endpoints` and the dev-server endpoints in `dev_server.py`.

The only middleware in front today is `CORSMiddleware` + `_OriginCheckMiddleware` + `_DefaultAppRewriteMiddleware` — none of which authenticate the caller. When an operator binds the server to a network-reachable address (`--host 0.0.0.0`, container host, LAN demo, etc.) and the deployment has no upstream auth layer, any network-reachable caller hits those routes unauthenticated.

This PR provides the smallest opt-in mitigation that turns the existing unauth surface into authenticated endpoints without breaking any current deployment.

## How

New `_BearerAuthMiddleware` in `src/google/adk/cli/api_server.py` mirrors the shape of the existing `_OriginCheckMiddleware` and `_DefaultAppRewriteMiddleware` ASGI middlewares in the same file.

Behavior matrix:

| `ADK_API_TOKEN` env | Path | `Authorization` header | Result |
| --- | --- | --- | --- |
| unset | any | any | passes through (current behavior) |
| set | `/health` or `/version` | any | passes through |
| set | other | absent | 401 |
| set | other | `Bearer <wrong-token>` | 401 |
| set | other | `Bearer <correct-token>` | passes through |
| set | non-HTTP (websocket, lifespan) | any | passes through |

The middleware is wired in `ApiServer.get_fast_api_app()` right after `_DefaultAppRewriteMiddleware`. No CLI, public-API, or config-schema changes.

## Tests

`tests/unittests/cli/test_bearer_auth_middleware.py` covers each row of the matrix above by exercising the middleware directly through the ASGI protocol (no FastAPI test client needed). Token-unset, token-set-no-header, token-set-wrong-bearer, token-set-correct-bearer, token-set-public-paths (parameterised over `/health` + `/version`), non-HTTP scopes, and session-route guard.

## Backward compatibility

`ADK_API_TOKEN` is unset by default, in which case the middleware is a pass-through. Every existing deployment continues to behave exactly as before. Adopting the new behavior is a one-line environment variable change at deploy time.

## Reference

Filed as a patch against Google bug-tracker issue 509219988 (the reviewing team there asked for a public patch alongside the report).

### DIFF
--- a/src/google/adk/cli/api_server.py
+++ b/src/google/adk/cli/api_server.py
@@ -310,6 +310,71 @@ class _DefaultAppRewriteMiddleware:
     await self._app(scope, receive, send)
 
 
+class _BearerAuthMiddleware:
+  """ASGI middleware that enforces Bearer token auth when ADK_API_TOKEN is set.
+
+  Behavior:
+    * If ADK_API_TOKEN is unset (the default), the middleware passes every
+      request through unchanged. This preserves the existing behavior for
+      deployments that already gate access at a different layer (reverse
+      proxy, sidecar, network policy, IAP, etc).
+    * If ADK_API_TOKEN is set, every request other than to the public health
+      and version endpoints must carry an ``Authorization: Bearer <token>``
+      header whose token equals ADK_API_TOKEN, otherwise the request is
+      rejected with HTTP 401.
+
+  The middleware exists because the ApiServer registers a number of routes
+  that reach in-process code execution (``/run``, ``/run_sse``) and per-user
+  session state (``/apps/.../users/.../sessions/...``) without per-route
+  authentication. When an operator binds the server to a network-reachable
+  address without an upstream auth layer, setting ADK_API_TOKEN is the
+  smallest mitigation that turns those routes into authenticated endpoints
+  without breaking any existing deployment.
+  """
+
+  _PUBLIC_PATHS = frozenset({"/health", "/version"})
+
+  def __init__(self, app: Any, token: Optional[str] = None) -> None:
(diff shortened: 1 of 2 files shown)


f4x0rz:
Closing in favor of #5981 which is opened from my CLA-linked account. Same commit content, with a Testing Plan section and live integration evidence added per @adk-bot's request. Sorry for the noise.

ghost:
Closing in favor of #5981, opened from my CLA-linked account. Same commit content, plus a Testing Plan section and live integration evidence added per @adk-bot's request. Sorry for the noise.

PR #5982, opened 2026-06-05
fix(auth): use branch-scoped events for auth responses

gezw:
## Summary

- read auth-response events through the current-branch event view
- ignore auth responses that are hidden by branch filtering
- add regression coverage for branch-scoped auth response handling

## Details

The auth preprocessor already consumes prior session events to decide whether an auth response should be applied to the next LLM request. This change makes that lookup use `InvocationContext._get_events(current_branch=True)` instead of reading the full session event list directly, so auth handling follows the same branch visibility semantics used by the rest of the invocation context.

## Validation

- `uv run --extra test pytest -q tests/unittests/auth/test_auth_preprocessor.py`
- `uv run --extra dev pyink --check src/google/adk/auth/auth_preprocessor.py tests/unittests/auth/test_auth_preprocessor.py`
- `uv run --extra test python -m py_compile src/google/adk/auth/auth_preprocessor.py tests/unittests/auth/test_auth_preprocessor.py`
- `git diff --check`


rohityan:
Hi @gezw , Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

gezw:
@googlebot rescan

gezw:
Thanks for the follow-up. The Google CLA check is passing now on this PR.

gezw:
I refreshed the branch against the latest `main`. The Google CLA check is passing, and the current reported checks are green or skipped as expected. Please let me know if there is anything else I should adjust for review.

PR #5987, opened 2026-06-05
fix: expose Gemini Live API client override

### LINKED ISSUE #5862: Make _live_api_client a public, documented override point (like api_client)

The `Gemini` class has a private `_live_api_client` property (leading underscore) that subclasses must override to customize the `genai.Client` used for Live API connections. This is contradictory — Python convention treats `_`-prefixed names as implementation details not intended for external override, yet Live API customization requires it.

The existing docstring documents how to override `api_client` for non-Live usage, but there's no public equivalent for Live API.

Rename `_live_api_client` to `live_api_client` (remove leading underscore) and update the docstring to document the override pattern, similar to the existing `api_client` example:

```python
class LiveGemini(Gemini):
    @cached_property
    def live_api_client(self):
        return genai.Client(enterprise=True, location="europe-central2")
```

Update the class docstring to include something like:

```
Customizing the underlying Live Client:
    To set ``google.genai.Client`` options for the Live API connection,
    subclass ``Gemini`` and override the ``live_api_client`` property::

        from functools import cached_property
        from google.genai import Client

        class LiveGemini(Gemini):
            @cached_property
            def live_api_client(self):
                return Client(enterprise=True, location="europe-central2")
```

### Impact on your work

We deploy Gemini agents on Vertex AI in a specific region. Without overriding `_live_api_client`, all Live API connections default to `us-central1`. We had to discover this private property by reading source code — it's not discoverable from the public API.

### Willingness to contribute

Yes, happy to submit a PR if the change direction is approved.

---

- Overriding `api_client` only: doesn't work, Live API uses `_live_api_client`, not `api_client`.
- Setting `location` via `http_options` on the model config: ADK doesn't expose a field for this on Live API connections.

In `google_llm.py`:

1. Rename `_live_api_client` → `live_api_client` (both the `@cached_property` and its usage in `connect()`).
2. Add the docstring section shown above.
3. Keep `_live_api_client` as a deprecated alias pointing to `live_api_client` for backward compatibility.

### Additional Context

The `api_client` property already follows this public-override pattern successfully. The `_live_api_client` property was introduced later (for the Live API code path) but its usage pattern is identical — subclasses need to override it — yet it was mistakenly made private.

he-yufeng:
Fixes #5862.

## Summary

`Gemini` already exposes `api_client` as a public override point for normal model calls, but Live API connections still use the private `_live_api_client` property. That makes regional Vertex AI Live setups depend on overriding a private member even though this is a real extension point.

This adds a public `live_api_client` property, documents the override pattern, and switches `connect()` to use it. The old `_live_api_client` property remains as a compatibility alias. Existing subclasses that already override `_live_api_client` are still honored by the default `live_api_client` implementation.

## Validation

```console
python -m py_compile src\google\adk\models\google_llm.py tests\unittests\models\test_google_llm.py
python -m pytest tests\unittests\models\test_google_llm.py -q -k "live_api_client or connect_with_custom_headers or connect_without_custom_headers"
python -m pytest tests\unittests\models\test_google_llm.py -q -k "live_api_client or live_api_version or speech_config"
python -m pyink --check src\google\adk\models\google_llm.py tests\unittests\models\test_google_llm.py
git diff --check upstream/main...HEAD
```

Note: my local environment initially had `google-genai==1.65.0`; this branch was tested after installing the repository-required `google-genai>=2.4,<3`.


he-yufeng:
Updated the branch to cover the mTLS Gemini endpoint in the Live API base URL test as well. This should satisfy the hardcoded googleapis.com check for the changed test file.

Validated locally:
- `python -m py_compile src/google/adk/models/google_llm.py tests/unittests/models/test_google_llm.py`
- `uv run pyink --check src/google/adk/models/google_llm.py tests/unittests/models/test_google_llm.py`
- `.venv\Scripts\python -m pytest -q tests\unittests\models\test_google_llm.py -k "live_api_client_uses_api_version_from_google_base_url or live_api_client_properties or api_client_uses_api_version_from_google_base_url" --basetemp .tmp\pytest-5987-models -p no:cacheprovider`

Focused model tests passed: 4 passed, 61 deselected.

he-yufeng:
Closing this lower-signal Live API client override PR for now. It has not moved beyond the automated analysis path, and I want to keep the open ADK queue focused on branches with clearer maintainer or user validation.

PR #6003, opened 2026-06-07
fix: add SSRF protection to RestApiTool._request()

Neelagiri65:
Re-opening of #5760 after signing the Google Individual CLA. Branch rebased onto current main, merge conflict resolved, and the failing CI addressed.

## What

`RestApiTool._request()` passed URLs to `httpx.AsyncClient.request()` with no validation. After #4368 fixed SSRF in `load_web_page` (hostname blocking, DNS pre-resolution, IP range filtering, scheme restriction), `RestApiTool` was never given the same protection.

A malicious OpenAPI spec with `servers[].url` pointing at `127.0.0.1` or `169.254.169.254` reached the HTTP client untouched.

## Changes

New private module `src/google/adk/tools/_ssrf_protection.py`. It carries the SSRF checks for the httpx-based request path used by `RestApiTool`.

Public surface:

- `validate_url(url) -> ValidatedTarget` — scheme check, hostname pattern block, DNS resolution, IP range filter. Returns the resolved addresses.
- `send_pinned_async(client, target, **params) -> httpx.Response` — issues the request against the validated IP literal, preserves the Host header, sets TLS SNI via `request.extensions["sni_hostname"]`.
- `is_blocked_hostname`, `is_blocked_address`, `resolve_host_addresses`, `rewrite_url_host` — helpers.

`RestApiTool._request()` now validates the URL and pins the connection before any request goes out. The `httpx_client_factory` parameter and `ssl_verify` semantics are untouched. Factory-supplied clients still manage their own SSL.

## Scope note

An earlier version of this PR also migrated `load_web_page.py` to consume this shared module, for a single source of truth. That is dropped now. `main` has since given `load_web_page.py` its own inline SSRF implementation with a dedicated test suite that patches module-local names (`load_web_page.socket` and friends). Migrating that file breaks those tests and creates a standing conflict every time the file changes upstream. So this PR leaves `load_web_page.py` exactly as it is on `main` and only adds protection to `RestApiTool`. The two paths use different HTTP stacks anyway (requests vs httpx).

## What gets blocked

`is_blocked_address` uses `ipaddress.is_global`, so one rule covers IPv4 and IPv6 without hand-maintained ranges:

- Loopback (`127.0.0.0/8`, `::1`)
- Link-local (`169.254.0.0/16`, `fe80::/10`), which covers cloud metadata endpoints
- Private (`10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `fc00::/7`)
- Multicast, reserved, unspecified
- `localhost`, `*.localhost`, with case folding and trailing-dot handling
- Non-http(s) schemes: `file://`, `ftp://`, `gopher://`, and the rest

## DNS rebinding

Validating the URL just before sending leaves a TOCTOU window. Even if `getaddrinfo` returns a public IP at validation, the second resolution at connect time could return a private IP. The fix pins to the pre-resolved IP:

- `validate_url` rejects the hostname if any record points to a non-globally-routable IP. An attacker cannot sneak past with `[8.8.8.8, 127.0.0.1]`.
- `send_pinned_async` rewrites the URL to the validated IP literal, sets `Host` to the original hostname for routing, and sets `sni_hostname` (consumed by httpcore) so TLS cert verification still works. The socket goes to the IP that was vetted, not whatever DNS returns at connect time.

## Testing plan

Repro of the original vulnerability on pre-fix `main`:

```json
{
  "openapi": "3.0.0",
  "info": {"title": "Internal API", "version": "1.0.0"},
  "servers": [{"url": "http://169.254.169.254"}],
  "paths": {
    "/latest/meta-data/iam/security-credentials/": {
      "get": {"operationId": "leak_creds", "responses": {"200": {"description": "ok"}}}
    }
  }
}
```

On pre-fix `main` the request goes to `169.254.169.254` unimpeded. On this branch:

```
ValueError: Blocked host: 169.254.169.254
```

The exception is raised by `validate_url` before any HTTP request hits the wire. Same outcome for `127.0.0.1`, `10.0.0.1`, `localhost`, `192.168.x.x`, and IPv6 loopback / link-local.

Running the tests:

```
pip install -e .
pytest tests/unittests/tools/test_ssrf_protection.py -v
pytest tests/unittests/tools/openapi_tool/openapi_spec_parser/test_rest_api_tool.py -v
```

`test_ssrf_protection.py` covers hostname blocking, address blocking across IPv4/IPv6 classes, host resolution, `validate_url` for every blocked class plus the DNS-rebinding case, URL rewriting, and `send_pinned_async` pinning behaviour. `test_rest_api_tool.py` adds five tests that block localhost, loopback, the metadata endpoint, a private IP, and the `file://` scheme through `_request`.

Note for reviewers: the SSRF tests use `8.8.8.8` / `8.8.4.4` as the stand-in public addresses. Recent CPython reclassifies the RFC 5737 documentation ranges (`203.0.113.0/24` and friends) as `is_global=False`, so the earlier version of the tests that used those addresses was correctly blocked by the production code and failed. Real globally-routable literals avoid that. All of them are mocked, no request leaves the process.

Related: #4368, #5760


rohityan:
Hi @Neelagiri65 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please address the failing CI tests.

Neelagiri65:
Thanks @rohityan. Rebased onto current main and sorted the failing checks.

- Merge conflict was in `load_web_page.py`. Main grew its own inline SSRF implementation there with tests that patch module-local names, so the earlier migration of that file no longer fits. I reverted `load_web_page.py` to match main exactly and scoped this PR to the `RestApiTool` fix plus the shared `_ssrf_protection` module. No behaviour change to `load_web_page`.
- The unit test failures came from the SSRF tests using RFC 5737 documentation IPs (`203.0.113.x`) as stand-in public addresses. Recent CPython reports those as `is_global=False`, so the production check was correctly blocking them. Switched the tests to real globally-routable literals. All mocked, no traffic leaves the process.
- Ran pre-commit (pyink, isort, ruff, addlicense, compliance) and the tool unit tests locally. Green.

Happy to adjust if you'd prefer the shared module folded directly into `rest_api_tool.py` instead of a separate file.

PR #6007, opened 2026-06-07
fix: Make EUC request args JSON-serializable

### LINKED ISSUE #6006: adk_request_credential FunctionCall args contain non-JSON-serializable SecuritySchemeType enum
## 🔴 Required Information

**Describe the Bug:**

`build_auth_request_event` in `src/google/adk/flows/llm_flows/functions.py` builds the `adk_request_credential` (EUC) `FunctionCall.args` with `AuthToolArguments(...).model_dump(exclude_none=True, by_alias=True)`. The default `mode="python"` leaves the auth scheme's `type` field as a live `SecuritySchemeType` enum member inside the args dict, rather than its string value. The resulting `Event.content` is therefore not JSON-serializable: any consumer that calls `json.dumps` on the in-memory event content raises `TypeError: Object of type SecuritySchemeType is not JSON serializable`.

This is latent for most flows because the session DB launders enums to strings on write (pydantic JSON serialization), and ADK's own auth preprocessor re-validates the args from either form. It surfaces for any consumer that serializes the in-memory event before it round-trips through the DB. A concrete trigger is memory ingestion: `VertexAiMemoryBankService` hands the in-memory `event.content` to the Vertex SDK, which `json.dumps`'s it.

**Steps to Reproduce:**

Run the snippet below against google-adk 2.2.0. It performs the same `model_dump` call as `build_auth_request_event` and then the same `json.dumps` the Vertex SDK performs during memory ingestion.

```python
import json
from google.adk.auth.auth_tool import AuthConfig, AuthToolArguments
from google.adk.auth.auth_credential import AuthCredential, AuthCredentialTypes, OAuth2Auth
from google.adk.auth.auth_schemes import OAuth2
from fastapi.openapi.models import OAuthFlows, OAuthFlowAuthorizationCode

scheme = OAuth2(flows=OAuthFlows(authorizationCode=OAuthFlowAuthorizationCode(
    authorizationUrl="https://example/authorize", tokenUrl="https://example/token", scopes={})))
cfg = AuthConfig(auth_scheme=scheme,
                 raw_auth_credential=AuthCredential(auth_type=AuthCredentialTypes.OAUTH2,
                                                    oauth2=OAuth2Auth(client_id="x", client_secret="y")))

# The exact dump build_auth_request_event performs (functions.py, default mode="python")
args = AuthToolArguments(function_call_id="abc", auth_config=cfg).model_dump(
    exclude_none=True, by_alias=True)
print(repr(args["authConfig"]["authScheme"]["type"]))  # <SecuritySchemeType.oauth2: 'oauth2'> — live enum
json.dumps(args)  # TypeError: Object of type SecuritySchemeType is not JSON serializable

# The fix: mode="json" coerces the enum to its string value, same dump otherwise
fixed = AuthToolArguments(function_call_id="abc", auth_config=cfg).model_dump(
    mode="json", exclude_none=True, by_alias=True)
print(repr(fixed["authConfig"]["authScheme"]["type"]))  # 'oauth2'
json.dumps(fixed)  # passes
```

**Expected Behavior:**

The `adk_request_credential` `FunctionCall.args` are JSON-serializable, consistent with how the same content is stored and re-read from the session DB. `model_dump(mode="json", exclude_none=True, by_alias=True)` coerces the enum to its `"oauth2"` string while preserving aliases; ADK's parse-back is unaffected because pydantic validation accepts the string form (exactly what it already does for DB-read events).

**Observed Behavior:**

`json.dumps` on the in-memory event content raises:

```text
TypeError: Object of type SecuritySchemeType is not JSON serializable
```

In a real flow (memory ingest on the EUC turn) this surfaces as a background task failure:

```text
ERROR vertex_ai_memory_bank_service.py | Background ingest_events task failed: Object of type SecuritySchemeType is not JSON serializable
```

**Environment Details:**

- ADK Library Version (pip show google-adk): 2.2.0 (latest release; identical on `main`, the dump is at `functions.py:305`, SHA `9670ce2`)
- Desktop OS: macOS
- Python Version (python -V): 3.13

**Model Information:**

- Are you using LiteLLM: No
- Which model is being used: N/A, the bug is in event construction and model-independent

---

## 🟡 Optional Information

**Regression:**

Not a recent regression. The python-mode dump has been in place across releases (present in the 1.x line and unchanged on the current 2.x `main`); it only becomes observable when a consumer serializes the in-memory EUC event before DB round-trip.

**Logs:**

```text
TypeError: Object of type SecuritySchemeType is not JSON serializable
```

**Additional Context:**

Sibling call sites in the same file use the same python-mode `model_dump` and may carry the same hazard if their models contain enums (like the tool-confirmation args around `functions.py:368`/`:371`, and `event.actions.model_dump(...)` near line 1241). Worth a glance during the fix, but the auth-request path is the confirmed, reproducible one. Discovered downstream while validating an OAuth resume flow; the in-memory event reached `json.dumps` via memory ingestion before any DB round-trip.

Inline in Steps to Reproduce above; self-contained against a clean `pip install google-adk==2.2.0`.

- Always (100%), deterministic for any consumer that `json.dumps`'s the in-memory EUC event content.

doughayden:

- Closes: #6006

**Problem:**

Two call sites dump `AuthToolArguments` with the default `mode="python"`, leaving the auth scheme's `type` field as a live `SecuritySchemeType` enum inside the `adk_request_credential` FunctionCall args:

- `build_auth_request_event` in `flows/llm_flows/functions.py` (the LLM-flow EUC path)
- `create_auth_request_event` in `workflow/utils/_workflow_hitl_utils.py` (the v2 workflow HITL path)

The in-memory `Event.content` is therefore not JSON-serializable: any consumer that `json.dumps` it before it round-trips the session DB raises `TypeError: Object of type SecuritySchemeType is not JSON serializable`. Memory ingestion (`VertexAiMemoryBankService`) is a concrete trigger. It stays hidden in most flows because the session DB launders enums to strings on write and the auth preprocessor re-validates from either form.

**Solution:**

Dump with `mode="json"` at both sites, which coerces the enum to its `"oauth2"` string while preserving `by_alias`/`exclude_none`. Parse-back is unaffected: pydantic validation already accepts the string form, exactly what DB-read events carry. Two one-line changes, each with a regression test.

### Testing Plan

**Unit Tests:**

Both tests build an EUC/auth request, assert the args are JSON-serializable, and assert the auth scheme type is the string `"oauth2"`. Each fails before its change and passes after.

- `test_function_request_euc_args_are_json_serializable` in `tests/unittests/flows/llm_flows/test_functions_request_euc.py` (LLM-flow path)
- `test_args_are_json_serializable` in `tests/unittests/workflow/utils/test_workflow_hitl_utils.py` (workflow HITL path)

```text
$ pytest tests/unittests/flows/llm_flows/test_functions_request_euc.py -q
4 passed, 24 warnings in 1.25s

$ pytest tests/unittests/workflow/utils/test_workflow_hitl_utils.py -q
12 passed, 5 warnings in 0.67s
```

The inline repro in the linked issue is the manual check: `json.dumps` on the args dict raises `TypeError` before the change and passes after. Verified against google-adk 2.2.0 (== current `main`).

### Checklist

### Additional context

The workflow HITL site was folded in per @surajksharma07's suggestion on the issue, so both auth-request builders are now covered. The remaining sibling `model_dump(exclude_none=True, by_alias=True)` call sites in `functions.py` (the tool-confirmation args near `:368`/`:371` and the event-actions dump near `:1241`) follow the same python-mode pattern and may carry the same hazard if their models contain enums, but they are a different feature path with no confirmed repro, so I left them out of scope. Happy to fold them in if you'd prefer consistency across the file.


rohityan:
Hi @doughayden, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors before we can proceed with the review.

doughayden:
Thanks @rohityan. The failing check is the check-file-contents hardcoded endpoints scan.

It's a clean fix for `test_workflow_hitl_utils.py`: the only googleapis URLs in the file were the two my test added, so I can swap them to `example.com` placeholders. The fixture only needs a well-formed OAuth scheme to pass the structural assertions.

I want your call on `test_functions_request_euc.py`. My new test contributes 2 googleapis lines following existing patterns, but the file already carried 20 from the other tests in that module. The scan reads the whole file, and swapping only mine still leaves it flagged. It feels like an overreach to rewrite 20 pre-existing fixtures I didn't author inside a serialization fix, but I'll do it if you give the green light. How would you prefer I handle it?

doughayden:
Not pushing changes to tests following https://github.com/google/adk-python/issues/6006#issuecomment-4721720414

rohityan:
Hi @doughayden , please fix the failing mypy-diff tests.

doughayden:
Hi @rohityan, the mypy failure in https://github.com/google/adk-python/actions/runs/27711711919/job/81974523407 is in a file this PR did not touch. The PR branch is behind `main` where the mypy fix already lives following today's 81b80671 (fix: api-registry to fetch all services). I'll update it with another merge commit to clear the mypy error. Note: the check-file-contents workflow will still fail until @surajksharma07 [patches it](https://github.com/google/adk-python/issues/6006#issuecomment-4721720414).

PR #6009, opened 2026-06-08
fix(live): keep streaming tool yields from completing turns

### LINKED ISSUE #5947: Streaming tool yields in live mode cause model to re-invoke the tool in a loop
## Description

When using an async generator streaming tool in live (bidi) mode, each `yield` from the tool causes the model to be interrupted and re-invoke the same tool, creating an infinite loop of duplicate tool calls.

## Root Cause

In `_process_function_live_helper` (`functions.py`), streaming tool results are sent back to the model via `live_request_queue.send_content()`:

```python
async for result in agen:
    updated_content = types.Content(
        role='user',
        parts=[
            types.Part.from_text(
                text=f'Function {tool.name} returned: {result}'
            )
        ],
    )
    invocation_context.live_request_queue.send_content(updated_content)
```

In `GeminiLlmConnection.send_content()`, this is forwarded as:

```python
await self._gemini_session.send_client_content(turns=[content], turn_complete=True)
```

The combination of `role='user'` and `turn_complete=True` makes the model treat each streaming update as a **new complete user turn**. The model:

1. Gets interrupted (stops current audio/text generation)
2. Receives what looks like a new user message: `"Function run_my_tool returned: {'status': 'running', ...}"`
3. Responds to it — often by calling the same tool again
4. The tool is already running, so it either starts a duplicate or returns "busy"
5. Repeat

This produces a rapid `interrupted → turn_complete → tool_call → interrupted → ...` loop visible in the logs, with the model never completing a coherent response.

## Steps to Reproduce

1. Define an async generator tool that yields progress updates:

```python
async def long_running_task(query: str):
    """A tool that takes time and yields progress updates."""
    yield {"status": "running", "message": "Step 1: fetching data..."}
    await asyncio.sleep(5)
    yield {"status": "running", "message": "Step 2: processing..."}
    await asyncio.sleep(5)
    yield {"status": "completed", "result": "Done!"}
```

2. Create an agent with this tool and run in live mode with a native-audio model
3. Send a message that triggers the tool
4. Observe: the model gets interrupted by each yield, re-calls the tool repeatedly, and never delivers a coherent spoken response

## Expected Behavior

Streaming tool updates should be delivered to the model in a way that does **not** interrupt the current generation or trigger re-invocation. The model should be able to narrate progress updates naturally between yields.

## Possible Fixes

- Send streaming updates with `turn_complete=False` so the model treats them as partial/background context rather than a new turn requiring a response
- Use a different role (e.g., `'tool'` or function_response) instead of `'user'` so the model doesn't interpret yields as user messages
- Queue streaming updates and let the model consume them after its current generation completes, rather than interrupting it
- Provide a mechanism to send out-of-band status updates that bypass the model entirely (e.g., via a callback to the application layer)

## Environment

- `google-adk`: 2.1.0
- Model: `gemini-live-2.5-flash-native-audio` (Vertex AI)
- Python: 3.12

he-yufeng:
## Summary

Fixes #5947.

This threads a `turn_complete` flag through the live request path so streaming tool updates can be sent as intermediate model input instead of being treated as completed user turns.

The change keeps the existing default behavior for normal `send_content()` calls:

- `LiveRequestQueue.send_content()` still defaults to `turn_complete=True`
- async generator tool yields now send `turn_complete=False`
- `_send_to_model()` forwards the flag to the live connection
- intermediate yields are not appended to session history
- Gemini only uses the Gemini 3.1 realtime text shortcut for completed turns

## Tests

```
PYTHONPATH=src python -m pytest tests/unittests/agents/test_live_request_queue.py tests/unittests/models/test_gemini_llm_connection.py tests/unittests/flows/llm_flows/test_base_llm_flow_realtime.py tests/unittests/streaming/test_streaming.py::test_live_streaming_text_content_persisted_in_session tests/unittests/flows/llm_flows/test_base_llm_flow.py -q
python -m pyink --check src/google/adk/agents/live_request_queue.py src/google/adk/models/base_llm_connection.py src/google/adk/models/gemini_llm_connection.py src/google/adk/flows/llm_flows/base_llm_flow.py src/google/adk/flows/llm_flows/functions.py tests/unittests/agents/test_live_request_queue.py tests/unittests/models/test_gemini_llm_connection.py tests/unittests/flows/llm_flows/test_base_llm_flow_realtime.py tests/unittests/testing_utils.py tests/unittests/workflow/testing_utils.py
git diff --check
```

### DIFF
--- a/src/google/adk/agents/live_request_queue.py
+++ b/src/google/adk/agents/live_request_queue.py
@@ -55,6 +55,9 @@ class LiveRequest(BaseModel):
   close: bool = False
   """If set, close the queue. queue.shutdown() is only supported in Python 3.13+."""
 
+  turn_complete: bool = True
+  """If set, content messages complete the current model turn."""
+
 
 class LiveRequestQueue:
   """Queue used to send LiveRequest in a live(bidirectional streaming) way."""
@@ -65,8 +68,10 @@ class LiveRequestQueue:
   def close(self):
     self._queue.put_nowait(LiveRequest(close=True))
 
-  def send_content(self, content: types.Content):
-    self._queue.put_nowait(LiveRequest(content=content))
+  def send_content(self, content: types.Content, turn_complete: bool = True):
+    self._queue.put_nowait(
+        LiveRequest(content=content, turn_complete=turn_complete)
+    )
 
   def send_realtime(self, blob: types.Blob):
     self._queue.put_nowait(LiveRequest(blob=blob))
--- a/src/google/adk/flows/llm_flows/base_llm_flow.py
+++ b/src/google/adk/flows/llm_flows/base_llm_flow.py
@@ -803,9 +803,9 @@ class BaseLlmFlow(ABC):
         is_function_response = content.parts and any(
             part.function_response for part in content.parts
         )
-        if not is_function_response:
-          if not content.role:
-            content.role = 'user'
+        if not is_function_response and not content.role:
+          content.role = 'user'
+        if not is_function_response and live_request.turn_complete:
           user_content_event = Event(
               id=Event.new_id(),
               invocation_id=invocation_context.invocation_id,
@@ -816,7 +816,9 @@ class BaseLlmFlow(ABC):
               session=invocation_context.session,
               event=user_content_event,
           )
-        await llm_connection.send_content(live_request.content)
+        await llm_connection.send_content(
+            live_request.content, turn_complete=live_request.turn_complete
+        )
 
   async def _receive_from_model(
       self,
--- a/src/google/adk/flows/llm_flows/functions.py
+++ b/src/google/adk/flows/llm_flows/functions.py
@@ -980,7 +980,9 @@ async def _process_function_live_helper(
             updated_content = _build_function_response_content(
                 tool, result, tool_context.function_call_id
             )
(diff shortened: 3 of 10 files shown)


he-yufeng:
Rebased onto current main and resolved the Gemini 3.1 text-send conflict by retaining the new cached model-capability flag while respecting turn_complete=False for intermediate tool yields. Six focused queue/flow/connection tests pass; pyink, py_compile, and diff check pass.

rohityan:

he-yufeng:
Rebased this on current `main` and re-ran the focused live-streaming checks:

- `PYTHONPATH=src python -m pytest tests/unittests/agents/test_live_request_queue.py tests/unittests/models/test_gemini_llm_connection.py tests/unittests/flows/llm_flows/test_base_llm_flow_realtime.py tests/unittests/streaming/test_streaming.py::test_live_streaming_text_content_persisted_in_session tests/unittests/flows/llm_flows/test_base_llm_flow.py -q --basetemp .tmp/pytest-6009 -p no:cacheprovider` -> 97 passed
- `python -m py_compile` on the touched live-flow files

All passed locally.

wuliang229:
Merged in https://github.com/google/adk-python/commit/07aa1e09e2c2b3aebd9c3457f30e57b9f9b167b7

PR #6011, opened 2026-06-08
fix: warn on deprecated Gemini model alias

### LINKED ISSUE #6010: Deprecated model alias (gemini-flash-latest) causes confusing 404 NOT_FOUND with no actionable guidance
## Summary

When deploying an ADK application to Cloud Run (or any environment where a deprecated Gemini model alias is configured), all `LlmAgent` calls silently fail with a confusing `404 NOT_FOUND` error that does not clearly indicate the root cause.

## Environment

- `google-adk`: 2.1.0
- `google-genai`: latest
- Platform: Google Cloud Run (asia-northeast1)
- Python: 3.12

## Steps to Reproduce

1. Set `GEMINI_MODEL=gemini-flash-latest` in the environment (common in older ADK projects or tutorials that referenced this alias)
2. Pass it to `LlmAgent`:
   ```python
   MODEL = os.environ.get("GEMINI_MODEL", "gemini-flash-latest")
   agent = LlmAgent(name="MyAgent", model=MODEL, ...)
   ```
3. Deploy to Cloud Run and invoke the agent

## Actual Error

```
google.genai.errors.ClientError: 404 NOT_FOUND.
{'error': {'code': 404, 'message': 'This model models/gemini-2.0-flash is
no longer available. Please update your code to use a newer model for the
latest features and improvements.', 'status': 'NOT_FOUND'}}
```

## Problem

The error message mentions `models/gemini-2.0-flash` — the **resolved** underlying model ID — but the user configured `gemini-flash-latest`. This disconnect makes diagnosis difficult:

- The user sees `gemini-2.0-flash` in the error but never wrote that name anywhere
- There is no hint that `gemini-flash-latest` is the problematic alias
- The error surface is deep in the agent invocation chain, far from the configuration

This was particularly confusing during a hackathon where the app appeared to work locally (older cached responses / different routing) but failed 100% on Cloud Run.

## Root Cause

`gemini-flash-latest` is a non-versioned alias that previously resolved to `gemini-2.0-flash`. Google has since deprecated `gemini-2.0-flash` (the specific model behind the alias), causing the alias to produce `404` errors.

Note: ADK `2.1.0` correctly uses `DEFAULT_MODEL = 'gemini-2.5-flash'` — the issue occurs when users explicitly pass an old alias via environment variable or hardcoded string.

## Suggested Improvements

**Option A — Startup validation warning**
Before agent invocation, validate the `model` string against a known deprecated-alias list and emit a `logging.warning`:
```
DeprecationWarning: model alias 'gemini-flash-latest' is no longer available.
Please use 'gemini-2.5-flash' or a versioned model ID.
```

**Option B — Improved error interception**
Catch `404 NOT_FOUND` errors from the Gemini API and re-raise with a user-friendly message:
```
ModelDeprecatedError: The model alias 'gemini-flash-latest' resolved to
'gemini-2.0-flash' which has been deprecated. Please update your model
configuration to 'gemini-2.5-flash'.
```

**Option C — Documentation**
Add a "Model Version Management" section to the ADK docs / README warning users against non-versioned aliases and recommending pinned model IDs.

## Workaround

Replace any occurrence of `gemini-flash-latest` (or other non-versioned aliases) with a specific versioned model ID:

```python
# Before (breaks silently when alias is deprecated)
MODEL = os.environ.get("GEMINI_MODEL", "gemini-flash-latest")

# After (stable, explicit)
MODEL = os.environ.get("GEMINI_MODEL", "gemini-2.5-flash")
```

This was discovered while building a multi-agent privacy tool (AI Shield Persona Agent) for the ADK Hackathon. Sharing in case it helps other hackathon participants and ADK users.

he-yufeng:
## Summary
- warn when `LLMRegistry.new_llm()` sees the deprecated `gemini-flash-latest` alias
- keep behavior unchanged: the configured model string is still passed through as-is
- add regression coverage for both plain and `gemini:`-prefixed alias usage

Fixes #6010.

## To verify
- `PYTHONPATH=src python -m pytest tests\unittests\models\test_models.py -q`
- `python -m pyink --check src\google\adk\models\registry.py tests\unittests\models\test_models.py`
- `git diff --check`

### DIFF
--- a/src/google/adk/models/registry.py
+++ b/src/google/adk/models/registry.py
@@ -31,6 +31,9 @@ logger = logging.getLogger('google_adk.' + __name__)
 
 _LazyEntry = tuple[str, str]
 _llm_registry_dict: dict[str, Union[type['BaseLlm'], _LazyEntry]] = {}
+_deprecated_model_aliases = {
+    'gemini-flash-latest': 'gemini-2.5-flash',
+}
 
 
 class LLMRegistry:
@@ -48,6 +51,7 @@ class LLMRegistry:
     """
 
     prefix, actual_model = LLMRegistry._parse_model(model)
+    LLMRegistry._warn_if_deprecated_alias(actual_model)
     cls = LLMRegistry.resolve(model)
 
     if prefix and LLMRegistry._match_prefix(prefix, cls.__name__):
@@ -66,6 +70,21 @@ class LLMRegistry:
       prefix, actual_model = model.split(':', 1)
       return prefix, actual_model
     return None, model
+
+  @staticmethod
+  def _warn_if_deprecated_alias(model: str) -> None:
+    replacement = _deprecated_model_aliases.get(model)
+    if replacement is None:
+      return
+
+    logger.warning(
+        (
+            'Model alias %r is deprecated and may resolve to an unavailable'
+            ' Gemini model. Use %r or another versioned model id instead.'
+        ),
+        model,
+        replacement,
+    )
 
   @staticmethod
   def _match_prefix(prefix: str, class_name: str) -> bool:
--- a/tests/unittests/models/test_models.py
+++ b/tests/unittests/models/test_models.py
@@ -133,6 +133,22 @@ def test_new_llm_with_prefix(mocker):
   mock_class.assert_called_once_with(model='gpt-4')
 
 
+@pytest.mark.parametrize(
+    'model_name',
+    ['gemini-flash-latest', 'gemini:gemini-flash-latest'],
+)
+def test_new_llm_warns_for_deprecated_model_alias(caplog, model_name):
+  """Test that deprecated Gemini aliases surface a useful warning."""
+  caplog.set_level('WARNING', logger='google_adk.google.adk.models.registry')
+
+  llm = models.LLMRegistry.new_llm(model_name)
+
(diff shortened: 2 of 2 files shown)


pj0201:
Hit this exact issue while building a multi-agent app for the ADK Hackathon — `gemini-flash-latest` silently mapping to a deprecated model caused confusing 404s on Cloud Run that were hard to trace back to the alias.

The startup warning in `LLMRegistry.new_llm()` is exactly the right fix — catches it early, no behavior change. Clean and minimal. Hope this gets merged soon!

rohityan:

he-yufeng:

- `PYTHONPATH=src python -m pytest tests\unittests\models\test_models.py -q --basetemp .tmp\pytest-6011 -p no:cacheprovider` -> 28 passed
- `python -m py_compile` on the touched files
- `git diff --check origin/main..HEAD`

All passed locally.

GWeale:
Thanks for tackling #6010! We're going to pass on this one. We'd rather not maintain a static alias list, and the dynamic models.list() approach doesn't work as a signal (it only returns fully-versioned IDs, so it false-flags valid aliases like gemini-2.0-flash and *-latest, and varies by backend/region). We'll rely on the backend 404, which already reports
the model as unavailable.

PR #6032, opened 2026-06-09
fix(runners): honor before_run_callback early-exit for Workflow runs

### LINKED ISSUE #6013: ADK 2.0 ignores before_run callback for Workflow
ADK 2.0 ignores before_run callback returning a value for workflows. In this case the execution should stop and no further llm calls should be executed.

Problem is in runner.py / _run_node_async where
```python
await ic.plugin_manager.run_before_run_callback(invocation_context=ic)
```
swallows returned value and never checks if execution should continue

**Steps to Reproduce:**
Use Workflow as root_agent
Implement plugin and override before_run_callback

**Expected Behavior:**
Execution should stop if before_run_callback returns a value as described here:
"Returning a value to halt execution of the runner and ends the runner with that event"

**Environment Details:**

 - ADK Library Version (pip show google-adk): 2.2
 - Desktop OS: Linux

**Model Information:**

 - Are you using LiteLLM: No

---

 - Always (100%)

garyzava:
## Summary

In the Workflow run path (`Runner._run_node_async`), the `before_run` plugin
callback was awaited but its return value was discarded. Per the plugin contract,
returning a `Content` from `before_run_callback` should halt the run and end it
with that content — but for Workflow root agents the value was ignored and
execution continued into the workflow.

This captures the callback result and, when it is a `Content`, emits it as the
early-exit event and stops — matching the existing behavior of the non-workflow
run path (`run_async`).

Fixes #6013

## What changed

* `src/google/adk/runners.py` — in `_run_node_async`, capture
  `run_before_run_callback(...)`'s result; if it's a `types.Content`, build the
  early-exit `Event`, append it (subject to `_should_append_event`), yield it, and
  return — instead of unconditionally proceeding. Mirrors the non-workflow path.
* `tests/unittests/workflow/test_workflow_failures.py` — regression test: a plugin
  whose `before_run_callback` returns `Content` halts the workflow and the node
  never executes.

## Why this approach

The non-workflow path already implements the documented early-exit contract
(capture the result, emit an early-exit event, skip execution). This change makes
the Workflow path consistent with it rather than introducing new behavior, so the
fix is minimal and the intended semantics are unambiguous.

## How I tested

```
$ pytest tests/unittests/workflow/test_workflow_failures.py -q
23 passed
```

The new test (`test_workflow_halts_when_before_run_callback_returns_content`)
fails before the change (the node executes; no early-exit event) and passes
after. Existing workflow-failure tests and the non-workflow plugin tests
(`tests/unittests/test_runners.py -k plugin`) still pass. `pyink` and `isort`
report no changes.

## Backward compatibility

No behavior change when `before_run_callback` returns `None` (the common case) —
the workflow runs exactly as before. Only the documented halt-on-`Content` case is
affected, and it now matches the non-workflow path.


rohityan:
Hi @garyzava , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. Can you meanwhile address the adk bot nits suggestions.

rohityan:

garyzava:
Hi @rohityan, thanks for taking a look! The bot's nit (missing type annotations on the `before_run_callback` test mock) is addressed in 5a7c9b5  the mock now annotates `invocation_context: InvocationContext` and the `types.Content` return type. The branch is also up to date with `main`, and the workflow suite passes locally (`pytest tests/unittests/workflow/test_workflow_failures.py` → 23 passed). Happy to make any further changes.

Dratatus:
Thank you for this fix — we hit the same bug while hardening a production app (a guardrail plugin silently bypassed for a root LlmAgent). Since this PR no longer merges cleanly and #6013 was closed by the stale bot, we rebased your change onto current main and extended it with a root-LlmAgent regression test: #6829 (credit to you in the commit message). Happy to hand it back or co-iterate if you prefer to revive this one instead.

PR #6042, opened 2026-06-09
fix(flows): Reset reconnect attempts on connection success

wukath:
Reset the reconnect attempt counter immediately when a live connection is established successfully, rather than waiting for the first event to be received. This prevents reconnection failures if the connection succeeds but no events are received before a transient disconnect occurs.


wuliang229 (approved):

wissamblue69-dotcom:
> Reset the reconnect attempt counter immediately when a live connection is established successfully, rather than waiting for the first event to be received. This prevents reconnection failures if the connection succeeds but no events are received before a transient disconnect occurs.

PR #6074, opened 2026-06-11
fix(code_executors): harden ContainerCodeExecutor sandbox by default

adilburaksen:
## Summary

`ContainerCodeExecutor` runs model-generated code, which can be influenced by untrusted input (e.g. via prompt injection). It starts the container with default Docker networking and no capability restrictions, so the executed code can reach the cloud metadata endpoint (`169.254.169.254`) — which yields the host service-account token — reach internal services, or escalate privileges.

This is inconsistent with the isolation posture of every other ADK code executor:

- `GkeCodeExecutor` runs under gVisor with `cap_drop: ["ALL"]`, non-root, read-only root filesystem, and a strict security context.
- `BuiltInCodeExecutor` / `VertexAiCodeExecutor` / `AgentEngineSandboxCodeExecutor` run in managed server-side sandboxes.
- `UnsafeLocalCodeExecutor` is explicitly documented as unsafe.

`ContainerCodeExecutor` was the only executor running code with full network access and no isolation flags or warning.

## Change

- Start the container with `network_disabled=True` by default. This is exposed as a configurable `network_disabled` field — set it to `False` to re-enable networking when the executed code is trusted.
- Drop all Linux capabilities (`cap_drop=["ALL"]`) and forbid privilege escalation (`security_opt=["no-new-privileges"]`), matching `GkeCodeExecutor`.
- Document the security posture in the class docstring and point users to the sandboxed executors for untrusted code.
- Add unit tests covering the hardened defaults and the opt-in network path.

## Compatibility

Code that legitimately needs network access can opt back in with `ContainerCodeExecutor(..., network_disabled=False)`. Dropping capabilities and `no-new-privileges` do not affect normal Python code execution.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -193,6 +193,7 @@ optional-dependencies.test = [
   "anthropic>=0.78",                                                 # For anthropic model tests; 0.78 introduced ThinkingConfigAdaptiveParam (required for Claude Opus 4.7).
   "anyio>=4.9,<5",
   "crewai[tools]; python_version>='3.11' and python_version<'3.12'", # For CrewaiTool tests; chromadb/pypika fail on 3.12+
+  "docker>=7",                                                       # For ContainerCodeExecutor tests
   "e2b>=2,<3",
   "gepa>=0.1",
   "google-antigravity>=0.1,<0.2",
--- a/src/google/adk/code_executors/container_code_executor.py
+++ b/src/google/adk/code_executors/container_code_executor.py
@@ -37,6 +37,15 @@ DEFAULT_IMAGE_TAG = 'adk-code-executor:latest'
 class ContainerCodeExecutor(BaseCodeExecutor):
   """A code executor that uses a custom container to execute code.
 
+  Security note: this executor runs model-generated code, which may be
+  influenced by untrusted input (e.g. via prompt injection). By default the
+  container is started with networking disabled and all Linux capabilities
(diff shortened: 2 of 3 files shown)


rohityan:
Hi @adilburaksen, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing unit tests before we can proceed with the review.

adilburaksen:
@rohityan fixed and rebased onto latest main. Summary of this push:

**Root cause of the failing tests:** the new test module imports `ContainerCodeExecutor`, whose module does a top-level `import docker`. `docker` was only in the `extensions` optional group, not the `test` group, so it wasn't installed in the unit-test env and collection failed with `ModuleNotFoundError: No module named 'docker'`.

**Changes:**
- Added `docker>=7` to the `test` dependency group (mirrors how `kubernetes` is provided for the GKE executor tests) so the module is importable in CI.
- Fixed the license-header year on the new test file (2025 → 2026), which clears `header-check`.
- Rebased onto current main, so the branch is no longer `BEHIND`.

**Testing plan / results** — ran the new tests locally with `docker` installed:
```
$ pytest tests/unittests/code_executors/test_container_code_executor.py -q
2 passed in 3.50s
```
The two tests assert that, by default, `containers.run` is called with `network_disabled=True`, `cap_drop=['ALL']`, and `security_opt=['no-new-privileges']`, and that `network_disabled=False` is honored when explicitly set.

PR #6079, opened 2026-06-11
Add ignore_args support for tool trajectory evaluation

### LINKED ISSUE #4794: Add ignore_args option to tool trajectory evaluation
## Problem

When evaluating tool trajectories, the current `TrajectoryEvaluator` always compares both tool **names** and **arguments** (`actual.args == expected.args`). This applies to all three match types: `EXACT`, `IN_ORDER`, and `ANY_ORDER`.

There are common scenarios where users care only that the correct tools were called (and optionally in the right order), but **not** about the exact arguments passed. For example:

- Tool arguments are non-deterministic (e.g., generated queries, timestamps)
- The eval focuses on verifying the agent's tool selection strategy, not argument correctness
- Arguments are complex objects where minor formatting differences cause false negatives

## Proposed Solution

Add an `ignore_args` boolean option to `ToolTrajectoryCriterion` that, when set to `True`, skips argument comparison entirely and only matches on tool names.

### Usage example

```json
{
  "tool_trajectory_avg_score": {
    "threshold": 0.8,
    "match_type": "IN_ORDER",
    "ignore_args": true
  }
}
```

### Scope of changes

1. **`ToolTrajectoryCriterion`** — add `ignore_args: bool = False` field
2. **`TrajectoryEvaluator`** — read the flag and conditionally skip arg comparison in all three match methods (`EXACT`, `IN_ORDER`, `ANY_ORDER`)
3. **Tests** — add coverage for `ignore_args=True` across all match types, including failure cases and dict-based config

allenjoshua16:

- Closes: #4794

**Problem:**

Tool trajectory evaluation currently compares both tool names and argument values when matching expected and actual tool calls.

This can cause valid evaluations to fail when tool arguments contain dynamic or non-deterministic values such as timestamps, generated queries, or other runtime-generated content.

**Solution:**

Added an `ignore_args` option to `ToolTrajectoryCriterion`.

When enabled, tool trajectory matching compares only tool names and ignores argument values. This behavior is supported across all existing match modes:

- `EXACT`
- `IN_ORDER`
- `ANY_ORDER`

Updated `TrajectoryEvaluator` to pass this configuration into the matching logic while preserving the existing behavior by default (`ignore_args=False`).

Added unit tests covering:
- `ignore_args` configuration loading
- matching behavior with different tool arguments
- existing matching behavior remaining unchanged

### Testing Plan

**Unit Tests:**

Tested with:

```bash
pytest tests/unittests/evaluation/test_trajectory_evaluator.py -q

### CODE REVIEW

**Comment by rohityan** [comment]
Hi @allenjoshua16 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing mypy-diff tests before we can proceed with the review.

**Comment by allenjoshua16** [comment]
Thanks for the feedback. I've addressed the mypy-diff failure and pushed an additional commit to the PR.

The issue was caused by mypy reporting a `no-any-return` error in `trajectory_evaluator.py`. I updated the implementation to return an explicit `bool` value and verified that the trajectory evaluator tests still pass locally:

```bash
pytest tests/unittests/evaluation/test_trajectory_evaluator.py -q
# 35 passed
```

Please let me know if there are any other issues I should address.

allenjoshua16:
Hi maintainers,

Just following up on this PR. I addressed the previously reported mypy-diff issue and pushed the requested fix. The PR checks now appear to be passing on my side.

Please let me know if there is any additional clarification or change needed from me. Whenever someone has time, I'd appreciate a review.

Thanks for your time and for maintaining the project.

rohityan:
Hi @allenjoshua16 , can you please resolve the conflicts.

i-yliu:
Closed with https://github.com/google/adk-python/pull/4795

PR #6100, opened 2026-06-13
fix: preserve single-turn structured output

### LINKED ISSUE #6089: message_as_output=True unconditionally set, destroying structured event.output for single_turn agents with output_schema
### Describe the Bug

When a `single_turn` `LlmAgent` is configured with an `output_schema`, the validated structured output is lost before it reaches downstream consumers.

The issue occurs because `process_llm_agent_output()` in `_llm_agent_wrapper.py` unconditionally sets:

```python
event.node_info.message_as_output = True
```

even when the agent is producing structured output through `output_schema`.

Later, `_consume_event_queue()` in `runners.py` interprets `message_as_output=True` as a signal that the message content itself represents the node output and clears `event.output` to avoid duplication.

As a result, the structured output stored in `event.output` is overwritten with `None`.

The implementation also appears to contradict the documented intent in `_consume_event_queue()`, whose comments explicitly describe `message_as_output` as applying only to the **"no output_schema"** case.

### Steps to Reproduce

1. Create an `LlmAgent` with:

   * `mode="single_turn"` (or use it as a workflow node)
   * an `output_schema` based on a Pydantic model

2. Execute the agent through a runner instance:

```python
runner = InMemoryRunner(agent=agent)

events = await runner.run_async(...)
```

or:

```python
runner = Runner(...)

events = await runner.run_async(...)
```

3. Inspect the emitted events.

4. Observe that:

```python
event.output
```

is always `None`, even when the model successfully produces valid structured output.

### Expected Behavior

When an `output_schema` is configured and validation succeeds:

```python
event.output
```

should contain the validated structured result.

Downstream consumers should be able to access the structured output through the event stream.

### Observed Behavior

For `single_turn` agents using `output_schema`:

```python
event.output
```

is always `None`.

The structured output is generated correctly but is later removed by `_consume_event_queue()` because `message_as_output` was set unconditionally.

### Environment Details

* **ADK Library Version:** Latest `main` branch
* **Desktop OS:** N/A (identified via static code analysis)
* **Python Version:** N/A

### Model Information

* **Using LiteLLM:** No
* **Model:** Any model used with `output_schema`

---

### Additional Context

The issue involves two locations.

#### 1. `src/google/adk/workflow/_llm_agent_wrapper.py`

`process_llm_agent_output()` sets:

```python
event.output = output
event.node_info.message_as_output = True
```

The flag is enabled regardless of whether the output is plain text or a validated structured result.

#### 2. `src/google/adk/runners.py`

`_consume_event_queue()` contains logic similar to:

```python
if not event.partial:
    if event.node_info.message_as_output and event.content is not None:
        event = event.model_copy()
        event.output = None
```

This removes the structured output before the event is returned.

### Documentation / Implementation Mismatch

The comment in `_consume_event_queue()` (approximately lines 788–794) states:

> When an LlmAgent node uses message_as_output (no output_schema), the wrapper sets both event.content and event.output to the same text. event.output is cleared here to avoid rendering the same value twice.

This comment explicitly limits the behavior to the **no output_schema** case.

However, the current implementation sets `message_as_output = True` regardless of whether an `output_schema` exists.

### Suggested Fix

Only mark the message as the output when no structured output schema is configured:

```python
event.output = output

if not agent.output_schema:
    event.node_info.message_as_output = True
```

### Minimal Reproduction

```python

class WeatherOutput(BaseModel):
    temperature: float
    condition: str

agent = LlmAgent(
    name="weather_agent",
    model="gemini-2.0-flash",
    output_schema=WeatherOutput,
    instruction="Return weather data as structured output.",
)

runner = InMemoryRunner(agent=agent)

events = await runner.run_async(
    user_id="test",
    session_id="test",
    new_message=types.Content(
        parts=[types.Part(text="What is the weather in Tokyo?")]
    ),
)

for event in events:
    print(event.output)  # Always None
```

### Frequency

* **Always reproducible (100%)**

he-yufeng:
Fixes #6089.

## Summary
- Keep `message_as_output` for plain single-turn agent text output.
- Leave structured `output_schema` results as `event.output` instead of marking them as message output.
- Add direct tests for both plain text and structured single-turn output handling.

## To verify
- `$env:PYTHONPATH=(Resolve-Path .\src).Path; python -m pytest tests\unittests\workflow\test_llm_agent_as_node.py -q`
- `python -m py_compile src\google\adk\workflow\_llm_agent_wrapper.py tests\unittests\workflow\test_llm_agent_as_node.py`
- `python -m ruff check src\google\adk\workflow\_llm_agent_wrapper.py`
- `git diff --check`

### DIFF
--- a/src/google/adk/workflow/_llm_agent_wrapper.py
+++ b/src/google/adk/workflow/_llm_agent_wrapper.py
@@ -350,7 +350,8 @@ def process_llm_agent_output(
     ctx.actions.state_delta[agent.output_key] = output
 
   event.output = output
-  event.node_info.message_as_output = True
+  if not agent.output_schema:
+    event.node_info.message_as_output = True
 
 
 async def run_llm_agent_as_node(
--- a/tests/unittests/workflow/test_llm_agent_as_node.py
+++ b/tests/unittests/workflow/test_llm_agent_as_node.py
@@ -39,6 +39,7 @@ from google.adk.tools.long_running_tool import LongRunningFunctionTool
 from google.adk.workflow import _llm_agent_wrapper as agent_wrapper
 from google.adk.workflow import START
 from google.adk.workflow._workflow import Workflow
+from google.adk.workflow._llm_agent_wrapper import process_llm_agent_output
 from google.adk.workflow.utils._workflow_graph_utils import build_node
 from google.genai import types
 from pydantic import BaseModel
@@ -171,6 +172,11 @@ def _new_workflow_runner(wf, test_name):
   return testing_utils.InMemoryRunner(app=app)
 
 
+async def _make_context(test_name: str, agent: LlmAgent) -> Context:
+  invocation_context = await create_parent_invocation_context(test_name, agent)
+  return Context(invocation_context, node=agent)
+
+
 # --- Validation ---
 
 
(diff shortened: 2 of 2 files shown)


rohityan:
Hi @he-yufeng , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing unit tests before we can proceed with the review.

he-yufeng:
@rohityan Done, and sorry for the long silence on your June ask. Rebased onto current main (the test file got a union merge with the branch-coverage tests landed since) and the workflow unit suite is green locally: 735 passed, 11 skipped. The one thing I could not replay is the original CI failure, since the June run logs are gone; if a specific test still fails on your side, name it and I'll fix that one directly.

he-yufeng:
Still applies on main: `process_llm_agent_output` sets `message_as_output=True` unconditionally, overriding what the caller asked for. Could a maintainer re-review when there's a chance?

PR #6118, opened 2026-06-14
fix: replace O(n^2) regex with linear string search in code block extraction (ReDoS)

### LINKED ISSUE #5992: Code executor: catastrophic O(n^2) regex backtracking in extract_code_and_truncate_content hangs on large delimiter-free responses
**Version**: google-adk 1.28.0 (the regex is unchanged on the 2.x line as well)

## What happens

When an `LlmAgent` has a `code_executor`, `_CodeExecutionResponseProcessor` runs `CodeExecutionUtils.extract_code_and_truncate_content` on **every** model response. The regex (in `google/adk/code_executors/code_execution_utils.py`):

```python
pattern = re.compile(
    rf'(?P<prefix>.*?)({leading_delimiter_pattern})(?P<code>.*?)({trailing_delimiter_pattern})(?P<suffix>.*?)$'.encode(),
    re.DOTALL,
)
pattern_match = pattern.search(response_text.encode())
```

is run via `search()` on the full joined response text. When the text contains **no** leading delimiter (e.g. a large structured/JSON response returned by a `before_model_callback` short-circuit, which still flows through the response post-processors), every start position rescans the whole string looking for the delimiter → **O(n²)**. On a few-hundred-KB response this stalls for tens of minutes — effectively a hang, and no per-turn timeout catches it because it is a synchronous CPU/regex call inside the worker thread.

## Reproduce

1. Attach any `BaseCodeExecutor` (e.g. `UnsafeLocalCodeExecutor`) to an `LlmAgent`.
2. Have the model (or a `before_model_callback`) emit a single text part of ~300KB that contains **no** ```` ```python ```` / ```` ```tool_code ```` fence (e.g. a large Plotly figure JSON).
3. The post-processor hangs in `pattern.search` while extracting code.

## Suggested fix

Before compiling/searching, short-circuit when no leading delimiter substring is present — the regex can only return `None` in that case, so this is behavior-preserving and makes the no-match path O(n):

```python
text_parts = [p for p in content.parts if p.text]
if text_parts:
    response_text = '\n'.join(p.text for p in text_parts)
    if not any(d[0] in response_text for d in code_block_delimiters):
        return None  # no opening delimiter -> no code block
```

(The default delimiters `'```tool_code\n'` and `'```python\n'` are literal substrings, so a plain `in` check is sufficient.)

Ashutosh0x:
## Summary
Fix for #5992 — Replace catastrophic O(n²) regex backtracking in `extract_code_and_truncate_content` with a linear-time string search.

## Problem
The regex pattern at line 153 of `code_execution_utils.py` uses multiple `.*?` groups with `re.DOTALL`:

\\\python
rf'(?P<prefix>.*?)({leading_delimiter_pattern})(?P<code>.*?)({trailing_delimiter_pattern})(?P<suffix>.*?)\$'
\\\

When the input is large and contains no matching delimiters (or partial delimiters), the regex engine tries all possible combinations of how the lazy quantifiers can match, causing **O(n²) backtracking** that hangs the process.

**CWE-1333:** Inefficient Regular Expression Complexity (ReDoS)

## Fix
Replaced the regex with a simple `str.find()`-based approach:
1. For each delimiter pair, find the first occurrence of the leading delimiter
2. Find the corresponding trailing delimiter after it
3. Pick the earliest match

This runs in **O(n × d)** time where d = number of delimiter pairs (typically 2-3), which is effectively O(n).

## Testing
The fix preserves the same behavior — extracting the first code block and truncating content after it. The string search approach handles the same edge cases:
- No delimiters found → returns None
- Empty code block → returns None
- Multiple code blocks → picks the earliest one

Fixes #5992


Ashutosh0x:
Hi @surajksharma07 — this fixes the ReDoS (CWE-1333) reported in #5992.

The regex in `extract_code_and_truncate_content()` uses multiple `.*?` groups with `re.DOTALL`, causing O(n²) backtracking on large inputs without matching delimiters. Replaced with a simple `str.find()` loop that runs in O(n) time.

Behavioral parity maintained — same inputs produce same outputs, just without the hang.

rohityan:
Hi @Ashutosh0x , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors before we can proceed with a review.

Issue #6099, opened 2026-06-13
Decision ledger: persist the authority, refusal, and approval behind each tool call


ainaomotayo:
### Is your feature request related to a specific problem?

Yes. ADK's observability answers what happened: OpenTelemetry spans for agent invocation, model calls, and tool execution; persisted session Events for function calls, responses, and confirmation actions; the BigQuery Agent
Analytics plugin streaming structured events at the runner level.

Incident response and compliance for production agents need a different artifact, what was decided:
- why this tool was selected for this step
- under which authority the call ran (agent-auth vs user-auth, and which principal)
- what a guardrail or plugin refused, and on what rule
- what confirmation was required, and what approval made the action permissible

Each of those decisions passes through an ADK surface (plugin callbacks, tool confirmation, per-tool auth config), but no component writes a unified record of them. A team with all five security layers configured still cannot reconstruct, from framework output alone, the moment an agent crossed from suggestion into action. 
Credit to Alex Shev for naming this in a community discussion of indirect prompt injection defenses: the layer most often missing is the decision ledger between policy and action.

### Describe the Solution You'd Like

An opt-in, runner-level decision ledger plugin (so every agent on a runner inherits it) that emits one structured, persistable entry per decision point: tool call, refusal, confirmation request, confirmation resolution. Deterministic fields (authority, policy outcome, confirmation payload) are attested by the framework; the model's stated rationale is recorded but typed as untrusted, since a hijacked agent will narrate a plausible justification. Schema and pseudo-code below.

To my knowledge, no framework persists this unified record today. In ADK all four decision surfaces already exist, so this is integration work, not new machinery.

### Impact on your work

I am a Google Developer Expert for AI producing developer education on ADK agent security. The recurring question from production teams in those discussions is not "how do I block the attack" but "how do I reconstruct what my agent decided when something gets through." Today my honest answer is that every team must build its own ledger from plugin callbacks, and none of those one-off schemas interoperate. A first-class answer would close the most common gap I see between teams configuring ADK's security layers and teams being able to audit them. No hard timeline; this is an ecosystem maturity request, not a personal blocker.

### Willingness to contribute

Yes. I am willing to contribute the plugin implementation and a schema RFC if maintainers are open to the direction.

---

### Describe Alternatives You've Considered

- Persisted session Events: capture function calls, responses, and confirmation actions with timestamps, but not the authority a tool ran under, the rule behind a refusal, or selection rationale. The ledger is the decision-semantics layer above the event log.
[...]

Issue #5913, opened 2026-05-30
Spanner search tool: validate structural SQL identifiers (defense-in-depth for prompt-injection-driven SQLi)


pk009900:
## Summary

`google.adk.tools.spanner.search_tool` interpolates several LLM-controllable
parameters into SQL via f-strings without validation:

- `table_name`, `columns`, `embedding_column_to_search`, `additional_filter`,
  `top_k` in `_generate_sql_for_knn` / `_generate_sql_for_ann`
- `spanner_googlesql_embedding_model_name` in `_generate_googlesql_for_embedding_query`
- `vertex_ai_embedding_model_endpoint` (inside `'...'`) in
  `_generate_postgresql_for_embedding_query`

These are not standard untrusted-input SQLi (Spanner's `execute_sql` is
single-statement; stacked queries fail at parse), but `similarity_search` is
registered as a `GoogleTool` in `SpannerToolset.get_tools()` (spanner_toolset.py:122),
so the LLM populates these parameters at runtime. A successful prompt injection
on the agent's input can lead the LLM to call the tool with malicious values,
enabling single-statement primitives like UNION-based read exfiltration and
INFORMATION_SCHEMA enumeration against any table the service account can SELECT.

## Threat model

```
prompt-injected user message
        │
        ▼
LLM populates similarity_search(table_name=..., columns=..., additional_filter=..., ...)
        │
        ▼
_generate_sql_for_knn renders the values into an f-string SQL template
        │
        ▼
snapshot.execute_sql(sql, params={embedding: ...})  -- structural identifiers unbound
```

The `query` parameter is correctly bound (`@query` / `$1`), but every structural
identifier above flows in unvalidated.

## Reproduction (single-statement primitives that work)

Each example below shows the LLM-emitted tool-call arguments and the resulting
SQL rendered by `_generate_sql_for_knn`. All three parse as a single Spanner
GoogleSQL statement and exercise the documented surface of `similarity_search`.

### 1. UNION-based exfiltration via `additional_filter`

```python
from google.adk.tools.spanner import search_tool
from google.cloud.spanner_admin_database_v1.types import DatabaseDialect

sql = search_tool._generate_sql_for_knn(
    dialect=DatabaseDialect.GOOGLE_STANDARD_SQL,
    table_name="documents",
    embedding_column_to_search="embedding",
    columns=["content"],
    additional_filter="1=1 UNION ALL SELECT password, 0.0 FROM admin_credentials",
    distance_type="COSINE",
[...]

PR #6130, opened 2026-06-15
Samples: add ATR security guardrail plugin

eeee2345:
This adds a sample plugin, contributing/samples/plugin/plugin_atr_guardrail/, that enforces an agent security policy globally across a Runner using the open-source Agent Threat Rules engine (pip install pyatr).

It demonstrates three fail-closed enforcement points, each returning a value that short-circuits the lifecycle:
- before_run_callback halts the run on a malicious user message (before any model call, so the block is observable without model credentials)
- before_model_callback skips the model call if the assembled prompt still carries a threat (defense in depth)
- before_tool_callback returns an error dict instead of executing a tool whose arguments match a rule

It is sample-only: no package dependency or public API change; pyatr is imported inside the sample, and the rule count is read at runtime rather than hard-coded.

Testing plan
- Plugins are not supported in adk web yet, so it runs via: python3 -m contributing.samples.plugin.plugin_atr_guardrail.main
- The injection prompt is halted at before_run_callback (no credentials needed); the benign prompt uses the model and needs standard ADK model credentials.
- Verified locally that all three callbacks fire correctly against the live engine (injection halts / skips / blocks; benign passes).

Local checks: pyink, isort (profile google), and mdformat-gfm all clean on the new files; Apache license headers present; no logging.getLogger(__name__) pattern. Motivated by maintainer interest in tool-call / prompt-injection defense (#5913, #6099).

Disclosure: I maintain the ATR project referenced here. This is a sample only and is not registered in the public plugins package.

### DIFF
--- a/contributing/samples/plugin/plugin_atr_guardrail/README.md
+++ b/contributing/samples/plugin/plugin_atr_guardrail/README.md
@@ -1,0 +1,66 @@
+# ADK Security Guardrail Plugin (Agent Threat Rules)
+
+This sample shows how to enforce a security policy across an entire ADK
+application with a single [Plugin](https://google.github.io/adk-docs/plugins/),
+backed by [Agent Threat Rules (ATR)](https://github.com/Agent-Threat-Rule/agent-threat-rules)
+— an open, MIT-licensed detection ruleset for AI-agent threats such as prompt
+injection, instruction override, and data exfiltration.
+
+A Plugin is registered once on the `Runner`, and its callbacks apply globally to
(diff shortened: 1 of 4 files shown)


eeee2345:
@googlebot I signed it!

rohityan:
Hi @eeee2345, Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Closing this PR here as it belongs to adk-python-community repo.
We highly recommend releasing the feature as a standalone package that we will then share through: https://google.github.io/adk-docs/integrations/

Issue #3111, opened 2025-10-08
Eval fails for non-English languages


lifnaja:
**Describe the bug**
When evaluating text in other languages (e.g., Thai, etc.), the eval logic incorrectly returns mismatches (Match score: 0)— even when the evaluated expression should be valid or identical in meaning.

**To Reproduce**

```
root_agent = Agent(
    name="sample_agent",
    model="gemini-2.5-flash",
    description="Greeting agent",
    instruction="""
    You are an assistant that will reply to the user with only the word 'สวัสดี'
    """,
)
```

**Expected behavior**
Eval should correctly handle input in other languages.

**Screenshots**

<img width="884" height="149" alt="Image" src="https://github.com/user-attachments/assets/a23e727a-6077-4733-8852-c14129971300" />

**Desktop (please complete the following information):**
 - OS: macOS
 - Python : 3.12.8
 - ADK version: 1.15.1

 **Model Information:**
 - Are you using LiteLLM: No
 - Which model is being used : gemini-2.5-flash

**Additional context**

hangfei:
Convert this to a feature request for more language support.

G26karthik:
I've submitted a PR to fix this issue: #3503 

The fix enhances the `FinalResponseMatchV2Evaluator` prompt to explicitly instruct the LLM judge to handle non-English languages including Thai, Chinese, Japanese, Korean, Arabic, and other non-Latin scripts.

**Changes:**
- Added clear i18n guidance to the evaluation prompt
- Instructed the evaluator to treat identical strings in ANY language as valid
- Added recognition for language-specific punctuation (。vs ., ؟ vs ?, etc.)

This also addresses #3162 (Chinese string evaluation issue) with the same fix.

Testing can be done by reproducing the original Thai "สวัสดี" test case - it should now return `score=1.0` instead of `score=0.0`.

ftnext:
I wrote the same issue https://github.com/google/adk-python/issues/4122 and maru0804-san proposed https://github.com/google/adk-python/pull/4143

guptaishaan:
Put together a fix in https://github.com/google/adk-python/pull/6401.

Yashas1600:
Put together a fix in https://github.com/google/adk-python/pull/6441.

guptaishaan:
Opened https://github.com/google/adk-python/pull/6443 with a fix for this.

guptaishaan:
Put together a fix in https://github.com/google/adk-python/pull/6444.

[...]

PR #6136, opened 2026-06-15
fix: ROUGE-1 eval returns 0 for non-English languages (ASCII-only tokenizer)

tcconnally:
## Problem

When evaluating text in non-Latin scripts (Thai, Chinese, Japanese, Arabic, etc.), the v1 ROUGE-1 evaluator returns scores of 0.0 even when the response matches the expected output exactly.

**Root cause:** The `rouge_score` library's default tokenizer uses `re.findall(r'\\w+', text)` which only matches ASCII `[a-zA-Z0-9_]`. Non-Latin characters produce zero tokens → ROUGE-1 score of 0.0 regardless of correctness.

## Reproduction (from #3111)

```python
agent = Agent(
    model="gemini-2.5-flash",
    instruction='Reply with only the word "สวัสดี"',
)
# Agent responds "สวัสดี" → ROUGE-1 score: 0.0 (should be 1.0)
```

## Fix

Added `_unicode_tokenize` function that:
1. Uses `re.UNICODE` flag for ASCII-majority text (preserves existing behavior)
2. Splits on Unicode whitespace/punctuation for non-ASCII text
3. Falls back to character-level tokens for scripts without word boundaries (Chinese, Japanese)

Closes #3111

### DIFF
--- a/src/google/adk/evaluation/final_response_match_v1.py
+++ b/src/google/adk/evaluation/final_response_match_v1.py
@@ -92,6 +92,30 @@ def _get_eval_status(score: float, threshold: float) -> EvalStatus:
   return EvalStatus.PASSED if score >= threshold else EvalStatus.FAILED
 
 
+def _unicode_tokenize(text: str):
+  """Tokenizes text using Unicode-aware word boundaries.
+
+  The default RougeScorer tokenizer uses r'\\w+' which only matches ASCII
+  [a-zA-Z0-9_]. For non-Latin scripts (Thai, Chinese, Japanese, Arabic, etc.),
+  this returns zero tokens, causing ROUGE scores of 0.0 on matching responses.
+
+  This tokenizer uses re.UNICODE for ASCII-majority text and falls back to
+  character-level tokenization for non-ASCII text.
+  """
+  import re
+  # For primarily non-ASCII text, tokenize by Unicode-aware patterns
+  ascii_chars = sum(1 for c in text if ord(c) < 128)
+  if ascii_chars > len(text) * 0.5:
+    return re.findall(r'[\\w]+', text.lower(), re.UNICODE)
+  # For non-Latin scripts, use whitespace splitting with Unicode support
+  tokens = re.split(r'[\\s\\p{P}]+', text, flags=re.UNICODE)
+  tokens = [t.lower() for t in tokens if t]
+  if tokens:
+    return tokens
+  # Character-level fallback for scripts without word boundaries
+  return list(text.lower())
+
+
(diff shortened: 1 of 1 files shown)


rohityan:
Hi @tcconnally , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Please fix formatting errors.

tcconnally:
Fixed the pre-commit formatting issue (pyink). Rebased on main.

tcconnally:
Pushed a follow-up commit that hardens the tokenizer — I found two issues in the previous version while validating it against the `rouge-score` library:

1. **The char-level fallback was unreachable.** It split non-ASCII text on whitespace first, but scripts without spaces (Chinese, Japanese, Thai) collapse to a single token, so `list(text)` never ran. Two different CJK strings that share characters scored `0.0` instead of partial credit — contradicting the docstring's promise of character-level tokens.
2. **Stemming was silently dropped.** Passing a custom `tokenizer=` makes `rouge-score` ignore `use_stemmer`, so existing English evals regressed (`"running"` no longer matched `"run"`).

The updated `_UnicodeTokenizer` now delegates ASCII-majority text to rouge-score's `DefaultTokenizer` (preserving Porter stemming and existing behavior exactly), and for non-ASCII text keeps Latin/digit runs as words while splitting remaining word characters individually.

Verified against the library:

| Case | Before | After |
|---|---|---|
| Thai exact (`สวัสดี`) | 1.0 | 1.0 |
| CJK exact (`你好世界`) | 0.0 | 1.0 |
| CJK partial (`你好世界` vs `你好朋友`, share 你好) | 0.0 | 0.5 |
| English stemming (`running fast` vs `run fast`) | 0.5 | 1.0 |
| Latin sanity (`the cat sat` vs `the cat`) | — | matches default tokenizer |

Ready for another look, @wyf7107 — thanks for your patience.

tcconnally:
Withdrawing this — the fix doesn't actually run. It passes a plain function as `tokenizer=` to `RougeScorer`, but the scorer calls `self._tokenizer.tokenize(text)`; a bare function has no `.tokenize`, so every ROUGE eval (English included) would raise `AttributeError`. It also uses `\p{P}` in `re.split`, which Python's built-in `re` module doesn't support (raises `re.error: bad escape`).

The underlying bug is real (the default tokenizer strips non-Latin script → ROUGE 0.0 for Thai/CJK), but the fix needs to be a proper `Tokenizer` subclass exposing `.tokenize`, using the `regex` module or explicit Unicode ranges, with tests across scripts. Apologies for the premature PR — closing, and I'll rework it properly if we pursue it.

Issue #5889, opened 2026-05-29
Feature request: BigQueryLoggerConfig.export_internal_spans flag to disable plugin span export


rannn505:
**Context**

`BigQueryAgentAnalyticsPlugin` (current as of ADK 1.34.1) emits internal spans `agent`, `llm_request`, `tool` via its `TraceManager`. Per the source docstring at `google/adk/plugins/bigquery_agent_analytics_plugin.py:642-647`, these spans intentionally do NOT attach to ambient OTel context to avoid corrupting framework span hierarchy.

**Problem**

When a global `BatchSpanProcessor` is configured for an external exporter (e.g., Cloud Trace), those internal spans STILL get exported — `SpanProcessor.on_start` / `on_end` fire for every span the tracer creates, regardless of context attachment.

Result in Cloud Trace UI:
- Generic spans `agent`, `llm_request`, `tool` (no agent name) appear alongside the readable ADK spans `invoke_agent {name}` / `execute_tool {name}`
- Because they don't share OTel parent context, they appear as orphan top-level spans with "(Missing span ID xxxx)" parent references
- Hard to read

**Current workarounds**

1. Subclass `SpanProcessor` to drop spans with `instrumentation_scope.name == "google.adk.plugins.bigquery_agent_analytics"` at export time. Works but burns CPU creating spans we then discard, and requires maintenance across ADK bumps
2. Disable the plugin entirely — loses the analytics writes which is the plugin's primary purpose

**Proposal**

Add a config flag on `BigQueryLoggerConfig`:

```python
class BigQueryLoggerConfig:
    ...
    export_internal_spans: bool = True
    """If False, plugin's internal spans (agent/llm_request/tool) are recorded for BigQuery row generation
    but NOT exported via the global TracerProvider's SpanProcessors. Useful when an external trace
    backend like Cloud Trace already receives readable spans from ADK's own _instrumentation.py."""
```

Implementation: gate the `tracer.start_span(...)` calls in `TraceManager.push_span` to use a tracer whose spans bypass external SpanProcessors when the flag is False. One approach: maintain a separate `TracerProvider` internal to the plugin with no SpanProcessor for export — only feeds the plugin's own `_SpanRecord` stack.

**Related issues**

- #5722 — contextvars leak across async generator boundaries (root cause of "Calling end() on an ended span" warnings this plugin also surfaces)
- #5310 — orphan parent_span_id (partially addressed but still visible with this plugin)
[...]

PR #6137, opened 2026-06-16
fix(models): gate Gemini cache creation on cacheable prefix tokens

### LINKED ISSUE #5847: gemini_context_cache_manager.create gate uses request total tokens, not cacheable prefix tokens
## Summary

`GeminiContextCacheManager._create_new_cache_with_contents` decides whether to call `caches.create()` using a token count that represents the **full request** (system_instruction + tools + all chat history + user turn), but the actual `caches.create()` call only sends the **cacheable prefix** (system_instruction + tools + `contents[:cache_contents_count]`).

This mismatch causes `caches.create()` to fail with `400 INVALID_ARGUMENT` ("The cached content is of <N> tokens. The minimum token count to start caching is 4096.") on every cold-prefix request whose chat history happens to push the full token count over 4096 while the prefix itself stays below.

Worse, on failure the manager returns `CacheMetadata(fingerprint=..., contents_count=..., cache_name=None)` (no `cache_name`), so the next request with the same fingerprint hits the same failure path and tries to create again. Infinite create-fail loop, zero cache hits, flood of `Failed to create cache: 400 INVALID_ARGUMENT` warnings.

## Affected version

- `google-adk == 1.34.1`

## Environment

- Model: `gemini-3.5-flash` on Vertex AI
- `ContextCacheConfig(cache_intervals=10, ttl_seconds=1800, min_tokens=4096)`
- App-level config (`App.context_cache_config`)
- ~24h prod observation: hundreds of failures/min, `cache_hit_pct ~= 0%` across all routed models

## Reproduction (code path walkthrough)

1. `google/adk/flows/llm_flows/context_cache_processor.py:81`
   ```python
   llm_request.cacheable_contents_token_count = previous_token_count
   ```
   `previous_token_count` is the **prior response's** `usage_metadata.prompt_token_count` — i.e. the full prompt the model just saw (incl. chat history + tools + system_instruction + the latest user turn).

2. `google/adk/models/gemini_context_cache_manager.py:312-324` (gate inside `_create_new_cache_with_contents`):
   ```python
   if cacheable_contents_token_count < ctx_cache_config.min_tokens:
       logger.info("Previous request too small to cache: ...")
       return None
   if cacheable_contents_token_count < _GEMINI_MIN_CACHE_TOKENS:
       logger.info("Token count below Gemini minimum: ...")
       return None
   ```
   This uses the full prompt count from step 1 as the gate.

3. `google/adk/models/gemini_context_cache_manager.py:372-441` (`_create_gemini_cache`):
   ```python
   cache_contents = llm_request.contents[:cache_contents_count]   # line 388
   ...
   cache_config = CreateCachedContentConfig(
       system_instruction=llm_request.config.system_instruction,
       tools=llm_request.config.tools,
       contents=cache_contents,
       ...
   )
   await self.genai_client.aio.caches.create(model=..., config=cache_config)   # line 423
   ```
   The actual `caches.create()` call only sends the prefix slice. The server measures *that* prefix, finds it < 4096, and rejects with 400.

4. On failure (`gemini_context_cache_manager.py:~133`), the manager returns:
   ```python
   CacheMetadata(fingerprint=..., contents_count=cache_contents_count)
   ```
   with no `cache_name`. The next request matches the same fingerprint, sees no `cache_name`, and goes back through `_create_new_cache_with_contents` — same gate, same failure.

## Observed log spam (production)

```
WARNING Failed to create cache: 400 INVALID_ARGUMENT.
{'error': {'code': 400, 'message': 'The cached content is of 1820 tokens. The minimum token count to start caching is 4096.', 'status': 'INVALID_ARGUMENT'}}
```

Token counts in real failures: 1820, 1857, 1991, 1829 — all well below 4096 because the prefix (system_instruction + tools + few-content sub-agents) is small even when the full chat history pushes the gate's measurement above 4096.

## Suggested fix

The gate must be measured on the same payload `caches.create()` will receive. Two options:

1. **Estimate prefix tokens at gate time.** Compute `system_instruction + tools + contents[:cache_contents_count]` token count locally and compare against `min_tokens` / `_GEMINI_MIN_CACHE_TOKENS`. This avoids the extra `count_tokens` API round trip if you can use the on-device tokenizer; otherwise one `client.models.count_tokens` call is cheap relative to letting the request fail at the server.
2. **Cache the negative result by fingerprint.** Even with the gate fix, transient API rejections still happen; refusing to retry the same fingerprint for `ttl_seconds` would prevent the infinite-loop log flood.

Either fix on its own breaks the loop; both together would be ideal.

## Workaround for users

Set `App.context_cache_config = None` to disable caching entirely until a fix lands. There is no per-agent disable in 1.34.1 — `App.context_cache_config` is global (`llm_agent.py:266`), and the only safe workaround is full disable.

nyxst4ck:

- Closes: #5847

**Problem:**

`GeminiContextCacheManager._create_new_cache_with_contents` decides whether a
request is large enough to cache by comparing
`llm_request.cacheable_contents_token_count` against Gemini's
`_GEMINI_MIN_CACHE_TOKENS` (4096). But `cacheable_contents_token_count` is set
(in `context_cache_processor.py`) to the **previous response's
`prompt_token_count`** — i.e. the token count of the *entire* previous prompt
(system instruction + tools + every content, including the trailing user turn).

The cache, however, only stores the **prefix** `contents[:cache_contents_count]`
(plus system instruction and tools — see `_create_gemini_cache`), where
`cache_contents_count` excludes the last continuous batch of user contents.

On a long conversation the full-prompt count can clear the 4096 gate while the
actual cached prefix is far below it, so `caches.create()` is invoked with a
sub-minimum payload and Gemini returns **`400 INVALID_ARGUMENT`**. The reporter
confirmed this in production, and a maintainer confirmed the root cause in the
issue thread.

**Solution:**

Gate cache creation on the size of the prefix that is *actually* cached rather
than on the full previous prompt.

Since the only accurate token count we have is for the whole prompt, the new
`_estimate_cacheable_prefix_tokens` scales that accurate count by the prefix's
estimated share of the request (`_estimate_request_tokens` is reused — now
accepting an optional `cache_contents_count` — to estimate both the full request
and the prefix). When the prefix already spans the whole request the scale
factor is `1.0` and the accurate count is used unchanged, so existing behavior
is preserved; when the prefix is a strict subset the estimate shrinks
accordingly and a too-small prefix is correctly skipped before any API call.

Scope note: this fixes the `400 INVALID_ARGUMENT` (the reported bug). The
separate latency concern discussed later in the thread (synchronous
`caches.create()` / cleanup on the request path, tracked in #5889) is
intentionally out of scope. The user-configured `min_tokens` gate is left on the
full-prompt count, since the 4096 prefix gate is what prevents the invalid API
call; happy to extend `min_tokens` to the prefix too if maintainers prefer.

### Testing Plan

**Unit Tests:**

Added `test_create_cache_gates_on_prefix_not_full_prompt`: a tiny cacheable
prefix followed by a huge trailing user turn with
`cacheable_contents_token_count = 75000`. It asserts no cache is created and
`caches.create` is never called. This test is **red on `main`** (`caches.create`
is called once — the 400-inducing path) and **green with this change**.

Updated `test_fingerprint_only_metadata_transitions_to_active_cache` to use a
realistically large cacheable prefix, since its previous setup relied on the
buggy assumption that a tiny prefix would still be cached because the injected
full-prompt count cleared the gate.

```
$ pytest tests/unittests/agents/test_gemini_context_cache_manager.py \
         tests/unittests/agents/test_context_cache_config.py \
         tests/unittests/flows/llm_flows/test_context_cache_processor.py -q
59 passed
```

The fix is exercised entirely through the unit test above (it mocks
`genai_client.aio.caches.create` and asserts it is not called for a
below-minimum prefix), which mirrors the production failure mode reported in
#5847 (`400 INVALID_ARGUMENT` from `caches.create` on a long conversation).

### Checklist


nyxst4ck:
@googlebot I signed it!

rohityan:
Hi @nyxst4ck ,Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

Issue #206, opened 2025-04-15
Multimodal response generation support


inspire-boy:
Some models support generate image in context. 
## Google doc
Documentation for Gemini: https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal-response-generation
## Model
- gemini-2.0-flash-exp
- gemini-2.0-flash-exp-image-generation
## Request raw content sample
```
{
  "contents": [
    {
      "parts": [
        {
          "text": "Create a image of a red cat."
        }
      ],
      "role": "user"
    }
  ],
  "generationConfig": {
    "responseMimeType": "text/plain",
    "responseModalities": [
      "Text",
      "Image"
    ]
  }
}
```
## Response raw content sample
```
{
      "content": {
        "parts": [
          {
            "inlineData": {
              "mimeType": "image/png",
              "data": "[------IMAGE BASE64------]"
            }
          }
        ],
        "role": "model"
      },
      "index": 0
    },
}
```

hangfei:
What's your request or question?

inspire-boy:
> What's your request or question?
Thank your reply.I want to use multimodal output to generate images, does adk currently have this capability and does it require any special configuration?
![Image](https://github.com/user-attachments/assets/e514d0fe-e351-4d43-be6d-78b7259f67fb)

boyangsvl:
I think this should work already as long as you:
* use the right model
* set text and image as response modalities

Please reopen the issue if it doesn't work.

inspire-boy:
@boyangsvl 
I use the custom app code example:https://google.github.io/adk-docs/get-started/quickstart-streaming/#5.-build-custom-app
Just appened "IMAGE" item in ***response_modalities***
![Image](https://github.com/user-attachments/assets/e41f7bc0-a794-47bf-9f8a-d89e2a1ea097)
Then execute:`uvicorn main:app --reload` start a service.
Error Trace log:
```
INFO:     Will watch for changes in these directories: ['D:\\py\\adk\\demo\\adk_streaming\\app']
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [103584] using StatReload
INFO:     Started server process [79244]
INFO:     Waiting for application startup.
INFO:     Application startup complete.   
INFO:     ('127.0.0.1', 61876) - "WebSocket /ws/85863365" [accepted]
Client #85863365 connected
[...]

PR #6140, opened 2026-06-16
feat(bigquery-analytics): hot-path same-session pause_orphan resolution

caohy1988:
## What

Resolves `pause_orphan` on the long-running `TOOL_COMPLETED` row (emitted when a paused tool resumes via a `function_response` in a user message) from the **in-memory session history only**.

`_resolve_same_session_pause_orphan(invocation_context, function_call_id)`:
- **`False`** — the originating paused call is present in this session's in-memory history → definitely not an orphan.
- **`None`** — unknown / not yet settled: no id to pair on, no history, or the pause wasn't found in the (possibly trimmed) in-memory history.

The emit path stamps `attributes.adk.pause_orphan` **only when the resolver returns a real bool**; on `None` the key is omitted so consumers read it as SQL NULL ("not yet determined").

## Why this shape (hot-path-safe split)

This is the resume path. The resolver is a **zero-I/O** in-memory scan: it adds no latency and introduces **no session-service or BigQuery read** into a plugin that is otherwise write-only.

The resolver **never returns `True`**. Declaring a true orphan requires a *settled* reconciliation — an off-hot-path session-service / BigQuery read after a settling delay — which is intentionally **out of scope here**. Adding a 30s settling sleep + a BQ read to the resume callback would be the wrong product tradeoff (resume latency) and a real architectural change (write-only producer → producer that also reads). A persistent `None` is exactly what that future fallback upgrades to `True`. Follow-up tracked separately (see #206 "settled pause-orphan reconciliation").

## Tests

`TestPauseOrphanSameSession`: resolver returns `False` only on a found pause; `None` for not-found / empty / missing-id / no-session / id-without-matching-part; never `True`; and does no `session_service` read (an exploding session_service proves it). Two emit-path tests assert `pause_orphan=False` when the pause is in history and the key is omitted otherwise.

> **Note on local verification:** I was unable to run the suite in my local environment — it's offline and the checked-out `main` requires a newer `opentelemetry-semconv` (`GEN_AI_TOOL_DEFINITIONS`) than is installed, and `uv sync` is blocked by an unrelated platform-incompatible dep (`google-antigravity`, no macOS-x86_64 wheel). Both changed files are syntax-validated; **CI is the verification gate for the test run.** Happy to iterate on any setup detail CI surfaces.

### DIFF
--- a/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
+++ b/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
@@ -3241,6 +3241,53 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
 
   # --- UPDATED CALLBACKS FOR V1 PARITY ---
 
+  def _resolve_same_session_pause_orphan(
+      self,
+      invocation_context: "InvocationContext",
+      function_call_id: Optional[str],
+  ) -> Optional[bool]:
+    """Same-session resolution of ``pause_orphan`` for a long-running resume.
+
+    Scans the in-memory session history for the originating paused
+    function call. This is a hot-path-safe, zero-I/O check: it never
+    issues a session-service or BigQuery read, so it adds no latency to
+    the resume path.
+
+    Returns:
+      * ``False`` — the originating paused call is present in this
+        session's in-memory history, so this completion is definitely
+        not an orphan.
+      * ``None`` — unknown / not yet settled: there is no id to pair on,
+        no session history is available, or the pause was not found in
+        the (possibly trimmed) in-memory history. Callers omit the
+        ``pause_orphan`` attribute in this case so consumers read it as
+        SQL NULL ("not yet determined").
+
+    Never returns ``True``. Declaring a true orphan requires the settled
+    fallback (an off-hot-path session-service / BigQuery reconciliation),
+    which is intentionally out of scope here so the resume path stays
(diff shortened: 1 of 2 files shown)


caohy1988 (commented):
Findings: none.

I reviewed this against the locked #206 split decision. The PR keeps the resume path hot-path-safe: no added `sleep`, no `get_session` / session-service read, no BigQuery read/query, and `_resolve_same_session_pause_orphan()` never returns `True`. The only positive resolution is `False` when the same session history contains a matching `TOOL_PAUSED` origin (`function_call.id` present in `event.long_running_tool_ids`), and unknown cases omit `attributes.adk.pause_orphan` so consumers see SQL NULL. That matches the agreed first PR scope; the settled orphan reconciliation remains correctly deferred.

I also checked that the pause producer and resolver use the same pair key, the emit path preserves existing `pause_kind` / `function_call_id`, and the new tests cover found, not found, no session/history/id, id-without-matching-part, no session-service read, emit false, and emit omitted. `git diff --check` is clean.

One operational note: the PR is currently `BEHIND` `origin/main`. A merge-tree check against current main shows no conflict in the touched files, but it should be synced/rebased before final merge so required checks rerun on the current base. Local focused pytest is blocked on this machine by an unrelated OpenTelemetry semconv import mismatch in the adk-python environment, so I relied on the diff/contract checks plus GitHub checks for execution.

PR #6157, opened 2026-06-18
feat: bump google-genai dependency to <3

wukath:
Support history_config parameter in LiveConnectConfig. Upgraded google-genai upper bound to <3 to allow v2 releases (the history_config change is in v2.9.0rc1).


wuliang229:
Another change was merged to v1 branch already.

PR #6175, opened 2026-06-21
fix(openapi): skip non-numeric response keys in generate_return_doc

### LINKED ISSUE #6174: OpenAPI tool: generate_return_doc crashes on non-numeric response keys (default, NXX)
**Describe the bug**

`PydocHelper.generate_return_doc` builds its sorted response list with
`sorted(responses.items(), key=lambda item: int(item[0]))`. OpenAPI 3.x allows
non-numeric keys in the Responses Object — `default` and the range patterns
`1XX`/`2XX`/`3XX`/`4XX`/`5XX`. When any of those keys is present, `int(item[0])`
raises `ValueError`, so generating a tool from an otherwise valid OpenAPI spec
crashes.

**To Reproduce**

```python
from fastapi.openapi.models import Response, Schema, MediaType
from google.adk.tools.openapi_tool.common.common import PydocHelper

responses = {
    "200": Response(description="OK", content={"application/json": MediaType(schema=Schema(type="string"))}),
    "default": Response(description="Error response"),
}
PydocHelper.generate_return_doc(responses)
# ValueError: invalid literal for int() with base 10: 'default'
```

**Expected behavior**

Non-numeric response keys should be ignored for return-type hinting (the code
already keeps only `2xx` responses with content on the very next line), so the
function should return the doc for the `200` response instead of crashing.

**Desktop**
- google-adk 2.1.0 and later; still present on `main` as of 2026-06-10

**Additional context**

`operation_parser` already filters `startswith('2')` before consuming the keys;
applying the same guard here keeps the two paths consistent.

garyzava:
## Summary

`PydocHelper.generate_return_doc` sorts the OpenAPI responses with
`int(item[0])`, which crashes with `ValueError: invalid literal for int() with
base 10` on valid non-numeric response keys like `default` or range patterns
(`2XX`, `4XX`). This skips non-numeric keys in the sort so a tool can be
generated from such specs.

Fixes #6174

## What changed

* `src/google/adk/tools/openapi_tool/common/common.py` - exclude non-numeric
  keys from the sort in `generate_return_doc`.
* `tests/unittests/tools/openapi_tool/common/test_common.py` - add regression
  tests for a `default` key, a range key, and a responses dict with only
  non-numeric keys.

## Why

The line after the sort already keeps only responses whose key starts with `2`
and that have content, so non-numeric keys (`default`, `1XX`-`5XX`) are never
used anyway. Dropping them from the sort removes the crash without changing any
valid behavior. This matches `operation_parser`, which filters `startswith('2')`
before using the keys.

## How I tested

```
$ pytest tests/unittests/tools/openapi_tool/common/test_common.py -k generate_return_doc -q
9 passed
```

The three new tests fail before the fix with
`ValueError: invalid literal for int() with base 10: 'default'` (and `'4XX'`)
and pass after. `pyink` and `isort` report no changes on both files.

## Backward compatibility

No API change. Specs with only numeric response keys behave exactly as before;
specs with `default`/range keys now work instead of crashing.


rohityan:
Hi @garyzava, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

PR #6186, opened 2026-06-22
feat(utils): add opt-in Jinja2 instruction templating via use_jinja2 flag

### LINKED ISSUE #2942: [Feat]: Introduce Jinja2-based Instruction Templating with a Feature Flag
This issue proposes enhancing the inject_session_state function by introducing Jinja2-based templating engine. 

The current regex-based approach is limited, lacking support for control flow like conditionals and loops. A new, dedicated function `_render_with_jinja2` will be created to handle Jinja2 rendering.

A feature flag, `use_jinja2`, will be added to the public inject_session_state function, allowing users to explicitly choose which engine to use during a transition period. This change maintains backward compatibility while enabling more complex and dynamic instruction templates for future development.

## Motivation

The existing regex-based templating may be insufficient for advanced use cases, for examples 

* Dynamically include or exclude parts of an instruction based on session state.
* Apply formatting to injected variables using filters.

While some of these features could be implemented via callbacks, native Jinja2 support provides a cleaner, more maintainable solution. Jinja2 also provides a robust and well-documented solution for all these needs, making agent instructions more expressive and adaptable.

## Proposed Changes

* Refactor inject_session_state: Extract the current logc into a private function `_render_with_regex`, 
add `_render_with_jinja2`, A new, private asynchronous function will be implemented to handle Jinja2 rendering. This function will set up a Jinja2 environment and pass session state and artifact accessors to the template.

* Introduce use_jinja2 Flag: The public inject_session_state function will be updated to accept a boolean argument, use_jinja2. If True, it will call _render_with_jinja2; otherwise, it will default to _render_with_regex.

* Update Documentation: The inject_session_state docstrings and the project's contributing guide will be updated to explain the new use_jinja2 flag and the benefits of the Jinja2 templating syntax.

## Expected Impact
No breaking changes: Existing codebases that do not use the use_jinja2 flag will continue to function as before.

## Testing Plan
Unit Tests: New unit tests will be added to cover the `_render_with_jinja2` function, ensuring it correctly handles:

* Basic variable substitution ({{ var_name }}).
* Conditional logic ({% if ... %}).
* For loops ({% for ... %}).
* Asynchronous artifact loading ({{ await artifact(...) }}).

vaibhav-patel:
## Summary

Adds an opt-in `use_jinja2` flag to `inject_session_state` (`src/google/adk/utils/instructions_utils.py`). When enabled, instruction templates are rendered with a sandboxed Jinja2 environment, unlocking control flow (`{% if %}`, `{% for %}`) and filters in addition to the existing `{var}` substitution. The default (`use_jinja2=False`) is fully backward compatible.

## Changes

- Extract the existing regex logic unchanged into `_render_with_regex`.
- Add `_render_with_jinja2`, which renders via `jinja2.sandbox.SandboxedEnvironment(enable_async=True)` and exposes:
  - `state` — the session state mapping (`{{ state['var'] }}`).
  - `artifact` — an async accessor (`{{ artifact('file') }}`), auto-awaited by the async environment; missing artifacts render as empty string.
- Import `jinja2` lazily so the base install isn't forced to depend on it; add a `jinja` optional extra (`pip install google-adk[jinja]`).
- Use a sandboxed environment because instruction templates may carry user/session data.

## Notes for reviewers

- **New dependency surface:** `jinja2` is not a core dependency today (only in the `eval`/`test` extras). I added a dedicated `jinja` optional extra rather than touching core deps. Happy to fold it into core deps or another extra if you prefer.
- The original feature request showed `{{ await artifact(...) }}`; that syntax is invalid in Jinja2. With `enable_async=True` the coroutine is awaited automatically, so the supported syntax is `{{ artifact('file') }}`.

## Testing

Added unit tests for variable substitution, conditionals, loops, filters, artifact loading, the uninitialized-artifact-service error, sandbox rejection of unsafe attribute access, and an explicit check that the default path is unchanged. All `tests/unittests/utils/test_instructions_utils.py` pass (24).

Fixes #2942.


vaibhav-patel:
@googlebot I signed it!

vaibhav-patel:
HI @rohityan 
The CI is fixed.

rohityan:
Hi @vaibhav-patel , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

GWeale:
this landed on main via #6593 and shipped in v2.7.0, so use_jinja2 is already on inject_session_state and this branch conflicts now. one thing your version has that the merged one doesn't: it renders through jinja2.sandbox.SandboxedEnvironment where main uses a plain jinja2.Environment, so a template can still reach __class__ and walk the mro on 2.8.0. a small follow-up pr swapping the sandbox in would be welcome.

PR #5636, opened 2026-05-07
fix(models): surface error when model returns STOP with empty content

Oppong08:
Tighten LlmResponse.create() so a Gemini candidate with empty parts and finish_reason=STOP no longer passes through as a successful empty response. It now routes to the error branch with error_code='MODEL_RETURNED_NO_CONTENT' and a descriptive error_message, so callers see an actionable error event instead of a silent empty final agent output. Reproduces against gemini-2.5-flash-lite when the second turn after a tool call returns zero output tokens.

Also broadens the skip-empty guard in BaseLlmFlow._postprocess_async to treat Content(parts=[]) as no-content (defense in depth) and updates the two existing tests that codified the old behavior.

**Please ensure you have read the [contribution guide](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md) before creating a pull request.**

### Link to Issue or Description of Change

**1. Link to an existing issue (if applicable):**

- Closes: #5631

**2. Or, if no issue exists, describe the change:**

**Problem:**

With `gemini-2.5-flash-lite` and an `LlmAgent` that calls a tool, the run can sometimes terminate with `final_output: ""`.

The reported flow is:

1. The model returns a `function_call`, such as a `python_executor` tool call.
2. ADK executes the tool successfully and emits the function-response event.
3. The follow-up model response returns `Content(role="model", parts=[])` with `finish_reason=STOP` and zero output tokens.
4. ADK treats that empty model response as the final event, causing the agent's final output to become an empty string.

This happened because `LlmResponse.create()` accepted `finish_reason=STOP` as a successful response even when `content.parts` was empty. In addition, the skip-empty guard in `BaseLlmFlow._postprocess_async` only checked whether `llm_response.content` existed, so a `Content` object with `parts=[]` could still pass through as a final response.

**Solution:**

This PR tightens `LlmResponse.create()` so a Gemini candidate with empty parts and `finish_reason=STOP` no longer passes through as a successful empty response.

Instead, it routes to the error branch with:

- `error_code="MODEL_RETURNED_NO_CONTENT"`
- a descriptive `error_message`

This gives callers an actionable error event instead of a silent empty final agent output.

This PR also broadens the skip-empty guard in `BaseLlmFlow._postprocess_async` to treat `Content(parts=[])` as no content unless an error is present. This acts as defense in depth and prevents empty content objects from being emitted as meaningful final responses.

[...]

PR #5512, opened 2026-04-28
fix(models): populate finish_reason on AnthropicLlm responses

sebastienc:
## Summary

Fixes #5394

_Disclaimer: I'm using Claude Code to generate the following text. I find the summary is better than what I would've wrote._

`AnthropicLlm` was calling `to_google_genai_finish_reason()` but never wiring its result into `LlmResponse`, so `finish_reason` was always `None`. This fix connects it end-to-end and expands the mapping to cover all stop reasons the current Anthropic API can return.

- **`to_google_genai_finish_reason`** — extended to cover `pause_turn` → `STOP` (streaming interruption, e.g. server-side tool use) and `refusal` → `SAFETY` (model refusal); all other unknown values continue to return `FINISH_REASON_UNSPECIFIED`
- **`message_to_generate_content_response`** — passes `stop_reason` through `to_google_genai_finish_reason` and sets it on the returned `LlmResponse`
- **`_generate_content_streaming`** — reads `stop_reason` from the `message_delta` event and sets `finish_reason` on the final aggregated `LlmResponse`

## Testing plan

10 new tests added to `tests/unittests/models/test_anthropic_llm.py`. Full suite: 69 passed.

```
tests/unittests/models/test_anthropic_llm.py::test_to_google_genai_finish_reason[end_turn-STOP] PASSED
tests/unittests/models/test_anthropic_llm.py::test_to_google_genai_finish_reason[stop_sequence-STOP] PASSED
tests/unittests/models/test_anthropic_llm.py::test_to_google_genai_finish_reason[tool_use-STOP] PASSED
tests/unittests/models/test_anthropic_llm.py::test_to_google_genai_finish_reason[max_tokens-MAX_TOKENS] PASSED
tests/unittests/models/test_anthropic_llm.py::test_to_google_genai_finish_reason[pause_turn-STOP] PASSED
tests/unittests/models/test_anthropic_llm.py::test_to_google_genai_finish_reason[refusal-SAFETY] PASSED
tests/unittests/models/test_anthropic_llm.py::test_to_google_genai_finish_reason[None-FINISH_REASON_UNSPECIFIED] PASSED
tests/unittests/models/test_anthropic_llm.py::test_non_streaming_sets_finish_reason PASSED
tests/unittests/models/test_anthropic_llm.py::test_non_streaming_finish_reason_max_tokens PASSED
tests/unittests/models/test_anthropic_llm.py::test_streaming_sets_finish_reason PASSED
```

Issue #5394, opened 2026-04-19
AnthropicLlm never populates finish_reason on LlmResponse


sebastienc:
`AnthropicLlm` has a correctly implemented `to_google_genai_finish_reason()` helper but it is commented out and never called, so `LlmResponse.finish_reason` is always `None` — for both streaming and non-streaming responses.

## Current behaviour

Non-streaming (`message_to_generate_content_response`):
```python
# TODO: Deal with these later.
# finish_reason=to_google_genai_finish_reason(message.stop_reason),
```

Streaming: the final `LlmResponse` is yielded with no `finish_reason`. The `message_delta` event captures `output_tokens` but ignores `delta.stop_reason`.

This means agent code checking `response.finish_reason == FinishReason.MAX_TOKENS` (or any other reason) will never trigger for Anthropic responses — a token-limit cutoff looks identical to a normal completion.

## Expected behaviour

- Non-streaming: uncomment and call `to_google_genai_finish_reason(message.stop_reason)`
- Streaming: capture `event.delta.stop_reason` in the `message_delta` handler and set it on the final `LlmResponse`
- Add the two missing mappings:
  - `pause_turn` → `STOP` (used when extended thinking pauses a turn)
  - `refusal` → `SAFETY` (model declined the request)

Full Anthropic `stop_reason` values: `end_turn`, `max_tokens`, `stop_sequence`, `tool_use`, `pause_turn`, `refusal`.

sebastienc:
I'd like to work on this. Can I be assigned?

llalitkumarrr:
Hello @sebastienc ,

Thank you for your interest in contributing. Do you already have a solution in mind? If so please feel free to submit a PR. Once it passes all checks and is approved by reviewer, it will be merged.

sebastienc:
Yes. I can submit a PR. Thanks.

PR #6195, opened 2026-06-23
fix(models): surface empty LiteLlm streaming completions as error event

kevin-hs-sohn:
## Why

Streaming completions where the provider returns a `finish_reason` but no text + no tool calls currently produce **zero** yielded `LlmResponse` events: `aggregated_llm_response` only gets set when `(text or reasoning_parts)` is truthy, and `aggregated_llm_response_with_tool_call` needs a function_call. With neither, the loop exits and the downstream Runner observes a silent successful empty stream.

This pattern is reported across multiple stalled fix attempts:

- #5394 — AnthropicLlm never populates finish_reason on LlmResponse
- #5006 — retry with resume message when model returns empty response
- #5636 — surface error when model returns STOP with empty content
- #3618 / #3699 — Handle empty message in LiteLLM response

It hits providers under real conditions: anthropic `content_filter`, gemini 2.5-flash-lite STOP-with-empty after tool calls, 0-token completions under safety, model_not_found responses normalized to `stop`. From the user's perspective the agent \"successfully\" ends a turn with no visible output. Downstream agent frameworks have no actionable signal to retry / surface / escalate.

## Change

- Track `last_finish_reason` + `last_model_version` across the stream.
- After both `aggregated_llm_response` and `aggregated_llm_response_with_tool_call` checks, if BOTH are `None` AND a `finish_reason` was observed, yield ONE `LlmResponse` with `error_code` set to the mapped finish_reason, `error_message` describing the failure mode, and `model_version` preserved. `usage_metadata` + `grounding_metadata` attach to that response.
- Minimum-surface: the guard only fires when stream produced no aggregated response AND a `finish_reason` was observed. Streams that genuinely yield nothing (test doubles, empty iterators) stay byte-identical.

## Tests

`tests/unittests/models/test_litellm.py` (4 new):

- `content_filter`-empty → surfaces with SAFETY error_code
- `stop`-empty → surfaces with STOP finish_reason + error_message
- Normal text stream → empty-guard does NOT fire (regression guard)
- Literally-empty stream (no chunks, no finish_reason) → byte-identical zero responses

**281 lite_llm tests pass + 1 skip; 0 regressions.**

## Context

Filed after several months of stalled PRs in this area (#5512 closed without merge, #5636 closed without merge, #5006 / #3699 open). Submitting a fresh attempt that addresses the same class of bugs with minimum-surface logic + thorough regression coverage. Happy to revise based on review.

Co-Authored-By: Kevin Sohn <kevin@openmagi.ai>


kevin-hs-sohn:
I signed the CLA. Re-pushed without the Co-Authored-By trailer so the check only sees my GitHub-account email.

rohityan:
Hi @kevin-hs-sohn, Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

PR #6201, opened 2026-06-23
fix(mcp): Preserve existing auth in mTLS transport

### LINKED ISSUE #6200: MCP mTLS transport may inject ADC Authorization when an existing auth header is lowercase
## 🔴 Required Information

**Describe the Bug:**

In the MCP mTLS transport path, ADK may inject ADC / execution-environment Service Account credentials even when an `Authorization` header is already present on the MCP request.

The immediate cause is that `_RefreshableAsyncCredentials.before_request()` checks for an existing Authorization header case-sensitively:

```python
if 'Authorization' in headers:
```

However, in the mTLS path, the MCP SDK uses `httpx.AsyncClient`, and ADK's `_GoogleAuthAsyncTransport.handle_async_request()` builds a plain dict from `httpx.Request.headers` before passing it to `AsyncAuthorizedSession.request()`:

```python
headers_dict = dict(request.headers)
```

`httpx` lowercases header names internally, so an existing `Authorization` header may be passed as `authorization`. In that case, the case-sensitive check misses the existing header and ADC credentials may be refreshed and injected into the same request.

As a result, an MCP server can receive an opaque access token for the runtime Service Account instead of the intended user / 3LO token.

**Steps to Reproduce:**

1. Use ADK v2.3.0 or current `main`.
2. Configure `McpToolset` with `StreamableHTTPConnectionParams` or `SseConnectionParams`.
3. Ensure the MCP request already has a user / 3LO `Authorization` header, for example through MCP auth credential handling or `header_provider`.
4. Run an MCP tool call in an environment where the mTLS transport path is enabled:
   - `GOOGLE_API_USE_CLIENT_CERTIFICATE` is unset or set to `true`
   - `google.auth.aio` is available
   - ADC is available
   - ADK can create `_GoogleAuthAsyncTransport` in `_get_mtls_transport()`
5. In `_GoogleAuthAsyncTransport.handle_async_request()`, `dict(request.headers)` can produce a lowercase `authorization` key.
6. `_RefreshableAsyncCredentials.before_request()` checks `'Authorization' in headers` and misses the lowercase key.

**Expected Behavior:**

HTTP header names are case-insensitive. `_RefreshableAsyncCredentials.before_request()` should treat `Authorization`, `authorization`, and any other casing as the same header.

If any Authorization header is already present, ADK should not inject ADC / Service Account credentials.

**Observed Behavior:**

An existing Authorization header can be missed when its key is lowercase. In that case, ADK may refresh ADC credentials and inject an execution-environment token into the MCP request.

This can cause the MCP server to receive a Service Account / ADC token instead of the intended user / 3LO token.

**Environment Details:**

- ADK Library Version (pip show google-adk): v2.3.0 and current `main`
- Desktop OS: Linux
- Python Version (python -V): Python 3.11

**Model Information:**

- Are you using LiteLLM: N/A
- Which model is being used: N/A. This issue occurs in the MCP HTTP transport path before model behavior is relevant.

---

## 🟡 Optional Information

**Regression:**

This appears related to the mTLS transport work introduced in v2.3.0. The issue is not about `list_tools` / discovery auth propagation. It affects the MCP HTTP request path when the mTLS transport bridges `httpx` requests into `google-auth-aio` `AsyncAuthorizedSession`.

**Logs:**

N/A. The behavior can be demonstrated at the header-normalization level:

```text
dict(httpx.Request(..., headers={"Authorization": "Bearer user_token"}).headers)
```

contains `authorization`, not `Authorization`.

**Screenshots / Video:**

N/A.

**Additional Context:**

The existing `Authorization` header is not only a user-supplied static header. ADK has MCP header propagation paths intended to pass user / 3LO credentials to the MCP server during MCP tool execution.

After a 3LO flow completes, ADK resolves the user access token as an `AuthCredential`. In the MCP integration, that credential is converted into an HTTP header by `McpToolset` / `McpTool`. For OAuth2 credentials, the implementation generates a header like this:

```python
headers = {"Authorization": f"Bearer {credential.oauth2.access_token}"}
```

For tool execution, the flow is roughly:

1. `McpTool._run_async_impl()` extracts auth headers from the resolved credential.
2. If `header_provider` is configured, it also obtains dynamic headers from `ReadonlyContext`.
3. The auth headers and dynamic headers are merged.
4. The merged headers are passed to `MCPSessionManager.create_session(headers=final_headers)`.
5. `MCPSessionManager` passes those headers to the MCP SDK's `streamablehttp_client` / `sse_client`.

The discovery path (`McpToolset.get_tools()` / `list_tools`) has a similar flow through `McpToolset._execute_with_session()`, where headers from `header_provider` and exchanged credentials are merged and passed to `create_session(headers=...)`.

Therefore, after a successful 3LO flow, the MCP HTTP request is expected to already contain:

```http
Authorization: Bearer <user_3lo_access_token>
```

That header is required by the MCP server to identify the user context. If the later mTLS transport / Google auth bridge injects ADC or Service Account credentials into the same request, the principal observed by the MCP server can change from the user to the runtime Service Account.

This issue is not about failing to obtain the user token, nor is it about the MCP header propagation path failing to set a header. The problem occurs after the user / 3LO Authorization header has already reached `create_session(headers=...)`: in the mTLS transport path, `dict(request.headers)` can represent it as lowercase `authorization`, and `_RefreshableAsyncCredentials.before_request()` does a case-sensitive check for `Authorization`.

This bug report focuses especially on the MCP tool-call request path. Independently of `list_tools` / discovery behavior, the tool-call HTTP request path should preserve and respect an existing Authorization header case-insensitively.

Suggested fix:

```python
if any(key.lower() == "authorization" for key in headers):
    logger.debug("Authorization header already present, not overwriting")
    return
```

However, the header handling itself should still be fixed because HTTP header names are case-insensitive.

Related issues / PRs:

- #5361: MCP mTLS support discussion / predecessor context
- #6150: v2.3.0 release including the mTLS transport work
- #3675: MCP header propagation feature, related but not the same bug
- #6146: ADC credentials being attached in another component, related security theme but different code path

```python
import httpx

request = httpx.Request(
    "POST",
    "https://example.com/mcp",
    headers={"Authorization": "Bearer user_token"},
)

print(dict(request.headers))
# contains "authorization", not "Authorization"
```

- Always (100%) when the mTLS transport path is active and the existing Authorization header is represented as lowercase `authorization`.

h-tsuboi918:

- Closes: #6200

N/A

**Problem:**

The MCP mTLS auth bridge checked only for an exact `Authorization` header key before adding ADC credentials. `httpx.Request.headers` can be converted into a plain dict with lowercase `authorization`, so the existing user / 3LO Authorization header could be missed.

When that happens, ADK may inject ADC / runtime Service Account credentials into the same MCP request, causing the MCP server to observe the runtime principal instead of the intended user principal.

**Solution:**

Make `_RefreshableAsyncCredentials.before_request()` check for an existing Authorization header case-insensitively before refreshing and injecting credentials.

This matches HTTP header semantics and preserves existing behavior for correctly cased `Authorization` headers while also handling lowercase `authorization` from the `httpx` request path.

### Testing Plan

**Unit Tests:**

Passed locally:

```text
uv run pytest tests/unittests/tools/mcp_tool/test_mcp_session_manager.py
51 passed, 8 warnings
```

Additional checks:

```text
git diff --check
git diff --cached --check
uv run pyink --check src/google/adk/tools/mcp_tool/mcp_session_manager.py tests/unittests/tools/mcp_tool/test_mcp_session_manager.py
uv run isort --check-only src/google/adk/tools/mcp_tool/mcp_session_manager.py tests/unittests/tools/mcp_tool/test_mcp_session_manager.py
```

Not run. This is covered by a focused unit test for the credential injection guard. The changed behavior is isolated to preserving an existing lowercase `authorization` header before ADC refresh/injection.

### Checklist

### Additional context

The issue is specific to the MCP mTLS transport path where ADK bridges `httpx` requests into `google-auth-aio` `AsyncAuthorizedSession`. The existing Authorization header may originate from MCP auth credential handling or a header provider, including user / 3LO token propagation.

No new code comments were added because the code change is limited to the existing Authorization header guard. There are no dependent downstream changes.


rohityan:
Hi @h-tsuboi918 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing ci test before we can proceed with a review.

h-tsuboi918:
Hi @rohityan ,Thanks for checking this.

I looked into the failed CI job:

https://github.com/google/adk-python/actions/runs/28122333977/job/83277473901?pr=6201

The failing step is `File Content Compliance` / `Check for hardcoded googleapis.com endpoints`.

The relevant workflow logic is here:

- `.github/workflows/continuous-integration.yml`, [lines 244-258](https://github.com/h-tsuboi918/adk-python/blob/fix/mcp-mtls-auth-header-case/.github/workflows/continuous-integration.yml#L244)
- It computes the changed Python files with:

```bash
CHANGED_FILES=$(git diff --diff-filter=ACMR --name-only origin/${GITHUB_BASE_REF}...HEAD | grep -E '\.py$' || true)
```

- It then scans the whole changed file for any `googleapis.com` URL matching:

```bash
FILES_WITH_ENDPOINTS=$(grep -lE 'https?://[a-zA-Z0-9.-]+\.googleapis\.com' $CHANGED_FILES)
```

- It fails if that file does not also contain `.mtls.googleapis.com`:

```bash
FILES_MISSING_MTLS=$(grep -L '.mtls.googleapis.com' $FILES_WITH_ENDPOINTS)
```

The job log reports this file:

```text
src/google/adk/tools/mcp_tool/mcp_session_manager.py
```

This PR does touch that file, but the matched `googleapis.com` string is not a hardcoded API endpoint introduced by this PR. It is an existing OAuth scope string:

- `src/google/adk/tools/mcp_tool/mcp_session_manager.py`, [line 637](https://github.com/h-tsuboi918/adk-python/blob/01c3cf004695d6488f76404b950f069adfce0039/src/google/adk/tools/mcp_tool/mcp_session_manager.py#L637)

```python
scopes = ['https://www.googleapis.com/auth/cloud-platform']
```

That value is an OAuth scope, not a service endpoint that can or should have a corresponding `.mtls.googleapis.com` variant. So my understanding is that this CI failure is a false positive caused by applying an endpoint-oriented grep to the entire changed file.

If we want to fix this, I think the right place is the CI check rather than adding a dummy `.mtls.googleapis.com` string in this PR. More specifically, the candidate fix location is around `.github/workflows/continuous-integration.yml`, lines 244-258, especially the logic that builds `FILES_WITH_ENDPOINTS` / `FILES_MISSING_MTLS`.

For example, before treating a changed file as a file containing a `googleapis.com` endpoint, the check could classify matches at the URL level and exclude OAuth scopes first. Conceptually, this could look like:

```bash
FILES_WITH_ENDPOINTS=$(
  grep -HEo 'https?://[a-zA-Z0-9.-]+\.googleapis\.com[^"'\''[:space:]]*' $CHANGED_FILES \
    | grep -vE 'https://www\.googleapis\.com/auth(/|$)' \
    | cut -d: -f1 \
    | sort -u || true
)
```

Then the existing `.mtls.googleapis.com` counterpart check would run only on files that contain non-scope `googleapis.com` URLs.

The intended behavior would be:

- still fail on hardcoded service endpoints such as `https://foo.googleapis.com/...`
- do not fail on OAuth scopes such as `https://www.googleapis.com/auth/cloud-platform`
- avoid requiring `.mtls.googleapis.com` for strings that are not endpoints

The actual change in this PR is limited to preserving an existing `Authorization` header in `_RefreshableAsyncCredentials.before_request` and adding a regression test for lowercase `authorization`:

- `src/google/adk/tools/mcp_tool/mcp_session_manager.py`, line 264
- the added regression test in `tests/unittests/tools/mcp_tool/test_mcp_session_manager.py`

For that reason, I do not think adding a dummy `.mtls.googleapis.com` string to `mcp_session_manager.py` would be the right fix for this PR, because the detected value is not a service endpoint. Should I include the CI check fix in this PR as well?

h-tsuboi918:
Hi @rohityan, I opened a separate PR for the CI false positive discussed above: #6245.
That PR targets the  check itself and excludes OAuth scope URLs such as  from the  endpoint/mTLS counterpart scan. This keeps the auth-header fix in this PR scoped to MCP, while addressing the unrelated CI check issue separately.
Related issue: #6238

h-tsuboi918:
Hi @rohityan, could you please re-trigger the checks for this PR?

The previous `File Content Compliance` failure came from the inline GitHub Actions compliance job, specifically the `googleapis.com` endpoint check. Since then, `main` changed in 8a7656b857c11777311024ec87ff5f5fd19e5f4c: the inline  compliance job was removed from `continuous-integration.yml`, and the compliance checks were consolidated into `scripts/compliance_checks.py` as a pre-commit hook.

Because of that base change, I expect the old `File Content Compliance` CI failure to no longer reproduce in the same way after a fresh run.

I also updated #6245 to target the new pre-commit-based implementation.

h-tsuboi918:
Hi @wukath @rohityan ,Thanks for fixing #6200.

I saw that the issue was closed by 38700324134f8d392a577f0cda78e4c4400e69d2 while this PR remained open. The landed change seems to address the same core issue this PR was intended to fix.

Could you clarify the preferred workflow for future external contributions to this repository? I understand that changes may go through an internal Piper / Copybara process, so I’d like to know whether GitHub PRs are expected to be reviewed and merged directly, or whether they are mainly used as proposals / references for internal changes.

I’m asking so that I can align future contributions with the expected process.

PR #6202, opened 2026-06-23
fix(models): Drop stale thought signatures

### LINKED ISSUE #3693: Remove unnecessary thought signatures from content sent to the Gemini models
Gemini models only look at the the latest thought signature. All thought signatures prior to the latest one can be removed from the LLM request content to reduce input tokens and API request size.

its-amann:
Fixes #3693

## Summary
- Keeps only the newest `thought_signature` in outgoing request contents.
- Clears earlier signatures before cache handling or request dispatch, so the payload does not carry values the next request no longer uses.
- Adds a regression test that builds multiple signed parts across the request history and verifies only the newest signature is sent.

## Testing
- `uv run --extra dev pyink src\google\adk\models\google_llm.py tests\unittests\models\test_google_llm.py`
- `uv run --extra dev ruff check src\google\adk\models\google_llm.py tests\unittests\models\test_google_llm.py`
- `uv run python -m pytest tests\unittests\models\test_google_llm.py::test_generate_content_async_keeps_only_latest_thought_signature`
- `uv run python -m pytest tests\unittests\models\test_google_llm.py::test_generate_content_async_keeps_only_latest_thought_signature tests\unittests\models\test_google_llm.py::test_generate_content_async tests\unittests\models\test_google_llm.py::test_generate_content_async_stream tests\unittests\models\test_google_llm.py::test_generate_content_async_stream_preserves_thinking_and_text_parts`


rohityan:
Hi @its-amann , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

GWeale:
gemini 3 validates the thought signature on every function-call step of the current turn and returns a 400 when one is missing. _remove_old_thought_signatures clears every signature but the newest across the whole history, so a turn with two sequential tool calls loses the first one. adk now keeps every signature part on purpose. open a new issue if you want us to look at it again.

PR #5974, opened 2026-06-05
feat(skills): Allow opt-in session state injection in skill instructions

lwangverizon:
### Link to Issue or Description of Change

**1. Link to an existing issue:**

- Closes: #5973

**Problem:**

Agent instructions support injecting dynamic session state via template
variables — `{var_name}`, `{app:var}`, `{artifact.file_name}`, and the optional
`{var?}` form — through `inject_session_state` (see
`flows/llm_flows/instructions.py`). Skill instructions returned by the
`load_skill` tool, however, are emitted verbatim. As a result a skill's
`SKILL.md` body cannot reference per-session state the way a system prompt can,
which is inconvenient for skills that need to adapt to the current user,
artifacts, or session context.

**Solution:**

Add an **opt-in** frontmatter metadata flag `adk_inject_session_state`, mirroring
the existing `adk_additional_tools` metadata convention. When set to `true`,
`LoadSkillTool` routes the skill instructions through the existing
`inject_session_state` utility before returning them; otherwise behavior is
unchanged.

The flag defaults to **disabled** on purpose: skill markdown routinely contains
literal braces (code samples, JSON, templates), and `inject_session_state`
raises `KeyError` on an unknown `{identifier}`. Making injection always-on would
break existing skills, so authors must explicitly opt in per skill. This keeps
the change fully backward-compatible.

A missing **required** variable returns a `STATE_INJECTION_ERROR` tool result
(rather than raising), consistent with the toolset's other error responses and
its `_detect_error_in_response` telemetry hook.

Example `SKILL.md`:

```yaml
---
name: greeter
description: Greets the current user by name.
metadata:
  adk_inject_session_state: true
---
Greet {user_name} warmly.{optional_note?}
```

Scope is limited to `load_skill` instructions; `load_skill_resource` content is
intentionally out of scope and can follow the same pattern later if desired.

**Files changed:**

- `src/google/adk/skills/models.py` — validate `adk_inject_session_state` is a
  bool; document the key in the `metadata` attribute docstring.
- `src/google/adk/tools/skill_toolset.py` — inject session state into skill
  instructions when the flag is set; return `STATE_INJECTION_ERROR` on a missing
  required variable.

### Testing Plan

**Unit Tests:**

- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

New tests:

- `tests/unittests/tools/test_skill_toolset.py`
  - `test_load_skill_no_injection_by_default` — braces preserved verbatim when
    the flag is absent (backward-compat).
[...]

PR #6231, opened 2026-06-28
fix(flows): scope skip_summarization text-append to AgentTool

### LINKED ISSUE #6230: skip_summarization text-append in __build_response_event is not scoped to AgentTool
## Description

The `skip_summarization` text-append in `__build_response_event`
(`src/google/adk/flows/llm_flows/functions.py`) is not scoped to `AgentTool`,
so it fires for **every** tool that sets `skip_summarization`. For non-AgentTool
tools whose function response is an internal acknowledgement, this surfaces the
raw response payload to the user as visible text.

## Background

PR #5974 (closing #3881) added this block so an `AgentTool`'s output is not lost
in UIs that don't render function responses:

```python
# src/google/adk/flows/llm_flows/functions.py  (current main)
has_displayable_result = display_result is not None and display_result != {
    'result': None
}
if (
    tool_context.actions.skip_summarization
    and 'error' not in function_result
    and has_displayable_result
):
    if isinstance(display_result, str):
        result_text = display_result
    else:
        result_text = json.dumps(display_result, ensure_ascii=False, default=str)
    content.parts.append(types.Part.from_text(text=result_text))
```

Issue #3881 is specifically about `AgentTool` (its title: *"The agent doesn't
work as expected when skip_summarization is set to True in AgentTool"*), and the
PR title is *"ensure AgentTool text output when skip_summarization is True"*. But
the condition only checks `tool_context.actions.skip_summarization` — there is no
`isinstance(tool, AgentTool)` guard, so the behavior applies to all tools.

## Why this is a problem

`skip_summarization` is also set by tools for the **opposite** reason: their
function response is an internal acknowledgement that should *not* be summarized
or shown. UI / widget-rendering tools are the common case — they return
something like `{"status": "ok"}` and set `skip_summarization=True` because the
user-facing result is a rendered widget, not text.

With the unguarded append, that ack is force-converted into a text `Part`. A
text part then bypasses the typical UI/SSE filters that strip
`functionResponse` / `functionCall` / `thought` parts, so the raw payload
(`{"status": "ok", ...}`) is surfaced to the user as visible text. We observed
this as a consistent, user-visible regression after upgrading to a build that
contains #5974: render-style tools began leaking their ack JSON into the chat
transcript.

## Steps to reproduce

1. Define a non-AgentTool (e.g. a `FunctionTool`) that sets
   `tool_context.actions.skip_summarization = True` and returns a dict such as
   `{"status": "ok"}`.
2. Invoke it through the normal function-calling flow.
3. The resulting function-response event contains an extra `Part.from_text(...)`
   with the serialized dict, in addition to the `functionResponse` part.

Expected: only the `functionResponse` part is present (the ack is not shown as
text). Actual: a text part carrying the serialized response is appended.

## Proposed fix

Guard the append with `isinstance(tool, AgentTool)` so it matches the original
intent — `AgentTool` keeps its text output; other tools that opt out of
summarization do not have their function response duplicated as text.

A PR implementing this (with a regression test) is attached.

## Environment

- ADK: current `main`
- Affected since: PR #5974 / commit "fix: ensure AgentTool text output when
  skip_summarization is True"

lwangverizon:
# fix(flows): scope skip_summarization text-append to AgentTool

Fixes #6230

## Problem

PR #5974 (issue #3881) added a text part to the function-response event when
`skip_summarization` is True, so an `AgentTool`'s output stays visible in UIs
that do not render function responses. The condition in `__build_response_event`
(`src/google/adk/flows/llm_flows/functions.py`) is not guarded by tool type, so
it fires for **every** tool that sets `skip_summarization`:

```python
if (
    tool_context.actions.skip_summarization
    and 'error' not in function_result
    and has_displayable_result
):
    ...
    content.parts.append(types.Part.from_text(text=result_text))
```

Tools other than `AgentTool` often set `skip_summarization` for the opposite
reason: their function response is an internal acknowledgement (e.g. a
UI/widget-rendering tool returning `{"status": "ok"}`) that should not be
summarized or shown to the user. Force-converting that ack into a text `Part`
makes it bypass UI/SSE filters that only strip
`functionResponse`/`functionCall`/`thought` parts, so the raw payload is
surfaced to the user as visible text.

Issue #3881 and PR #5974 are both explicitly about `AgentTool`, so this is a
scope/regression issue rather than an intended behavior change for all tools.

## Change

Guard the text-append with `isinstance(tool, AgentTool)`:

```python
if (
    isinstance(tool, AgentTool)
    and tool_context.actions.skip_summarization
    and 'error' not in function_result
    and has_displayable_result
):
    ...
```

- `AgentTool` keeps its text output (the #3881 behavior is preserved).
- Other tools that opt out of summarization no longer have their function
  response duplicated as a visible text part.
- `AgentTool` is imported lazily inside the function to avoid the
  `agents -> flows -> tools -> agents` circular import (the codebase already
  lazy-imports `AgentTool` elsewhere for the same reason).

## Tests

- Adds `test_skip_summarization_non_agent_tool_appends_no_text_part` in
  `tests/unittests/flows/llm_flows/test_functions_simple.py`: a `FunctionTool`
  that sets `skip_summarization=True` and returns a dict produces only a
  `function_response` part and no text part.
- The existing AgentTool skip_summarization tests
  (`tests/unittests/tools/test_agent_tool.py`, added by #5974) continue to pass,
  confirming the intended behavior is preserved.

## Backwards compatibility

Narrows an over-broad behavior introduced in #5974 back to its stated scope
(`AgentTool`). No public API or signature change.


rohityan:
Hi @lwangverizon , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing ci formatting tests.

rohityan:

PR #6236, opened 2026-06-29
feat(sessions): add merge_state for race-free server-side atomic state merges

lwangverizon:
# feat(sessions): add `merge_state` for race-free server-side atomic state merges

## Problem

ADK's only path to persist state is `append_event`, which couples a commutative
state merge to **event-log append + whole-session optimistic concurrency (OCC)**.
`DatabaseSessionService.append_event` takes a per-session lock, compares the
session's revision marker against `sessions.update_time`, raises
`"The session has been modified in storage since it was loaded…"` on mismatch,
writes an `events` row, and does a Python read-modify-write dict union into the
state column.

So two workers updating **independent** state keys on the same logical session
race on the session-level marker: one spuriously hits the stale-session error
and must reload + retry, even though the keys never collide. There is no
official "merge one key without OCC" path today.

## Change

Add `merge_state` — a state-only write path that performs a server-side atomic
JSON merge with **no events row** and **without advancing the session OCC
marker**.

```python
async def merge_state(
    self, *, app_name: str, user_id: str, session_id: str,
    delta: dict[str, Any],
) -> None
```

- `delta` keys use the existing prefix convention and are routed via the
  existing `_session_util.extract_state_delta()`: `app:` → `app_states`,
  `user:` → `user_states`, unprefixed → `sessions`.
- `temp:` keys raise `ValueError` (temp state is never persisted). Empty/`None`
  delta is a no-op.
- **No events row** is written and **`sessions.update_time` is never advanced**,
  so a concurrently held in-memory `Session` does not go stale and its next
  `append_event` still succeeds.

### Implementation

- `BaseSessionService.merge_state` — default raises `NotImplementedError`,
  mirroring the existing `get_user_state` pattern.
- `DatabaseSessionService` — per-dialect atomic merge issued via
  `sqlalchemy.text` so that **only the `state` column** is written (this is what
  keeps `update_time`'s `onupdate=func.now()` from firing and bumping the OCC
  marker):
  - PostgreSQL: `state = state || CAST(:delta AS JSONB)`
  - MySQL / MariaDB: `state = JSON_MERGE_PATCH(state, CAST(:delta AS JSON))`
  - SQLite: `state = json_patch(state, :delta)`
  App/user rows are ensured via the existing `_get_or_create_state` (handles
  concurrent inserts), then merged; missing **session** rows raise
  `ValueError("Session … not found.")`. Dialects with no known JSON-merge
  function raise `ValueError`.
- `SqliteSessionService` — reuses the existing `json_patch` upsert helpers; adds
  `_merge_session_state_in_db` (a copy of `_update_session_state_in_db` minus the
  `update_time=?` clause) to honor the no-OCC-bump contract.
- `InMemorySessionService` — routed `dict.update` into the app/user/session
  stores without bumping `last_update_time`.
- `VertexAiSessionService` — raises `NotImplementedError` (the Agent Engine API
  has no server-side merge), mirroring its `get_user_state`.

### Merge-semantics caveat (documented in the docstrings)

For flat scalar values (counters, flags, balances — the intended use case) all
backends behave identically. For non-scalar values they diverge: PostgreSQL `||`
is a shallow top-level merge that stores an explicit `None` as JSON `null` (key
kept), matching `append_event`'s Python `dict` union; MySQL/MariaDB/SQLite follow
RFC 7396 (`JSON_MERGE_PATCH` / `json_patch`) — nested objects merge recursively
and a `None` value deletes the key. The docstrings direct callers to
`append_event` when they need uniform dict-union semantics. (SQLite's null-delete
behavior is pre-existing via `SqliteSessionService`.)

## Tests

`tests/unittests/sessions/test_session_service.py` — parameterized across
IN_MEMORY / IN_MEMORY (light copy) / DATABASE / SQLITE:

- session / app / user scoped merges; cross-scope routing in one call
- `temp:` rejection (incl. mixed delta rejected atomically)
- empty-delta no-op; missing-session raises for session-scoped keys
- app/user rows auto-created without a prior `create_session`
- merges never create an event row

`DatabaseSessionService`-specific:

- OCC marker unchanged after session-scoped and app/user-scoped merges, and a
  previously-held session can still `append_event` afterwards (not stale)
- no `events` row created
- concurrent merges to independent keys both land (no lost update), looped

`tests/unittests/sessions/test_vertex_ai_session_service.py` —
`merge_state` raises `NotImplementedError`.

Run:

```
pytest tests/unittests/sessions/test_session_service.py
```

## Backwards compatibility

Strictly additive. New method with a `NotImplementedError` default on the base
class; no existing signature, behavior, or schema changes. `append_event`
semantics are untouched.


rohityan:
Hi @lwangverizon , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing ci mypy check tests before we can proceed with the review.

GWeale:
closing this for now. on 1 july we asked for the mypy failures to be fixed, and the branch has only had merges from main since; the 24 july run still reports new errors in in_memory_session_service.py and sqlite_session_service.py, and it now conflicts with main. if you pick it up again, open an issue first: a state write that skips the event log and the session staleness check is a design change we would want to agree on before code.

PR #6237, opened 2026-06-29
fix(cli): resolve uncaught PEP 479 RuntimeError during generator exhaustion in stream_query

ejoslin:
### Description

This PR resolves a silent runtime crash occurring at the end of streaming reasoning engine responses. 

When consuming a synchronous iterator inside `AdkApp.stream_query` using FastAPI's `StreamingResponse`, the generator exhaustion sequence throws a traceback to server `stderr` because the `StopIteration` exception escaping the `_aiter_from_iter` coroutine is automatically converted by Python's runtime into `RuntimeError: generator raised StopIteration` (conforming to PEP 479).

Since `StopIteration` is converted into a `RuntimeError` before escaping the `await` expression, the outer `except StopIteration:` block is bypassed, abruptly terminating the connection and dumping a stack trace on the server side despite returning an initial `HTTP 200 OK` to the client.

### Proposed Changes

- Replaced the direct `next(iterator)` call with the safe `next(iterator, _sentinel)` pattern inside `run_in_threadpool`.
- This ensures that when the generator is exhausted, it returns a sentinel object rather than throwing an exception across the asynchronous threadpool boundary.
- The thread runner checks for the sentinel identity (`chunk is _sentinel`) and terminates cleanly (`break`), completely bypassing the PEP 479 `RuntimeError` transformation.

### How Was This Tested?

- Verified locally with test cases returning synchronous generators.
- Verified that server-side `stderr` logs are 100% clean and free of `RuntimeError` tracebacks on generator completion.

### DIFF
--- a/src/google/adk/cli/fast_api.py
+++ b/src/google/adk/cli/fast_api.py
@@ -837,6 +837,11 @@ def get_fast_api_app(
           " tracing."
       )
 
+          " tracing."
+      )
+
+      )
+
     @app.middleware("http")
     async def context_propagation(
         request: Request, call_next: Callable[[Request], Awaitable[Any]]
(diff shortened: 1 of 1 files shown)


GWeale:
closing, this landed in aa9c187f with the same sentinel approach plus a test

Issue #6267, opened 2026-07-02
Emit agent/node lifecycle events (start/finish) in runner.run_async


ferponse:
### Context

We maintain the **AG-UI ↔ ADK middleware** (`ag-ui-adk`, part of [ag-ui-protocol/ag-ui](https://github.com/ag-ui-protocol/ag-ui)). To emit AG-UI's `STEP_STARTED` / `STEP_FINISHED` events — which bracket the nodes/phases of a run — from ADK workflows, the middleware must know when each node/agent **starts and finishes**.

### Problem

The `runner.run_async` stream carries text / tool-call / state / reasoning events, but **no explicit "agent X started / finished" boundaries**. Consumers that want to bracket work per node/agent (workflows, multi-agent, loops) must **infer** boundaries from `author` / `branch` transitions, which is lossy:

- exact **parallel-branch end** can't be observed (only inferred at the run boundary), and
- a `ParallelAgent` inside a `LoopAgent` can't be split per iteration — the events are identical across iterations (see #6266).

### ADK already has the lifecycle internally

- `before_agent_callback` / `after_agent_callback` fire **once per agent invocation** (including once per `LoopAgent` iteration).
- The 2.0 Workflow graph engine has `_node_status` / `node_tracing` / `_node_runner`.

It is simply not **surfaced in the public event stream**.

### Ask

Optionally (behind a flag/config) emit lightweight **lifecycle events** — agent/node **started** and **finished** — into `runner.run_async`, authored by the agent, so any consumer can bracket nodes exactly: including per `LoopAgent` iteration and exact parallel-branch boundaries.

### Prior art / parity

Other agent frameworks already emit these, and AG-UI integrations map them directly to `STEP_STARTED` / `STEP_FINISHED`:

- **CrewAI** → `MethodExecutionStartedEvent` / `MethodExecutionFinishedEvent`
- **AWS Strands** → `multiagent_node_start` / `multiagent_node_stop`

Surfacing lifecycle events in the stream would let the ADK integration reach the same precision, without consumers having to install their own plugin/callback to observe boundaries.

### Motivation

We need this to faithfully emit **all** AG-UI event types from ADK; `STEP_STARTED`/`STEP_FINISHED` is currently the one category the ADK integration can only approximate. See the AG-UI PR: ag-ui-protocol/ag-ui#2076.

### Related

#6266 is a minimal, loop-scoped subset of this (just expose the loop iteration index). **This issue is the general form and would subsume #6266.**

[...]

PR #6268, opened 2026-07-02
feat(agents): annotate events with the LoopAgent iteration

### LINKED ISSUE #6266: LoopAgent: surface the current iteration index in the event stream
### Context

We maintain the **AG-UI ↔ ADK middleware** (`ag-ui-adk`, part of [ag-ui-protocol/ag-ui](https://github.com/ag-ui-protocol/ag-ui)). AG-UI is a protocol between agents and frontends; one of its event categories is `STEP_STARTED` / `STEP_FINISHED`, which bracket the phases/nodes of a run (the spec recommends `stepName = node name`).

To emit these from ADK workflows (`SequentialAgent` / `ParallelAgent` / `LoopAgent` / the 2.0 `Workflow` graph), the middleware infers node boundaries from the `runner.run_async` event stream using each event's `author` (node name) and `branch` (`ParallelAgent` assigns a distinct `par.<sub_agent>` per branch via `_create_branch_ctx_for_sub_agent`). This works cleanly for serial and parallel topologies.

### Problem

There is **one gap: a `ParallelAgent` nested inside a `LoopAgent`.** The concurrent branches keep the **same `author`, `branch`, and `invocation_id`** across iterations, and the flat event stream carries **no iteration-boundary signal** — so a consumer cannot tell iteration 1 from iteration 2. The events are identical except for their unique `id`. (Verified on `google-adk 2.3.0`.)

`LoopAgent(max_iterations=2, sub_agents=[ParallelAgent(sub_agents=[p, q])])` emits:

```
author='p'  branch='par.p'  invocation_id='e-…'   ← iteration 1
author='q'  branch='par.q'  invocation_id='e-…'
author='p'  branch='par.p'  invocation_id='e-…'   ← iteration 2 (indistinguishable)
author='q'  branch='par.q'  invocation_id='e-…'
```

As a result, per-iteration steps cannot be reconstructed for this topology.

### ADK already tracks the iteration

`LoopAgent._run_async_impl` already maintains `times_looped`, and on the resumable path it even yields a state event carrying it (`self._create_agent_state_event(ctx)`). It's simply not surfaced on the events produced *within* an iteration.

### Ask (minimal, additive)

Surface the current loop iteration in the event stream, either:

- a **read-only field** on events produced within a `LoopAgent` iteration (e.g. `Event.loop_iteration`, or in event metadata), or
- a lightweight **iteration-boundary marker** event emitted at each iteration start (generalizing the existing resumable-path state event to always fire).

This is purely additive and does not change execution behavior.

### Why not put the iteration into `branch`?

That would be convenient for consumers, but `branch` governs context/event isolation — changing it per iteration could affect how loop iterations **share state** (the whole point of a loop). So we're explicitly asking to expose the iteration **separately from** `branch`.

### Motivation

We need this to faithfully emit **all** AG-UI event types from ADK. `STEP_STARTED`/`STEP_FINISHED` is currently the one category the ADK integration can only approximate (for loop-of-parallel). See the AG-UI PR: ag-ui-protocol/ag-ui#2076.

### Related

A more general alternative — emitting explicit agent/node **lifecycle** events in the stream — is filed as the companion issue #6267, which would subsume this. This issue (#6266) is the minimal, lowest-risk subset scoped to loops.

ferponse:
## What

`LoopAgent` re-runs its sub-agents each iteration, but events produced in different iterations are **indistinguishable** in the stream — same `author`, `branch`, and `invocation_id`. So a consumer cannot tell iteration 1 from iteration 2. This is most visible with a `ParallelAgent` nested in a `LoopAgent`, whose branch ids (`par.p`, `par.q`) repeat every iteration.

This records the current 0-based iteration under `Event.custom_metadata['loop_iteration']` as each sub-agent event bubbles up through the loop.

Fixes #6266.

## Why

We're building the AG-UI ↔ ADK middleware (`ag-ui-adk`), which maps ADK workflow nodes to AG-UI `STEP_STARTED` / `STEP_FINISHED` events. It infers node boundaries from the `runner.run_async` stream (`author` = node, `branch` = structural path). This works for every serial and parallel topology **except a parallel block inside a loop**, where iterations are indistinguishable, so per-iteration steps can't be reconstructed. Verified on `google-adk 2.3.0`.

## How

- New `LoopAgent.LOOP_ITERATION_KEY = 'loop_iteration'` constant.
- `_annotate_loop_iteration(event, times_looped)` stamps `event.custom_metadata` via `setdefault`, so the nearest enclosing loop wins when loops are nested.
- Called on each sub-agent event before it is yielded.

Additive and behavior-preserving: no new events, no schema change (`custom_metadata` is the existing custom bucket).

`LoopAgent(max_iterations=2, sub_agents=[ParallelAgent(sub_agents=[p, q])])` — after this change:

```
author='p' branch='par.p'  custom_metadata['loop_iteration']=0   ← iteration 1
author='q' branch='par.q'  custom_metadata['loop_iteration']=0
author='p' branch='par.p'  custom_metadata['loop_iteration']=1   ← iteration 2 (now distinguishable)
author='q' branch='par.q'  custom_metadata['loop_iteration']=1
```

## Tests

`tests/unittests/agents/test_loop_agent.py`: added `test_run_async_annotates_loop_iteration` (iterations 0/1/2) and `test_run_async_loop_iteration_survives_nested_loops` (nested loops — inner wins). The full `test_loop_agent.py` passes.

## Notes

- `LoopAgent` is `@deprecated` in favor of `Workflow`; this fixes the still-widely-used classic loop. The same iteration signal would be valuable from the Workflow graph engine (out of scope here).
- Companion issue #6267 proposes the general form (explicit agent/node lifecycle events in the stream), which would subsume this.


rohityan:
Hi @ferponse , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

Issue #2936, opened 2025-09-12
set_model_response escape latin characters in final response


amenegola:
When defining output_schema in a LlmAgent with other tools, output schema processor executes json.dumps with ensure_ascii = True, making the agent respond with escaped latin characters, and therefore presenting responses not human readable:

https://github.com/google/adk-python/blob/168c7248664229fa99501c0de942493b29795a0e/src/google/adk/flows/llm_flows/_output_schema_processor.py#L110

This line could be changed with ensure_ascii=False:

`json.dumps(func_response.response, ensure_ascii=False) `

amenegola:
Hello @seanzhou1023 

Following up on this issue, I've opened a Pull Request (#2937) with the proposed fix. The changes involve explicitly setting `ensure_ascii=False` in the `json.dumps` call, which I've tested locally to correctly handle Latin characters.

Please feel free to review it when you have a moment, and let me know if there's anything else I can do to help move this forward.

Thanks for your time!

PR #6282, opened 2026-07-03
fix(workflow): preserve non-ASCII characters in LLM agent node input

### LINKED ISSUE #6279: Input Schema: Non ASCII characters are being escaped to Unicode
**Describe the Bug:**
When defining an Agent and including input_schema in order to serialize a Pydantic object non-latin characters are being escaped to unicode format. This behavior causes token bloating, slow LLM responses and in some cases LLM is not able to return a readable response and returns symbols blended with the origin language of the prompt. In my case, for a RAG solution I sent context in Hebrew to an Agent and each letter was replaced with 6 Unicode ascii letters which led to high number of prompt tokens and inconsistency with LLM response.

**Steps to Reproduce:**
- `pip install google-adk==2.2.0 google-genai==2.8.0 `
- Sample of python code causing the issue (I added before_model_callback to verify that sent data is in Unicode). Any Agent using input_schema will reproduce that.  adding some code to reference
```py

class SearchAttempt(BaseModel):
    query: str
    surrounding_chunks: int
    result_count: int
    error_message: str | None = None

class NormalizedChunk(BaseModel):
    retrieval_index: int
    document_bucket: str
    document_file_name: str
    chunk_index: int
    distance: float | None = None
    title: str | None = None
    publication_date: str | None = None
    section_title: str | None = None
    section_summary: str | None = None
    text: str | None = None

class SearchResult(BaseModel):
    status: Literal["success", "error"]
    query: str
    surrounding_chunks: int
    normalized_chunks: list[NormalizedChunk] = Field(default_factory=list)
    error_message: str | None = None
    results_count: int

class EvaluationInput(BaseModel):
    original_query: str
    search_result: SearchResult
    attempts: list[SearchAttempt]



def debug_rag_evaluator_request(
    callback_context: CallbackContext,
    llm_request: LlmRequest,
) -> None:
    print("========== raw parts ==========")
    for content in llm_request.contents or []:
        for part in content.parts or []:
            if part.text:
                print(part.text[:5000])
    print("========== end raw parts ==========")

rag_evaluator_agent = Agent(
    name="rag_evaluator_agent",
    model=_agent_model(),
    input_schema=EvaluationInput,
    output_schema=RetrievalDecision,
    include_contents="none",
    before_model_callback=debug_rag_evaluator_request,
    generate_content_config=types.GenerateContentConfig(
        responseMimeType="application/json",
        temperature=0.0,
        max_output_tokens=8192
    ),
    instruction="""
......
""",
)
```

**Expected Behavior:**
It seems that for `response_schema` this issue was fixed on [#2936](https://github.com/google/adk-python/issues/2936) by putting ensure_ascii=False for output schema dumps logic. The same should be done for input schema or users should at least have the option to choose whether to use that flag.
The bad behavior arises from: [_node_input_to_content](https://github.com/google/adk-python/blob/main/src/google/adk/workflow/_llm_agent_wrapper.py) which calls json.dumps and model_dump_json without ensure_ascii=False flag. I ended up monkey-patching the function in my service which causes ADK to send request with Hebrew letters properly but of course this is discouraged.

**Observed Behavior:**
Request for LLM becomes bloated (~6 times more tokens than with non escaped letters). LLM responses were very slow and not predictable (responses were huge with many repeating symbols, signs were blended with hebrew letters, etc).

**Environment Details:**

 - ADK Library Version (pip show google-adk): 2.2.0
 - Desktop OS:** [e.g., macOS, Linux, Windows]: Linux (WSL)

**Model Information:**

 - Are you using LiteLLM: No

 - Always (100%)

anneheartrecord:
## Summary

`input_schema` inputs containing non-Latin characters (Hebrew, Chinese, etc.) reach the LLM as `\uXXXX` escapes, which bloats prompt tokens (~6x for Hebrew) and degrades model responses, as reported in #6279.

The escaping comes from `json.dumps` being called with its default `ensure_ascii=True` on the LLM-bound input text. Two paths were affected:

- `workflow/_llm_agent_wrapper.py` – `_node_input_to_content()` for `dict`/`list` node input.
- `flows/llm_flows/contents.py` – `_build_task_input_user_content()`, which rebuilds a delegated task's function-call args as the sub-agent's first user turn. This path takes priority over the wrapper fallback, so it affects the common chat/root → task sub-agent delegation case.

Both now serialize with `ensure_ascii=False`, matching how the output-schema path already serializes responses (`_output_schema_processor.py`), fixed earlier in #2936/#2937.

Note: the `BaseModel` branch (`model_dump_json()`) is intentionally left unchanged — Pydantic v2 already emits raw UTF-8 (does not escape non-ASCII), which the added regression test confirms.

## Testing

- `tests/unittests/workflow/test_llm_agent_as_node.py` – `dict`/`list`/`BaseModel` node input preserves non-ASCII.
- `tests/unittests/flows/llm_flows/test_contents.py` – delegated task FC args preserve non-ASCII.

Both new tests fail before the change and pass after.

Fixes #6279

### DIFF
--- a/src/google/adk/flows/llm_flows/contents.py
+++ b/src/google/adk/flows/llm_flows/contents.py
@@ -386,7 +386,7 @@ def _build_task_input_user_content(
         try:
           import json as _json
 
-          text = _json.dumps(dict(fc.args))
+          text = _json.dumps(dict(fc.args), ensure_ascii=False)
         except (TypeError, ValueError):
           text = str(fc.args)
         parts = [types.Part(text=text)]
--- a/src/google/adk/utils/content_utils.py
+++ b/src/google/adk/utils/content_utils.py
@@ -64,7 +64,7 @@ def to_user_content(value: Any) -> types.Content:
     deep-copied)
   - str -> single text part
   - BaseModel -> model_dump_json() text part
-  - dict/list -> json.dumps() text part
+  - dict/list -> json.dumps() text part (non-ASCII preserved, not escaped)
   - anything else -> str() text part
   """
   if isinstance(value, types.Content):
@@ -74,7 +74,7 @@ def to_user_content(value: Any) -> types.Content:
   elif isinstance(value, BaseModel):
     text = value.model_dump_json()
   elif isinstance(value, (dict, list)):
-    text = json.dumps(value)
+    text = json.dumps(value, ensure_ascii=False)
   else:
     text = str(value)
   return types.Content(role='user', parts=[types.Part(text=text)])
--- a/tests/unittests/flows/llm_flows/test_contents.py
+++ b/tests/unittests/flows/llm_flows/test_contents.py
@@ -1894,3 +1894,41 @@ def test_recover_compacted_parallel_call_reinjects_sibling_response():
       if part.function_response
   }
   assert response_ids == {"lr-1", "reg-1"}
+
+
+def test_task_input_user_content_preserves_non_ascii():
+  """Delegated task input must not escape non-ASCII FC args (issue #6279).
+
+  A chat coordinator delegates to a task sub-agent via a function call; the
+  task agent's first user turn is rebuilt from the FC args. Escaping non-Latin
+  characters to ``\\uXXXX`` there bloats prompt tokens and degrades responses.
+  """
+  fc_id = "fc_task_1"
+  events = [
+      Event(
+          invocation_id="inv1",
+          author="coordinator",
+          content=types.Content(
+              role="model",
+              parts=[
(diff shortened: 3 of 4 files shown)


anneheartrecord:
@googlebot I signed it!

anneheartrecord:
@googlebot I signed it!

rohityan:
Hi @anneheartrecord , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

PR #6294, opened 2026-07-04
fix(workflow): treat RetryConfig(max_attempts=0) as no retries

### LINKED ISSUE #6293: RetryConfig(max_attempts=0) runs 5 attempts instead of disabling retries
### Description

`RetryConfig.max_attempts` documents that **"If 0 or 1, it means no retries. If not specified, default to 5."** and defaults to `None`. But `_should_retry_node` resolves the default with a falsy-coalescing `or`:

```python
# src/google/adk/workflow/utils/_retry_utils.py:35
max_attempts = retry_config.max_attempts or 5
```

`0` is falsy, so an explicit `RetryConfig(max_attempts=0)` (documented as "no retries") is silently replaced with `5`, giving 4 retries instead of none. `max_attempts=1` works (`1 or 5 == 1`); only the documented `0` case is broken. The intended "unset" sentinel is `None`, which must still map to 5.

### Reproduction

```python
from google.adk.workflow._retry_config import RetryConfig
from google.adk.workflow._node_state import NodeState
from google.adk.workflow.utils._retry_utils import _should_retry_node

cfg = RetryConfig(max_attempts=0)  # documented: no retries
# node on its first (original) attempt:
_should_retry_node(RuntimeError(), cfg, NodeState(attempt_count=1))
# -> True  (bug; a workflow node then runs up to 5 attempts total)
```

### Expected

`max_attempts=0` (and `1`) → no retries; `None` → default of 5.

### Fix

```python
max_attempts = (
    retry_config.max_attempts if retry_config.max_attempts is not None else 5
)
```

`Workflow`/`RetryConfig` are public (both in `__all__`) and this drives every workflow node's retry loop (`_NodeRunner._attempt_retry`). Happy to send a small PR.

Osamaali313:
Fixes #6293

## Problem

`RetryConfig.max_attempts` is documented as *"Maximum number of attempts, including the original request. If 0 or 1, it means no retries. If not specified, default to 5."* and defaults to `None`. But `_should_retry_node` resolved the default with a falsy-coalescing `or`:

```python
# src/google/adk/workflow/utils/_retry_utils.py:35
max_attempts = retry_config.max_attempts or 5
```

`0` is falsy, so `RetryConfig(max_attempts=0)` — documented as "no retries" — is silently replaced with `5`, running 4 retries instead of none. `max_attempts=1` happens to work (`1 or 5 == 1`); only the documented `0` case is broken. The intended "unset" sentinel is `None`, which must still map to `5`.

`Workflow` and `RetryConfig` are public API (both in `__all__`, no `@experimental` gate), and this drives every workflow node's retry loop via `_NodeRunner._attempt_retry`.

## Fix

```python
max_attempts = (
    retry_config.max_attempts if retry_config.max_attempts is not None else 5
)
```

## Tests

Added `TestShouldRetryNode` to `tests/unittests/workflow/utils/test_retry_utils.py` (the file previously covered only `_get_retry_delay`), asserting:

- `max_attempts` of `0` and `1` disable retries
- `max_attempts=3` retries until the limit
- unset (`None`) defaults to 5

Before the fix, the `max_attempts=0` case fails; after, it passes. The full `tests/unittests/workflow/` suite passes (619 passed, no regressions).


Osamaali313:
Pushed a small formatting commit — `pyink` (line-length 80) wanted the long `_should_retry_node` assertions in `test_retry_utils.py` wrapped; I ran `pyink` with the repo config and it's now clean (`pyink --check` passes, tests still green: 10 passed).

Heads up on the other red check: the **Mypy** failure isn't from this PR's diff — the "new" errors it lists are all in `src/google/adk/models/lite_llm.py` (`_finish_reason_to_error_message` arg-type), which this PR doesn't touch. Both baseline and PR report 3375 errors; it looks like the baseline-diff surfaced a pre-existing `lite_llm.py` typing issue after the `main` merge. Happy to help if you'd like, but it's outside this change.

rohityan:
Hi @Osamaali313 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

PR #6302, opened 2026-07-05
fix(memory): do not leak temp file when RAG upload fails

agharsallah:
## Problem

`VertexAiRagMemoryService.add_session_to_memory` creates a
`tempfile.NamedTemporaryFile(delete=False)` and only removes it via
`os.remove` on the success path. Two failure paths leak the file
permanently:

1. The `ValueError("Rag resources must be set.")` raised when no rag
   resources are configured — the temp file is already on disk by then.
2. Any exception from `rag.upload_file` — the `os.remove` at the end is
   never reached.

## Fix

- Move the rag-resource validation **before** the temp file is created,
  so nothing is written to disk when the config is invalid.
- Wrap the upload loop and `os.remove` in `try/finally` so the temp file
  is always deleted, even when the upload raises.

## Tests

Added two cases to `test_vertex_ai_rag_memory_service.py`:
- `test_add_session_removes_temp_file_when_upload_fails` — mocks
  `rag.upload_file` to raise and asserts the temp file is gone afterward.
- `test_add_session_does_not_create_temp_file_without_rag_resources` —
  asserts no temp file is created when rag resources are unset.

All tests pass.

### DIFF
--- a/src/google/adk/memory/vertex_ai_rag_memory_service.py
+++ b/src/google/adk/memory/vertex_ai_rag_memory_service.py
@@ -145,6 +145,9 @@ class VertexAiRagMemoryService(BaseMemoryService):
 
   @override
   async def add_session_to_memory(self, session: Session) -> None:
+    if not self._vertex_rag_store.rag_resources:
+      raise ValueError("Rag resources must be set.")
+
     with tempfile.NamedTemporaryFile(
         mode="w", delete=False, suffix=".txt"
     ) as temp_file:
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @agharsallah , Thank you for your contribution through this pull request!  This PR has merge conflicts that require changes from your end. Could you please rebase your branch with the latest main branch to address these? Once this is complete, please let us know so we can proceed with the review.

agharsallah:
@rohityan Rebased

DeanChensj:
Hi @agharsallah ,  thanks for this PR!  The same issue has already been addressed in https://github.com/google/adk-python/commit/80a05b7f639216c8bf60a870598f0d44efcba339, so we will close this one for now. Feel free to send out a new PR if you find any issue.

PR #6308, opened 2026-07-05
fix(runners): widen Runner.__init__ agent param to accept BaseNode

### LINKED ISSUE #6270: Runner.__init__ 'agent' param typed Optional[BaseAgent] but class field and runtime accept BaseNode (e.g. Workflow) roots
## Summary

`Runner`'s class-level field is typed to accept node roots:

```python
# runners.py:153
agent: Optional[BaseAgent | 'BaseNode'] = None
```

but the `__init__` parameter was not widened to match:

```python
# runners.py:175
agent: Optional[BaseAgent] = None,
```

The runtime explicitly supports a `BaseNode` root (e.g. a `Workflow`):

```python
# runners.py:1023-1024
if isinstance(self.agent, BaseNode) and not isinstance(self.agent, BaseAgent):
    ...
```

So passing a `Workflow` as the root works at runtime, but type checkers (pyright) reject `Runner(agent=my_workflow)` against the narrower `__init__` signature.

## Expected

The `__init__` `agent` parameter annotation matches the class field, e.g. `Optional[BaseAgent | BaseNode]`.

## Version

`google-adk==2.3.0`

## Note

I only verified the `__init__` `agent` path. There may be similar narrowing on related entry points (`node=`, `run_live`), but I have not confirmed those.

anxkhn:

- Closes: #6270

**Problem:**

The `Runner.agent` class field is typed `Optional[BaseAgent | 'BaseNode']`
(`runners.py:158`) and the runtime explicitly branches on
`isinstance(self.agent, BaseNode)` in both `run_async` and `run_live`, so a
`BaseNode` root (for example a `Workflow`) is fully supported at execution time.

However, the `Runner.__init__` `agent` parameter (`runners.py:180`) was still
typed `Optional[BaseAgent]`, narrower than the field it initializes. As a result,
static type checkers reject `Runner(agent=my_workflow)` even though it runs
correctly. The parameter annotation was simply out of sync with the class field
and with the runtime's own `BaseNode` handling.

This is a static typing correctness issue, not a runtime one (Python does not
enforce annotations at runtime, so the call already works). It shows up in ADK's
own codebase: with the current annotation, `mypy` flags the internal call site
in `a2a/utils/agent_to_a2a.py` where a plain `BaseAgent` is forwarded into
`Runner`:

```
src/google/adk/a2a/utils/agent_to_a2a.py:169: error: Argument 1 to "Runner"
  has incompatible type ...; expected "BaseAgent | None"  [arg-type]
```

**Solution:**

Widen the `__init__` `agent` annotation to `Optional[BaseAgent | 'BaseNode']` so
it matches the class field byte-for-byte. This mirrors the existing
forward-reference style already used for the field (`runners.py:158`) and for the
`run_live` `node` parameter (`runners.py:455`); `BaseNode` is kept as a lazy
forward reference (the module uses `from __future__ import annotations` and
imports `BaseNode` inside the methods that need it), so no new import is added.

This is a type-only change with no runtime behavior change. The narrower internal
helper `_resolve_app` was intentionally left as-is: this PR is scoped to the
public `__init__` surface that #6270 reports.

### Testing Plan

**Unit Tests:**

Added `TestRunnerResolveApp::test_resolve_app_with_base_node_via_agent_param`,
which constructs `Runner(agent=<BaseNode>, app_name=...)` and asserts the node is
accepted and wrapped (`runner.app.root_agent is node`, `runner.agent is node`).
This is a behavioral contract guard for the accepted-and-wrapped path; the
annotation itself is validated by the type checker (below), since Python does not
enforce annotations at runtime.

`pytest` summary:

```
$ pytest tests/unittests/test_runners.py -q
70 passed
```

Type check (the actual validation of a type-only change): with this change,
`mypy` no longer reports the `expected "BaseAgent | None"` error at the
`a2a/utils/agent_to_a2a.py` `Runner(...)` call site, and no new `mypy` errors are
introduced elsewhere.

Not applicable. This is a type-annotation-only change with no runtime behavior
change; it is covered by the unit test and the static type check above.

### Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas. (No comment needed; one-line annotation change mirroring the existing field.)
- [ ] I have manually tested my changes end-to-end. (Not applicable; type-only change.)
- [ ] Any dependent changes have been merged and published in downstream modules. (None.)

### Additional context

The change is a single line plus one test:

```diff
-      agent: Optional[BaseAgent] = None,
+      agent: Optional[BaseAgent | 'BaseNode'] = None,
```

### DIFF
--- a/src/google/adk/runners.py
+++ b/src/google/adk/runners.py
@@ -239,7 +239,7 @@ class Runner:
       *,
       app: Optional[App] = None,
       app_name: Optional[str] = None,
-      agent: Optional[BaseAgent] = None,
+      agent: Optional[BaseAgent | BaseNode] = None,
       node: BaseNode | None = None,
       plugins: Optional[List[BasePlugin]] = None,
       artifact_service: Optional[BaseArtifactService] = None,
@@ -320,7 +320,7 @@ class Runner:
   def _resolve_app(
       app: Optional[App],
       app_name: Optional[str],
-      agent: Optional[BaseAgent],
+      agent: Optional[BaseAgent | BaseNode],
       node: BaseNode | None,
       plugins: Optional[List[BasePlugin]],
   ) -> App:
--- a/tests/unittests/test_runners.py
+++ b/tests/unittests/test_runners.py
@@ -1777,6 +1777,22 @@ class TestRunnerResolveApp:
     assert runner.app_name == "test_app"
     assert runner.agent is self.root_agent
 
+  def test_resolve_app_with_base_node_via_agent_param(self):
+    """Test that a BaseNode root passed via `agent` is accepted."""
+    from google.adk.workflow._base_node import BaseNode
+
+    node = BaseNode(name="test_node")
+    runner = Runner(
+        app_name="test_app",
+        agent=node,
+        session_service=self.session_service,
+        artifact_service=self.artifact_service,
+    )
+    assert runner.app is not None
+    assert runner.app.root_agent is node
+    assert runner.app_name == "test_app"
+    assert runner.agent is node
+
   def test_resolve_app_with_node_wraps_in_app(self):
     """Test that a bare node is wrapped into an App."""
     from google.adk.workflow._base_node import BaseNode


rohityan:

rohityan:
Hi @wyf7107 , can you please review this.LGTM

jazhang00:
Hi @anxkhn, thanks for creating this PR. Currently a runner could be initialized by passing `BaseNode` into `node`. Will this resolve the linked issue?

anxkhn:
good question. yes, `Runner(node=...)` works today and is the natural entry point for node roots, but it doesn't resolve #6270 as filed: that issue is about the `__init__` annotation being narrower than the field it initializes. the class field accepts node roots (typed `BaseNode`, which `BaseAgent` extends) and both `run_async` and `run_live` branch on `isinstance(self.agent, BaseNode)`, so a basenode passed via `agent=` already executes correctly and lands in `self.agent` via `self.agent = app.root_agent`. only the `__init__` param annotation disagreed with the runtime.

the two params also aren't interchangeable: `agent=` requires an explicit `app_name`, while `node=` defaults it to `node.name`. so code that forwards a value which might be a node into `Runner(agent=...)` can't just switch to `node=` without changing that behavior, and type checkers keep flagging it until the param matches the field. this PR is a type-only fix with no runtime change.

DeanChensj:
Thanks for opening this PR and for the contribution!
`BaseNode` (including workflows and custom nodes) is already supported in `Runner`, but it is designed to be passed via the `node` parameter rather than `agent`:
```python
# Pass a BaseNode directly via the `node` argument:
runner = Runner(node=my_workflow, session_service=session_service)
# Or wrap it in an App (recommended):
runner = Runner(app=App(name="my_app", root_agent=my_workflow), session_service=session_service)
```
For backwards compatibility reasons (as well as differing argument handling, such as `app_name` resolution), `Runner.__init__` maintains two separate mutually exclusive parameters:
- `agent`: specifically for `BaseAgent`
- `node`: specifically for `BaseNode`
Since `BaseNode` is already supported through `Runner(node=...)`, we prefer keeping `agent` typed strictly for `BaseAgent` to maintain this parameter separation. You can resolve the type checker issue by passing your workflow/node to `Runner(node=...)` instead.
Thanks again for looking into this!

PR #6345, opened 2026-07-08
fix(cli): remove shadowing local json import breaking A2A setup

### LINKED ISSUE #6332: `adk api_server --a2a` silently mounts no A2A routes: `UnboundLocalError` on `json` in `get_fast_api_app`
## 🔴 Required Information

**Describe the Bug:**

Running `adk api_server --a2a` against an agent directory that contains an
`agent.json` (agent card) fails to set up the A2A endpoint and starts the server
**without any A2A routes**. The failure is swallowed by a `try/except` and only
appears as one `ERROR` log line, so from the user's side `--a2a` silently does
nothing.

Root cause is a shadowed import in `src/google/adk/cli/fast_api.py`:
`get_fast_api_app()` imports `json` at module scope (line 19) but **also** has a
function-local `import json` (line 748, in the `gemini_enterprise_app_name`
block). Because that local import exists anywhere in the function, Python treats
`json` as local for the **entire** function, so the earlier `json.load(f)` that
reads `agent.json` (line 720) raises `UnboundLocalError` before the local import
is ever reached.

**Steps to Reproduce:**

1. Create an agent folder `agents/hello/` with `agent.py` exposing
   `root_agent = LlmAgent(...)` and an `agents/hello/agent.json` (any valid A2A
   `AgentCard` JSON).
2. Run `adk api_server --a2a agents`.
3. Probe the A2A card endpoint:
   ```bash
   curl -s -o /dev/null -w "%{http_code}\n" \
     http://127.0.0.1:8000/a2a/hello/.well-known/agent-card.json
   ```
4. Observe the startup `ERROR` log (below) and a `404` from the card endpoint.

**Expected Behavior:**

`adk api_server --a2a` mounts the A2A routes (`/a2a/<app>` and
`/a2a/<app>/.well-known/agent-card.json`) for every agent folder that has an
`agent.json`.

**Observed Behavior:**

The server starts (`Uvicorn running`) but A2A setup fails and the routes are
never added — the card endpoint returns `404`, and there is no A2A JSON-RPC
endpoint. The ADK REST endpoints (`/list-apps`, `/run_sse`) work normally.
Startup log:

```text
ERROR - fast_api.py:739 - Failed to setup A2A agent hello:
cannot access local variable 'json' where it is not associated with a value
```

**Environment Details:**

 - ADK Library Version (pip show google-adk): 2.2.0 (verified against the
   pristine `google_adk-2.2.0-py3-none-any.whl`; `fast_api.py` lines 19 / 720 / 748)
 - Desktop OS: Linux
 - Python Version (python -V): 3.13

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: N/A — CLI/serving bug, independent of the model

---

## 🟡 Optional Information

**Additional Context:**

The `--a2a` A2A-setup block and the redundant import live in the same function
(`get_fast_api_app`):

```python
# src/google/adk/cli/fast_api.py  (google-adk 2.2.0)
import json                                   # line 19  (module scope)

def get_fast_api_app(...):
    ...
    # A2A setup, per agent folder with an agent.json:
    with (p / "agent.json").open("r", encoding="utf-8") as f:
        data = json.load(f)                   # line 720 -> UnboundLocalError
        agent_card = AgentCard(**data)
    ...
    if gemini_enterprise_app_name:
        ...
        import inspect
        import json                           # line 748 -> makes `json` function-local
```

Suggested fix — remove the redundant function-local import (the module-level
`import json` at line 19 already covers it):

```diff
     import inspect
-    import json

     from google.adk.agents import Agent
```

Confirmed locally: with line 748 removed, `--a2a` mounts both routes and the card
endpoint returns `200`.

This is **distinct from** the a2a-sdk 1.x `ModuleNotFoundError: No module named
'a2a.server.apps'` startup crash noted in #6220 — that is a version-pin problem;
this is a pure Python scoping bug that fails even on the supported a2a-sdk 0.3.x
line (repro'd with a2a-sdk 0.3.26). Surfaced while building a hybrid Agent Engine
deployment (streamQuery + A2A); see companion issue #<HYBRID_ISSUE>.

The fix is the one-line diff above; the trigger is any agent folder with an
`agent.json` served via `adk api_server --a2a`.

 - Always (100%) — deterministic whenever an `agent.json` is present under `--a2a`.

AnayGarodia:
Fixes #6332

Running `adk api_server --a2a` (or `get_fast_api_app(..., a2a=True)`) silently mounts no A2A routes. The `gemini_enterprise_app_name` block near the end of `get_fast_api_app` has a redundant function-local `import json`. Because of that local import, `json` is a local variable for the entire function, so the earlier `json.load(f)` in the A2A discovery loop raises `UnboundLocalError: cannot access local variable 'json'`. That exception is caught by the loop's `except Exception`, logged as "Failed to setup A2A agent", and swallowed, so every A2A agent silently fails to register and no `/a2a/...` routes are added.

`json` is already imported at module scope, so this removes the redundant local import. (The sibling `import inspect` in that block is kept: `inspect` is not imported at module scope.)

The existing A2A test only checks `/list-apps`, so it passed despite the broken route mounting. The added regression test drives `get_fast_api_app(a2a=True)` over a temp agent directory and asserts the A2A application is actually constructed (i.e. execution gets past the `json.load` call). It fails on `main` with the `UnboundLocalError` path and passes with this change; the existing A2A tests continue to pass.

### DIFF
--- a/tests/unittests/cli/test_fast_api.py
+++ b/tests/unittests/cli/test_fast_api.py
@@ -2137,6 +2137,88 @@ def test_openapi_json_schema_accessible(test_app):
         "0.3.x-only: mocks server.apps.A2AStarletteApplication (gone in 1.x)"
     ),
 )
+def test_a2a_setup_reaches_application_build(
+    mock_session_service,
+    mock_artifact_service,
+    mock_memory_service,
+    mock_agent_loader,
+    mock_eval_sets_manager,
+    mock_eval_set_results_manager,
+    temp_agents_dir_with_a2a,
+    monkeypatch,
+):
+  """Regression test for the A2A setup silently failing.
+
+  A function-local ``import json`` made ``json`` a local for the whole
+  ``get_fast_api_app`` function, so the earlier ``json.load(agent.json)`` in
+  the A2A loop raised ``UnboundLocalError``. The surrounding ``except`` swallowed
+  it, so no A2A routes were ever mounted. This asserts the loop gets past the
+  ``json.load`` call and actually builds the A2A application.
+  """
+  with (
+      patch("signal.signal", return_value=None),
+      patch(
+          "google.adk.cli.fast_api.create_session_service_from_options",
+          return_value=mock_session_service,
+      ),
+      patch(
+          "google.adk.cli.fast_api.create_artifact_service_from_options",
+          return_value=mock_artifact_service,
+      ),
+      patch(
+          "google.adk.cli.fast_api.create_memory_service_from_options",
+          return_value=mock_memory_service,
+      ),
+      patch(
+          "google.adk.cli.fast_api.AgentLoader",
+          return_value=mock_agent_loader,
+      ),
+      patch(
+          "google.adk.cli.fast_api.LocalEvalSetsManager",
+          return_value=mock_eval_sets_manager,
+      ),
+      patch(
+          "google.adk.cli.fast_api.LocalEvalSetResultsManager",
+          return_value=mock_eval_set_results_manager,
+      ),
+      patch(
+          "google.adk.cli.fast_api._create_task_store_from_options",
+          return_value=MagicMock(),
+      ),
+      patch("google.adk.a2a.executor.a2a_agent_executor.A2aAgentExecutor"),
+      patch("a2a.server.request_handlers.DefaultRequestHandler"),
+      patch("a2a.types.AgentCard", return_value=MagicMock()),
+      patch("a2a.server.apps.A2AStarletteApplication") as mock_a2a_app,
+  ):
+    mock_app_instance = MagicMock()
+    mock_app_instance.routes.return_value = []
+    mock_a2a_app.return_value = mock_app_instance
+
+    monkeypatch.chdir(temp_agents_dir_with_a2a)
+
+    get_fast_api_app(
+        agents_dir=".",
+        web=True,
+        session_service_uri="",
+        artifact_service_uri="",
+        memory_service_uri="",
+        allow_origins=["*"],
+        a2a=True,
+        host="127.0.0.1",
+        port=8000,
+    )
+
+    # If json.load raised UnboundLocalError, execution never reaches the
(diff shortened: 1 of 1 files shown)


AnayGarodia:
@googlebot I signed it!

AnayGarodia:
@googlebot I signed it!

AnayGarodia:
@googlebot I signed it!

rohityan:
Hi @AnayGarodia , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:
Hi @wyf7107 , can you please review this. LGTM

AnayGarodia:
Rebased onto `main` to clear the conflict.

Heads-up from the rebase: the underlying `UnboundLocalError` is already resolved on `main` — commit 38587d7f (the a2a-sdk compatibility-shim refactor) removed the function-local `import json` as a side effect. So this PR now carries just the regression test that locks the behavior in. That fix landed without a test, and a future function-local `import json` would silently reintroduce the shadowing, so the guard still earns its place.

The test is skipped on the a2a-sdk v1 path (`_compat.IS_A2A_V1`), matching the sibling A2A tests that mock `A2AStarletteApplication`. Happy to retitle to `test(cli): ...` if you'd prefer given the fix itself is now upstream.

rohityan:

AnayGarodia:
rebased onto latest main, no conflicts. the src fix already landed on main so this branch now just carries the regression test. fast_api tests pass locally

AnayGarodia:
rebased onto latest main, no conflicts. fast_api tests all pass locally. the src fix already landed upstream so this is just the regression test now (skips itself on a2a 1.x)

GWeale:
thanks for the three rebases, and sorry this sat so long. the fix landed in 38587d7f and test_a2a_request_handler_uses_push_config_store already covers that json.load path, so i'm closing this rather than add a second guard.

PR #6347, opened 2026-07-08
fix: block SSRF via IPv6 addresses embedding a non-global IPv4 (NAT64/IPv4-compatible)

adilburaksen:
### Summary

`load_web_page` vets every resolved address with `_is_blocked_address`, which currently rejects an address only when `not address.is_global`. For IPv6, `is_global` on the *outer* address does not reflect the reachability of an **embedded IPv4** target:

| Address | `is_global` | Embedded IPv4 |
|---|---|---|
| `64:ff9b::169.254.169.254` (NAT64, RFC 6052) | `True` | `169.254.169.254` |
| `::169.254.169.254` (IPv4-compatible) | `True` | `169.254.169.254` |

On a network with NAT64 (common in IPv6-only / DNS64 clusters, e.g. GKE), a URL such as
`http://[64:ff9b::169.254.169.254]/computeMetadata/v1/` passes the `is_global` check and is
routed by NAT64 to the internal `169.254.169.254` metadata endpoint — an SSRF that can reach the
cloud metadata server / other link-local and private ranges.

The existing IPv4-mapped (`::ffff:a.b.c.d`) and 6to4 (`2002::/16`) forms are already reported as
non-global by `ipaddress`, so they were handled; NAT64 and the deprecated IPv4-compatible form are
the gaps.

### Fix

Add `_embedded_ipv4()` which extracts the embedded IPv4 for IPv4-mapped, 6to4, NAT64
(`64:ff9b::/96`) and IPv4-compatible addresses, and reject the outer address when the embedded
IPv4 is not globally routable. Public NAT64 targets (e.g. `64:ff9b::8.8.8.8`) remain allowed, so
there is no over-blocking of legitimate destinations.

### Scope / honesty

Reaching the internal target requires the deployment network to actually route the embedded form
(NAT64 present, or a stack that routes IPv4-compatible IPv6). Where it doesn't route, this is
defense-in-depth completing the existing SSRF filter; where NAT64 is present it closes a real
metadata-SSRF path. Added regression tests for both forms.

### DIFF
--- a/src/google/adk/tools/load_web_page.py
+++ b/src/google/adk/tools/load_web_page.py
@@ -158,8 +158,42 @@ def _is_blocked_hostname(hostname: str) -> bool:
   )
 
 
+_NAT64_WELL_KNOWN_PREFIX = ipaddress.ip_network('64:ff9b::/96')
+
+
+def _embedded_ipv4(address: _ResolvedAddress) -> ipaddress.IPv4Address | None:
+  """Returns the IPv4 address embedded in an IPv6 address, if any.
+
+  ``is_global`` on the outer IPv6 address does not reflect the reachability of
+  the embedded IPv4 target for IPv4-mapped (``::ffff:a.b.c.d``), IPv4-compatible
+  (``::a.b.c.d``), 6to4 (``2002::/16``) and NAT64 (``64:ff9b::/96``) addresses.
+  For example ``64:ff9b::169.254.169.254`` is reported as global but, on a
+  network with NAT64, routes to the internal ``169.254.169.254`` metadata
+  endpoint. Returning the embedded IPv4 lets the caller vet it directly.
+  """
+  if not isinstance(address, ipaddress.IPv6Address):
+    return None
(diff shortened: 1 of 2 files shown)


adilburaksen:
Note: the same `not address.is_global` vetting (which this PR refines to also decode embedded IPv4) is being introduced for `RestApiTool` in #6003, so it will have the same NAT64/IPv4-compatible blind spot. It may be worth factoring `_embedded_ipv4` (or the whole address-vetting routine) into a shared helper so both `load_web_page` and `RestApiTool` benefit from a single, complete implementation.

rohityan:
Hi @adilburaksen , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:

xuanyang15:
Merged with other changes in https://github.com/google/adk-python/commit/6f18257117847947953458af83788dc61c65ffbb.

PR #6370, opened 2026-07-10
fix: use mTLS transport for outbound A2A calls to Google endpoints

### LINKED ISSUE #6365: A2A fails using Agent Identity in Agent Runtime
**Describe the Bug:**

When deploying an orchestrator agent to Agent Runtime using a traditional Service Account, Agent-to-Agent (A2A) remote communication via `AgentRegistry.get_remote_a2a_agent` works as expected. However, when switching the orchestrator agent to use Agent Identity, all outbound remote A2A calls fail with a `401 Unauthenticated` error. In both scenarios, the underlying IAM permissions configured for the identities are identical.

The mTLS fixes introduced in commit 03671c6 addressed token binding for the internal `AgentRegistry` endpoints, but this transport mechanism does not appear to carry over when invoking custom remote URLs via the A2A client layer.

**Steps to Reproduce:**

1. Deploy a `google-adk` (v2.4.0) orchestrator agent using Agent Identity.
2. Deploy a `google-adk` (v2.4.0) sub-agent using Agent Identity.
3. Attempt to invoke the sub-agent from the orchestrator using `AgentRegistry.get_remote_a2a_agent(AGENT_NAME, httpx_client=httpx_client)`.
4. Observe a consistent `401 Unauthenticated` failure.
5. Switch the orchestrator agent back to a standard Service Account and observe a successful invocation.

**Expected Behavior:**

The library should natively handle Agent Identity credentials for outbound A2A communication, or the library should document how to securely bridge the bound mTLS token context to external remote A2A agents without dropping token-sharing prevention (`GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES` described [here](https://docs.cloud.google.com/iam/docs/troubleshoot-auth-manager#401-error))

**Observed Behavior:**

All outbound remote A2A calls fail with a 401 Unauthenticated error.

**Error Log:**

```
ERROR:google_adk.google.adk.agents.remote_a2a_agent:A2A request failed: HTTP Error 401: Client error '401 Unauthorized' for url https://<region>-aiplatform.googleapis.com/reasoningEngines/v1/projects/<project>/locations/<region>/reasoningEngines/<id>/api/a2a/app
```

**Environment Details:**

- google-adk==2.4.0
- Local OS: Windows
- Deployment Target: Google Agent Runtime
- Python version: 3.13.14

**Model Information:**

- Are you using LiteLLM: No

**Regression:**

- Did this work in a previous version of ADK? No.
- How often has this issue occurred?: Always (100%)

ShresthSamyak:
## Summary

A2A remote calls made via `AgentRegistry.get_remote_a2a_agent(...)` fail with
`401 UNAUTHENTICATED` when the orchestrator uses **Agent Identity**, while
working fine with a standard Service Account. This routes outbound A2A traffic
to Google endpoints over a mutual-TLS channel so channel-bound credentials are
accepted.

Fixes #6365.

## Root cause

`get_remote_a2a_agent()` builds a `RemoteA2aAgent` with only an optional
`httpx_client`, and `RemoteA2aAgent` creates a **plain** `httpx.AsyncClient` for
the actual A2A RPCs. Google context-aware access policies (used by Agent
Identity) issue access tokens that are cryptographically bound to a mutual-TLS
channel; presenting a bound token over a plain connection is rejected with
`401 UNAUTHENTICATED`.

The mTLS work in 03671c6 hardened only the AgentRegistry **control-plane** calls
(`_make_request` via `AuthorizedSession`). The A2A **data-plane** client never
carried the client certificate or a channel-bound token, so:

| Identity | Token | Over plain client | Result |
| --- | --- | --- | --- |
| Service Account | unbound bearer | accepted anywhere | works |
| Agent Identity | mTLS/DPoP channel-bound | binding check fails | 401 |

`get_mcp_toolset()` already injects Google auth for Google MCP endpoints;
`get_remote_a2a_agent()` had no equivalent for A2A. This aligns them.

## Changes

- **New `utils/_async_mtls_transport.py`** — the async google-auth mTLS
  transport (`_RefreshableAsyncCredentials`, `_GoogleAuthAsyncTransport`,
  `_SharedAsyncTransport`, `_GoogleAuthAsyncByteStream`) extracted from
  `mcp_session_manager.py`, plus a `create_google_auth_mtls_transport(url)`
  factory. Shared by both the MCP and A2A paths.
- **`mcp_session_manager.py`** — imports the transport classes from the shared
  module instead of defining them locally. No behavior change.
- **`remote_a2a_agent.py`** — new `enable_google_auth_mtls` flag; when set and
  no `httpx_client` is supplied, the default client is built on the mTLS
  transport, falling back to a plain client if mTLS cannot be negotiated.
- **`agent_registry.py`** — enables the flag for Google A2A endpoints when a
  client certificate is configured and `GOOGLE_API_USE_MTLS_ENDPOINT != "never"`.
  A caller-supplied `httpx_client` is never overridden; if it would suppress
  required mTLS, a warning is logged so the 401 is diagnosable.

## Behavior / compatibility

- Non-Google endpoints and Service Account flows are unaffected.
- A user-provided `httpx_client` is always respected (with a warning when it
  would suppress required mTLS).
- The mTLS path requires `google.auth.aio` (`google-auth[aiohttp]`); if
  unavailable, it degrades gracefully to a plain client.
- No public API removed; `enable_google_auth_mtls` is an additive,
  default-`False` parameter.

## Testing plan

Unit tests (`pytest`), all passing (230 passed):

- `tests/unittests/utils/test_async_mtls_transport.py` — factory success /
  non-mTLS / exception / no-aio paths, and token host-scoping.
- `tests/unittests/integrations/agent_registry/test_agent_registry.py` — mTLS
  enabled for Google endpoints; not enabled for non-Google, without a client
  cert, with a caller-supplied client, or when the endpoint is set to `never`;
  and that the suppressed-mTLS warning fires (and stays silent for non-Google).
- `tests/unittests/agents/test_remote_a2a_agent.py` — default client is plain
  when disabled, uses the mTLS transport when enabled, and falls back when mTLS
  is unavailable.
- `tests/unittests/tools/mcp_tool/test_mcp_session_manager.py` — unchanged MCP
  mTLS behavior after the extraction.

Additionally verified locally that `create_google_auth_mtls_transport()`
negotiates a real mutual-TLS channel and presents the client certificate to a
server that requires it, with the bearer token attached over that channel
(returns 200; a cert-less client is rejected at the handshake).

Live validation against a deployed Agent Runtime + Agent Identity setup is
pending; the success path emits
`Successfully configured mTLS using AsyncAuthorizedSession` at INFO.


ShresthSamyak:
Thanks for taking a look, @rohityan, and for merging `main` in.

I've pushed two follow-up commits to address the CI failures from this change:

- `e0157c9` — fixes the mypy and compliance checks. For mypy, I added `__all__` to the new `_async_mtls_transport` module so `AsyncAuthorizedSession` is treated as an explicit re-export, and a `# type: ignore` on the `google.auth.aio` import fallback shim. For compliance, the new tests were using incidental `googleapis.com` hostnames, which I changed to `example.com`.
- `3bebd5e` — removes a stray trailing blank line in `.agents/skills/adk-style/references/imports.md` that was tripping the `end-of-file-fixer` hook.

A few of the remaining red checks appear unrelated to this PR's diff, but please let me know if I'm missing something:

- **Unit Tests**: the failures are all `test_verify_snippets.py` erroring at collection with `FileNotFoundError` for `open_source_workspace/.agents/skills/adk-verify-snippets/scripts/verify_md.py`. That path looks specific to the internal checkout layout, so it may not resolve on the public runner — the rest of the suite passes.
- **Mypy**: the one remaining flagged line is in `models/lite_llm.py`, which this PR doesn't modify.
- **Do Not Merge on GitHub / Copybara**: the standard external-PR marker.

Happy to make any adjustments you'd like on the fix itself. Thanks again!

HonzaKopecky:
Hey guys, I am super happy to see some work being done on Agent Identity. The description of the PR says that `get_mcp_toolset()` already injects Google auth for Google MCP endpoints but I am actually experiencing exactly the same problem when attempting to use BigQuery MCP server via `AgentRegistry`. Also getting `401 UNAUTHENICATED` until I set `GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES=False`, then all is fine. Below is the source code of the agent. Changing the environment variable makes it work or break.

```python
import datetime
from zoneinfo import ZoneInfo

import os
from google.adk.integrations.agent_registry import AgentRegistry

PROJECT = os.environ.get("GOOGLE_CLOUD_PROJECT")
LOCATION = os.environ.get("GOOGLE_CLOUD_LOCATION", "global")

def project_header_provider(_) -> dict:
    return {
        "x-goog-user-project": PROJECT
    }

def get_weather(query: str) -> str:
    """Simulates a web search. Use it get information on weather.

    Args:
        query: A string containing the location to get weather information for.

    Returns:
        A string with the simulated weather information for the queried location.
    """
    if "sf" in query.lower() or "san francisco" in query.lower():
        return "It's 60 degrees and foggy."
    return "It's 90 degrees and sunny."

registry = AgentRegistry(
    project_id=PROJECT,
    location=LOCATION,
    header_provider=project_header_provider,
)

# server name of the BigQuery MCP server registered in Agent Registry
mcp_server_name = "..."
my_mcp_toolset = registry.get_mcp_toolset(mcp_server_name=mcp_server_name)

root_agent = Agent(
    name="root_agent",
    model=Gemini(
        model="gemini-flash-latest",
        retry_options=types.HttpRetryOptions(attempts=3),
    ),
    instruction="You are a helpful AI assistant designed to provide accurate and useful information.",
    tools=[get_weather, get_current_time, my_mcp_toolset],
)

app = App(
    root_agent=root_agent,
    name="app",
)
```

ShresthSamyak:
Thanks @HonzaKopecky — really useful data point.

This PR is scoped to the A2A path (`get_remote_a2a_agent`), but `get_mcp_toolset` already routes through the same underlying google-auth mTLS transport (`MCPSessionManager._get_mtls_transport`, which this PR extracts into the shared `_async_mtls_transport` module), and it attempts mTLS by default (`GOOGLE_API_USE_CLIENT_CERTIFICATE` defaults to `true`). So the fact that you still get a 401 unless you set `GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES=False` suggests the mTLS channel isn't actually being established in your runtime — i.e. the bound token is falling back to a plain connection and getting rejected. Disabling token-sharing prevention works around that by unbinding the token, but at the cost of the binding protection that Agent Identity is providing.

To help narrow it down, with INFO logging enabled (`logging.getLogger("google_adk").setLevel(logging.INFO)`), do you see `Successfully configured mTLS using AsyncAuthorizedSession` or the `Failed to configure mTLS ...` fallback in the logs? And is `GOOGLE_API_USE_CLIENT_CERTIFICATE` set in your deployment?

If it's hitting the fallback path, then the client-certificate discovery in Agent Runtime is likely the real blocker for both MCP and A2A, which would be worth surfacing to the maintainers separately from this change.

callumscoby:
Hey @ShresthSamyak, thanks so much for looking at this. 

I've tested it in Agent Runtime with Agent Identity but I still observe a 401 Unauthenticated error. I added some more verbose logging which showed that the `enable_google_auth_mtls` flag is being correctly set and that `_GoogleAuthAsyncTransport` is being used, but it's still failing.

Even though we're configured to use mTLS we look to still be calling the standard API endpoint (`https://REGION-aiplatform.googleapis.com`) rather than the mTLS version (`https://REGION-aiplatform.mtls.googleapis.com`) - as described [here](https://docs.cloud.google.com/gemini-enterprise-agent-platform/scale/runtime/agent-gateway-runtime-deploy#route-traffic).

If we overwrite the endpoint in `remote_a2a_agent.py` to use the mTLS endpoint, we observe a successful invocation e.g.

```
async def _ensure_resolved(self) -> None:
....
  if not self.description and self._agent_card.description:
    self.description = self._agent_card.description

# Resolve to use mTLS endpoint
# --- START ---
if self._enable_google_auth_mtls and self._agent_card:
  card_url = _compat.agent_card_url(self._agent_card)
  if card_url:
      parsed_url = urlparse(str(card_url))
      hostname = parsed_url.hostname
      if (
          hostname
          and hostname.endswith("googleapis.com")
          and ".mtls." not in hostname
      ):
          mtls_hostname = hostname.replace(
              "googleapis.com", "mtls.googleapis.com"
          )
          rewritten_url = str(card_url).replace(hostname, mtls_hostname)
  
          self._agent_card.url = rewritten_url
# --- END ---

# Initialize A2A client
if not self._a2a_client:
...
```

Does this sound logical?

callumscoby:
Hey @ShresthSamyak

I think this is similar to the work in [#37ca6fb](https://github.com/google/adk-python/commit/37ca6fbeee6b8f206dc206178ed7a5b60ecfc1be) released in [v2.4.0](https://github.com/google/adk-python/releases)

Except that instead of using `connectors.googleapis.com` / `connectors.mtls.googleapis.com` we should be using `aiplatform.googleapis.com` / `aiplatform.mtls.googleapis.com`

We look to do similar in `agent_registry.py` but not in `remote_a2a_agent.py`

What do you think?

ShresthSamyak:
Great catch, @callumscoby — and thank you for testing it end-to-end on Agent Runtime. Your diagnosis is exactly right: the transport was presenting the client cert and the bound token, but still dialing the standard `*-aiplatform.googleapis.com` host, so the channel-bound token was rejected. It needs to go to the `.mtls.googleapis.com` endpoint, as in the `37ca6fb` precedent you linked.

I've pushed a fix (`57140e9`) that does this. When the mTLS transport initializes for a Google A2A endpoint, it rewrites the agent card's RPC URL to its `.mtls.googleapis.com` variant and binds the transport (including its token host-scoping) to that same endpoint. Rather than rewriting inline, it reuses the existing `_mtls_utils.effective_googleapis_endpoint` helper (the same one used elsewhere), so it stays consistent with `37ca6fb`, is a no-op for non-Google hosts and when `GOOGLE_API_USE_MTLS_ENDPOINT=never`, and only applies when mTLS actually negotiates (otherwise it keeps the plain client and original endpoint). It also handles the card URL across a2a-sdk 0.3.x and 1.x.

When you get a chance, could you re-test against your Agent Runtime + Agent Identity setup on this latest commit? I'd love to confirm the 401 is resolved without needing the `GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES` workaround. Thanks again for the detailed investigation — it made this straightforward to pin down.

callumscoby:
Hey @ShresthSamyak, it works! The 401 is resolved without needing the `GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES`  workaround. Thanks a lot!

ShresthSamyak:
Amazing — thanks so much for confirming, @callumscoby! Really appreciate you taking the time to test it end-to-end on Agent Runtime + Agent Identity, both before and after the fix. Your diagnosis of the `.mtls.googleapis.com` endpoint routing is what made this straightforward to land, and it's great to have it verified in a real deployment without needing the token-sharing workaround.

Thanks also to @HonzaKopecky for the related MCP data point, and to @rohityan for keeping the branch in sync.

To summarize where this stands for reviewers: the change is verified to resolve #6365 in production, and the PR's own checks (compliance, mypy, unit tests for the changed code) are green. The remaining red checks are pre-existing and unrelated to this diff (the `test_verify_snippets.py` collection error tied to the internal `open_source_workspace/` path, and the standard Copybara external-PR marker).

@GWeale — this is ready for review whenever you have a chance. Happy to make any adjustments you'd like.

HonzaKopecky:
Hey @ShresthSamyak, thanks for looking into MCP too even though this PR is not directly about MCP. Sorry for the delay, I only work on this project once a week. 

I managed to test the code with logging enabled and I can see the `INFO:google_adk.google.adk.tools.mcp_tool.mcp_session_manager:Successfully configured mTLS using AsyncAuthorizedSession` message being logged in when the agent starts. Even after upgrading to ADK `v2.5.0` I am getting `401 UNAUTHENTICATED`.

I now tested even with the HEAD of this PR and the problem is still present. `GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES=False` still works around the problem with both versions of ADK.

`GOOGLE_API_USE_CLIENT_CERTIFICATE` is not present in my environment. 

Is there anything more I can do to help with this issue?

ShresthSamyak:
Thanks for testing with logging, @HonzaKopecky — that pins it down, and it corrects my earlier guess. The `Successfully configured mTLS using AsyncAuthorizedSession` line means the client cert and bound token are being set up correctly, so cert discovery isn't the problem. It's the same issue @callumscoby found on the A2A path: even with mTLS negotiated, the request is still sent to the standard `*.googleapis.com` endpoint rather than `*.mtls.googleapis.com`, so the channel-bound token is rejected.

On the MCP side specifically, `get_mcp_toolset` builds `StreamableHTTPConnectionParams(url=...)` with the endpoint as registered (a plain `*.googleapis.com` host) and doesn't resolve it to the mTLS variant.

Since this PR is scoped to (and verified for) the A2A path, I've opened a focused follow-up for MCP so it doesn't hold this one up: #6459. It routes the MCP endpoint through the same `_mtls_utils.effective_googleapis_endpoint` helper when a client cert is configured. If you get a chance, it'd be really helpful if you could test your BigQuery MCP setup against that branch and confirm the 401 is resolved without the `GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES` workaround — the same way @callumscoby verified the A2A fix here. Thanks again for the detailed investigation!

Issue #4233, opened 2026-01-22
Event streaming in run_async triggers repeated Pydantic serialization


ohs30359-nobuhara:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**
When running an agent via `async for event in runner.run_async(...)`,
`pydantic.BaseModel.model_dump_json()` is invoked an unexpectedly large number of times within a single user request.

This results in a noticeable performance impact (CPU time and latency), especially when large objects such as `LlmRequest`, `GenerateContentConfig`, and `Schema` are repeatedly serialized even though they appear to be immutable within the same run.

I am not certain whether this behavior is expected by design or if it is an unintended side effect of the event-driven architecture, so I am reporting this as a potential performance issue / design concern rather than a definitive bug.

**To Reproduce**
Below is a minimal reproduction that triggers the behavior.

```
from google.adk.runners import Runner
from src.agents.agent import sample_agent

runner = Runner(agent=sample_agent)

async def main():
    async for event in runner.run_async(
        input="hello",
    ):
        pass  # simply consuming events

# run with asyncio.run(main())
```

To measure the issue, I monkey-patched BaseModel.model_dump_json to record:

- call count
- latency
- serialized size

**Observed behavior**
For a single user request, `model_dump_json()` is called thousands of times.
Example aggregated metrics (per single request):

```
google.adk.events.event.Event                        n=    177
google.adk.models.llm_request.LlmRequest             n=   1886
google.adk.models.llm_response.LlmResponse           n=   3772
google.adk.sessions.session.Session                  n=     69
google.genai.types.Content                           n=     61
google.genai.types.FunctionDeclaration               n=   7048
google.genai.types.Schema                            n=  49380
google.genai.types.GenerateContentConfig             n=   1886
```
Large immutable objects (e.g. schema and config) appear to be repeatedly serialized during event emission.

**Expected behavior**
One of the following (or clarification if current behavior is intentional):
- Fewer repeated `model_dump_json()` calls for immutable objects within a single run
- Caching or memoization of serialized representations
- Documentation clarifying that this level of serialization is expected when using `run_async` with event streaming

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Linux
 - Python version(python -V): 3.13
 - ADK version(pip show google-adk): 1.21.0

 **Model Information:**
 - Are you using LiteLLM: Yes
[...]

PR #6371, opened 2026-07-10
perf(telemetry): skip non-recording LLM trace work

dexhunter:

- Related: #4233

**Problem:**

`trace_call_llm` builds telemetry configuration and serializes request and response content before writing span attributes, even when the supplied or current span is not recording. A non-recording span cannot retain those values, so large LLM payloads incur avoidable CPU work.

**Solution:**

Resolve the span first and return when `span.is_recording()` is false. Recording spans keep the existing telemetry and attribute path unchanged. The regression test covers both explicitly supplied and current spans and proves that neither request nor response serialization runs on the non-recording path.

This addresses one narrow source of the repeated serialization reported in #4233; it does not claim to eliminate every serialization path in that issue.

### Performance

The strict same-workload evaluator enables content capture and uses a large deterministic response. Each reported run is itself the median of nine benchmark batches.

| Revision | Three repeated evaluator results (us/call) | Median (us/call) |
|---|---:|---:|
| Current `main` | 153.056, 153.571, 154.278 | 153.571 |
| This PR | 0.122033, 0.122233, 0.125897 | 0.122233 |

That is a 99.92% reduction in this helper's non-recording-span overhead. This is a helper-level CPU benchmark, not an end-to-end agent or model latency claim. The evaluator also verifies that recording-span attributes and serialization behavior remain unchanged.

The early-guard direction was identified through autoresearch with Weco. The comparable baseline, patch, and hardened regression are recorded in this [public autoresearch trajectory](https://dashboard.weco.ai/share/6LpUWhPmgXt--gGyCoa0P_7v0cHGbB-M).

### Testing Plan

**Unit Tests:**

- [x] All focused unit tests pass locally.

Results on the final rebased patch:

- `pytest tests/unittests/telemetry/test_spans.py -q`: 80 passed, 1 skipped
- strict evaluator: three repeated upstream and candidate runs passed all recording/non-recording correctness gates
- `uvx ruff check --fix src/google/adk/telemetry/tracing.py tests/unittests/telemetry/test_spans.py`: passed
- repository hooks and `git diff --check`: passed

Not applicable for this tracing guard. The focused unit suite and strict evaluator exercise explicit and current spans, recording and non-recording behavior, request and response serialization, telemetry-disabled cases, tools/no-tools cases, and serialization fallback behavior without an external model request.

### Checklist

- [x] New and existing focused unit tests pass with my changes.
- [ ] I have manually tested my changes end-to-end. (Not applicable; see above.)

### Additional context

The patch is intentionally limited to the non-recording trace path; recording-span behavior remains covered and unchanged.

### DIFF
--- a/src/google/adk/telemetry/tracing.py
+++ b/src/google/adk/telemetry/tracing.py
@@ -329,10 +329,13 @@ def trace_call_llm(
     llm_request: The LLM request object.
     llm_response: The LLM response object.
   """
+  span = span or trace.get_current_span()
+  if not span.is_recording():
+    return
+
   telemetry_config = _telemetry_config_from_invocation_context(
       invocation_context
   )
-  span = span or trace.get_current_span()
   # Special standard Open Telemetry GenaI attributes that indicate
   # that this is a span related to a Generative AI system.
   span.set_attribute("gen_ai.system", "gcp.vertex.agent")
--- a/tests/unittests/telemetry/test_spans.py
+++ b/tests/unittests/telemetry/test_spans.py
@@ -41,8 +41,6 @@ from google.adk.telemetry.tracing import use_inference_span
 from google.adk.tools.base_tool import BaseTool
 from google.adk.tools.tool_context import ToolContext
 from google.genai import types
-from mcp import ClientSession as McpClientSession
-from mcp import ListToolsResult as McpListToolsResult
 from mcp import Tool as McpTool
 from opentelemetry._logs import LogRecord
 from opentelemetry.semconv._incubating.attributes.gen_ai_attributes import GEN_AI_AGENT_NAME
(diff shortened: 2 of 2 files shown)


dexhunter:
Rebased onto current `main` at `fa884e3d`. The feature diff remains limited to `tracing.py` and `test_spans.py`.

Validation after the rebase:
- `81 passed` in `tests/unittests/telemetry/test_spans.py`
- `git diff --check upstream/main` passed
- the PR branch is clean

The remaining repository-wide failures reproduce outside this patch on current `main`:
- `tests/unittests/test_verify_snippets.py` points to the missing `open_source_workspace/.agents/skills/adk-verify-snippets/scripts/verify_md.py` path, while the script is currently under `.agents/...`
- `end-of-file-fixer` modifies the current-main `.agents/skills/adk-style/references/imports.md`
- the mypy baseline comparison can report the same `lite_llm.py` enum union in a different literal order as a new error, despite identical source and equal error counts

I left those unrelated upstream issues out of this focused telemetry PR.

PR #5921, opened 2026-05-31
fix(_function_parameter_parse_util): return [] not None for empty schema properties

devteamaegis:
## What's broken

`_get_required_fields()` in `src/google/adk/tools/_function_parameter_parse_util.py` returns `None` (bare `return` statement, line 437) when `schema.properties` is falsy. The caller in `_automatic_function_calling_util.py` assigns the result directly to `declaration.parameters.required`. Any code that then iterates over `.required` raises `TypeError: 'NoneType' object is not iterable`. This fires for any function whose properties dict is empty — for example, a function that only takes `tool_context` once it's been added to `ignore_params`.

## Why it happens

`return` with no value in a `list[str]`-annotated function returns `None` instead of the expected empty list.

## Fix

Line 437: changed `return` to `return []`. One character change.

## Test

Added `test_empty_properties_required_is_empty_list_not_none` in `TestBuildFunctionDeclarationLegacy`. It calls `_get_required_fields` with an empty-properties schema and asserts the result is `[]` and is iterable without raising.

Fixes #5920

PR #6377, opened 2026-07-11
fix: Return empty list instead of None from _get_required_fields for schemas without properties

### LINKED ISSUE #5920: BUG: _get_required_fields returns None instead of [] for empty properties, causing TypeError
## What breaks

`_get_required_fields()` in `src/google/adk/tools/_function_parameter_parse_util.py` has a bare `return` statement (line 437) that returns `None` instead of an empty list when `schema.properties` is falsy (empty dict or `None`). The return value is assigned to `declaration.parameters.required` in `_automatic_function_calling_util.py`. Any subsequent code that iterates over `.required` — validation, serialization, downstream processing — raises `TypeError: 'NoneType' object is not iterable`.

This triggers whenever a function has no non-ignored parameters (e.g. a function that only accepts `tool_context` with all params in `ignore_params`).

```
TypeError: 'NoneType' object is not iterable
  File ..., line 43, in <module>
    _ = [r for r in schema.required]
                    ^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

## Root cause

Line 437: `return` (returns `None`) instead of `return []`.

## Fix

One-character change: `return` → `return []` on line 437 of `_function_parameter_parse_util.py`.

ItsMacto:

- Closes: #5920
- Related: #5921 (earlier PR proposing the same change, stalled on an unsigned CLA since May; this is independent work with regression tests)

**Problem:**
`_get_required_fields` in `tools/_function_parameter_parse_util.py` is annotated `-> list[str]` but has a bare `return` when the schema has no properties, so it returns `None`. Any caller that relies on the declared contract (e.g. iterating the result, as reported in #5920) gets `TypeError: 'NoneType' object is not iterable`. A maintainer confirmed the bare return as the culprit in the issue thread and endorsed `return []`.

**Solution:**
Return `[]` instead, honoring the declared return type. One-line change.

One honest scoping note: the two current callers in `_automatic_function_calling_util.py` guard the call behind a truthy properties dict, so today the empty branch is only reachable by calling the util directly (a function whose sole parameter is `tool_context` produces `parameters=None` rather than `required=None`). The fix corrects the contract violation at its source and hardens the util for present and future callers; the regression test pins the contract at the util level, and an added declaration-path test documents the `tool_context`-only behavior.

### Testing Plan

**Unit Tests:**

Three tests added to `tests/unittests/tools/test_build_function_declaration.py`:

- `_get_required_fields` returns `[]` (not `None`) for a schema without properties — fails on the unfixed code (`assert None == []`), passes with the fix.
- `_get_required_fields` returns the correct non-empty list for a schema with properties (guards against over-correction).
- A `tool_context`-only function's declaration has `parameters is None` (documents the actual public-path behavior).

```
# before fix:
FAILED test_get_required_fields_no_properties_returns_empty_list - assert None == []

# with fix:
$ pytest tests/unittests/tools/test_build_function_declaration.py -q
41 passed
```

`pyink --check` and `isort --check-only` are clean on both touched files.

Not applicable beyond the unit level — the change is a pure return-value correction in a private utility with no runtime surface of its own; the declaration-path test covers the integrated behavior.

### Checklist

### Additional context

None.


ItsMacto:
Note on the Mypy check failures: the flagged "new" errors are in `src/google/adk/models/lite_llm.py`, which this PR does not touch. The baseline-vs-PR error diff appears to be unstable for that file (it also only fails on 3.10–3.13, passing on 3.14). All unit tests, the linter, and A2A checks pass. Happy to rebase/retrigger if that helps.

rohityan (dismissed):

rohityan:
Hi @ItsMacto , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:
Hi @wuliang229 , can you please review this. LGTM.

PR #6379, opened 2026-07-12
fix(models): stop CompletionsHTTPClient leaking its httpx client via atexit

anxkhn:

- N/A (no existing issue). Described below following the bug report template, per
  CONTRIBUTING ("you may instead describe the bug or feature directly within the
  PR description, following the structure of our issue templates").

**Problem:**

`CompletionsHTTPClient._client` (the OpenAI-compatible HTTP client used by
`ApigeeLlm`, in `src/google/adk/models/apigee_llm.py`) registers its per-instance
`httpx.AsyncClient` for cleanup with `atexit.register(self._cleanup_client,
client)` on first access. `atexit` holds its callbacks (and their arguments) for
the entire life of the process, so this keeps a strong reference to the
`AsyncClient` and its connection pool in the process-global `atexit` registry.
Nothing ever calls `atexit.unregister`, so `close()`, `aclose()`, and `__del__`
cannot actually release the client before interpreter exit. A long-running
deployment that constructs many `CompletionsHTTPClient` / `ApigeeLlm` instances
therefore grows the `atexit` registry and leaks one open connection pool per
instance, without bound.

There is also a smaller latent issue in `_cleanup_client`: it schedules
`loop.create_task(client.aclose())` without keeping a reference to the resulting
task, so the task can be garbage collected before it finishes ("Task was
destroyed but it is pending").

**Steps to Reproduce:**

1. `pip install google-adk`
2. Run the snippet below (it exercises only the client's lifecycle; no Apigee
   credentials or network calls are needed):

```python
import gc, weakref
from google.adk.models.apigee_llm import CompletionsHTTPClient

refs = []
for _ in range(5):
    c = CompletionsHTTPClient(base_url="https://localhost")
    httpx_client = c._client   # triggers the atexit registration
    refs.append(weakref.ref(httpx_client))
    c.close()                  # explicit close
    del httpx_client
    del c

gc.collect()
alive = sum(1 for r in refs if r() is not None)
print(f"{alive}/5 httpx clients still alive after close() + del + gc")
```

**Observed Behavior (before this change):**

```text
5/5 httpx clients still alive after close() + del + gc
```

Every client (and its connection pool) is pinned by the `atexit` registry even
after an explicit `close()`, an owner drop, and a full GC.

**Expected Behavior:**

After `close()` (or once the owner is dropped and finalized), the client should
be releasable and its connection pool freed; the process should not accumulate
one retained `AsyncClient` per constructed instance.

**Solution:**

- Register the cleanup callback against `weakref.proxy(client)` so the `atexit`
  registry no longer keeps the client alive. This mirrors the pattern already
  used in `bigquery_agent_analytics_plugin.py`
  (`atexit.register(self._atexit_cleanup, weakref.proxy(batch_processor))`).
- Store the bound callback per instance
  (`self._atexit_callback = partial(self._cleanup_client, weakref.proxy(client))`)
  and unregister exactly that callback in `close()` / `aclose()`. A per-instance
  `partial` is required because `_cleanup_client` is a shared `@staticmethod`, and
  `atexit.unregister(func)` removes every registration of `func`; unregistering
  the unique partial removes only this instance's handler and leaves other
  instances untouched.
- Tolerate a dead proxy in `_cleanup_client` (catch `ReferenceError`), matching
  the same in-repo precedent.
- Retain the fire-and-forget `aclose()` task in a module-level set until it
  completes, so it is not garbage collected while pending.

**Observed Behavior (after this change):**

```text
0/5 httpx clients still alive after close() + del + gc
```

The same snippet reports `0/5` alive. The del-only path (drop the owner without
calling `close()`, relying on `__del__` + the non-pinning proxy) also reports
`0/5`.

### Testing Plan

**Unit Tests:**

Added `tests/unittests/models/test_completions_http_client.py` regression tests
covering: the register/unregister pairing on `close()` and `aclose()`, the no-op
when no client was ever created, that the client is releasable (weakref goes to
`None`) after `close()` + GC, that cleanup tolerates a dead `weakref.proxy`, and
that the pending `aclose()` task is retained until it finishes.

```
$ pytest tests/unittests/models/test_completions_http_client.py \
         tests/unittests/models/test_apigee_llm.py -q
82 passed
```

The reproduction snippet above is the manual check: `5/5` retained clients before
the change, `0/5` after. It needs no Apigee setup because it only drives the HTTP
client's construction/cleanup lifecycle.

### Checklist

### Additional context

The correct non-pinning pattern already exists in this codebase
(`bigquery_agent_analytics_plugin.py`); this change brings
`CompletionsHTTPClient` in line with it.

### DIFF
--- a/src/google/adk/models/apigee_llm.py
+++ b/src/google/adk/models/apigee_llm.py
@@ -459,7 +459,7 @@ def _validate_model_string(model: str) -> bool:
 
 # Keeps a strong reference to fire-and-forget cleanup tasks so they are not
 # garbage collected before they finish (see CompletionsHTTPClient._cleanup_client).
-_CLEANUP_TASKS: set[asyncio.Task] = set()
+_CLEANUP_TASKS: set[asyncio.Task[None]] = set()
 
 
 class CompletionsHTTPClient:
--- a/src/google/adk/models/apigee_llm.py
+++ b/src/google/adk/models/apigee_llm.py
@@ -20,6 +20,7 @@ import base64
 import collections.abc
 import enum
 from functools import cached_property
+from functools import partial
 import json
 import logging
 import os
@@ -28,6 +29,7 @@ from typing import AsyncGenerator
 from typing import Generator
 from typing import Optional
 from typing import TYPE_CHECKING
+import weakref
 
 from google.adk import version as adk_version
 from google.genai import types
@@ -455,6 +457,11 @@ def _validate_model_string(model: str) -> bool:
   return False
 
 
+# Keeps a strong reference to fire-and-forget cleanup tasks so they are not
+# garbage collected before they finish (see CompletionsHTTPClient._cleanup_client).
+_CLEANUP_TASKS: set[asyncio.Task] = set()
+
+
 class CompletionsHTTPClient:
   """A generic HTTP client for completions, compatible with OpenAI API."""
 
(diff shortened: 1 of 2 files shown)


rohityan:
Hi @anxkhn , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Can you please fix the failing CI check tests before we can proceed with the review.

anxkhn:
rebased onto current main and fixed the new mypy error by typing the retained cleanup tasks as asyncio.Task[None]. the focused tests pass (25 passed), and pre-commit passes on both changed files. @rohityan, could you please take another look?

Issue #1807, opened 2025-07-06
`_find_agent_to_run` selects an older agent instead of the most recently active one


devevignesh:
Description
The _find_agent_to_run method in the Google ADK Runner exhibits incorrect behavior when determining which agent should handle a new user message. It selects an agent from earlier in the conversation history instead of the most recently active agent, causing the conversation to be routed incorrectly.

Root Cause
The core of the issue is the current implementation of the _find_agent_to_run method.

Current Behavior:
The method searches backward through all events in the session history. It returns the first agent it encounters that meets the following criteria:

It is the root agent, OR

It is an LlmAgent that can transfer to other agents (i.e., _is_transferable_across_agent_tree returns True).

The Problem:
In a session where multiple transferable agents have been active (e.g., a x_agent followed by an y_agent), the backward search encounters the x_agent's events first because they are earlier in the history. Since both agents are transferable LlmAgents, the search incorrectly concludes that the x_agent should handle the message, even though the y_agent was the most recent one to interact with the user.

The search logic does not prioritize the most recently active agent.

Expected Behavior
The _find_agent_to_run method should identify and return the most recently active agent that is capable of handling the new message. The search should prioritize recency to ensure the conversation continues with the correct context.

**Desktop (please complete the following information):**
 - Python version(python -V): 3.13.3
 - ADK version(pip show google-adk): 1.5.0

 **Model Information:**
 For example, which model is being used.

**Additional context**

devevignesh:
@DeanChensj could you please look into this?

hangfei:
It should be fixed. Please retry with latest version. If you still find this issue, please re-open.

PR #6381, opened 2026-07-12
fix(tools): sanitize anyOf schemas for Vertex AI function declarations

### LINKED ISSUE #6373: output_schema + tools + BuiltInPlanner (thinking) fails on Vertex AI backend but works on AI Studio — anyOf schema rejection + dropped function-call result in SSE streaming
🔴 Required Information

Describe the Bug:
The same agent app runs correctly on the Gemini Developer API (AI Studio) but fails on Vertex AI (GOOGLE_GENAI_USE_VERTEXAI=1), due to two distinct problems:

Schema validation on Vertex: Pydantic output_schema and tool parameters containing Optional/Union/Literal fields serialize to JSON Schema anyOf. Vertex AI rejects these with: Unable to submit request because ... functionDeclaration ... schema didn't specify the schema type field. The identical request succeeds on AI Studio. Matches googleapis/python-genai#1807.
Dropped function calls in streaming: When output_schema is combined with function-calling tools and BuiltInPlanner thinking in SSE mode, the streaming loop can terminate before the function-call response arrives, leaving an empty string that fails Pydantic validation. Matches #3599 (currently closed) and the output_schema + tools incompatibility in #3025.

Steps to Reproduce:

Install google-adk 1.18.0 (with google-genai 1.75.0)
Define an LlmAgent with output_schema = a Pydantic model containing at least one Optional[list[str]] field, one function tool, and planner=BuiltInPlanner(thinking_config=ThinkingConfig(include_thoughts=True)) — see Minimal Reproduction Code below
Run with GOOGLE_GENAI_USE_VERTEXAI=0 (AI Studio) and send a query via SSE streaming → succeeds
Run the same query with GOOGLE_GENAI_USE_VERTEXAI=1 + GOOGLE_CLOUD_PROJECT/GOOGLE_CLOUD_LOCATION set → fails with the schema type error above (and, when tools are invoked, an empty final response that fails Pydantic validation)

Expected Behavior:
Parity between the AI Studio and Vertex AI backends for the same agent definition — Vertex should accept anyOf schemas (or ADK should emit Vertex-compatible schemas), and output_schema + tools + thinking should return the function-call result under streaming.

Observed Behavior:
Works on AI Studio; fails on Vertex AI:

Schema rejection: Unable to submit request because ... functionDeclaration ... schema didn't specify the schema type field.
When tools are called under SSE streaming with thinking enabled, the final response arrives as an empty string, which fails Pydantic validation against output_schema.

Environment Details:

ADK Library Version (pip show google-adk): 1.18.0 (google-genai 1.75.0)
Desktop OS: macOS 26.4.1 (arm64 / Apple Silicon)
Python Version (python -V): 3.12.13

Model Information:

Are you using LiteLLM: No
Which model is being used: gemini-2.5-flash

🟡 Optional Information

Regression:
Unknown — this is our first attempt at running the agent on the Vertex AI backend; the AI Studio backend has always worked with the same definition.

Logs:
Full traceback to be attached shortly — the JSON snippet in Additional Context below is the exact schema construct Vertex rejects per googleapis/python-genai#1807.

Additional Context:
Architecture: one orchestrator LlmAgent with output_schema=CoordinatorResponse, plus 4 sub-agents exposed as tools via AgentTool (each with its own output_schema); BuiltInPlanner(ThinkingConfig(include_thoughts=True)); SSE streaming.

Evidence — the Optional[list[str]] field serializes to exactly the anyOf-without-top-level-type construct that Vertex rejects per googleapis/python-genai#1807 (verified locally with pydantic's model_json_schema()):

json"sources": {
  "anyOf": [
    {"items": {"type": "string"}, "type": "array"},
    {"type": "null"}
  ],
  "default": null,
  "title": "Sources"
}

Related issues:

googleapis/python-genai#1807 — Vertex AI anyOf schema validation
#3599 — reasoning + output_schema + tools under streaming
#3025 — tools + output_schema incompatibility
#1018 — thinking_config on 2.5 models

Minimal Reproduction Code:

pythonfrom pydantic import BaseModel
from google.adk.planners import BuiltInPlanner

class CoordinatorResponse(BaseModel):
    answer: str
    sources: list[str] | None = None  # serializes to anyOf -> rejected by Vertex

def get_weather(city: str) -> dict:
    """Return a dummy weather forecast for a city."""
    return {"city": city, "forecast": "sunny"}

root_agent = LlmAgent(
    name="coordinator",
    model="gemini-2.5-flash",
    instruction="Answer the question. Use tools when needed.",
    tools=[get_weather],
    output_schema=CoordinatorResponse,
    planner=BuiltInPlanner(
        thinking_config=types.ThinkingConfig(include_thoughts=True)
    ),
)

# Run via `adk web` (SSE streaming) or a Runner.
# GOOGLE_GENAI_USE_VERTEXAI=0             -> works (AI Studio)
# GOOGLE_GENAI_USE_VERTEXAI=1 + project   -> fails (Vertex AI)

How often has this issue occurred?:

Always (100%)

ShauryaaSharma:
Pydantic serializes Optional[X]/Union fields as anyOf, which Vertex AI rejects because the wrapping schema has no top-level type field (googleapis/python-genai#1807), while AI Studio accepts it unmodified. Flatten the common Optional[X] -> anyOf: [X, {type: null}] pattern into X + nullable: true for parameters_json_schema and response_json_schema when targeting Vertex AI, gated on GoogleLLMVariant.VERTEX_AI so AI Studio behavior is unchanged.

Fixes #6373

- Closes: #6373

**Problem:**
Pydantic serializes `Optional[X]`/`Union[X, None]` fields as `anyOf`, e.g.:

```json
{"anyOf": [{"type": "array", "items": {"type": "string"}}, {"type": "null"}], "default": null}
```

Vertex AI rejects this with `functionDeclaration ... schema didn't specify the schema type field`, because the wrapping schema itself has no top-level `type` (see [googleapis/python-genai#1807](https://github.com/googleapis/python-genai/issues/1807)). The Gemini Developer API (AI Studio) accepts the same schema unmodified — this is why the same agent definition works on AI Studio but fails on Vertex, as reported in #6373.

**Solution:**
Add a Vertex-only sanitization pass that flattens the common `Optional[X]` -> `anyOf: [X, {type: null}]` pattern into `X` merged with `"nullable": true`, which Vertex accepts:

```json
{"type": "array", "items": {"type": "string"}, "default": null, "nullable": true}
```

This is applied in `build_function_declaration_with_json_schema` (`src/google/adk/tools/_function_tool_declarations.py`) to `parameters_json_schema`, `response_json_schema`, and the Pydantic-`BaseModel`-as-function path, gated strictly on `get_google_llm_variant() == GoogleLLMVariant.VERTEX_AI` so AI Studio's existing (working) behavior is unchanged. True multi-variant unions (e.g. `Union[int, str]`, no `None` branch) can't be losslessly flattened this way and are intentionally left as-is.

Scope note: this PR covers tool/function declaration schemas only — the exact construct reported in the issue, and fully within ADK's control. The `output_schema` -> `GenerateContentConfig.response_schema` path hands a raw Pydantic class to the `google-genai` SDK, which converts it internally; that's better tracked against the upstream `googleapis/python-genai#1807` issue rather than patched here. The second symptom in #6373 (dropped function-call results under SSE + `output_schema` + `BuiltInPlanner` thinking) was investigated separately — no bug was found in ADK's streaming aggregator (`streaming_utils.py`), so it's left out of this PR pending a live repro against Vertex.

### Testing Plan

**Unit Tests:**

Added `TestVertexAiAnyOfSanitization` to `tests/unittests/tools/test_function_tool_declarations.py`, covering:

- `Optional[list[str]]` primitive field -> flattened to array + `nullable: true`
- `Optional[Address]` (nested Pydantic model, `$ref`-based) -> flattened, `$ref` preserved
- A `BaseModel` passed directly (mirrors `output_schema`-shaped classes) with an `Optional[list[str]]` field -> flattened
- A true `int | str` union (no `None` branch) -> left untouched, confirming we don't overreach

All new tests are gated behind an autouse fixture setting `GOOGLE_GENAI_USE_ENTERPRISE=true`, so existing AI Studio-path tests (which assert raw `anyOf` is present) are unaffected.

```
$ pytest tests/unittests/tools/test_function_tool_declarations.py -q
44 passed, 4 warnings in 5.26s
```

Ran the full `tests/unittests/tools/` directory:

```
$ pytest tests/unittests/tools/ -q
1607 passed, 1 skipped, 5 failed
```

The 5 failures (`test_edit_file_tool.py`, `test_read_file_tool.py`) are pre-existing Windows CRLF-handling issues, unrelated to this change — confirmed identical on `main` prior to this change.

Also ran the full `tests/unittests` suite (excluding one pre-existing, unrelated collection error in `test_verify_snippets.py` that expects a repo layout not present in this checkout — reproduces identically on `main`):

```
$ pytest tests/unittests --ignore=tests/unittests/test_verify_snippets.py -q
8385 passed, 79 failed, 69 skipped, 24 xfailed, 10 xpassed
```

Spot-checked two of the 79 failures (`bigquery_agent_analytics_plugin`, `test_samples.py::test_sample[...parallel_functions...]`) against `main` and confirmed both fail identically without this change — none of the 79 touch tool/schema declaration code.

Ran `pre-commit` (`isort`, `pyink`, `addlicense`, `ruff`) against the changed files — clean (`isort` auto-fixed import ordering; re-verified tests still pass after).

Not completed — I don't have a GCP project/Vertex AI credentials available in my dev environment to run the exact repro from #6373 (an `LlmAgent` with `output_schema` + a tool + `BuiltInPlanner` thinking, run with `GOOGLE_GENAI_USE_VERTEXAI=1`) against the live API. The unit tests assert the exact schema transformation Vertex requires per the linked upstream issue, but this hasn't been confirmed against a real Vertex API call yet. If a maintainer or the issue reporter can verify against their Vertex project, that would close the loop.

### Checklist

### Additional context

`tox` multi-version testing (required by CONTRIBUTING.md before a PR) is blocked in my environment by the pre-existing `test_verify_snippets.py` collection error mentioned above, present in all 5 supported Python versions independent of this change. I ran the equivalent full-suite `pytest` invocation above as the closest substitute.


xuanyang15:
Hi @ShauryaaSharma, thanks for creating this PR! https://github.com/google/adk-python/issues/6373 is already closed, could you please verify if the issue is still reproducible with the latest version?

ShauryaaSharma:
Hi @xuanyang15, thanks for checking in.

Looking at the issue history: #6373 was closed as "completed" after a maintainer noted they couldn't reproduce it with `google-adk 2.4.0` on Vertex AI and asked the reporter to upgrade, there's no follow-up from the reporter confirming it was actually fixed.

I checked directly against current `main`: Pydantic still serializes `Optional[X]` fields as a wrapping `anyOf` schema with no top-level `type` field (same shape as originally reported), and `_function_tool_declarations.py` on `main` has no code today that sanitizes this before sending to Vertex AI. So the underlying bug is still structurally present — I don't have Vertex credentials to reconfirm the live API rejection, but the code path this PR targets is unchanged and still needed.

Separately: the Mypy/Unit Test failures on this PR were stale, not caused by this change. This branch was last synced with `main` on 2026-07-17, before two unrelated fixes landed upstream (a fix for a flaky antigravity test, and a `lite_llm.py` change that was tripping the mypy baseline-diff check). I've merged latest `main` into this branch and pushed.

percymcn:
Reimplemented `_flatten_optional_any_of` / `_sanitize_json_schema_for_vertex` verbatim from this diff and ran them on real `pydantic==2.13.4` output before commenting. The flattening works exactly as described for the scalar and container cases, and the recursion order (children first, then the parent) is right. Two things I measured that I think matter for review, both about the PR's own stated invariant — *every (sub)schema declares a top-level `type`*.

**1. `Optional[SomeModel]` still violates the invariant after the fix.**

Pydantic renders an optional nested model as a `$ref` branch, so the flatten produces a node with no `type`:

| field | raw | after `_sanitize_json_schema_for_vertex` | has top-level `type`? |
|---|---|---|---|
| `Optional[str]` | `{"anyOf":[{"type":"string"},{"type":"null"}],…}` | `{"type":"string",…,"nullable":true}` | ✅ |
| `Optional[list[str]]` | `{"anyOf":[{"type":"array",…},{"type":"null"}],…}` | `{"type":"array","items":{…},"nullable":true}` | ✅ |
| `Optional[Inner]` | `{"anyOf":[{"$ref":"#/$defs/Inner"},{"type":"null"}],…}` | `{"$ref":"#/$defs/Inner","default":null,"nullable":true}` | ❌ |
| `Optional[Union[int,str]]` | `{"anyOf":[{"type":"integer"},{"type":"string"},{"type":"null"}],…}` | unchanged (3 variants) | ❌ |

So the sanitizer runs, removes the `anyOf`, and hands back a node that still trips the condition #1807 describes. An optional sub-model is probably the most common `Optional[...]` in a real agent, so this is likely to be the shape people hit first.

**2. The test for that case can't fail on it.**

`test_optional_pydantic_model_field_flattened` covers `Optional[Address]` and asserts:

```python
self.assertNotIn("anyOf", address_schema)
self.assertIn("$ref", address_schema)
self.assertTrue(address_schema["nullable"])
```

All three pass on `{"$ref": …, "default": null, "nullable": true}` — the output above. The test asserts the *shape* of the rewrite and never asserts the *property* the rewrite exists to establish, so it currently documents the still-rejected output as correct. Worth adding `assertIn("type", …)` to each case in this class; that one line is what would have surfaced this.

**A fix for the `$ref` case, verified.** When the branch is a local `$ref`, its target is always an object schema — `$defs.Inner` is `{"properties":{…},"required":[…],"title":"Inner","type":"object"}` — so supplying `"type": "object"` alongside the `$ref` satisfies the invariant losslessly, without inlining the definition or touching `$defs`.

**And a case that genuinely has no fix here, which I'd suggest saying in the PR body.** The docstring correctly notes multi-variant unions are left untouched, but that reads as "out of scope" when the user-visible outcome is the same rejection. I checked the obvious rescue — drop the null variant and set `nullable` — and it does not help: `{"anyOf":[{"type":"integer"},{"type":"string"}],"nullable":true}` still has no top-level `type`. So `Optional[Union[int,str]]` cannot be made Vertex-safe by this approach at all, and `test_true_union_left_untouched` asserting it is untouched will keep passing while that case keeps failing.

Caveat on scope: I have no Vertex credentials, so none of the above is a fresh observation of Vertex's validator — it is all measured against the invariant this PR states and the pydantic output it operates on.

PR #6392, opened 2026-07-15
fix(code_executors): stop ContainerCodeExecutor leaking via atexit

anxkhn:

- N/A (no existing issue; described below per the issue template structure).

**Problem:**

`ContainerCodeExecutor.__init__` registers its cleanup with
`atexit.register(self.__cleanup_container)`. Because `self.__cleanup_container`
is a bound method, it holds a strong reference to the executor instance, and the
process-global `atexit` registry keeps that reference for the entire interpreter
lifetime. As a result every `ContainerCodeExecutor` that is ever created is
retained until the process exits, even after it is no longer used, and its
long-lived Docker container is only stopped at interpreter shutdown rather than
when the executor is discarded. In a long-running server that constructs an
executor per app/agent/session, both the `atexit` registry and the number of
running containers grow without bound.

**Solution:**

Register the exit handler through a `weakref.proxy(self)` instead of a bound
method, so the `atexit` registry no longer keeps the executor alive; a discarded
executor (and its container reference) can be garbage collected normally, and
`atexit` still stops the container at shutdown if the executor is still in use.
`__cleanup_container` becomes a `@staticmethod` that reads the executor through
the proxy and guards `ReferenceError`, so if the executor has already been
collected by the time the interpreter exits the handler is a safe no-op.

This mirrors the pattern already used in this repository for the same situation:
`src/google/adk/plugins/bigquery_agent_analytics_plugin.py` registers its
`atexit` cleanup with `weakref.proxy(...)` and guards the callback against
`ReferenceError`.

The change is limited to `container_code_executor.py` (add `import weakref`,
switch the registration, convert the cleanup to a `ReferenceError`-guarded
staticmethod) plus two regression tests. Behavior for a live executor is
unchanged: the container is still stopped and removed at exit.

### Testing Plan

**Unit Tests:**

Added to `tests/unittests/code_executors/test_container_code_executor.py`:

- `test_executor_is_not_retained_by_atexit`: constructs an executor (Docker
  mocked), drops it, forces a `gc.collect()`, and asserts a `weakref.ref` to it
  is `None`, i.e. the `atexit` registration no longer retains it. This test
  fails on the previous bound-method registration and passes with the fix.
- `test_cleanup_stops_and_removes_container`: invokes the cleanup while the
  executor is still referenced and asserts the container is still stopped and
  removed, confirming the exit behavior is preserved.

`pytest` results (`uv run --no-sync pytest`, Python 3.11):

```
tests/unittests/code_executors/test_container_code_executor.py .... [100%]
4 passed

tests/unittests/code_executors/  ->  80 passed
```

This is an object-lifecycle fix with no user-facing behavior change, so it is
covered by unit tests rather than a manual `adk web`/`runner` flow. The leak can
be observed directly on the previous code with Docker mocked: after constructing
and deleting a `ContainerCodeExecutor` and running `gc.collect()`, a
`weakref.ref` to it is still alive (retained by the `atexit` registry); with the
fix the reference is cleared. Running `atexit._run_exitfuncs()` after the
executor has been collected completes without raising, confirming the
`ReferenceError` guard makes a dead-proxy callback a safe no-op at shutdown.

### Checklist

- [ ] I have manually tested my changes end-to-end. (Covered by unit tests; see above.)
- [x] Any dependent changes have been merged and published in downstream modules. (None.)

### Additional context

The weakref approach is used deliberately rather than `atexit.unregister`:
`atexit.unregister` matches on the exact registered `(func, args, kwargs)` and
cannot precisely target a registration made with bound arguments, so a
weakref-based registration (observable via garbage collection) is the reliable
fix and is also the idiom already established in this repository.

### DIFF
--- a/src/google/adk/code_executors/container_code_executor.py
+++ b/src/google/adk/code_executors/container_code_executor.py
@@ -19,6 +19,7 @@ import logging
 import os
 from typing import Any
 from typing import cast
+import weakref
 
 import docker
 from docker.client import DockerClient
@@ -226,8 +227,15 @@ class ContainerCodeExecutor(BaseCodeExecutor):
     # Initialize the container.
     self.__init_container()
 
-    # Close the container when the on exit.
-    atexit.register(self.__cleanup_container)
+    # Close the container on interpreter exit. Register a weakref.proxy so the
+    # process-global atexit registry does not keep a strong reference to this
+    # executor: otherwise every ContainerCodeExecutor ever created (and its
+    # long-lived Docker container) would be retained until interpreter exit,
+    # even after the executor is no longer used. A server that builds executors
+    # per app/agent/session would leak them unboundedly.
+    atexit.register(
+        ContainerCodeExecutor.__cleanup_container, weakref.proxy(self)
+    )
 
   @override
   def execute_code(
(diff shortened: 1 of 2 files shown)


rohityan:

rohityan:
Hi @sasha-gitg , can you please review this. LGTM.

Issue #6398, opened 2026-07-15
Make EventsCompactionConfig sliding-window fields optional (enable a single trigger without a sentinel)


hiroakis:
### Is your feature request related to a specific problem?
`EventsCompactionConfig` has two independent compaction triggers:

- token-threshold: `token_threshold` + `event_retention_size` (both `Optional`, must be set together)
- sliding-window: `compaction_interval` + `overlap_size` (both **required** `int`)

Because the sliding-window fields are required, there is no clean way to enable **only** the token-threshold trigger and turn the sliding-window trigger off. Today I have to set `compaction_interval` to a large sentinel value (e.g. `1_000_000`) so the guard `len(new_invocation_ids) < compaction_interval` never fires — a hack that hides intent and is easy to get wrong (`0` does the opposite: fire every invocation).

For a cost-driven use case, the token-threshold trigger is precisely coupled to the signal we care about (prompt size), whereas the invocation-count-based sliding-window trigger fires regardless of token size and can summarize unnecessarily. So being able to enable just one trigger is genuinely useful.

### Describe the Solution You'd Like
Make `compaction_interval` + `overlap_size` an **optional pair**, mirroring the existing `token_threshold` + `event_retention_size` pair:

- `compaction_interval: Optional[int] = Field(default=None, gt=0)`
- `overlap_size: Optional[int] = Field(default=None, ge=0)`
- A `model_validator` requiring the two to be set together (or both `None`), matching `_validate_token_params`.
- Require that at least one trigger (token-threshold pair OR sliding-window pair) is configured.

This lets users opt into a single trigger system without a sentinel value:

```python
# token-threshold only
EventsCompactionConfig(
    summarizer=...,
    token_threshold=160_000,
    event_retention_size=50,
)  # no compaction_interval / overlap_size needed
```

This is a public-API change, so raising it here for direction before opening a PR. (A separate, smaller bug report covers adding `gt=0` / `ge=0` validation to the currently-required fields.)

### Impact on your work
We run an ADK-based gateway where a few long tool-heavy sessions drive the vast majority of Gemini input-token cost. We want token-threshold compaction only. Today we must pass a `1_000_000` sentinel to disable the sliding-window trigger, which is unclear to readers and a foot-gun (`0` silently means "every invocation"). An optional pair would make the intended configuration self-documenting and safe. Not time-critical, but it would simplify our config and likely help others with cost-oriented setups.

sanketpatil06:
[...]

PR #6399, opened 2026-07-15
fix(apps): validate EventsCompactionConfig.compaction_interval > 0

### LINKED ISSUE #6397: EventsCompactionConfig.compaction_interval accepts 0/negative, silently compacting on every invocation
## 🔴 Required Information

**Describe the Bug:**
`EventsCompactionConfig.compaction_interval` and `overlap_size` are declared as plain required `int`s with no value constraints, unlike their sibling fields `token_threshold` (`Field(gt=0)`) and `event_retention_size` (`Field(ge=0)`).

As a result, `compaction_interval=0` (or a negative value) is silently accepted and makes the sliding-window compaction trigger fire on **every** invocation. In `_run_compaction_for_sliding_window` (`src/google/adk/apps/compaction.py`) the guard

```python
if len(new_invocation_ids) < config.compaction_interval:
    return None  # Not enough new invocations to trigger compaction.
```

can never be true when `compaction_interval <= 0`, so compaction runs after every invocation instead of being throttled.

This is also confusing because `0` means different things for the two sliding-window fields: `overlap_size=0` is a valid "no overlap", but `compaction_interval=0` is a degenerate "fire every time" (not "disable").

**Steps to Reproduce:**
1. `pip install google-adk`
2. Run the snippet below (or construct an `App` with `events_compaction_config=EventsCompactionConfig(compaction_interval=0, overlap_size=0, summarizer=...)`).
3. Observe that construction succeeds with `compaction_interval=0`.
4. With such a config, sliding-window compaction fires on every invocation.

**Expected Behavior:**
`compaction_interval=0` (and negative values) should be rejected at construction time with a `ValidationError`, consistent with `token_threshold` (`gt=0`) and `event_retention_size` (`ge=0`). `overlap_size` should allow `0` (no overlap) but reject negatives.

**Observed Behavior:**
`EventsCompactionConfig(compaction_interval=0, overlap_size=0)` is accepted with no error, and downstream causes compaction to run on every invocation.

**Environment Details:**

 - ADK Library Version (pip show google-adk): 2.4.0
 - Desktop OS: macOS
 - Python Version (python -V): 3.10

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: N/A (configuration-validation bug, model-agnostic)

---

## 🟡 Optional Information

**Regression:**
N/A — appears to have been present since `EventsCompactionConfig` was introduced.

**Additional Context:**
Suggested fix (small, mirrors the existing validated sibling fields):
- `compaction_interval: int = Field(gt=0)`
- `overlap_size: int = Field(ge=0)`  (0 = no overlap, still valid)

Happy to send a PR with tests.

Separately (tracked as a feature request, not this bug): because `compaction_interval` / `overlap_size` are required, there is currently no clean way to enable only the token-threshold trigger without setting `compaction_interval` to a large sentinel value.

```python
from google.adk.apps.app import EventsCompactionConfig

# Accepted today (no error); makes sliding-window compaction fire every invocation.
cfg = EventsCompactionConfig(compaction_interval=0, overlap_size=0)
print(cfg.compaction_interval)  # -> 0

# Negative is also accepted today.
EventsCompactionConfig(compaction_interval=-1, overlap_size=0)
```

 - Always (100%)

hiroakis:
## Summary

Add validation to `EventsCompactionConfig` so `compaction_interval` must be `> 0` and `overlap_size` must be `>= 0`, consistent with the sibling fields `token_threshold` (`gt=0`) and `event_retention_size` (`ge=0`).

Currently `compaction_interval=0` (or negative) is silently accepted and makes the sliding-window trigger fire on every invocation, because the guard `len(new_invocation_ids) < config.compaction_interval` in `_run_compaction_for_sliding_window` can never hold.

Fixes #6397

## Changes

- `_configs.py`: `compaction_interval: int = Field(gt=0)`, `overlap_size: int = Field(ge=0)`
- Tests: reject `compaction_interval=0/-1` and `overlap_size=-1`; allow `overlap_size=0`

## Test

`pytest tests/unittests/apps/test_compaction.py` — 52 passed. pre-commit hooks pass.

## Note

This PR intentionally keeps the fields required and only adds bounds. Making the sliding-window pair optional (to disable it without a sentinel) is tracked separately in #6398.


rohityan:
Hi @hiroakis , closing this since it has been fixed in this commit https://github.com/google/adk-python/commit/ec42cfa237bc2613f62759962454aed6dea43130

PR #6292, opened 2026-07-04
fix(evaluation): Support non-English responses in ROUGE-1 matching

agharsallah:
**Please ensure you have read the [contribution guide](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md) before creating a pull request.**

### Link to Issue or Description of Change

**1. Link to an existing issue (if applicable):**

- Closes: #3111

**Problem:**

The `response_match_score` metric (`RougeEvaluator`) always returns 0 for responses in non-Latin scripts, even when the actual and expected responses are identical. The root cause is the default `rouge_score` tokenizer, which lowercases the text and then replaces every character outside `[a-z0-9]` with a space — so Thai, Chinese, Arabic, Japanese, Cyrillic, etc. tokenize to an empty token list and every comparison scores 0.

Reproduction on `main`:

```python
from rouge_score import rouge_scorer
scorer = rouge_scorer.RougeScorer(["rouge1"], use_stemmer=True)
scorer.score("สวัสดี", "สวัสดี")["rouge1"].fmeasure  # 0.0 — identical strings
scorer.score("hello", "hello")["rouge1"].fmeasure    # 1.0
```

**Solution:**

Pass a Unicode-aware tokenizer to `RougeScorer` in `final_response_match_v1.py`:

- Word characters are detected with `str.isalnum()` plus Unicode combining-mark categories (`Mn`/`Mc`), so scripts with combining vowel signs (e.g. Thai `สวัสดี`, Devanagari matras) stay intact as single tokens.
- Pure-ASCII tokens are delegated to `rouge_score`'s own `DefaultTokenizer`, so lowercasing, Porter stemming, and scores for English text are **unchanged** (guarded by an equivalence test).
- No new dependencies; `tokenizers` is re-exported through the existing `google.adk.dependencies.rouge_scorer` indirection, consistent with how `rouge_scorer` is imported today.

Known limitation (noted for reviewers): languages written without spaces (Thai, Chinese) are matched at phrase granularity rather than word granularity, since proper word segmentation would require a language-specific segmenter. This PR fixes the "identical/overlapping text scores 0" bug without adding dependencies.

### Testing Plan

**Unit Tests:**

- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

New tests in `tests/unittests/evaluation/test_final_response_match_v1.py`:

- identical non-English text scores 1.0 (Thai, Chinese, Arabic, Japanese, Russian)
[...]

PR #6401, opened 2026-07-15
fix: use Unicode-aware tokenizer in ROUGE-1 scorer to support non-English eval

guptaishaan:
The `rouge_score` library's default tokenizer strips all non-ASCII characters using the regex `[^a-z0-9]+`, which causes any non-English text (Thai, Chinese, Arabic, etc.) to be reduced to empty token lists. As a result, the ROUGE-1 f-measure always returned 0.0 for non-English text, even when the candidate and reference strings were identical.

The fix introduces `_UnicodeFriendlyTokenizer`, a custom `rouge_score.tokenizers.Tokenizer` subclass that tokenizes by simple whitespace splitting and lowercasing. This correctly handles Unicode text from any language while preserving the existing behavior for ASCII text (whitespace-split unigrams). The `tokenizers` module is now also re-exported from the `rouge_scorer` dependency shim to keep the import path consistent.

Two regression tests are added to `test_final_response_match_v1.py`: one asserting that identical Thai text scores 1.0, and one asserting that partially-overlapping Thai text yields the expected partial score.

Fixes #3111.


ftnext:
It looks like this PR substantially overlaps with #6292, which also introduces a Unicode-aware tokenizer for the ROUGE-1 evaluator.

To me, #6292 appears to preserve the existing ASCII tokenization behavior more carefully and includes broader regression coverage across multiple languages. It may be worth coordinating with that PR or clarifying the advantages of this approach before proceeding.

rohityan:
Hi @guptaishaan , Thank you for your contribution! We appreciate you taking the time to submit this pull request. 
Closing this in favor or #6292 .

Issue #6342, opened 2026-07-08
Expose session ttl / expire_time through the api_server REST layer and get_fast_api_app


roanny:
### Problem

v2.4.0 added `ttl` / `expire_time` support to `VertexAiSessionService.create_session(**kwargs)` (commit 49d4441). However, deployments that serve ADK through `get_fast_api_app` / `adk api_server` (e.g. a Cloud Run service with a web frontend that manages sessions over the REST endpoints) have **no way to set a session TTL at all**:

1. The REST create-session endpoints do not accept it. `POST /apps/{app_name}/users/{user_id}/sessions/{session_id}` accepts only `state`, and `POST /apps/{app_name}/users/{user_id}/sessions` accepts only `state` / `session_id` / `events` (`cli/api_server.py` — `create_session_with_id` / `create_session`). Neither forwards `ttl`/`expire_time` kwargs to the session service.
2. The `--auto_create_session` path creates sessions implicitly on `/run` without any TTL either.
3. There is no service-level default to fall back on: the `VertexAiSessionService` constructor takes no TTL parameter, `get_fast_api_app` accepts only a `session_service_uri` **string** (so a pre-configured/subclassed service instance cannot be injected), and the `agentengine://` URI has no TTL query parameter.

The net effect: REST-surface users cannot adopt the new TTL capability without monkey-patching private internals (the REST handler or the service factory).

### Proposed solution

Any of the following (in order of preference — they are complementary):

1. Accept optional `ttl` / `expire_time` fields in the REST create-session request bodies and forward them as kwargs to `session_service.create_session(...)`.
2. Support a **default TTL at service construction** — a `VertexAiSessionService` constructor kwarg, surfaced through a `session_service_uri` query parameter (e.g. `agentengine://<id>?ttl=7200s`) and/or a `get_fast_api_app` parameter. This would also cover the `--auto_create_session` path, which has no request body to extend.

### Alternatives considered

- Subclassing `VertexAiSessionService` with a TTL default and registering it through the service registry (`services.py` in `agents_dir` + `get_service_registry().register_session_service(...)`, which `create_session_service_from_options` consults first): **works**, but it is per-deployment custom code — a private-ish factory each deployment must maintain and re-verify on every ADK bump. First-class `ttl` support in the REST body and/or a construction-time default would serve every integration without that custom surface.
- Patching the REST handler: private API, breaks on every upgrade.
[...]

PR #6408, opened 2026-07-16
feat: Expose session ttl/expire_time through the api_server REST create-session endpoint

ItsMacto:

- Addresses: #6342 (implements option 1 from the issue, as proposed in [this comment](https://github.com/google/adk-python/issues/6342#issuecomment-4988271484); option 2, a construction-time default TTL that would also cover `--auto_create_session`, is left as a follow-up so this PR stays small)

**Problem:**
`VertexAiSessionService.create_session(**kwargs)` supports `ttl` and `expire_time`, but deployments that serve ADK through `adk api_server` / `get_fast_api_app` cannot use them. The REST create-session endpoint only accepts `state` / `session_id` / `events` and forwards nothing else, so there is no way to create an expiring session over HTTP without patching private internals.

**Solution:**
Add optional `expire_time` and `ttl` fields to `CreateSessionRequest` and forward them to the session service, with the following behavior:

- Values are passed through verbatim as strings (RFC 3339 timestamp for `expire_time`, duration like `"7200s"` for `ttl`), matching the documented kwarg usage. The Vertex API stays the authority on value formats.
- The two fields are mutually exclusive. A request setting both fails validation (422) with the same message `VertexAiSessionService` itself raises for that case.
- The fields are only forwarded when actually set, so existing callers and backends see a byte-identical service call.
- If expiration is requested but the configured session service does not accept expiration kwargs (in-memory, database), the endpoint returns 400 with a clear message instead of silently ignoring the request. Support is detected by checking whether the service's `create_session` accepts `**kwargs`, so custom services registered through the service registry work without the server having to know about them.
- The deprecated `POST .../sessions/{session_id}` endpoint is intentionally unchanged.

I kept the REST surface to an explicit whitelist of the two expiration fields rather than an open kwargs pass-through. The underlying Vertex config also accepts keys like `http_options` and `wait_for_completion` that should not be reachable by HTTP clients, and named fields keep the OpenAPI schema meaningful. Safe user-facing keys such as `display_name` and `labels` could be added the same way later if there is interest.

### Testing Plan

**Unit Tests:**

Five tests added to `tests/unittests/cli/test_fast_api.py`, using a small `InMemorySessionService` subclass that accepts and records expiration kwargs:

- `expireTime` is forwarded to a supporting session service as `expire_time`.
- `ttl` is forwarded as `ttl`.
- Setting both `ttl` and `expireTime` returns 422.
- Requesting `ttl` against a plain `InMemorySessionService` returns 400 with the not-supported message.
- Existing create-session tests cover the unchanged default path (no expiration fields, in-memory backend).

```
$ pytest tests/unittests/cli/test_fast_api.py -q
90 passed
```

`pyink --check` and `isort --check-only` are clean on both touched files, and `mypy` reports no new errors on `api_server.py` compared to main.

Not run against a live Agent Engine backend. The Vertex-side handling of `ttl` / `expire_time` is already covered by the existing `VertexAiSessionService` unit tests; this change only adds the REST plumbing in front of it, and the new unit tests exercise that seam directly through the FastAPI test client, including the 400 path.

### Checklist

### Additional context

The mutual-exclusion error message intentionally reuses the exact wording from `VertexAiSessionService.create_session` so the two layers cannot drift apart.


rohityan:
Hi @ItsMacto , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

ItsMacto:
@rohityan Wanted to bump this again

PR #6419, opened 2026-07-17
fix: handle Windows paths in `adk eval`

### LINKED ISSUE #6415: bug: `adk eval` misparses Windows absolute eval-set paths as case selectors
### Describe the bug

On Windows, `adk eval` incorrectly parses absolute eval-set file paths.

The CLI supports optionally selecting individual cases using:

```text
<eval-set path or ID>:<comma-separated case IDs>
```

The parser treats every colon as the case-selector separator, so the colon in a Windows drive prefix such as `C:\\` is mistaken for that separator.

For example, `C:\tmp\agent\eval.evalset.json` is parsed approximately as:

```python
{"C": [r"\tmp\agent\eval.evalset.json"]}
```

instead of:

```python
{r"C:\tmp\agent\eval.evalset.json": []}
```

The problem also prevents case selectors from working with Windows absolute paths.

### Steps to reproduce

```powershell
uv run pytest -q -p no:cacheprovider `
  tests/unittests/cli/utils/test_cli_tools_click.py::test_cli_eval_with_eval_set_file_path
```

The existing regression fails because the CLI writes no eval result:

```text
assert len(eval_set_results) == 1
E assert 0 == 1
```

Direct parser demonstration:

```powershell
uv run python -c "from google.adk.cli.cli_eval import parse_and_get_evals_to_run as parse; inputs=[r'C:\tmp\agent\eval.evalset.json',r'C:\tmp\agent\eval.evalset.json:case1',r'C:\tmp\agent\eval.evalset.json:case1,case2']; [print(repr(value), '->', repr(parse([value]))) for value in inputs]"
```

### Actual result

```text
'C:\\tmp\\agent\\eval.evalset.json'
-> {'C': ['\\tmp\\agent\\eval.evalset.json']}

'C:\\tmp\\agent\\eval.evalset.json:case1'
-> {'C': ['\\tmp\\agent\\eval.evalset.json']}

'C:\\tmp\\agent\\eval.evalset.json:case1,case2'
-> {'C': ['\\tmp\\agent\\eval.evalset.json']}
```

The drive letter becomes the eval-set identifier and requested case IDs are lost.

### Expected result

```python
{r"C:\tmp\agent\eval.evalset.json": []}
{r"C:\tmp\agent\eval.evalset.json": ["case1"]}
{r"C:\tmp\agent\eval.evalset.json": ["case1", "case2"]}
```

Existing behavior should remain unchanged:

```python
{"/tmp/agent/eval.evalset.json": ["case1", "case2"]}
{"my_eval_set": ["case1", "case2"]}
```

### Environment

- OS: Windows 10, build 19045
- Python: 3.11.15
- uv: 0.11.29
- Local `google-adk`: 2.4.0
- Latest upstream `main`
- Commit: `4a84d8a459853b039d3774da27e00b57b4d51aa5`

No model, credentials, or external service is involved.

### Root cause

`parse_and_get_evals_to_run()` uses unrestricted splitting:

```python
eval_set = input_eval_set.split(":")[0]
evals = input_eval_set.split(":")[1].split(",")
```

The first colon in a Windows path is the drive-letter separator. When a selector colon is also present, that selector is beyond index 1 and is ignored.

### Proposed narrow solution

Make parsing aware of a leading Windows drive prefix:

1. Preserve the drive-designator colon.
2. Split only at a subsequent selector delimiter.
3. Preserve the full drive-qualified path when no selector is present.
4. Keep POSIX-path and eval-set-ID behavior unchanged.

The solution should not depend solely on the file already existing, so nonexistent paths retain useful file-not-found diagnostics.

### Regression-test plan

Add platform-independent parser tests for Windows absolute paths without selectors, with one selector, and with multiple selectors, plus POSIX paths and eval-set IDs with selectors. Use Windows-style strings even on Linux so existing CI covers the regression. Retain the existing CLI-level test shown above.

### Conflict check

I searched open and closed issues and pull requests for the parser name, Windows absolute eval-set paths, `adk eval` Windows paths, drive-letter colon parsing, and the focused test name. I found no issue or PR addressing this root cause.

PR #6166 is a separate Windows wildcard-expansion change and explicitly notes that this eval test fails independently on unmodified `main`. Issue #2602 concerns pytest-driven result persistence and has a different root cause.

### Contribution

I am willing to contribute a small, focused fix with these regression tests after maintainer approval.

sgoel2be24-cyber:
## Summary

- detect Windows drive-qualified eval-set paths before parsing case selectors
- search for the selector delimiter only after the drive prefix
- preserve existing POSIX-path, eval-set-ID, blank-selector, multiple-input,
  and duplicate-merging behavior
- add platform-independent regression coverage for Windows path variants

Fixes #6415

## Root cause

`parse_and_get_evals_to_run()` treated the colon in a Windows drive prefix
such as `C:\` as the eval-case selector separator. This reduced the eval-set
path to `C` and discarded selectors appearing after a later colon.

## Testing

- `9 passed` — focused parser regression cases
- `1 passed` — existing
  `test_cli_eval_with_eval_set_file_path` regression
- `11 passed` — complete parser test module
- `42 passed` — adjacent CLI test module
- Ruff passed
- isort check passed
- Pyink check passed
- `git diff --check` passed

The complete repository suite was not rerun because the native Windows
baseline contains known unrelated platform failures.

Two aggregate pre-commit hooks use POSIX executable/shebang paths that native
Windows could not launch directly. Their underlying checks passed through the
repository’s available Git Bash and uv routes; no hook configuration was
changed.

## Scope

The change is limited to the eval-set parser and focused unit tests. It does
not depend on filesystem existence and does not change evaluation execution,
result storage, or CLI syntax.

### DIFF
--- a/src/google/adk/cli/cli_eval.py
+++ b/src/google/adk/cli/cli_eval.py
@@ -117,15 +117,27 @@ def parse_and_get_evals_to_run(
   eval_set_to_evals: dict[str, list[str]] = {}
   for input_eval_set in evals_to_run_info:
     evals = []
-    if ":" not in input_eval_set:
+    drive_letter = input_eval_set[:1]
+    has_windows_drive_prefix = (
+        len(input_eval_set) >= 3
+        and drive_letter.isascii()
+        and drive_letter.isalpha()
+        and input_eval_set[1] == ":"
+        and input_eval_set[2] in ("\\", "/")
+    )
+    selector_separator_index = input_eval_set.find(
+        ":", 3 if has_windows_drive_prefix else 0
+    )
+    if selector_separator_index == -1:
       # We don't have any eval cases specified. This would be the case where the
       # the user wants to run all eval cases in the eval set.
       eval_set = input_eval_set
     else:
       # There are eval cases that we need to parse. The user wants to run
       # specific eval cases from the eval set.
-      eval_set = input_eval_set.split(":")[0]
-      evals = input_eval_set.split(":")[1].split(",")
+      eval_set = input_eval_set[:selector_separator_index]
+      selector_list = input_eval_set[selector_separator_index + 1 :]
+      evals = selector_list.split(":")[0].split(",")
       evals = [s for s in evals if s.strip()]
 
     if eval_set not in eval_set_to_evals:
--- a/tests/unittests/cli/utils/test_cli_eval.py
+++ b/tests/unittests/cli/utils/test_cli_eval.py
@@ -19,6 +19,67 @@ from __future__ import annotations
 from types import SimpleNamespace
 from unittest import mock
 
+import pytest
+
+
+@pytest.mark.parametrize(
+    ("input_eval_set", "expected"),
+    [
+        pytest.param(
+            r"C:\tmp\agent\eval.evalset.json",
+            {r"C:\tmp\agent\eval.evalset.json": []},
+            id="windows-backslash-path-without-selectors",
+        ),
+        pytest.param(
+            r"C:\tmp\agent\eval.evalset.json:case1",
+            {r"C:\tmp\agent\eval.evalset.json": ["case1"]},
+            id="windows-backslash-path-with-one-selector",
+        ),
+        pytest.param(
+            r"C:\tmp\agent\eval.evalset.json:case1,case2",
+            {r"C:\tmp\agent\eval.evalset.json": ["case1", "case2"]},
+            id="windows-backslash-path-with-multiple-selectors",
+        ),
+        pytest.param(
+            "C:/tmp/agent/eval.evalset.json:case1",
+            {"C:/tmp/agent/eval.evalset.json": ["case1"]},
+            id="windows-forward-slash-path",
+        ),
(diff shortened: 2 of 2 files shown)


sgoel2be24-cyber:
Thanks for updating the branch with the latest `main`.

I checked the resulting CI failures. All unit-test jobs fail on the same unrelated Antigravity test, `test_resumed_replayed_steps_are_skipped`, because its `sess_456_agy` fixture is shorter than the current 32-character `conversation_id` requirement.

Current upstream `main` fails the same test with the same validation error across Python 3.10–3.14, and this PR does not modify the affected Antigravity source or test files.

On the updated PR head, the contribution-specific checks still pass:

- 9 focused parser regression tests
- existing Windows CLI regression
- complete parser module: 11 tests
- adjacent CLI module: 42 tests
- Ruff, isort, Pyink, and `git diff --check`

The triage-agent failure also appears transient because Gemini returned `503 UNAVAILABLE`. I don’t see any contributor-side code change needed for this PR.

PR #6432, opened 2026-07-20
fix(sessions): order sqlite sessions deterministically by update_time

### LINKED ISSUE #6431: SqliteSessionService.list_sessions() has the same missing ORDER BY as #6272
Describe the bug:
 Same bug class as #6272, in the SQLite backend. SqliteSessionService.list_sessions() (src/google/adk/sessions/sqlite_session_service.py, L298–315) fetches sessions with two raw SELECT ... WHERE queries and no ORDER BY, so row order is whatever the query plan returns and sessions[-1] for "latest session" is undefined behavior. The existing sqlite tests compare order-blind, so this was never caught.

 #6276 fixes this for DatabaseSessionService only  its diff doesn't touch the sqlite service, leaving the two backends with inconsistent contracts.

 To Reproduce:
 Create sessions A, B, C in that order, then update A so it has the newest update_time. list_sessions() returns:
 Returned order:
   id=A  last_update_time=1784563266.526278
   id=B  last_update_time=1784563266.420852
   id=C  last_update_time=1784563266.473127

 Returned id order:            ['A', 'B', 'C']
 Expected ascending-time order: ['B', 'C', 'A']
 A is last (newest activity)?   False
Order matches update_time asc? False
 Order follows insertion (primary-key scan), not activity.

 Expected behavior:
 ORDER BY update_time ASC with stable tie-breakers, the contract per the discussion in #6272, implemented in #6276 (update_time rather than create_time, since orphan sessions can be newly created but stale in activity). update_time is already in the SELECT column list.

PR incoming.

Q01P:
Fixes #6431

## Summary
- order `SqliteSessionService.list_sessions()` by `update_time ASC` with stable `user_id` / `id` tie-breakers, appended to both the user-scoped and app-wide `SELECT` queries
- add a regression where a newly created but stale orphan session must sort before older-but-active sessions, and tied `update_time` values fall back to `id`

## Why
`list_sessions()` previously executed without an `ORDER BY`, so consumers using `sessions[-1]` for the most recently active session could receive an arbitrary row depending on the SQLite query plan (in practice, primary-key/insertion order).

This mirrors the ordering contract settled in #6272 and implemented in #6276 for `DatabaseSessionService`; #6276's diff doesn't touch the sqlite backend, leaving the two backends inconsistent. Ordering follows `update_time` (not `create_time`) because orphan sessions can be newly created while having stale activity — `update_time` maps to the returned `Session.last_update_time`. If that contract changes in review, this should follow.

## Testing
- RED before fix: `uv run --extra test pytest tests/unittests/sessions/test_session_service.py::test_sqlite_session_service_list_sessions_orders_by_update_time_then_id -q`
  ```
  E     AssertionError: assert ['active_a', ...le', 'orphan'] == ['orphan', 'm...', 'active_b']
  E       At index 0 diff: 'active_a' != 'orphan'
  1 failed, 4 warnings in 2.80s
  ```
- GREEN after fix: same focused test → `1 passed, 4 warnings in 2.02s`
- `uv run --extra test pytest tests/unittests/sessions/test_session_service.py -q` → `136 passed, 5 warnings`
- `uv run --extra dev pyink --check src/google/adk/sessions/sqlite_session_service.py tests/unittests/sessions/test_session_service.py` → `2 files would be left unchanged`
- `uv run --extra dev ruff check src/google/adk/sessions/sqlite_session_service.py tests/unittests/sessions/test_session_service.py` → `All checks passed!`
- `git diff --check` → clean


rohityan:
Hi @Q01P , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:
Hi @sasha-gitg , can you please review this. LGTM

DeanChensj:
Hi @Q01P , thanks for your PR! The same functionality has been achieved by https://github.com/google/adk-python/commit/53e1afbcb5b211ad097985594a038bab49e2f509 as a subset, thus closing this for now.  Any future contribution is still highly welcome!

PR #6441, opened 2026-07-22
fix: disable stemmer in RougeScorer to support non-English eval

Yashas1600:
The `_calculate_rouge_1_scores` function in `final_response_match_v1.py` was initializing `rouge_scorer.RougeScorer` with `use_stemmer=True`. The NLTK PorterStemmer used internally only handles ASCII/Latin characters; for non-English scripts such as Thai, Arabic, or Chinese, it strips or corrupts tokens, resulting in zero overlap and a `Match score: 0` even when the candidate and reference are identical.

Setting `use_stemmer=False` preserves the original tokens for all languages, allowing ROUGE-1 to correctly measure unigram overlap for non-English text.

Fixes #3111


rohityan:
Hi @Yashas1600 , Thank you for your contribution! We appreciate you taking the time to submit this pull request. However closing this PR in favor of #6292 .

PR #6443, opened 2026-07-22
fix(eval): use Unicode-aware tokenizer in ROUGE scorer to support non-English text

guptaishaan:
The `_calculate_rouge_1_scores` function used `rouge_scorer.RougeScorer` with `use_stemmer=True` and no custom tokenizer, which defaults to the library's `DefaultTokenizer`. That tokenizer strips all non-ASCII characters via the regex `[^a-z0-9]+`, so any non-Latin text (Thai, Chinese, Arabic, etc.) is reduced to empty tokens and always scores 0.

The fix introduces a `_UnicodeTokenizer` class that splits on whitespace and retains tokens containing at least one Unicode letter or digit (using `unicodedata.category`). This preserves Thai words (including their combining vowel marks), as well as text in other scripts, while still working correctly for ASCII languages.

A regression test `test_calculate_rouge_1_scores_non_english_identical` is added to verify that identical Thai text produces a perfect ROUGE-1 score of 1.0.

Fixes #3111


rohityan:
Hi @guptaishaan , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Closing this solution in favor of #6292 as it fixes Issue #3111 for all non-English languages while delegating ASCII tokens to `DefaultTokenizer`, preserving 100% backward compatibility for English evaluations.

PR #6444, opened 2026-07-22
fix: handle non-ASCII text in ROUGE-1 eval scorer

guptaishaan:
The `rouge_score` library's default tokenizer matches only ASCII alphanumeric
characters (`[a-z0-9]+`) and silently discards every other character. As a
result, text written in non-Latin scripts such as Thai (`สวัสดี`), Chinese,
Arabic, or any other language that uses non-ASCII code points is tokenized to
an empty list, which causes `_calculate_rouge_1_scores` to return a score of
0.0 even when the candidate and reference strings are identical.

The fix introduces a small `_UnicodeTokenizer` class that extends
`rouge_score.tokenizers.Tokenizer`. Its `tokenize` method uses the regex
`[a-z0-9]+|[^\x00-\x7f]+` to preserve contiguous runs of non-ASCII characters
as individual tokens while keeping the same ASCII-alphanumeric filtering for
Latin text. The `_calculate_rouge_1_scores` helper is updated to pass this
tokenizer to `RougeScorer` instead of using `use_stemmer=True` with the
default ASCII-only tokenizer. Two regression tests for identical and
non-matching Thai strings are added to the existing test module.

Fixes #3111


rohityan:
Hi @guptaishaan ,Thank you for your contribution! We appreciate you taking the time to submit this pull request. Closing this in favor of solution in PR #6292

PR #6446, opened 2026-07-22
fix(eval): use Unicode-aware tokenizer in RougeEvaluator to support non-English languages

guptaishaan:
The `rouge_score` library's default tokenizer filters tokens through the regex
`[^a-z0-9]+`, which strips every character outside the ASCII alphanumeric range.
As a result, text written in non-English scripts (Thai, Chinese, Arabic, etc.)
is reduced to an empty token list, causing `_calculate_rouge_1_scores` to always
return 0 regardless of whether the candidate and reference are identical.

This PR introduces `_UnicodeTokenizer`, a minimal tokenizer that splits on
whitespace and lowercases text while preserving Unicode characters.  It is
passed to `RougeScorer` via the existing `tokenizer` constructor parameter,
replacing the ASCII-only default behaviour.  No changes are required to
the `rouge_score` dependency version, and English-language evaluation is
unaffected.

Two regression tests are added: one verifying that identical Thai text scores
1.0, and one verifying that non-overlapping Thai tokens score 0.0.

Fixes #3111


rohityan:
Hi @guptaishaan , Thank you for your contribution! We appreciate you taking the time to submit this pull request.
Closing this in favor of solution in PR #6292 .

PR #6459, opened 2026-07-23
fix: route MCP calls to the .mtls.googleapis.com endpoint for Agent Identity

ShresthSamyak:
## Summary

MCP tools obtained via `AgentRegistry.get_mcp_toolset(...)` fail with
`401 UNAUTHENTICATED` when the app uses **Agent Identity**, unless
`GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES=False` is set (which
works around it by unbinding the token). This routes Google-hosted MCP servers
to the mutual-TLS endpoint so the channel-bound token is accepted.

Related to #6365 and follow-up to #6370 (which fixes the same class of bug on
the A2A path).

## Root cause

`MCPSessionManager` already negotiates an mTLS transport for Google MCP
endpoints (the `Successfully configured mTLS using AsyncAuthorizedSession` log
confirms the client certificate and channel-bound token are set up). However,
`get_mcp_toolset` builds `StreamableHTTPConnectionParams(url=endpoint_uri)` with
the endpoint exactly as registered — a standard `*.googleapis.com` host. A
channel-bound (Agent Identity) token is only honored on the corresponding
`*.mtls.googleapis.com` endpoint, so presenting it to the standard endpoint
still returns 401 even though mTLS was negotiated.

This is the same issue fixed for A2A in #6370, and mirrors the endpoint
resolution added for Application Integration tools in 37ca6fb.

## Change

`agent_registry.py` — in `get_mcp_toolset`, when a client certificate is
configured (`_use_client_cert_effective()`), resolve the MCP endpoint through
the existing `_mtls_utils.effective_googleapis_endpoint` helper so it uses the
`*.mtls.googleapis.com` variant. It is a no-op for non-Google hosts,
already-mTLS hosts, and when `GOOGLE_API_USE_MTLS_ENDPOINT=never`, so Service
Account and non-Google flows are unaffected.

## Testing plan

Unit tests (`pytest`), all passing:

- `test_get_mcp_toolset_uses_mtls_endpoint_for_google` — the connection targets
  the `.mtls.googleapis.com` endpoint when a client cert is configured.
- `test_get_mcp_toolset_keeps_plain_endpoint_without_client_cert` — the plain
  endpoint is kept when no client cert is configured.
- `test_get_mcp_toolset_keeps_non_google_endpoint` — non-Google endpoints are
  left unchanged.

Live validation against a deployed Agent Runtime + Agent Identity setup with a
registered BigQuery MCP server is welcome; the reporter in #6370 observed the
401 on this exact path.

### DIFF
--- a/src/google/adk/integrations/agent_registry/agent_registry.py
+++ b/src/google/adk/integrations/agent_registry/agent_registry.py
@@ -39,6 +39,7 @@ from google.adk.tools.mcp_tool.mcp_session_manager import SseConnectionParams
 from google.adk.tools.mcp_tool.mcp_session_manager import StdioConnectionParams
 from google.adk.tools.mcp_tool.mcp_session_manager import StreamableHTTPConnectionParams
 from google.adk.tools.mcp_tool.mcp_toolset import McpToolset
+from google.adk.utils._mtls_utils import effective_googleapis_endpoint
 import google.auth
 from google.auth.transport import mtls
 from google.auth.transport import requests as requests_auth
@@ -421,6 +422,15 @@ class AgentRegistry:
       raise ValueError(
           f"MCP Server endpoint URI not found for: {mcp_server_name}"
       )
+
+    # Route Google-hosted MCP servers to the mutual-TLS endpoint when a client
+    # certificate is configured. A channel-bound (Agent Identity) access token
+    # is only accepted on *.mtls.googleapis.com; presenting it to the standard
+    # endpoint returns 401 UNAUTHENTICATED even though the mTLS transport is
+    # negotiated. effective_googleapis_endpoint is a no-op for non-Google hosts,
+    # already-mTLS hosts, and when GOOGLE_API_USE_MTLS_ENDPOINT=never.
+    if _use_client_cert_effective():
+      endpoint_uri = effective_googleapis_endpoint(endpoint_uri)
 
     if mcp_server_id and not auth_scheme:
       try:
--- a/src/google/adk/integrations/agent_registry/agent_registry.py
+++ b/src/google/adk/integrations/agent_registry/agent_registry.py
@@ -223,6 +223,9 @@ class AgentRegistry:
           else None
       )
       self._session.configure_mtls_channel(client_cert_source)
+    # Retained so endpoint selection for MCP/A2A targets can mirror the base-URL
+    # policy: `auto` uses the mTLS endpoint only when a cert is actually present.
+    self._client_cert_source = client_cert_source
     self._base_url = _get_agent_registry_base_url(client_cert_source)
 
   def _get_auth_headers(self) -> Dict[str, str]:
@@ -423,13 +426,15 @@ class AgentRegistry:
           f"MCP Server endpoint URI not found for: {mcp_server_name}"
       )
 
-    # Route Google-hosted MCP servers to the mutual-TLS endpoint when a client
-    # certificate is configured. A channel-bound (Agent Identity) access token
-    # is only accepted on *.mtls.googleapis.com; presenting it to the standard
-    # endpoint returns 401 UNAUTHENTICATED even though the mTLS transport is
-    # negotiated. effective_googleapis_endpoint is a no-op for non-Google hosts,
-    # already-mTLS hosts, and when GOOGLE_API_USE_MTLS_ENDPOINT=never.
-    if _use_client_cert_effective():
+    # Route Google-hosted MCP servers to the mutual-TLS endpoint following the
+    # same policy as the registry base URL. A channel-bound (Agent Identity)
+    # access token is only accepted on *.mtls.googleapis.com; presenting it to
+    # the standard endpoint returns 401 even though mTLS was negotiated. Gating
+    # on the resolved client cert (not just the cert-use policy) avoids pointing
+    # a plain client at the mTLS host when `auto` is set but no cert is present.
+    # effective_googleapis_endpoint is itself a no-op for non-Google hosts and
+    # already-mTLS hosts.
+    if _should_use_mtls_endpoint(self._client_cert_source):
       endpoint_uri = effective_googleapis_endpoint(endpoint_uri)
 
     if mcp_server_id and not auth_scheme:
@@ -647,8 +652,15 @@ def _use_client_cert_effective() -> bool:
     return use_client_cert_str == "true"
 
 
-def _get_agent_registry_base_url(client_cert_source: Any | None = None) -> str:
-  """Returns the base URL based on mTLS configuration and cert availability."""
+def _should_use_mtls_endpoint(client_cert_source: Any | None) -> bool:
+  """Whether Google endpoints should resolve to their .mtls.googleapis.com host.
+
+  Follows the GOOGLE_API_USE_MTLS_ENDPOINT policy (AIP-4114): ``always`` selects
+  the mTLS endpoint unconditionally, ``auto`` only when a client certificate is
+  actually available (``client_cert_source is not None``), and ``never`` never.
+  Using cert *availability* rather than cert-use *policy* avoids pointing a
+  plain client (mTLS fell back) at an mTLS-only host.
+  """
   use_mtls_endpoint_str = os.getenv(
       "GOOGLE_API_USE_MTLS_ENDPOINT", _MtlsEndpoint.AUTO.value
   ).lower()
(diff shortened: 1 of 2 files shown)


src/google/adk/integrations/agent_registry/agent_registry.py
```diff
@@ -422,6 +423,15 @@ def get_mcp_toolset(
...
       )
 
+    # Route Google-hosted MCP servers to the mutual-TLS endpoint when a client
+    # certificate is configured. A channel-bound (Agent Identity) access token
+    # is only accepted on *.mtls.googleapis.com; presenting it to the standard
+    # endpoint returns 401 UNAUTHENTICATED even though the mTLS transport is
+    # negotiated. effective_googleapis_endpoint is a no-op for non-Google hosts,
+    # already-mTLS hosts, and when GOOGLE_API_USE_MTLS_ENDPOINT=never.
+    if _use_client_cert_effective():
+      endpoint_uri = effective_googleapis_endpoint(endpoint_uri)
```
gnanirahulnutakki, line 438: **Endpoint selection needs the resolved certificate/transport state, not only the cert-use policy.** `_use_client_cert_effective()` says whether certificate use is enabled; it does not prove that a certificate is available. I reproduced this with `GOOGLE_API_USE_CLIENT_CERTIFICATE=true`, `GOOGLE_API_USE_MTLS_ENDPOINT=auto`, and `mtls.has_default_client_cert_source() == false`: `AgentRegistry._base_url` correctly remains `https://agentregistry.googleapis.com/v1`, but the MCP connection becomes `https://us-central1-aiplatform.mtls.googleapis.com/mcp`. `MCPSessionManager` then falls back to its plain client when mTLS configuration returns no transport, leaving a non-mTLS client pointed at the mTLS host.

The opposite edge also misses the documented policy: with `GOOGLE_API_USE_MTLS_ENDPOINT=always` and client-certificate use disabled, this gate never calls the resolver and leaves the regular host. [AIP-4114](https://google.aip.dev/auth/4114) defines `auto` as mTLS only when a certificate is available, while `always` selects the mTLS endpoint independently.

Could this preserve the full matrix—`always` rewrite, `auto` rewrite only after mTLS actually negotiates, and `never` retain the original URL—and add regression cases for `auto`/no-cert and `always`/cert-disabled? The A2A follow-up in #6370 already keeps its original endpoint when transport negotiation returns `None`, which looks like the useful precedent here.

gnanirahulnutakki (commented):
I found one endpoint-selection edge case in the new routing. The focused Agent Registry and mTLS utility suites pass locally (80 tests); the current red unit-test matrix is failing on unrelated `google-antigravity`/protobuf version skew. Details and a suggested policy matrix are inline.

ShresthSamyak:
Thanks for the careful review, @gnanirahulnutakki — you're right on both counts, and the base-URL policy is the correct reference. I gated on `_use_client_cert_effective()` (the cert-*use* policy) instead of cert *availability*, which mishandled exactly the two cases you called out.

I've pushed a fix that routes the MCP endpoint through the same policy the registry base URL already uses:

- Extracted the base-URL decision into `_should_use_mtls_endpoint(client_cert_source)` and made `_get_agent_registry_base_url` delegate to it (so the two can't drift).
- `AgentRegistry.__init__` now retains the resolved `client_cert_source`, and `get_mcp_toolset` gates the endpoint rewrite on `_should_use_mtls_endpoint(self._client_cert_source)`.

That gives the full matrix:
- **always** → mTLS endpoint unconditionally (your cert-disabled case now rewrites);
- **auto** → mTLS endpoint only when a certificate is actually available (`client_cert_source is not None`), so `auto` + no cert now correctly stays on the standard host instead of pointing a plain client (after the mTLS fallback) at the mTLS host;
- **never** → original URL.

Added regression tests for all four Google cases (auto+cert, auto+no-cert, always+no-cert, never+cert) plus the non-Google no-op.

On the A2A precedent — agreed it's the right model; #6370 keys off actual transport negotiation returning `None`. Here I keyed off the resolved `client_cert_source` so MCP endpoint selection stays consistent with `_get_agent_registry_base_url` in the same class, which matches your reproduction (`has_default_client_cert_source() == false` → no rewrite). Happy to align them further if you'd prefer.

(The red unit-test matrix is unrelated — the `integrations/antigravity_agent` sample is failing on a google-antigravity Protobuf gencode/runtime version skew on current `main`; the Agent Registry and mTLS suites pass.)

tests/unittests/integrations/agent_registry/test_agent_registry.py
```diff
@@ -196,6 +196,69 @@ async def test_get_mcp_toolset_handles_missing_destination_id(
...
+      self, registry
+  ):
+    # always -> mTLS endpoint unconditionally, even without a cert.
+    self._stub_mcp_server(registry, self._GOOGLE_MCP_URL)
+    registry._client_cert_source = None
+    with patch.dict(os.environ, {"GOOGLE_API_USE_MTLS_ENDPOINT": "always"}):
+      toolset = registry.get_mcp_toolset("test-mcp-server")
+    assert toolset.connection_params.url == self._GOOGLE_MCP_MTLS_URL
+
+  def test_get_mcp_toolset_never_keeps_plain_endpoint_with_cert(
```
gnanirahulnutakki: Non-blocking formatter nit: the repository-pinned `pyink==25.12` reports this signature and rewrites it onto one line (`def test_get_mcp_toolset_never_keeps_plain_endpoint_with_cert(self, registry):`). The rest of the focused style checks pass.

gnanirahulnutakki (commented):
The updated endpoint policy now matches the full `always` / `auto` / `never` matrix. I checked the author delta at `c6d04107`, ran the complete focused Agent Registry suite (52 passed), and verified Ruff, isort, and diff checks. One non-blocking `pyink` formatting nit remains inline; I found no functional concerns.

src/google/adk/integrations/agent_registry/agent_registry.py
gnanirahulnutakki: Verified on `c6d04107`: the MCP path now shares the base-URL policy, `auto` is gated on the resolved certificate source, and the regression matrix covers `auto` with/without a cert plus `always`, `never`, and non-Google endpoints. The full focused Agent Registry suite passes locally (52 tests). This addresses the original finding—thank you for the thorough update.

ShresthSamyak:
Thanks @gnanirahulnutakki — and glad the endpoint policy matrix looks right now. Good catch on the formatting nit; I ran the pinned `pyink==25.12` and it collapsed that signature onto one line as you noted. Pushed in 2f0043e (`pyink` and `isort` now clean on the file).

gnanirahulnutakki (commented):
Rechecked the formatter follow-up at `2f0043ee`. It is exactly the `pyink` normalization noted in the prior review; the two changed files now pass Pyink, isort, Ruff, and diff checks, and the full focused Agent Registry suite remains green (52 passed). No remaining issues from my review—thank you for the quick update.

HonzaKopecky:
@ShresthSamyak I can confirm the fix works for both Google BigQuery and Google Drive MCP servers! Just tested it directly in Agent Runtime. Thanks a lot for all the work you put in!

ShresthSamyak:
That's great to hear, @HonzaKopecky — thank you for testing it directly in Agent Runtime, and for confirming it across both the BigQuery and Google Drive MCP servers. Really appreciate you following through on the verification; it's good to have it validated on real deployments without the token-sharing workaround.

gnanirahulnutakki (commented):
Revalidated the force-pushed head `4b115485`. The two PR files are byte-for-byte identical to `2f0043ee`, the last verified head. On this head, the full focused Agent Registry suite passes (52 tests), as do pinned Pyink 25.12, isort 8.0.1, Ruff 0.15.17, and `git diff --check`. I also built a conflict-free synthetic merge onto current `main` (`5a248de5`) and repeated the 52-test and style checks successfully. GitHub's only red check is the repository's expected `maintainers will submit via Copybara` handoff; I found no new code concern.

rohityan:
Hi @ShresthSamyak , Thank you for your contribution! We appreciate you taking the time to submit this pull request. Your PR has been received by the team and is currently under review. We will provide feedback as soon as we have an update to share.

rohityan:
Hi @wukath , can you please review this. LGTM

wukath:
Hi @ShresthSamyak , thank you for putting together this PR and adding unit tests for mTLS MCP endpoint resolution!
Upon reviewing the implementation in `agent_registry.py`, we noticed that `get_mcp_toolset()` delegates URI extraction to `_get_connection_uri()` ([line 338](https://github.com/google/adk-python/blob/main/src/google/adk/integrations/agent_registry/agent_registry.py#L338)):
```python
if url := i.get("url"):
  if self._use_mtls:
    url = _mtls_utils.effective_googleapis_endpoint(url)
  return url, protocol_version, mapped_binding
```
Because _get_connection_uri() already rewrites Google API URLs to their .mtls.googleapis.com variant whenever self._use_mtls is true, get_mcp_toolset() was already receiving the effective mTLS endpoint. Calling _mtls_utils.effective_googleapis_endpoint(endpoint_uri) again inside get_mcp_toolset() is therefore a no-op.

Closing this PR but let me know if you run into other issues

PR #6460, opened 2026-07-23
fix(litellm): strip embedded thought_signature from tool call id

### LINKED ISSUE #6454: 422 Error: Tool Call ID required on tool calls
**Describe the Bug:**
When a Gemini "thinking" model's `thought_signature` is only available embedded in the tool call `id` (the third fallback checked by `_extract_thought_signature_from_tool_call` in `google/adk/models/lite_llm.py`, used when the response doesn't expose the signature via `extra_content.google.thought_signature` or `provider_specific_fields`), `_message_to_generate_content_response` correctly extracts the signature for `part.thought_signature`, but never strips that same `__thought__<signature>` suffix from the id before assigning it to `part.function_call.id`:

```python
# google/adk/models/lite_llm.py, _message_to_generate_content_response
part = types.Part.from_function_call(
    name=tool_call.function.name,
    args=_parse_tool_call_arguments(tool_call.function.arguments),
)
part.function_call.id = tool_call.id          # <-- bug: never split on _THOUGHT_SIGNATURE_SEPARATOR
if thought_signature:
```

The resulting `function_call.id` is a several-hundred-character string containing raw (non-URL-safe) base64 characters (`/`, `+`). This id gets persisted verbatim into session history (confirmed via direct inspection of a `DatabaseSessionService`-backed Postgres `events` row — see Logs). On any later turn where the same agent that made that call is replayed (its own history isn't converted to descriptive text the way other agents' events are, since it's the current agent's own authored event), that malformed id is sent to the model as a literal `tool_call_id` and rejected by the downstream OpenAI-compatible endpoint/proxy — and because the corrupted id is now permanently part of the session's event history, every subsequent turn on that thread fails identically, permanently breaking the conversation.

This was originally introduced (not regressed) by the PR that added thought_signature preservation itself — see Regression section.

**Steps to Reproduce:**
1. Install `google-adk==2.5.0` (also confirmed present on `main` as of this writing) with the LiteLLM extra.
2. Construct a litellm tool-call response where the `thought_signature` is only present embedded in the tool call `id` via the `__thought__` separator (i.e. no `extra_content`/`provider_specific_fields` set — this is the shape some upstream proxy paths return for Gemini 2.5 "thinking" models).
3. Pass that message through `_message_to_generate_content_response`.
4. Inspect `response.content.parts[0].function_call.id` — see Minimal Reproduction Code below for a runnable snippet.
5. Observe the id is not stripped of the `__thought__<signature>` suffix.

**Expected Behavior:**
`part.function_call.id` should be the clean, original tool-call id (e.g. `call_789`), with `part.thought_signature` populated separately — matching the behavior when the signature arrives via `extra_content` or `provider_specific_fields` (paths 1–2), where the id is never touched.

**Observed Behavior:**
`part.function_call.id` retains the entire raw string, e.g. `call_789__thought__<base64 signature>`. When this id is later replayed to the model as a `tool_call_id`, the request is rejected. In our deployment (LiteLLM proxy in front of Gemini 2.5 Flash), this surfaces as:

```
litellm.exceptions.BadRequestError: litellm.BadRequestError: Litellm_proxyException - Error code: 422 - {'detail': {'error': '{"error":"Messages[6].ToolCallID: Tool Call ID required on tool calls"}'}}
```

**Environment Details:**

 - ADK Library Version (pip show google-adk): 2.4.0 (installed/reproduced here); confirmed the same code is present unchanged in the 2.5.0 tag and on `main`
 - Desktop OS: macOS (local repro); production occurrence observed on Linux (AKS) — the bug is in pure data-handling logic, not OS-dependent
 - Python Version (python -V): 3.12.7

**Model Information:**

 - Are you using LiteLLM: Yes
 - Which model is being used: gemini-2.5-flash, with thinking enabled, via a LiteLLM proxy in front of an internal OpenAI-compatible Gemini endpoint

---

## 🟡 Optional Information

**Regression:**
Not a regression from a previously-working version — this is an oversight in the PR that introduced `thought_signature` preservation for LiteLLM tool calls itself: [`ae565be` "fix: Preserve thought_signature in LiteLLM tool calls"](https://github.com/google/adk-python/commit/ae565be30e64249b2913ad647911061a8b170e21) (closes #4650). That commit added `_extract_thought_signature_from_tool_call()` and correctly used it to populate `part.thought_signature`, but never reused the same split to clean up `tool_call.id` on the same code path. The commit's own added test, `test_extract_thought_signature_from_id`, only asserts the extractor returns the correct signature — it never asserts `part.function_call.id` is clean afterward, which is presumably why this slipped through review.

**Logs:**
```text
  File ".../litellm/main.py", line 637, in acompletion
    response = await init_response
  ...
  File ".../litellm/litellm_core_utils/exception_mapping_utils.py", line 422, in _map_openai_exception
    raise BadRequestError(
litellm.exceptions.BadRequestError: litellm.BadRequestError: Litellm_proxyException - Error code: 422 - {'detail': {'error': '{"error":"Messages[6].ToolCallID: Tool Call ID required on tool calls"}'}}

During handling of the above exception, another exception occurred:

  File ".../google/adk/runners.py", line 558, in _drive_root_node
    await root_ctx._run_node_internal(...)
  File ".../google/adk/agents/context.py", line 603, in _run_node_internal
    raise DynamicNodeFailError(...)
google.adk.workflow._errors.DynamicNodeFailError: Dynamic node <agent_name> failed
```

Corresponding persisted event (from a Postgres-backed `DatabaseSessionService`, field values truncated/redacted where noted):
```json
{
  "author": "<agent_name>",
  "content": {
    "role": "model",
    "parts": [{
      "function_call": {
        "id": "call_657c83221e1643f791d271007ccc__thought__CjQBjz1rX40Rt88HYX9Ijpkt2vbI0oO3QO/0Lwxwn7SnNofgFt+7eWmYPKfpcdm3m/...(truncated, ~500 chars total)",
        "args": {"agent_name": "..."},
        "name": "transfer_to_agent"
      },
      "thought_signature": "CjQBjz1rX40Rt88HYX9Ijpkt2vbI0oO3QO_0Lwxwn7SnNofgFt-7eWmYPKfpcdm3m_...(same bytes, URL-safe base64)"
    }]
  },
  "model_version": "gemini-2.5-flash"
}
```
Note the `function_call.id` suffix (after `__thought__`) is byte-identical to the separately-stored `thought_signature` field (modulo standard vs. URL-safe base64 alphabet) — confirming this is exactly the id-embedding fallback path leaking into the persisted id.

**Screenshots / Video:**
N/A

**Additional Context:**
- This affects any multi-agent (`transfer_to_agent`) topology where the same agent that issued the original tool call is later reactivated: because the current agent's own history is replayed as raw `function_call`/`function_response` parts (only *other* agents' events get converted to descriptive text by `_present_other_agent_message` in `contents.py`), the corrupted id is resent verbatim on every future turn — permanently breaking that conversation thread, not just the one turn.
- We observed this as non-deterministic across otherwise-identical environments and deployments (all going through the same LiteLLM proxy) — it depends on which of the three signature-delivery shapes a given model response happens to use, not on any config difference we could identify. Other teams we've talked to have reported similarly "hit or miss" behavior, suggesting this isn't specific to our proxy setup.
- Suggested fix: in `_message_to_generate_content_response`, when `_THOUGHT_SIGNATURE_SEPARATOR in (tool_call.id or "")`, split on it and assign only the first segment to `part.function_call.id` — mirroring the split already performed inside `_extract_thought_signature_from_tool_call`. We're carrying this as a local plugin workaround (a `before_model_callback` that strips the suffix from any leaked id before the request is sent) but would prefer to drop it once fixed upstream.

```python
import base64

from google.adk.models.lite_llm import (
    _message_to_generate_content_response,
    _THOUGHT_SIGNATURE_SEPARATOR,
)
from litellm.types.utils import ChatCompletionMessageToolCall, Function
from litellm import ChatCompletionAssistantMessage

sig_b64 = base64.b64encode(b"test_signature").decode("utf-8")

# Simulates the response shape where thought_signature is only available
# embedded in the tool call id (no extra_content / provider_specific_fields).
    role="assistant",
    content=None,
    tool_calls=[
            type="function",
            id=f"call_789{_THOUGHT_SIGNATURE_SEPARATOR}{sig_b64}",
            function=Function(name="test_function", arguments="{}"),
        )
    ],
)

response = _message_to_generate_content_response(message)
fc = response.content.parts[0].function_call

print("function_call.id:", fc.id)
# Observed: "call_789__thought__dGVzdF9zaWduYXR1cmU="
# Expected: "call_789"
assert fc.id == "call_789", f"id was not stripped of thought_signature suffix: {fc.id!r}"
```

 - Intermittently (<50%) — depends on which of the three signature-delivery shapes a given Gemini "thinking" tool-call response happens to use; once it occurs on a given session, it recurs on every subsequent turn for that session (100% from that point forward).

saakshigupta2002:

- Closes: #6454
- Related: #6454

_N/A — this PR resolves the existing issue linked above._

**Problem:**

Gemini "thinking" models attach a `thought_signature` to their function-call
parts. Depending on the provider path, that signature can arrive in one of
three places, which `_extract_thought_signature_from_tool_call` already handles:

1. `extra_content.google.thought_signature` (OpenAI-compatible API)
2. `provider_specific_fields` on the tool call or its function (Vertex)
3. **Embedded in the tool call ID** via the `__thought__` separator
   (`call_789__thought__<base64-signature>`) — the shape some upstream proxy
   paths return for Gemini 2.5 thinking models.

For paths 1 and 2 the tool call ID is never touched. For path 3, however,
`_message_to_generate_content_response` in
`src/google/adk/models/lite_llm.py` extracted the signature onto
`part.thought_signature` but then assigned the **entire, unsplit ID**
(including the `__thought__<signature>` suffix) to `part.function_call.id`:

```python
part.function_call.id = tool_call.id          # <-- bug: suffix never stripped
if thought_signature:
```

Consequences:

- `function_call.id` becomes a several-hundred-character string containing raw,
  non-URL-safe base64 characters (`/`, `+`).
- That corrupted ID is **persisted verbatim** into session history (e.g. a
  `DatabaseSessionService`-backed events row).
- On any later turn where the authoring agent's own history is replayed, the
  malformed ID is sent back to the model as a literal `tool_call_id` and
  rejected by OpenAI-compatible endpoints / proxies:

  ```
  litellm.BadRequestError: Error code: 422 -
  {"error":"Messages[6].ToolCallID: Tool Call ID required on tool calls"}
  ```

- Because the corrupted ID is now a permanent part of the session's event
  history, **every subsequent turn on that thread fails identically**,
  permanently breaking the conversation.

This is consistent with the official guidance that a thought signature is a
**separate field**, distinct from the function-call identifier, and must be
preserved as such across turns
([Gemini thinking guide — thought signatures](https://ai.google.dev/gemini-api/docs/thinking#signatures)).
The bug was introduced by the change that added `thought_signature`
preservation itself; it is not a regression of previously correct behavior.

**Solution:**

Strip any embedded `__thought__` suffix from the tool call ID before assigning
it to `function_call.id`, so the persisted ID is the clean, original identifier
— while the signature continues to be preserved separately on
`part.thought_signature`.

A small helper mirrors the existing extraction logic (splitting on the first
`__thought__` separator, exactly as `_extract_thought_signature_from_tool_call`
does):

```python
def _strip_thought_signature_from_tool_call_id(
    tool_call_id: Optional[str],
) -> Optional[str]:
  if not tool_call_id:
    return tool_call_id
  return tool_call_id.split(_THOUGHT_SIGNATURE_SEPARATOR, 1)[0]
```

and is applied at the assignment site:

```python
part.function_call.id = _strip_thought_signature_from_tool_call_id(
    tool_call.id
)
```

This solution was chosen because:

- It fixes the corruption at the source (inbound response conversion), so no
  malformed ID is ever persisted, rather than papering over it downstream.
- It is **fully backward compatible**: IDs without the separator are returned
  unchanged, so the `extra_content` and `provider_specific_fields` paths and all
  plain tool-call IDs are unaffected.
- It keeps behavior symmetric with the existing signature-extraction path and
  leaves outbound LiteLLM re-embedding untouched.

### Testing Plan

**Unit Tests:**

Added to `tests/unittests/models/test_litellm.py`:

- `test_message_to_generate_content_response_strips_thought_signature_from_id`
  — the core regression: an embedded-ID tool call yields a clean
  `function_call.id` (`call_789`) while `thought_signature` is still preserved.
- `test_strip_thought_signature_from_tool_call_id_removes_suffix`
- `test_strip_thought_signature_from_tool_call_id_leaves_plain_id`
- `test_strip_thought_signature_from_tool_call_id_handles_empty` (`None` / `""`)
- `test_strip_thought_signature_from_tool_call_id_splits_once`

The core regression test was verified to **fail before the fix** (proving it
reproduces the bug) and **pass after the fix**.

Summary of passed `pytest` results:

```
# Targeted (thought_signature-related):
$ pytest tests/unittests/models/test_litellm.py -k "thought_signature or strip_thought" -q
17 passed, 335 deselected

# Full lite_llm suite (no regressions):
352 passed
```

The failure is in pure data-handling logic, so it is fully reproducible without
a live model. Minimal reproduction:

```python
import base64
from litellm.types.utils import (
    ChatCompletionAssistantMessage,
    ChatCompletionMessageToolCall,
    Function,
)

sig = base64.b64encode(b"embedded_sig").decode()
msg = ChatCompletionAssistantMessage(
    role="assistant",
    content=None,
    tool_calls=[
            type="function",
            id=f"call_789__thought__{sig}",
            function=Function(name="test_function", arguments="{}"),
        )
    ],
)

part = _message_to_generate_content_response(msg).content.parts[0]
print(part.function_call.id)      # before: 'call_789__thought__<sig>'  after: 'call_789'
print(part.thought_signature)     # b'embedded_sig' (preserved in both cases)
```

Before the fix, the printed ID retains the `__thought__<signature>` suffix;
after the fix it is the clean `call_789`, and the signature is still present on
`part.thought_signature`.

### Checklist

### Additional context

- Files changed:
  - `src/google/adk/models/lite_llm.py` — new
    `_strip_thought_signature_from_tool_call_id` helper and its use in
    `_message_to_generate_content_response`.
  - `tests/unittests/models/test_litellm.py` — regression and unit tests.
- Scope is intentionally minimal: only the inbound ID assignment is changed. The
  `extra_content` and `provider_specific_fields` extraction paths, and the
  outbound tool-call serialization, are left unchanged.

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -987,6 +987,33 @@ def _extract_thought_signature_from_tool_call(
     if len(parts) == 2:
       return _decode_thought_signature(parts[1])
 
+
+def _strip_thought_signature_from_tool_call_id(
+    tool_call_id: Optional[str],
+) -> Optional[str]:
+  """Strips an embedded thought_signature suffix from a tool call ID.
+
+  Some Gemini thinking model paths embed the thought_signature in the tool
+  call ID via the ``__thought__`` separator (see
+  ``_extract_thought_signature_from_tool_call``). The signature is surfaced
+  separately on ``part.thought_signature``, so the ID assigned to
+  ``function_call.id`` must be the clean, original ID without the suffix.
+  Otherwise the corrupted ID is persisted to session history and later
+  replayed to the model as a ``tool_call_id``, which downstream
+  OpenAI-compatible endpoints reject.
+
+  Args:
+    tool_call_id: The raw tool call ID, which may contain an embedded
+      thought_signature.
+
+  Returns:
+    The tool call ID with any ``__thought__`` suffix removed, or the value
+    unchanged when no separator is present.
+  """
+  if not tool_call_id:
+    return tool_call_id
+  return tool_call_id.split(_THOUGHT_SIGNATURE_SEPARATOR, 1)[0]
+
   return None
 
 
@@ -2182,7 +2209,12 @@ def _message_to_generate_content_response(
             name=tool_call.function.name,
             args=_parse_tool_call_arguments(tool_call.function.arguments),
         )
-        part.function_call.id = tool_call.id
+        # Strip any embedded thought_signature suffix so the persisted
+        # function_call.id stays the clean, original tool call ID. The
+        # signature is preserved separately on part.thought_signature.
+        part.function_call.id = _strip_thought_signature_from_tool_call_id(
+            tool_call.id
+        )
         if thought_signature:
           part.thought_signature = thought_signature
         parts.append(part)
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -1002,17 +1002,30 @@ def _strip_thought_signature_from_tool_call_id(
   replayed to the model as a ``tool_call_id``, which downstream
   OpenAI-compatible endpoints reject.
 
+  The suffix is only stripped when it actually decodes as a thought_signature,
+  mirroring ``_extract_thought_signature_from_tool_call``. If the text after the
+  separator does not decode, the separator is treated as part of the opaque
+  tool call ID and the ID is returned unchanged, so an ID is never mutated
+  unless a signature was genuinely extracted from it. This matches the Gemini
+  requirement to echo back the exact function-call ID.
+
   Args:
     tool_call_id: The raw tool call ID, which may contain an embedded
       thought_signature.
 
   Returns:
-    The tool call ID with any ``__thought__`` suffix removed, or the value
-    unchanged when no separator is present.
+    The tool call ID with the ``__thought__`` suffix removed when that suffix
+    is a decodable signature; otherwise the value unchanged.
   """
-  if not tool_call_id:
+  if not tool_call_id or _THOUGHT_SIGNATURE_SEPARATOR not in tool_call_id:
     return tool_call_id
-  return tool_call_id.split(_THOUGHT_SIGNATURE_SEPARATOR, 1)[0]
+  base_id, _, encoded_signature = tool_call_id.partition(
+      _THOUGHT_SIGNATURE_SEPARATOR
+  )
+  if _decode_thought_signature(encoded_signature) is None:
+    # The suffix is not a valid signature, so it belongs to the opaque ID.
+    return tool_call_id
+  return base_id
 
   return None
 
--- a/tests/unittests/models/test_litellm.py
+++ b/tests/unittests/models/test_litellm.py
@@ -57,6 +57,7 @@ from google.adk.models.lite_llm import _redirect_litellm_loggers_to_stdout
 from google.adk.models.lite_llm import _safe_json_serialize
 from google.adk.models.lite_llm import _schema_to_dict
 from google.adk.models.lite_llm import _split_message_content_and_tool_calls
+from google.adk.models.lite_llm import _strip_thought_signature_from_tool_call_id
 from google.adk.models.lite_llm import _THOUGHT_SIGNATURE_SEPARATOR
 from google.adk.models.lite_llm import _to_litellm_response_format
 from google.adk.models.lite_llm import _to_litellm_role
@@ -2935,6 +2936,67 @@ def test_message_to_generate_content_response_no_thought_signature():
   fc_part = response.content.parts[0]
   assert fc_part.function_call.name == "plain_tool"
   assert fc_part.thought_signature is None
+
+
+def test_message_to_generate_content_response_strips_thought_signature_from_id():
+  """function_call.id is cleaned when the signature is embedded in the ID.
+
+  Regression test: when the thought_signature is only available embedded in
+  the tool call ID (the __thought__ fallback), the original code assigned the
+  entire unsplit ID to function_call.id, corrupting session history and
+  breaking every subsequent turn once the malformed ID is replayed as a
+  tool_call_id.
+  """
+  sig_b64 = base64.b64encode(b"embedded_sig").decode("utf-8")
+  embedded_id = f"call_789{_THOUGHT_SIGNATURE_SEPARATOR}{sig_b64}"
+  message = ChatCompletionAssistantMessage(
+      role="assistant",
+      content=None,
+      tool_calls=[
+          ChatCompletionMessageToolCall(
+              type="function",
+              id=embedded_id,
+              function=Function(
+                  name="test_function",
+                  arguments='{"test_arg": "test_value"}',
+              ),
+          )
+      ],
+  )
+
+  response = _message_to_generate_content_response(message)
+  fc_part = response.content.parts[0]
+  # The ID is stripped back to the clean, original tool call ID ...
+  assert fc_part.function_call.id == "call_789"
+  assert _THOUGHT_SIGNATURE_SEPARATOR not in fc_part.function_call.id
+  # ... and the signature is still preserved separately.
+  assert fc_part.thought_signature == b"embedded_sig"
+
+
+def test_strip_thought_signature_from_tool_call_id_removes_suffix():
+  """The embedded __thought__ suffix is removed from the ID."""
+  sig_b64 = base64.b64encode(b"embedded_sig").decode("utf-8")
+  tool_call_id = f"call_789{_THOUGHT_SIGNATURE_SEPARATOR}{sig_b64}"
+  assert _strip_thought_signature_from_tool_call_id(tool_call_id) == "call_789"
+
+
+def test_strip_thought_signature_from_tool_call_id_leaves_plain_id():
+  """A plain ID with no separator is returned unchanged."""
+  assert _strip_thought_signature_from_tool_call_id("call_plain") == "call_plain"
+
+
+@pytest.mark.parametrize("value", [None, ""])
+def test_strip_thought_signature_from_tool_call_id_handles_empty(value):
+  """None and empty IDs are returned unchanged."""
+  assert _strip_thought_signature_from_tool_call_id(value) == value
+
+
+def test_strip_thought_signature_from_tool_call_id_splits_once():
+  """Only the first separator is used, mirroring signature extraction."""
+  tool_call_id = (
+      f"call_1{_THOUGHT_SIGNATURE_SEPARATOR}sig{_THOUGHT_SIGNATURE_SEPARATOR}x"
+  )
+  assert _strip_thought_signature_from_tool_call_id(tool_call_id) == "call_1"
 
 
 @pytest.mark.asyncio
--- a/tests/unittests/models/test_litellm.py
+++ b/tests/unittests/models/test_litellm.py
@@ -2991,12 +2991,43 @@ def test_strip_thought_signature_from_tool_call_id_handles_empty(value):
   assert _strip_thought_signature_from_tool_call_id(value) == value
 
 
-def test_strip_thought_signature_from_tool_call_id_splits_once():
-  """Only the first separator is used, mirroring signature extraction."""
-  tool_call_id = (
-      f"call_1{_THOUGHT_SIGNATURE_SEPARATOR}sig{_THOUGHT_SIGNATURE_SEPARATOR}x"
+def test_strip_thought_signature_from_tool_call_id_retains_undecodable_suffix():
+  """An ID whose __thought__ suffix is not a valid signature is untouched.
+
+  The separator can legitimately appear inside an opaque tool call ID. When the
+  text after it does not decode as a signature, no signature was extracted, so
+  the ID must be preserved exactly rather than truncated.
+  """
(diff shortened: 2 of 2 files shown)


src/google/adk/models/lite_llm.py
```diff
@@ -987,6 +987,33 @@ def _extract_thought_signature_from_tool_call(
...
+    tool_call_id: The raw tool call ID, which may contain an embedded
+      thought_signature.
+
+  Returns:
+    The tool call ID with any ``__thought__`` suffix removed, or the value
+    unchanged when no separator is present.
+  """
+  if not tool_call_id:
+    return tool_call_id
+  return tool_call_id.split(_THOUGHT_SIGNATURE_SEPARATOR, 1)[0]
```
gnanirahulnutakki: Please only strip the suffix when it was successfully recognized as an embedded signature. `_extract_thought_signature_from_tool_call()` rejects malformed base64, but this helper truncates every ID containing the separator. At this head I reproduced `call_keep__thought__not-base64` becoming `function_call.id == "call_keep"` while `part.thought_signature is None`. That mutates the opaque ID before it is persisted even though no signature was extracted; the Gemini guidance also requires returning the exact function-call ID in the response. Could the ID and signature be parsed together (or the raw ID retained when decoding fails), with a malformed-suffix regression test?

gnanirahulnutakki (commented):
Reviewed exact head `96aac45a`. The focused thought-signature tests pass (17/17) and the full LiteLLM model suite passes (352/352). I found one persistence-safety edge where an undecodable `__thought__` suffix still causes the opaque tool-call ID to be truncated; details and a minimal reproducer are inline.

src/google/adk/models/lite_llm.py
saakshigupta2002: > Please only strip the suffix when it was successfully recognized as an embedded signature. `_extract_thought_signature_from_tool_call()` rejects malformed base64, but this helper truncates every ID containing the separator. At this head I reproduced `call_keep__thought__not-base64` becoming `function_call.id == "call_keep"` while `part.thought_signature is None`. That mutates the opaque ID before it is persisted even though no signature was extracted; the Gemini guidance also requires returning the exact function-call ID in the response. Could the ID and signature be parsed together (or the raw ID retained when decoding fails), with a malformed-suffix regression test?

Good catch, thanks! You're right the separator can be part of the id itself. I changed it to only strip when the suffix actually decodes as a signature, otherwise the id is left alone. call_keep__thought__not-base64 stays intact now (signature None) and I added tests for it.

src/google/adk/models/lite_llm.py
```diff
@@ -987,6 +987,46 @@ def _extract_thought_signature_from_tool_call(
...
+  Returns:
+    The tool call ID with the ``__thought__`` suffix removed when that suffix
+    is a decodable signature; otherwise the value unchanged.
+  """
+  if not tool_call_id or _THOUGHT_SIGNATURE_SEPARATOR not in tool_call_id:
+    return tool_call_id
+  base_id, _, encoded_signature = tool_call_id.partition(
+      _THOUGHT_SIGNATURE_SEPARATOR
+  )
+  if _decode_thought_signature(encoded_signature) is None:
```
gnanirahulnutakki, line 1025: [P2] Preserve IDs when the decoded signature is empty

The malformed-base64 case is fixed, but an empty suffix still breaks the same invariant. `base64.b64decode("")` returns `b""`, so `call_keep__thought__` reaches `return base_id` here and becomes `call_keep`; later `_message_to_generate_content_response()` uses `if thought_signature:`, so that empty value is not persisted on `part.thought_signature`. I reproduced `_decode_thought_signature("") == b""` and `_strip_thought_signature_from_tool_call_id("call_keep__thought__") == "call_keep"` on this head. Please retain the raw ID when the decoded value is empty (or parse once and base stripping on the same value that will actually be stored), and add this boundary as a regression test.

gnanirahulnutakki (commented):
Re-reviewed exact head `15bbca823e48`. The malformed-base64 case is fixed, and the seven focused tests pass locally. One empty-signature boundary remains inline. Separately, `uv run pyink --check src/google/adk/models/lite_llm.py tests/unittests/models/test_litellm.py` reports that the test file still needs formatting.

rohityan:
Hi @saakshigupta2002, Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

saakshigupta2002:
> Hi @saakshigupta2002, Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

Hi @rohityan just signed the agreement! :)

PR #6463, opened 2026-07-24
fix: disable stemmer in ROUGE-1 scorer to support non-English text

guptaishaan:
The `_calculate_rouge_1_scores` function in `final_response_match_v1.py` was
calling `rouge_scorer.RougeScorer` with `use_stemmer=True`. The Porter stemmer
used by the `rouge-score` library is designed for English only; when applied to
non-Latin scripts (e.g., Thai, Chinese, Arabic), it garbles or drops tokens,
causing the ROUGE-1 f-measure to return 0 even when the candidate and reference
texts are identical.

Changing `use_stemmer=False` fixes the issue for all non-English languages while
keeping evaluation correct for English text, since the stemmer's effect on
English is a minor normalization that is not critical for match scoring.

A regression test that asserts a perfect score for identical Thai text is also
added.

Fixes #3111

### DIFF
--- a/src/google/adk/evaluation/final_response_match_v1.py
+++ b/src/google/adk/evaluation/final_response_match_v1.py
@@ -114,7 +114,7 @@ def _calculate_rouge_1_scores(candidate: str, reference: str):
   Returns:
       A dictionary containing the ROUGE-1 precision, recall, and f-measure.
   """
-  scorer = rouge_scorer.RougeScorer(["rouge1"], use_stemmer=True)
+  scorer = rouge_scorer.RougeScorer(["rouge1"], use_stemmer=False)
 
   # The score method returns a dictionary where keys are the ROUGE types
   # and values are Score objects (tuples) with precision, recall, and fmeasure.
--- a/tests/unittests/evaluation/test_final_response_match_v1.py
+++ b/tests/unittests/evaluation/test_final_response_match_v1.py
@@ -51,6 +51,14 @@ def _create_test_invocations(
   )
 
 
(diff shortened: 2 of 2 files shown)


i-yliu:
Hi @guptaishaan,
Thank you for the PR to fix #3111.

We looked into the root cause of #3111 in detail, setting `use_stemmer=False` doesn't seem to resolve the issue for non-English evaluations:

1. **Root Cause in `rouge_score`**: The zero-score issue for non-Latin scripts is caused by `rouge_score`'s default tokenizer (`DefaultTokenizer`), which applies a regex filter `re.sub(r"[^a-z0-9]+", " ", text)` during its initial tokenization pass. This strips out all characters outside `[a-z0-9]`, reducing non-Latin text (e.g., Thai `สวัสดี`, Chinese `你好`, Arabic, Japanese, etc.) to an empty token list `[]`.
2. **Why `use_stemmer=False` is insufficient**: Setting `use_stemmer=False` only skips the second-stage Porter Stemmer pass. The initial regex filtering (`[^a-z0-9]`) still runs and wipes out non-Latin characters before stemming would even occur. Additionally, globally disabling `use_stemmer` alters evaluation behavior for English responses where stemming normalization is expected.

To properly fix this without breaking English evaluation or stripping non-Latin scripts, we are addressing #3111 in **PR #6292** by implementing a `_UnicodeAwareTokenizer`. 

Since PR #6292 provides comprehensive tokenizer-level fix, we will close this PR in favor of #6292.

Thank you.

PR #6466, opened 2026-07-24
fix(plugins): add opt-in exactly-once delivery to BigQueryAgentAnalyticsPlugin

### LINKED ISSUE #6465: BigQueryAgentAnalyticsPlugin can duplicate rows when writes are retried
## Problem

`BigQueryAgentAnalyticsPlugin` can write duplicate rows when its internal BigQuery Storage Write API request is retried after an ambiguous failure.

This concerns retries of the **BigQuery write**, not retries by the agent, model, or tools.

The plugin writes to BigQuery's `_default` stream. `BatchProcessor._write_rows_with_retry` constructs one `AppendRowsRequest` and resends the same serialized batch following transient errors or its 30-second
client-side timeout.

If BigQuery accepts the first append but its acknowledgement is lost or delayed, the plugin retries the batch. Because `_default` provides at-least-once delivery, BigQuery can accept it again and produce duplicate rows.

Relevant code:

- [`BatchProcessor._write_rows_with_retry`](https://github.com/google/adk-python/blob/07455ee62cbd5c11625aed8c75380b88eba55e6d/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#L1693-L1818)
- [Default-stream selection](https://github.com/google/adk-python/blob/07455ee62cbd5c11625aed8c75380b88eba55e6d/src/google/adk/plugins/bigquery_agent_analytics_plugin.py#L3099-L3105)

## Impact

We observed byte-identical duplicate analytics rows in production, including identical timestamps, trace/span IDs, token usage, status, and content.

The issue is rare. Over seven days, six duplicate `LLM_RESPONSE` rows inflated a naive `SUM(usage.total)` by 182,555 tokens, approximately 0.31% of the total. This can silently over-count tokens, costs, latency, and other metrics derived from the analytics table. However this was for a poc agent with few users and can easily inflate with more use.

## Proposed fix

BigQuery does not allow append offsets on `_default`, and `BigQueryWriteAsyncClient` has no `exactly_once` option.

Exactly-once append behaviour requires an application-created `COMMITTED` stream with offsets:

1. Create one committed stream per `BatchProcessor`.
2. Include the current offset in each `AppendRowsRequest`.
3. Retry an ambiguous append using the same stream and offset.
4. Advance the offset after success.
5. Treat `ALREADY_EXISTS` at that offset as confirmation that the earlier
   attempt succeeded.
6. Handle unexpected `OUT_OF_RANGE` without blindly resending on a new stream.

A committed stream preserves immediate row visibility while making Storage Write API retries idempotent.

This fix would prevent duplicates caused by BigQuery write retries. It would not deduplicate genuinely separate events emitted by the agent or application.

For anyone familiar with Apache beam this is similar at a high level to their use_at_least_once=False option for BigQuery Storage Write API sinks. Beam provides the stream, offset, and durable checkpoint management behind that abstraction; because ADK uses BigQueryWriteAsyncClient directly, the plugin must manage its committed stream and append offsets itself.

References:

- [BigQuery streaming guarantees](https://docs.cloud.google.com/bigquery/docs/write-api-streaming)
- [`AppendRowsRequest.offset`](https://docs.cloud.google.com/python/docs/reference/bigquerystorage/latest/google.cloud.bigquery_storage_v1.types.AppendRowsRequest.html)

addenergyx:
Closes: #6465

`BatchProcessor._write_rows_with_retry` writes to BigQuery's shared `_default` Storage Write API stream, which is at-least-once and rejects an append offset. It resubmits the same batch after a client-side timeout or `UNAVAILABLE`/`INTERNAL`, so if the first append actually landed but the ack was lost, the retry can write a byte-identical duplicate row. 

Adds an opt-in `exactly_once_delivery` flag to `BigQueryLoggerConfig` (default `False`, no behavior change for existing users). When enabled, each event loop's `BatchProcessor` writes to its own `COMMITTED` stream with a tracked append offset instead of `_default`. A retry after an ambiguous failure resends the same offset; BigQuery returns `ALREADY_EXISTS` instead of writing it again, which is treated as a confirmed no-op rather than an error. An unexpected `OUT_OF_RANGE` (offset desync) drops the batch loudly rather than guessing a corrective offset. Similar in spirit to Apache Beam's `use_at_least_once=False` on `WriteToBigQuery`.

Added `TestExactlyOnceDelivery` (5 tests) covering the offset lifecycle, including the exact timeout-then-`ALREADY_EXISTS` sequence from production. Full suite (`tox`, py310-py314): 9231 passed, 0 failed on every version. Not yet tested against a live BigQuery project.


GWeale:
landed in 04b8b727, v2.7.0, down to the same private names, and #6465 is closed. head is a behavioral superset of what's here.

PR #6468, opened 2026-07-24
fix: route workflow IDs and retry jitter through platform seams

DABH:
## Describe the bug

`google.adk.platform` provides injectable time and uuid providers so that embedders can customize how generated values are produced - for example to make agent runs deterministic and reproducible for testing, simulation, or auditing. Event IDs, invocation IDs, timestamps, and client function-call IDs already route through them, but three call sites on the workflow HITL/retry path still call the stdlib directly, so they bypass any installed provider:

1. **`events/request_input.py`** - `RequestInput.interrupt_id` defaults to `uuid.uuid4()`. This is the ID that recorded user `FunctionResponse`s reference on resume; an ID minted outside the installed provider can't be reproduced, which breaks interrupt matching.
2. **`workflow/_tool_node.py`** - `ToolContext.function_call_id` is minted with `uuid.uuid4()` instead of `platform_uuid.new_uuid()` (which `flows/llm_flows/functions.py` already uses for the same purpose).
3. **`workflow/utils/_retry_utils.py`** - retry backoff jitter draws from the global `random` module. Jitter defaults to 1.0 for any `RetryConfig`, so every retried node computes an unreproducible delay.

## Describe the fix

- Add a **platform random seam** - `google.adk.platform._random` with `get_random()` / `set_random_provider()` / `reset_random_provider()`, exported from `google.adk.platform` per the visibility convention - mirroring the existing `time`/`uuid` providers (ContextVar-based, default = stdlib `random.Random()`).
- Route the three call sites through the platform seams. **Default behavior is unchanged** (same uuid4/uniform semantics when no provider is installed).
- Migrate the two existing jitter tests from `mock.patch('random.uniform', ...)` to injecting a mock via `set_random_provider`, exercising the new seam with the exact same assertions.

This follows the injectable-provider approach also adopted in google/adk-java#1255.

## Testing plan

New unit tests:
- `tests/unittests/platform/test_random.py` - default provider, custom provider, reset (mirrors `test_uuid.py`).
- `tests/unittests/events/test_request_input.py` - default `interrupt_id` is a uuid; minted via the platform id provider; explicit id preserved.
- `tests/unittests/workflow/test_tool_node.py::test_tool_node_function_call_id_uses_platform_id_provider` - the tool receives a provider-minted `function_call_id`.
- `tests/unittests/workflow/utils/test_retry_utils.py::test_jitter_uses_platform_random_provider` - a seeded provider makes the computed delay reproducible.

```
$ pytest tests/unittests/workflow/ tests/unittests/platform/ tests/unittests/events/
791 passed, 11 skipped, 10 xfailed in 6.95s
```

All pre-commit hooks (isort, pyink, addlicense, new-file-prefix, ADK compliance, codespell) pass.

### DIFF
--- a/src/google/adk/events/request_input.py
+++ b/src/google/adk/events/request_input.py
@@ -15,13 +15,13 @@ from __future__ import annotations
 
 from typing import Any
 from typing import Optional
-import uuid
 
 from pydantic import alias_generators
 from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Field
 
+from ..platform import uuid as platform_uuid
 from ..utils._schema_utils import SchemaType
 
 
@@ -39,7 +39,7 @@ class RequestInput(BaseModel):
           "The ID of the interrupt, usually a function call ID. This is used"
           " to identify the interrupt that the input is for."
       ),
-      default_factory=lambda: str(uuid.uuid4()),
+      default_factory=platform_uuid.new_uuid,
   )
   """The ID of the interrupt, usually a function call ID.
 
--- a/src/google/adk/platform/__init__.py
+++ b/src/google/adk/platform/__init__.py
@@ -11,3 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from ._random import get_random
+from ._random import reset_random_provider
+from ._random import set_random_provider
+
+__all__ = [
+    'get_random',
+    'reset_random_provider',
+    'set_random_provider',
+]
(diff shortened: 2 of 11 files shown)


DABH:
@DeanChensj The prior CI failure appears to just be because my PR was out of date with the main branch.  I've updated and we should be good to go.  Would you mind re-approving the CI run so things can retry?  Thanks again!!

DABH:
Ah, looks like `main` is broken on CI now - once `main` is green, I'll update here one more time and hopefully all will go smoothly (I've verified locally and it looks green basing this off an older commit from `main`.).  Keeping an eye out... thanks again!

DABH:
Hi @DeanChensj , the repo's CI is green again, could you please take another look at this PR and review at your convenience?  Would love to get this fix out in the next release.  Thanks!

PR #6476, opened 2026-07-25
fix: use character-level tokenization for non-ASCII text in ROUGE-1 eval

guptaishaan:
The `_calculate_rouge_1_scores` function in `final_response_match_v1.py` uses
`rouge_scorer.RougeScorer` with its default word-level tokenizer, which only
matches ASCII alphanumeric characters (`[a-z0-9]+`). For non-Latin scripts such
as Thai, Chinese, or Japanese — where words are not separated by spaces and
characters are outside the ASCII range — the tokenizer produces an empty token
list, causing ROUGE-1 to return a score of 0.0 even when the candidate and
reference strings are identical.

The fix detects when either the candidate or reference text contains non-ASCII
characters and switches to a character-level tokenizer (`_CharacterTokenizer`)
that splits text into individual Unicode code points. This produces correct
overlap scores for all non-ASCII languages while leaving the existing
word-level (with stemming) path fully intact for ASCII/English text.

Two regression tests are added to `test_final_response_match_v1.py`:
one verifying that identical Thai strings score 1.0 and one verifying that
two different Thai strings score strictly between 0 and 1.

Fixes #3111

### DIFF
--- a/src/google/adk/evaluation/final_response_match_v1.py
+++ b/src/google/adk/evaluation/final_response_match_v1.py
@@ -96,6 +96,18 @@ def _get_eval_status(score: float, threshold: float) -> EvalStatus:
   return EvalStatus.PASSED if score >= threshold else EvalStatus.FAILED
 
 
+class _CharacterTokenizer:
+  """Tokenizes text into individual characters.
+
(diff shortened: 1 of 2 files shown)


i-yliu:

PR #6480, opened 2026-07-25
fix(eval): wire App plugins through eval paths

### LINKED ISSUE #5503: # `cli_eval` bypasses `App.plugins`, breaking observability plugins (e.g., `BigQueryAgentAnalyticsPlugin`) during eval runs

doughayden:

- Closes: #5503

**Problem:**

Eval inference builds its `Runner` from the bare `root_agent`, so the wrapping `App`'s plugins never apply. `EvaluationGenerator._generate_inferences_from_root_agent` constructs `Runner(app_name=..., agent=root_agent, plugins=[<internal eval plugins>])`, which silently drops `app.plugins` for the duration of an eval run.

Three non-live surfaces reach that leaf, and all three build `LocalEvalService` without the `App`:

- the `adk eval` CLI (`cli_eval` / `cli_tools_click`), which is what #5503 originally reported
- `AgentEvaluator` (the `evaluate` / `evaluate_eval_set` path), the programmatic and pytest entry point, so a scripted or CI eval bypasses observability and guardrail plugins
- the `adk web` eval tab (`dev_server.run_eval`)

The result is that an eval scores a different agent than `adk web` chat and the deployed server, both of which run the full `App`. It also degrades the LLM-judge metrics that read `app_details` (developer instructions plus tool declarations), since those are shaped by plugins the eval never applied.

**Solution:**

Thread `app=` from all three callers into the shared inference leaf, which then builds the eval Runner from `app.model_copy(update={"plugins": list(app.plugins) + internal_eval_plugins, "root_agent": root_agent})`. The copy leaves the caller's `App` untouched, merges the internal eval plugins with the user's rather than replacing them, and overrides `root_agent` so the Runner still targets whatever agent the caller asked to evaluate, a sub-agent included. The App's `context_cache_config` and `resumability_config` come along with it. When no `App` is present, the bare-agent construction is unchanged.

This also matches where the Runner API is already heading, since `Runner(plugins=...)` is deprecated in favor of `app=`.

Runner construction is shared between the non-live and live leaves through `_build_eval_runner_kwargs`, so `use_live=True` carries the App the same way. Agent resolution in `cli_eval` gains `get_app_or_root_agent`, which surfaces the `App` alongside the root agent and keeps `get_root_agent` as a thin wrapper for existing callers.

### Testing Plan

**Unit Tests:**

New tests cover each surface: the CLI resolver (App present, App absent, `app` attribute that is not an `App`, `get_agent_async` modules, and the back-compat wrapper), `AgentEvaluator._get_agent_for_eval` returning the pair, `LocalEvalService` forwarding `app` on both the live and non-live branches, and the Runner construction itself asserting that `agent=`/`plugins=` are not passed when `app=` is and that the user's plugin survives the merge.

```
$ pytest tests/unittests/evaluation tests/unittests/cli/utils -q
851 passed, 96 warnings in 25.49s
```

A reproduction lives at [doughayden/adk-issue-examples/06-eval_drops_app_plugins](https://github.com/doughayden/adk-issue-examples/tree/main/06-eval_drops_app_plugins). An `LlmAgent` is wrapped in an `App` whose plugins include a `SentinelPlugin` that appends to a module-level list every time it runs inside a `Runner`, so "did the App's plugins apply during eval" is directly observable. The script runs one real eval through the public `AgentEvaluator.evaluate` entry point. Nothing is patched or stubbed in either run, so the second one exercises the full caller chain (`AgentEvaluator` to `LocalEvalService` to `EvaluationGenerator`).

Released `google-adk[eval]==2.5.0`, from the example directory:

```
$ uv run main.py
google-adk 2.5.0
App-aware eval inference: False
App-aware live eval inference: False

🐞 Running an ADK build without the fix.

App-level plugin invocations during eval inference: []
✅ Bug reproduced: the App's plugins were dropped — none ran during eval inference, so the eval scored the bare root_agent.
```

The same script against this branch, with only the ADK build swapped:

```
$ uv run --isolated --no-project \
    --with "google-adk[eval] @ git+https://github.com/doughayden/adk-python@fix/eval-app-aware-inference" \
    --with python-dotenv \
    main.py
google-adk 2.5.0
App-aware eval inference: True
App-aware live eval inference: True

🔧 Running an ADK build that carries the App through eval.

App-level plugin invocations during eval inference: ['sentinel']
✅ Fix verified end to end: the App's plugins ran during eval inference, so the eval scores the same agent production runs.
```

(An unrelated `vertexai.preview.rag` deprecation warning is elided from both outputs.)

The live path is covered by unit tests rather than here, since exercising it needs a live-API session.

### Checklist

### Additional context

This builds on #5534, which threads the `App` through `cli_eval` and `evaluation_generator` and fixes the CLI path. The work here keeps that approach and extends it to the two surfaces #5534 does not reach, `AgentEvaluator` and `dev_server`, plus the live inference leaf that landed after it was opened.

`get_root_agent` recently became async and gained `get_agent_async` support, so `get_app_or_root_agent` mirrors that shape and handles both entrypoints. Existing `get_root_agent` callers are unaffected.

### DIFF
--- a/src/google/adk/cli/cli_eval.py
+++ b/src/google/adk/cli/cli_eval.py
@@ -27,6 +27,7 @@ import click
 from google.genai import types as genai_types
 
 from ..agents.base_agent import BaseAgent
+from ..apps.app import App
 from ..evaluation.base_eval_service import BaseEvalService
 from ..evaluation.base_eval_service import EvaluateConfig
 from ..evaluation.base_eval_service import EvaluateRequest
@@ -75,18 +76,41 @@ def _get_agent_module(agent_module_file_path: str) -> ModuleType:
   return _import_from_path(module_name, file_path)
 
 
-async def get_root_agent(agent_module_file_path: str) -> BaseAgent:
-  """Returns root agent given the agent module."""
+async def get_app_or_root_agent(
+    agent_module_file_path: str,
+) -> tuple[Optional[App], BaseAgent]:
+  """Returns the (app, root_agent) pair for the given agent module.
+
+  If the module exposes an `App` instance via `app`, that App and its
+  `root_agent` are returned. Otherwise `app` is None and the root agent is
+  resolved the same way as `get_root_agent`. This lets eval flows participate
+  in the App's plugin / cache / resumability lifecycle when one is defined,
+  while preserving the bare-`root_agent` path for projects that don't use App.
+  """
   agent_module = _get_agent_module(agent_module_file_path)
   agent_module_with_agent = getattr(agent_module, "agent", agent_module)
+  app = getattr(agent_module_with_agent, "app", None)
+  if isinstance(app, App):
+    return app, cast(BaseAgent, app.root_agent)
   if hasattr(agent_module_with_agent, "root_agent"):
-    return cast(BaseAgent, agent_module_with_agent.root_agent)
+    return None, cast(BaseAgent, agent_module_with_agent.root_agent)
   elif hasattr(agent_module_with_agent, "get_agent_async"):
     root_agent, _ = await agent_module_with_agent.get_agent_async()
-    return cast(BaseAgent, root_agent)
+    return None, cast(BaseAgent, root_agent)
   raise ValueError(
       "Agent module should have either `root_agent` or `get_agent_async`."
   )
+
+
+async def get_root_agent(agent_module_file_path: str) -> BaseAgent:
+  """Returns root agent given the agent module.
+
+  Kept for backward compatibility. New callers should prefer
+  `get_app_or_root_agent`, which also surfaces the wrapping `App` (if any)
+  so plugins, context-cache, and resumability configs are honored.
+  """
+  _, root_agent = await get_app_or_root_agent(agent_module_file_path)
+  return root_agent
 
 
 def try_get_reset_func(agent_module_file_path: str) -> Any:
--- a/src/google/adk/cli/cli_tools_click.py
+++ b/src/google/adk/cli/cli_tools_click.py
@@ -1022,7 +1022,7 @@ def cli_eval(
     from ..evaluation.simulation.user_simulator_provider import UserSimulatorProvider
     from .cli_eval import _collect_eval_results
     from .cli_eval import _collect_inferences
-    from .cli_eval import get_root_agent
+    from .cli_eval import get_app_or_root_agent
     from .cli_eval import parse_and_get_evals_to_run
     from .cli_eval import pretty_print_eval_result
   except ModuleNotFoundError as mnf:
@@ -1032,7 +1032,7 @@ def cli_eval(
   print(f"Using evaluation criteria: {eval_config}")
   eval_metrics = get_eval_metrics_from_config(eval_config)
 
-  root_agent = asyncio.run(get_root_agent(agent_module_file_path))
+  app, root_agent = asyncio.run(get_app_or_root_agent(agent_module_file_path))
   app_name = os.path.basename(agent_module_file_path)
   agents_dir = os.path.dirname(agent_module_file_path)
   eval_sets_manager = None
@@ -1124,6 +1124,7 @@ def cli_eval(
         eval_set_results_manager=eval_set_results_manager,
         user_simulator_provider=user_simulator_provider,
         metric_evaluator_registry=metric_evaluator_registry,
+        app=app,
     )
 
     inference_results = asyncio.run(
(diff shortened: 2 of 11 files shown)


doughayden:
Reopening this after closing it by mistake. Nothing changed on the branch, head is still fa2dc3b and the checks are green.

The underlying problem in #5503 is still open and I don't see anything on main that wires App.plugins through the eval paths, so this is still current.

PR #6483, opened 2026-07-26
fix: handle non-ASCII languages in ROUGE-1 eval tokenizer

guptaishaan:
The `_calculate_rouge_1_scores` function in `final_response_match_v1.py` uses
`rouge_scorer.RougeScorer` with its default tokenizer, which strips every
character that does not match `[a-z0-9]`. For languages such as Thai, Chinese,
or Arabic—where words contain no ASCII characters—every token is removed and
the scorer always returns an F-measure of 0.0, causing eval to report a
mismatch even for identical strings.

The fix introduces a `_UnicodeTokenizer` that is used automatically whenever
either the candidate or reference text contains non-ASCII characters. It splits
on whitespace and, for each non-ASCII word, emits individual Unicode characters
as tokens so that ROUGE-1 overlap is computed at the character level—a
well-established fallback for scripts that lack whitespace word delimiters.
ASCII text continues to use the existing stem-based tokenizer unchanged.

Two regression tests are added to `test_final_response_match_v1.py`: one
asserting that identical Thai text scores 1.0, and one asserting that entirely
different Thai text scores 0.0.

Fixes #3111.

### DIFF
--- a/src/google/adk/evaluation/final_response_match_v1.py
+++ b/src/google/adk/evaluation/final_response_match_v1.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import re
 from typing import Optional
 
 from google.genai import types as genai_types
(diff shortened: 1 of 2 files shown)


i-yliu:

ankursharmas:
https://github.com/google/adk-python/pull/6292 already addresses the issue.

Issue #4754, opened 2026-03-07
`find_context_parameter` fails to detect context-typed parameters with postponed annotations


gautamvarmadatla:
**Describe the Bug:**

If a tool function is defined in a module that uses `from __future__ import annotations`, ADK can fail to recognize `Context` / `ToolContext` parameters because `find_context_parameter()` looks at the raw annotation string instead of resolved type hints. When that happens, `find_context_parameter()` misses the real context parameter. `FunctionTool` then falls back to the literal name `tool_context`, so a custom-named context parameter like `ctx` is not excluded from declaration building and eventually triggers a misleading `ValueError` saying it must be named `tool_context`. This is easy to miss because a parameter already named `tool_context` happens to match the fallback and masks the bug.

**Steps to Reproduce:**
1. Install google-adk from source (main branch)
2. Create a file with the following code:
``` python
  from __future__ import annotations

  from google.adk.tools.function_tool import FunctionTool
  from google.adk.tools.tool_context import ToolContext

  def my_tool(query: str, ctx: ToolContext) -> str:
      """A tool that searches for something."""
      return f"Result for {query}"

  tool = FunctionTool(my_tool)
  print(tool._get_declaration())
```
3. Run the file with `uv run python my_tool.py`

**Expected Behavior:**
The tool should be created successfully with ctx excluded from the schema, identical to the behavior without `from  __future__ import annotations`

**Observed Behavior:**
```
  ValueError: `ToolContext` parameter must be named as `tool_context`. Found `ctx` instead in function `my_tool`.
```

**Environment Details:**

 - ADK Library Version (pip show google-adk): 1.26.0
 - Desktop OS: Windows 11
 - Python Version (python -V): 3.10.19

---

surajksharma07:
@gautamvarmadatla Thanks for the detailed report and for already putting up PR #4755 — the root cause is spot on: `find_context_parameter()` uses an identity check (`annotation is Context`) that breaks when `from __future__ import annotations` turns annotations into strings.

As an immediate workaround until the PR lands, renaming the context parameter to `tool_context` will unblock you since `FunctionTool` falls back to that name — could you confirm if that works for your case in the meantime?

Once you've validated the fix in PR #4755 holds up across your use cases, it'd be great to get that reviewed and merged.

Team is looking more into it.

gautamvarmadatla:
hi @surajksharma07, thanks so much for the prompt response.
[...]

PR #6488, opened 2026-07-27
fix: build function declaration when a param uses a TYPE_CHECKING-only annotation

NishchayMahor:
## Summary

A `FunctionTool` fails to build its declaration when the tool is defined in a module using `from __future__ import annotations` and its context parameter is typed with an import guarded by `if TYPE_CHECKING:` — the pattern ADK's own docs recommend for context params:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:

def my_tool(x: int, tool_context: ToolContext) -> str:
    ...
```

Building the declaration raises `NameError: name 'ToolContext' is not defined` at the first LLM request — even though `tool_context` is excluded from the schema anyway.

## Root cause

`_get_function_fields` (and the return-type path in `_build_response_json_schema`) call `get_type_hints(func)`, which resolves **all** annotations at once, including the excluded context parameter. When that annotation is a `TYPE_CHECKING`-only forward reference, resolution raises `NameError` — but only `TypeError` was caught, so it aborts declaration building on the default JSON-schema path.

## Not a duplicate of #4754

#4754 / commit 22fc332 fixed `find_context_parameter` so the context param is correctly *identified* and excluded. This is a separate crash further down: schema and return-type resolution still resolve the excluded param's annotation. Verified the repro still fails on current main (which already contains 22fc332).

## Fix

Catch `NameError`/`AttributeError` alongside `TypeError` in both spots and fall back to the raw per-parameter annotation — logic that already exists directly below but was unreachable. When `get_type_hints` succeeds (the common case), behavior is unchanged.

## Testing

Added `test_type_checking_only_context_annotation_builds_declaration` to `tests/unittests/tools/test_function_tool_with_import_annotations.py`, asserting the declaration builds with the real param described and the context param excluded. Fails on main (NameError), passes with the fix.

```
python -m pytest tests/unittests/tools/test_function_tool_with_import_annotations.py -q
11 passed
python -m pytest tests/unittests/tools/test_build_function_declaration.py <this file> -q
52 passed
```
(Async suites in `test_function_tool.py` fail identically with and without this change — they need pytest-asyncio in my env — so no regression from this diff.) `pyink` + `isort` clean.

This fix was developed with AI assistance; I found the bug, verified the root cause and the fix end-to-end, and reviewed every line.

### DIFF
--- a/src/google/adk/tools/_function_tool_declarations.py
+++ b/src/google/adk/tools/_function_tool_declarations.py
@@ -64,8 +64,12 @@ def _get_function_fields(
   # Get type hints with forward reference resolution
   try:
     type_hints = get_type_hints(func)
-  except TypeError:
-    # Can happen with mock objects or complex annotations
+  except (TypeError, NameError, AttributeError):
+    # TypeError can happen with mock objects or complex annotations. NameError
+    # / AttributeError happen when an annotation is an unresolvable forward
+    # reference at runtime, e.g. a type imported only under `if TYPE_CHECKING:`
+    # (common for the context parameter, which is excluded from the schema
+    # anyway). Fall back to the raw per-parameter annotations below.
     type_hints = {}
 
   for name, param in sig.parameters.items():
@@ -162,7 +166,10 @@ def _build_response_json_schema(
     try:
       type_hints = get_type_hints(func)
       return_annotation = type_hints.get('return', return_annotation)
-    except TypeError:
+    except (TypeError, NameError, AttributeError):
+      # An unresolvable parameter annotation (e.g. a `TYPE_CHECKING`-only
+      # import) must not block resolving the return type; fall back to the
+      # raw return annotation, which pydantic resolves below.
       pass
 
   # Handle AsyncGenerator and Generator return types (streaming tools)
--- a/tests/unittests/tools/test_function_tool_with_import_annotations.py
+++ b/tests/unittests/tools/test_function_tool_with_import_annotations.py
@@ -17,12 +17,16 @@ from __future__ import annotations
 from typing import Any
 from typing import Dict
 from typing import Optional
+from typing import TYPE_CHECKING
 
 from google.adk.tools import _automatic_function_calling_util
 from google.adk.tools.function_tool import FunctionTool
 from google.adk.utils.variant_utils import GoogleLLMVariant
 from google.genai import types
 import pydantic
+
+if TYPE_CHECKING:
+  from google.adk.tools.tool_context import ToolContext
 
 
 def test_string_annotation_none_return_vertex():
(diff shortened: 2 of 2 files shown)


rohityan:
Hi @NishchayMahor , Thank you for your contribution! It appears you haven't yet signed the Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to complete the signing process. Once the CLA is signed, we'll be able to proceed with the review of your PR. Thank you!

NishchayMahor:
Thanks @rohityan — I've signed the CLA at https://cla.developers.google.com/ under nishchaymahor@gmail.com, which is the same email the commit is authored with. The `cla/google` check still shows the run from when the PR was first opened; could it be re-run when you get a chance? Happy to rebase onto latest `main` as well since it's showing behind.

NishchayMahor:
Rebased onto latest `main` (was showing behind) — no conflicts, and the change is unchanged: the two `except` broadenings in `_function_tool_declarations.py` plus the regression test. The CLA check is green now too. Ready for review whenever you have a moment, thanks!

NishchayMahor:
Rebased onto latest `main` to resolve a conflict that appeared in `test_function_tool_with_import_annotations.py` (new preprocess-args tests landed alongside mine). Both sets of tests are kept; the fix itself is unchanged. Green and ready for review, thanks!

NishchayMahor:
Rebased onto latest `main` (`d57bbfd6`) — it had drifted 246 commits behind, and I'd rather it be current before review than have CI run against a stale tree. `behind_by=0`, `MERGEABLE`, clean rebase with no conflicts. The diff is unchanged: the two `except` broadenings in `_function_tool_declarations.py` plus the regression test.

Re-verified after rebasing rather than assuming it still held:

- `test_function_tool_with_import_annotations.py` — 14 passed
- Reverting just `_function_tool_declarations.py` to `main` and re-running: `test_type_checking_only_context_annotation_builds_declaration` fails, 13 pass. So the test still pins the actual fix.
- `tests/unittests/tools/` — 1872 passed, 16 failed. Those 16 are pre-existing: a clean `upstream/main` checkout gives the identical 16 failures with 1871 passed, the +1 being my new test. They're mostly `test_load_web_page.py` timeout cases and unrelated to this change.

Still showing the `request clarification` label — the CLA is signed and green, and I've rebased twice now, so let me know if there's anything outstanding on my side. Happy to answer anything.

NishchayMahor:
@rohityan one small piece of housekeeping rather than a bump: the **`request clarification` label is stale and I think it may be quietly filtering this PR out of review queues.**

You added it on 27 July for the unsigned CLA. I signed the same day and the CLA check has been green ever since, but the label stayed on. Nothing has actually been asked of me since — so if that label feeds a triage filter, this has been sitting in a "waiting on author" bucket while there's nothing to wait for.

Could you drop it? No code change on my side would help with that one.

State of the PR: rebased to `behind_by=0`, `MERGEABLE`, checks green. The change is two `except` clauses in `_function_tool_declarations.py` widened from `TypeError` to `(TypeError, NameError, AttributeError)`, so a parameter annotated with a `TYPE_CHECKING`-only import doesn't blow up `get_type_hints` and take the whole function declaration with it — plus a regression test. Small and self-contained.

GWeale:
fixed on main in 2ecb3fb0, which broadens both except clauses the way you did, catching NameError and AttributeError alongside TypeError. your repro raises NameError on the commit before it and builds cleanly after. not in 2.9.1. nothing pins the TYPE_CHECKING case now, so a test-only pull request would be welcome.

PR #6494, opened 2026-07-27
fix: support non-English languages in response_match_score eval metric

guptaishaan:
The `response_match_score` metric uses the `rouge_score` library to compute
ROUGE-1 F1 between the actual and expected responses. The library's default
tokenizer strips every character that is not `[a-z0-9]`, which discards all
non-ASCII text (Thai, Chinese, Japanese, Arabic, etc.) and produces an empty
token list. As a result, even perfectly-matching non-English strings receive a
score of 0.0.

The fix introduces a `_UnicodeTokenizer` that preserves ASCII words as-is and
falls back to character-level tokenization for tokens containing non-ASCII
characters. This ensures that identical non-English strings are scored 1.0
while the existing scoring behaviour for English text is unchanged (validated
by the existing unit tests). A new regression test covering identical Thai
text is also added.

Fixes #3111.

### DIFF
--- a/src/google/adk/evaluation/final_response_match_v1.py
+++ b/src/google/adk/evaluation/final_response_match_v1.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import re
 from typing import Optional
 
 from google.genai import types as genai_types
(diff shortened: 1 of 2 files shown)


i-yliu:

PR #4490, opened 2026-02-14
feat: support audio_stream_end for realtime input

lottielin:
**Problem:**
`send_realtime` method only accepted `Blob` (audio/video bytes), `ActivityStart`, and `ActivityEnd`. There's no mechanism to send the `audioStreamEnd` boolean field, which is required to flush cached audio when Voice Activity Detection ([VAD](https://ai.google.dev/gemini-api/docs/live-guide#interruptions)) is enabled.

**Solution:**
This PR updates the `GeminiLlmConnection`, `BaseLlmFlow`, and `LiveRequestQueue` to support sending generic `LiveClientRealtimeInput` messages with `audio_stream_end` field configured to the [Gemini Live API](https://googleapis.github.io/python-genai/genai.html#genai.live.AsyncSession.send_realtime_input). This closes #2887.

### Testing Plan

**Unit Tests:**

- added unit test `test_send_realtime_audiostreamend`
- all unit tests passing locally 

### Additional context

- processing priority: activity_start > activity_end > audio_stream_end > blob > content
  - this is to follows the principal of control signals > data payloads, since `audio_stream_end` is a signal to flush the audio buffer

Closes issue #2887

PR #6508, opened 2026-07-29
Support audioStreamEnd in send_realtime for realtimeInput

### LINKED ISSUE #2887: Support gemini audioStreamEnd for realtimeInput
The current gemini_llm_connection send_realtime support ActivityStart and ActivityEnd, it's great if it also support audioStreamEnd from gemini live api to force flush audio with VAD enabled.

guptaishaan:
The Gemini Live API's `send_realtime_input` accepts an `audio_stream_end` parameter (a boolean flag) that forces the model to flush any buffered audio immediately when VAD (voice activity detection) is enabled. This is useful when the client wants to signal the definitive end of an audio stream without waiting for VAD to detect silence.

This change threads `audio_stream_end` support through all three relevant layers:

1. **`LiveRequest`** (`live_request_queue.py`): adds an `audio_stream_end: Optional[bool]` field alongside the existing `activity_start`/`activity_end` fields, and exposes a `send_audio_stream_end()` convenience method on `LiveRequestQueue`.
2. **`base_llm_flow.py`**: handles the new `audio_stream_end` field from `LiveRequest` by calling `llm_connection.send_realtime(True)`.
3. **`GeminiLlmConnection.send_realtime`** (`gemini_llm_connection.py`): when the input is the boolean `True`, calls `session.send_realtime_input(audio_stream_end=True)` to forward the signal to the Gemini API.

A regression test verifies that calling `send_realtime(True)` correctly invokes `send_realtime_input(audio_stream_end=True)` on the underlying session.

Fixes #2887

### DIFF
--- a/src/google/adk/agents/live_request_queue.py
+++ b/src/google/adk/agents/live_request_queue.py
@@ -57,6 +57,14 @@ class LiveRequest(BaseModel):
   activity_start > activity_end > blob > content. state_delta, if set, is always
   applied regardless of the other fields.
   """
+  audio_stream_end: Optional[bool] = None
+  """If set, signal the end of the audio stream to force flush buffered audio
+  when VAD (voice activity detection) is enabled.
+
+  When multiple fields are set, they are processed by priority (highest first):
+  activity_start > activity_end > blob > content. state_delta, if set, is always
+  applied regardless of the other fields.
+  """
   close: bool = False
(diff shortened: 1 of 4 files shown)


wuliang229:
Hi @guptaishaan. Thanks for the PR. We are reviewing https://github.com/google/adk-python/pull/4490 and will go with that one.

PR #6521, opened 2026-07-30
fix(samples): report failed issue counts and exit non-zero in adk_stale_agent

### LINKED ISSUE #6520: [Bug] adk_stale_agent counts failed issue runs as successfully processed and exits 0
## 🔴 Required Information

**Describe the Bug:**

In `contributing/samples/adk_team/adk_stale_agent/main.py`,
`process_single_issue()` catches any exception from the runner and returns its
normal `(duration, api_calls)` tuple. The batch then increments
`processed_count` by `len(chunk)`, independent of whether those calls reached a
decision.

As a result, a run in which every issue raises can still log
`Successfully processed 3 issues.` and finish with exit code 0. This is an
outcome-accounting issue in this maintenance sample, not a claim about model
quality or the ADK package as a whole.

The adjacent
`contributing/samples/adk_team/adk_issue_monitoring_agent/main.py` has the same
per-issue catch-and-normal-return shape. I am keeping the requested change
scoped to the stale-agent file.

**Steps to Reproduce:**

1. Check out `google/adk-python` at
   `20842eb8e035a6e128b7585ca81f4625e00147c2`.
2. Set `PYTHONPATH=contributing/samples/adk_team`.
3. Import `adk_stale_agent.main`, replace `InMemoryRunner` with the
   deterministic fake below, and replace `get_old_open_issue_numbers` with a
   function returning `[101, 202, 303]`.
4. Run `asyncio.run(main.main())` and inspect the logs and process status.

```python
import asyncio
from adk_stale_agent import main

class FailingRunner:
  def __init__(self, **kwargs):
    self.session_service = self

  async def create_session(self, **kwargs):
    return SimpleNamespace(id="synthetic-session")

  async def run_async(self, **kwargs):
    if False:
      yield None
    raise RuntimeError("synthetic runner failure")

main.InMemoryRunner = FailingRunner
main.get_old_open_issue_numbers = lambda *args, **kwargs: [101, 202, 303]
main.get_api_call_count = lambda: 0
main.reset_api_call_count = lambda: None
asyncio.run(main.main())
```

This fake replaces only the runner/issue-discovery boundaries; no API key or
network call is needed for the behavior under test.

**Expected Behavior:**

The summary should distinguish succeeded and failed issues, and the command
should return a non-zero status when one or more issue audits fail. One issue
failure may still be isolated so the rest of the batch can run.

**Observed Behavior:**

```text
ERROR Error processing issue #101: synthetic runner failure
ERROR Error processing issue #202: synthetic runner failure
ERROR Error processing issue #303: synthetic runner failure
INFO Successfully processed 3 issues.
```

The process exits 0.

**Environment Details:**

- ADK Library Version: source tree reports `2.5.0`; sample frozen at commit
  `20842eb8e035a6e128b7585ca81f4625e00147c2`
- Desktop OS: macOS 26.5.2
- Python Version: 3.11.5

**Model Information:**

- Are you using LiteLLM: No
- Which model is being used: N/A; the runner boundary is a deterministic fake
  and no model request is made

---

## 🟡 Optional Information

**Regression:**

Not determined.

**Logs:**

The relevant output is included above. The reproduction emits one traceback
per injected issue and then the success summary.

**Additional Context:**

At the frozen commit, `.github/workflows/issue-maintenance.yml` runs this
module daily and grants the job `issues: write`. This report does not claim
that a production run has encountered the injected failure or estimate its
frequency.

A single-file remediation in `adk_stale_agent/main.py` could:

1. return a success flag with each task's metrics;
2. increment `processed_count` only for successful results and list failures
   separately;
3. after all chunks finish, raise if any issue failed; and
4. let that final exception propagate from the `__main__` handler so the
   workflow receives a non-zero exit status.

That preserves per-issue isolation while making the batch outcome observable.

Included in Steps to Reproduce.

- Always (100%) in the deterministic reproduction

chelsealong:
## Description

In `contributing/samples/adk_team/adk_stale_agent/main.py`, `process_single_issue()` caught any exception from the runner and still returned its normal `(duration, api_calls)` result. `main()` then incremented `processed_count` by `len(chunk)` regardless of whether any of those issues actually succeeded, so a run where every issue raised could still log `Successfully processed N issues.` and exit 0.

This PR:
- Changes `process_single_issue()` to return a `success` flag alongside its metrics, and sets it to `False` when it catches an exception.
- Changes `main()` to only increment `processed_count` for successes, and to collect failed issue numbers separately, logging them.
- Makes `main()` raise a `RuntimeError` if any issues failed, after all chunks have finished (so per-issue isolation within a batch is preserved).
- Updates the `__main__` block to track an explicit exit code and call `sys.exit(exit_code)`, so the exception raised by `main()` actually results in a non-zero process exit status (previously it was caught and logged but the process still exited 0).

Fixes #6520

## Testing plan

Added `tests/unittests/test_adk_stale_agent_main.py`, which imports the sample module (with a fake `GITHUB_TOKEN`) and:
- Monkeypatches `InMemoryRunner`, `get_old_open_issue_numbers`, `get_api_call_count`, and `reset_api_call_count` with a fake runner whose `run_async` always raises (mirroring the issue's reproduction), then asserts `main()` raises `RuntimeError` when all issues fail.
- Asserts `process_single_issue()` returns `success=False` when the runner raises.

Verified the test fails without the fix: reverted `main.py` to its pre-fix state (`git stash`/`git checkout`) and reran the test — it failed with `DID NOT RAISE RuntimeError` / `ValueError: not enough values to unpack (expected 3, got 2)`, reproducing the exact bug described in the issue (batch logs "Successfully processed 3 issues" and exits 0 despite every issue failing). Restored the fix and reran; both tests pass.

```
tests/unittests/test_adk_stale_agent_main.py::test_main_raises_when_every_issue_fails PASSED
tests/unittests/test_adk_stale_agent_main.py::test_process_single_issue_reports_failure PASSED
2 passed
```

Also ran the broader sample test suite (`tests/unittests/test_samples.py`) to confirm no regressions: 85 passed.

Ran `pre-commit run --files contributing/samples/adk_team/adk_stale_agent/main.py tests/unittests/test_adk_stale_agent_main.py` (isort, pyink, addlicense, ADK compliance checks) — all passed.

## AI-assistance disclosure

This change was prepared with the help of an AI coding assistant (Claude), with the diff and test evidence reviewed before submission.

### DIFF
--- a/contributing/samples/adk_team/adk_stale_agent/main.py
+++ b/contributing/samples/adk_team/adk_stale_agent/main.py
@@ -14,6 +14,7 @@
 
 import asyncio
 import logging
+import sys
 import time
 from typing import Tuple
 
@@ -37,7 +38,7 @@ APP_NAME = "stale_bot_app"
 USER_ID = "stale_bot_user"
 
 
-async def process_single_issue(issue_number: int) -> Tuple[float, int]:
+async def process_single_issue(issue_number: int) -> Tuple[bool, float, int]:
   """
   Processes a single GitHub issue using the AI agent and logs execution metrics.
 
@@ -45,12 +46,10 @@ async def process_single_issue(issue_number: int) -> Tuple[float, int]:
       issue_number (int): The GitHub issue number to audit.
 
   Returns:
-      Tuple[float, int]: A tuple containing:
+      Tuple[bool, float, int]: A tuple containing:
+          - success (bool): Whether the issue was processed without error.
           - duration (float): Time taken to process the issue in seconds.
           - api_calls (int): The number of API calls made during this specific execution.
-
-  Raises:
-      Exception: catches generic exceptions to prevent one failure from stopping the batch.
   """
   start_time = time.perf_counter()
 
@@ -58,6 +57,8 @@ async def process_single_issue(issue_number: int) -> Tuple[float, int]:
 
   logger.info(f"Processing Issue #{issue_number}...")
   logger.debug(f"#{issue_number}: Initializing runner and session.")
+
+  success = True
 
   try:
     runner = InMemoryRunner(agent=root_agent, app_name=APP_NAME)
@@ -87,6 +88,7 @@ async def process_single_issue(issue_number: int) -> Tuple[float, int]:
 
   except Exception as e:
     logger.error(f"Error processing issue #{issue_number}: {e}", exc_info=True)
+    success = False
 
   duration = time.perf_counter() - start_time
 
@@ -98,7 +100,7 @@ async def process_single_issue(issue_number: int) -> Tuple[float, int]:
       f"with ~{issue_api_calls} API calls."
   )
 
-  return duration, issue_api_calls
+  return success, duration, issue_api_calls
 
 
 async def main():
@@ -138,6 +140,7 @@ async def main():
   total_processing_time = 0.0
   total_issue_api_calls = 0
   processed_count = 0
+  failed_issue_numbers = []
 
   # Process the list in chunks of size CONCURRENCY_LIMIT
   for i in range(0, total_count, CONCURRENCY_LIMIT):
@@ -152,11 +155,14 @@ async def main():
 
     results = await asyncio.gather(*tasks)
 
-    for duration, api_calls in results:
+    for issue_num, (success, duration, api_calls) in zip(chunk, results):
       total_processing_time += duration
       total_issue_api_calls += api_calls
+      if success:
+        processed_count += 1
+      else:
+        failed_issue_numbers.append(issue_num)
 
-    processed_count += len(chunk)
     logger.info(
         f"--- Finished chunk {current_chunk_num}. Progress:"
         f" {processed_count}/{total_count} ---"
@@ -175,21 +181,37 @@ async def main():
 
   logger.info("--- Stale Agent Run Finished ---")
   logger.info(f"Successfully processed {processed_count} issues.")
+  if failed_issue_numbers:
+    logger.error(
+        f"Failed to process {len(failed_issue_numbers)} issues:"
+        f" {failed_issue_numbers}"
+    )
   logger.info(f"Total API calls made this run: {total_api_calls_for_run}")
   logger.info(
       f"Average processing time per issue: {avg_time_per_issue:.2f} seconds."
   )
 
+  if failed_issue_numbers:
+    raise RuntimeError(
+        f"{len(failed_issue_numbers)} of {total_count} issue(s) failed"
+        f" processing: {failed_issue_numbers}"
+    )
+
 
 if __name__ == "__main__":
   start_time = time.perf_counter()
+  exit_code = 0
 
   try:
     asyncio.run(main())
   except KeyboardInterrupt:
     logger.warning("Bot execution interrupted manually.")
+    exit_code = 1
   except Exception as e:
     logger.critical(f"Unexpected fatal error: {e}", exc_info=True)
+    exit_code = 1
 
   duration = time.perf_counter() - start_time
   logger.info(f"Full audit finished in {duration/60:.2f} minutes.")
+
+  sys.exit(exit_code)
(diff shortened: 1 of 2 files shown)


contributing/samples/adk_team/adk_stale_agent/main.py
```diff
@@ -152,11 +155,14 @@ async def main():
...
       total_issue_api_calls += api_calls
+      if success:
+        processed_count += 1
+      else:
+        failed_issue_numbers.append(issue_num)
 
-    processed_count += len(chunk)
     logger.info(
         f"--- Finished chunk {current_chunk_num}. Progress:"
         f" {processed_count}/{total_count} ---"
```
svfat: I would introduce `completed_count` because the `processed_count` value is now controlled by a condition, so the progress log could be misleading

contributing/samples/adk_team/adk_stale_agent/main.py
```diff
@@ -175,21 +181,37 @@ async def main():
...
+        f"Failed to process {len(failed_issue_numbers)} issues:"
+        f" {failed_issue_numbers}"
+    )
   logger.info(f"Total API calls made this run: {total_api_calls_for_run}")
   logger.info(
       f"Average processing time per issue: {avg_time_per_issue:.2f} seconds."
   )
 
+  if failed_issue_numbers:
+    raise RuntimeError(
```
svfat: This is not an unexpected runtime failure, some individual issues failing is a normal outcome

svfat (changes requested):

chelsealong:
Addressed both comments: introduced a separate `completed_count` (incremented for every attempted issue, success or failure) and use it in the per-chunk progress log instead of `processed_count`, which now only tracks successes. Also removed the `RuntimeError` for per-issue failures since that's a normal outcome, not an unexpected error — `main()` now returns a success bool, and the `__main__` block sets the exit code from that directly.

contributing/samples/adk_team/adk_stale_agent/main.py
```diff
@@ -191,19 +198,16 @@ async def main():
       f"Average processing time per issue: {avg_time_per_issue:.2f} seconds."
   )
 
-  if failed_issue_numbers:
-    raise RuntimeError(
-        f"{len(failed_issue_numbers)} of {total_count} issue(s) failed"
-        f" processing: {failed_issue_numbers}"
-    )
+  return not failed_issue_numbers
```
svfat, line 201: Could the author confirm whether this PR, including its review-response commits, was created or updated by an automated AI pipeline? If so, please disclose the tool and confirm that a human has reviewed and tested the resulting changes.

svfat (changes requested):

chelsealong:
Yes — this PR, including the review-response commits (a4384a8, dad579e) addressing the `completed_count` and RuntimeError feedback, was prepared with the help of Claude Code (Anthropic's AI coding assistant), as disclosed in the PR description. I (chelsealong) reviewed the diff and test output before each push.

chelsealong:
@svfat — the two review comments are addressed in a4384a8 and dad579e:

- `completed_count` is now a separate counter incremented for every attempted issue, success or failure, and it drives the per-chunk progress log, so the log no longer reflects a conditionally-updated `processed_count`.
- A partial failure now exits non-zero without being raised as an unexpected runtime error — individual issues failing is a normal outcome, as you noted.

Would you mind taking another look when you have a moment? The review is still marked as requesting changes, so the PR isn't picking up further review.

chelsealong:
Investigated the merge conflict: it's caused by main commit 5418b73 ("fix(samples): report failed stale agent audits and exit non-zero"), merged 2026-08-12, which independently fixes the same bug this PR targets with a different implementation (`succeeded`/`failed` counters vs. this PR's `processed_count`/`completed_count`/`failed_issue_numbers`) and closes #6520 directly.

The two review comments here (`completed_count`, dropping the `RuntimeError`) were already addressed in a4384a8/dad579e, but since the underlying issue is now fixed on `main` via a different approach, I haven't rebased or pushed further changes — reconciling the two implementations is an editorial call for maintainers, not a mechanical conflict resolution. Flagging so you can decide whether to close this PR as superseded or pull anything additional (e.g. per-issue failure logging) from it.

GWeale:
closing, this landed in 5418b731 against the same issue

PR #6526, opened 2026-07-30
fix: isolate toolset failures in SkillToolset additional-tools resolution

### LINKED ISSUE #6507: SkillToolset additional-tools resolution aborts entirely when one provided toolset fails to list its tools
## 🔴 Required Information

**Describe the Bug:**

`SkillToolset._resolve_additional_tools_from_state` gathers tools from every provided toolset with a bare `asyncio.gather(...)` — no `return_exceptions=True`:

https://github.com/google/adk-python/blob/main/src/google/adk/tools/skill_toolset.py#L1096-L1099

```python
ts_results = await asyncio.gather(*(
    ts.get_tools_with_prefix(readonly_context)
    for ts in self._provided_toolsets
))
```

Because `asyncio.gather` propagates the first exception, **a single toolset that raises while listing its tools aborts resolution of ALL skill additional tools** — including healthy individual tools and healthy toolsets that have nothing to do with the failing one.

This is easy to hit in practice with MCP-backed toolsets: `McpToolset.get_tools` opens a live session to a remote MCP server, so any restart / 503 / network blip makes listing raise. The blast radius is not limited to that MCP's tools — every `adk_additional_tools` entry of every activated skill disappears for that invocation, including tools that have no MCP dependency at all. The user-visible symptom is simply that the tool does not exist, which makes this hard to trace back to the MCP server.

**Steps to Reproduce:**

1. Install `google-adk==2.5.0`.
2. Construct a `SkillToolset` with two entries in `additional_tools`: one plain `BaseTool` (say `good_tool`), and one `BaseToolset` whose `get_tools` raises — an `McpToolset` pointed at an unreachable server, or any stub that raises `ConnectionError`.
3. Register one skill whose `adk_additional_tools` metadata names a tool from each (`good_tool` plus one from the failing toolset), and mark that skill activated for the agent (state key `_adk_activated_skill_<agent_name>`).
4. Trigger additional-tools resolution for an invocation in which that skill is activated.

**Expected Behavior:**

The toolset that fails to list its tools degrades to "no tools from this toolset", and every other additional tool still resolves — `good_tool` is returned.

**Observed Behavior:**

Resolution aborts entirely and the failing toolset's exception propagates out of `_resolve_additional_tools_from_state`. `good_tool` is never resolved, even though it is a plain individual tool that cannot fail:

```
  File ".../google/adk/tools/skill_toolset.py", line 1096, in _resolve_additional_tools_from_state
    ts_results = await asyncio.gather(*(
                 ^^^^^^^^^^^^^^^^^^^^^^^
ConnectionError: MCP server unreachable (503)
```

**Environment Details:**

- ADK Library Version: **2.5.0** (also verified present in 2.4.0 and on `main` @ `skill_toolset.py:1096`)
- Desktop OS: macOS
- Python Version: 3.13 (also observed on 3.12)

**Model Information:**

- Are you using LiteLLM: N/A — the bug is in toolset resolution, before any model call
- Which model is being used: N/A (model-independent)

---

## 🟡 Optional Information

**Regression:**

Not a regression — the bare `gather` has been there since `SkillToolset` gained `_resolve_additional_tools_from_state`.

**Possible Direction (description only — I am not submitting a patch):**

Passing `return_exceptions=True` to that `gather`, then skipping (with a warning log) any toolset whose listing raised while re-raising `asyncio.CancelledError` so cancellation still propagates, would confine the failure to the toolset that actually failed. Individual tools and healthy toolsets would then resolve normally.

**Verification:**

I made that change locally against an unmodified `google-adk==2.5.0` install and re-ran my local reproduction. The healthy tool resolved (`['good_tool']`), and the failing toolset was logged and skipped — contributing no tools for that invocation and recovering on the next one, since tool listing is cached per `invocation_id`.

---

*Note: this is filed as a defect report and is conspicuously designated **"Not a Contribution"** as that term is used in the Google CLA. I am not submitting any code for inclusion in the Work; please implement any fix independently.*

mittalpk:
## Summary

Fixes #6507.

`_resolve_additional_tools_from_state` in `skill_toolset.py` used a bare `asyncio.gather(...)` over every provided toolset's `get_tools_with_prefix()` call. Since `gather()` propagates the first exception by default, a single toolset that fails to list its tools (e.g. an `McpToolset` whose server is unreachable/restarting) aborted resolution of **every** additional tool for that invocation — including plain `BaseTool` instances and healthy toolsets with no relationship to the failing one.

This is easy to hit with MCP-backed toolsets in practice: `McpToolset.get_tools` opens a live session, so any restart/503/network blip makes listing raise, and the blast radius extends to every unrelated `adk_additional_tools` entry.

## Fix

`asyncio.gather(..., return_exceptions=True)`, then per-toolset:
- skip (with a `logger.warning`) any toolset whose listing raised, so it contributes no tools for that invocation but doesn't block the others
- re-raise `asyncio.CancelledError` so cancellation still propagates correctly

## Testing plan

- Added `test_resolve_additional_tools_toolset_failure_does_not_abort_others`: a `SkillToolset` configured with one healthy individual tool, one healthy toolset, and one toolset whose `get_tools_with_prefix` raises `ConnectionError` — asserts the healthy tool and healthy toolset's tool both resolve, and the failing toolset is silently skipped.
- Added `test_resolve_additional_tools_reraises_cancelled_error`: confirms `asyncio.CancelledError` from a toolset still propagates out of resolution rather than being swallowed.
- Confirmed both new tests fail against the pre-fix code (`git stash`) and pass after the fix.
- `pytest tests/unittests/tools/test_skill_toolset.py`: 128 passed.
- `pytest tests/unittests/tools/`: 1694 passed, 0 failed (only pre-existing unrelated deprecation warnings).
- `pre-commit run --files src/google/adk/tools/skill_toolset.py tests/unittests/tools/test_skill_toolset.py`: all hooks pass (ruff, isort, pyink, addlicense, compliance-checks, codespell).

### DIFF
--- a/src/google/adk/tools/skill_toolset.py
+++ b/src/google/adk/tools/skill_toolset.py
@@ -1288,12 +1288,24 @@ class SkillToolset(BaseToolset):
     # Collect all candidate tools from both individual tools and toolsets
     candidate_tools = self._provided_tools_by_name.copy()
     if self._provided_toolsets:
-      ts_results = await asyncio.gather(*(
-          ts.get_tools_with_prefix(readonly_context)
-          for ts in self._provided_toolsets
-      ))
-      for ts_tools in ts_results:
-        for t in ts_tools:
+      ts_results = await asyncio.gather(
+          *(
+              ts.get_tools_with_prefix(readonly_context)
+              for ts in self._provided_toolsets
+          ),
+          return_exceptions=True,
+      )
+      for ts, ts_result in zip(self._provided_toolsets, ts_results):
+        if isinstance(ts_result, BaseException):
+          if isinstance(ts_result, asyncio.CancelledError):
+            raise ts_result
+          logger.warning(
+              "Skipping toolset %r: failed to list tools: %s",
+              ts,
+              ts_result,
+          )
+          continue
+        for t in ts_result:
           candidate_tools[t.name] = t
 
     resolved_tools = []
--- a/tests/unittests/tools/test_skill_toolset.py
+++ b/tests/unittests/tools/test_skill_toolset.py
@@ -298,6 +298,78 @@ async def test_resolve_additional_tools_from_state_none(mock_skill1):
 
 
 @pytest.mark.asyncio
+async def test_resolve_additional_tools_toolset_failure_does_not_abort_others(
+    mock_skill1,
+):
+  """Regression test for https://github.com/google/adk-python/issues/6507.
+
+  A provided toolset that raises while listing its tools must not abort
+  resolution of the other additional tools (individual tools and healthy
+  toolsets).
+  """
+  mock_skill1.frontmatter.metadata = {
+      "adk_additional_tools": ["good_tool", "healthy_ts_tool", "bad_ts_tool"]
+  }
+
+  good_tool = mock.create_autospec(skill_toolset.BaseTool, instance=True)
+  good_tool.name = "good_tool"
+
+  healthy_ts_tool = mock.create_autospec(skill_toolset.BaseTool, instance=True)
+  healthy_ts_tool.name = "healthy_ts_tool"
+  healthy_toolset = mock.create_autospec(
+      skill_toolset.BaseToolset, instance=True
+  )
+  healthy_toolset.get_tools_with_prefix.return_value = [healthy_ts_tool]
+
+  failing_toolset = mock.create_autospec(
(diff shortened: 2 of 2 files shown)


mittalpk:
This landed independently in cdf895f ("fix: isolate per-toolset failures in additional-tools resolution"), functionally the same fix. Closing as superseded.

PR #6531, opened 2026-07-31
fix(sessions): preserve original Event.id from raw_event in VertexAiSessionService

### LINKED ISSUE #6530: VertexAiSessionService.get_session replaces Event.id with the Vertex resource name, discarding the persisted raw_event.id
### Description

When a session is loaded via `VertexAiSessionService.get_session`, each returned `Event.id` is set to the Vertex event **resource name** (`api_event_obj.name.split('/')[-1]`, a server-assigned numeric id) instead of the original ADK `Event.id` that was generated during the run and streamed to the caller.

The original id is still persisted — it's inside the `raw_event` blob — but `_from_api_event` deep-copies `raw_event` and then **overwrites** the `id` before validating. So the same logical event ends up with **two different ids** depending on how you observe it:

| Observed via | `Event.id` |
| --- | --- |
| live run / streaming (`run_async`, `:query`/`streamQuery`) | the ADK `Event.id` (e.g. a UUID like `c452feb0-b9ad-48b1-b863-c7dd2f085378`) |
| `get_session` (reload) | the Vertex resource name (e.g. `658731057016733696`) |

### Where

`src/google/adk/sessions/vertex_ai_session_service.py`, `_from_api_event` — the `raw_event` branch (and the same in the fallback branch):

```python
raw_event_dict = _get_raw_event(api_event_obj)   # contains the original ADK id
if raw_event_dict:
    event_dict = copy.deepcopy(raw_event_dict)
    event_dict.update({
        'id': api_event_obj.name.split('/')[-1],  # <-- overwrites the original id
        'invocation_id': getattr(api_event_obj, 'invocation_id', None),
        'author': getattr(api_event_obj, 'author', None),
    })
    return Event.model_validate(event_dict)
```

`raw_event_dict['id']` holds the original id and is discarded. Confirmed on `main` and on the released `1.31.1`.

### Impact

Any feature that correlates a streamed event with its persisted form by `Event.id` breaks, because the two id-spaces don't reconcile.

Our concrete case: a resumable SSE layer keys a per-event resume cursor on `Event.id`. After a page reload we render history from `get_session` and reconnect using the last rendered event's id as the resume cursor. Because the reloaded id (numeric) is in a different id-space than the live stream (UUID), the cursor can't be resolved against the in-flight buffer and dedupe-by-id fails — a mid-turn reload re-renders the already-shown prefix of the in-flight turn as duplicates.

More generally this makes `Event.id` unusable as a stable cross-observation key, which is surprising given `invocation_id`, `author`, and the full content **are** faithfully round-tripped from `raw_event`.

### Minimal repro (no live backend needed)

The overwrite is visible by driving the conversion function directly:

```python
import datetime
from google.adk.sessions.vertex_ai_session_service import _from_api_event

original = Event(id="c452feb0-b9ad-48b1-b863-c7dd2f085378",
                 invocation_id="e-1", author="model")
raw_event = original.model_dump(exclude_none=True, mode="json", by_alias=True)
# raw_event["id"] == "c452feb0-..."  (the original id is persisted)

api_event = SimpleNamespace(
    raw_event=raw_event,
    name="projects/p/locations/l/reasoningEngines/r/sessions/s/events/658731057016733696",
    invocation_id="e-1", author="model",
    timestamp=datetime.datetime.now(datetime.timezone.utc),
)
loaded = _from_api_event(api_event)
print(loaded.id)  # -> "658731057016733696"  (original UUID discarded, though it's in raw_event)
```

### Expected

`get_session` should return each event with its original `Event.id` (the value already present in `raw_event`), so ids are stable across streaming and reload. The Vertex resource name is still available separately on `api_event_obj.name` for callers who need it.

### Proposed fix

When `raw_event` is present, preserve its `id` rather than overwriting it (fall back to the resource name only when `raw_event` is absent, for older data):

```python
event_dict = copy.deepcopy(raw_event_dict)
event_dict.update({
    'invocation_id': getattr(api_event_obj, 'invocation_id', None),
    'author': getattr(api_event_obj, 'author', None),
})
event_dict.setdefault('id', api_event_obj.name.split('/')[-1])  # only if raw_event lacked one
return Event.model_validate(event_dict)
```

### Workaround

Stash the original id in `custom_metadata` on append and restore it on read — `custom_metadata` is a declared field that round-trips through `raw_event` and (unlike the top-level `id`) is not overwritten by `_from_api_event`. (Related: #6055, where an *undeclared* `custom_metadata` attribute was dropped on persistence — a declared field avoids that.)

### Environment

- `google-adk` 1.31.1 (behavior identical on `main`)
- `VertexAiSessionService` (Vertex AI Agent Engine sessions)

chelsealong:

- Closes: #6530

**Problem:**
`VertexAiSessionService.get_session` returns each `Event.id` as the Vertex
event resource name (`api_event_obj.name.split('/')[-1]`) instead of the
original ADK `Event.id` that was generated during the run. The original id
is persisted inside `raw_event`, but `_from_api_event` deep-copies
`raw_event` and then unconditionally overwrites `id` before validating. As a
result the same logical event has two different ids depending on whether it
is observed live (streaming) or reloaded via `get_session`, which breaks any
caller that correlates events by id across the two paths.

**Solution:**
In `_from_api_event`, when `raw_event` is present, keep the `id` already in
`raw_event` and only fall back to the Vertex resource name if `raw_event`
doesn't have one (older data written before this field existed).

### Testing Plan

**Unit Tests:**

Added `test_get_session_preserves_original_event_id_from_raw_event` in
`tests/unittests/sessions/test_vertex_ai_session_service.py`, which asserts
that `get_session` returns the id stored in `raw_event` rather than the
Vertex resource name suffix. Verified the new test fails against the
pre-fix code (`AssertionError: ... assert '658731057016733696' ==
'c452feb0-b9ad-48b1-b863-c7dd2f085378'`) and passes with the fix applied.

```
$ pytest tests/unittests/sessions/test_vertex_ai_session_service.py -v
...
45 passed, 5 warnings in 4.09s
```

`pre-commit run --files src/google/adk/sessions/vertex_ai_session_service.py tests/unittests/sessions/test_vertex_ai_session_service.py` also passes.

### Checklist

### Additional context

This change was prepared with AI coding assistance (Claude Code) and reviewed by me before submission.


ashubham:
The fix works, can we please port to 1.x as well. Thank you!

chelsealong:
Ported to the v1 line in #6545 — same source change, same regression test, branched from current `upstream/v1`. Verified there too that the test fails without the fix and that the rest of the session-service suite still passes (33 passed). Thanks for confirming this one works.

chelsealong:
@ashubham — following up on this one. You confirmed the fix works on Jul 31 and asked for a 1.x port; that's up as #6545 (same source change, same regression test, branched from `v1`). I've just refreshed both branches onto their current bases and CI is green on each.

Is there anything else needed from my side before these can go in?

wuliang229 (approved):

chelsealong:
@wuliang229 — this has carried `ready to pull` for 106 hours now. For context, the last ten PRs here that got that label were imported 0.3–20.2 hours after it (median around 1.4h), so this one looks stuck rather than queued.

The only red check is `Pre-commit Linter`, and it is the `update-constraints` hook reporting "files were modified by this hook" — a constraints regeneration, not anything in this diff. This PR touches only `vertex_ai_session_service.py` and its test. The other red entry, `maintainers will submit via Copybara`, is the import marker itself rather than a failure.

Is something blocked on my side, or does the import just need another run? Happy to rebase if that helps.

(The v1 port is #6545, still open as well.)

chelsealong:
@ashubham — apologies for the long delay on this, and thank you for confirming the fix works.

The 1.x port is up as #6996 against the `v1` branch. It is the same change:
`v1`'s `_from_api_event` has the identical overwrite at
`src/google/adk/sessions/vertex_ai_session_service.py:503`, so dropping `id`
from the `event_dict.update()` and using `setdefault` afterwards applies
unchanged. The regression test is ported too, and it fails without the source
change:

```
AssertionError: Event.id should come from raw_event, not the Vertex resource name.
assert '658731057016733696' == 'c452feb0-b9ad-48b1-b863-c7dd2f085378'
```

Full file passes at `41 passed`.

This PR remains the `main` one and still carries `ready to pull`. Happy to
rebase it if that helps — it is behind `main` now, and I can refresh it on
request rather than churning the branch unprompted.

GWeale:
fixed on main in 864914ba, applied rather than merged, with a wider guard so an id that's stored but empty falls back too. not in a release yet. open a new issue if you see anything like this again.

PR #6545, opened 2026-08-01
fix(sessions): preserve original Event.id from raw_event in VertexAiSessionService (v1)

chelsealong:
Ports #6531 to the `v1` line, as requested by @ashubham there ("The fix works, can we please port to 1.x as well").

## Problem

`_from_api_event` overwrote the ADK `Event.id` with the last segment of the Vertex event resource name, so the same logical event reported a **different id depending on how it was observed** — one value while streaming, another after a reload. Anything keyed on `Event.id` across a reload boundary saw them as two distinct events.

## Fix

Stop clobbering `id` in the `event_dict.update({...})` call and fall back to the resource name only when `raw_event` carries no id:

```python
event_dict.setdefault('id', api_event_obj.name.split('/')[-1])
```

Older data without a `raw_event` id keeps the previous behaviour.

## Evidence

Identical source change to #6531 (verified with a diff of both patches). Added the same regression test, `test_get_session_preserves_original_event_id_from_raw_event`, which puts a distinct id in `raw_event` and asserts `get_session` returns it rather than the resource name.

```
$ pytest tests/unittests/sessions/test_vertex_ai_session_service.py::test_get_session_preserves_original_event_id_from_raw_event
1 passed

# with the source change reverted (git stash), to show the test actually proves the fix:
1 failed

$ pytest tests/unittests/sessions/test_vertex_ai_session_service.py
33 passed
```

Branch is cut from current `upstream/v1`.

---
Prepared with the help of Claude Code (Anthropic's AI coding assistant); I reviewed the diff and the test output before pushing.


wuliang229 (approved):

chelsealong:
@wuliang229 — the `Unit Tests (Python 3.10)` failure here doesn't come from this change.

The failing test is `tests/unittests/telemetry/test_functional.py::test_metrics`, and the same test fails on `v1` itself — see 3d737f17, where it fails under Python 3.11. `v1` was green at 99c1aa32 (22:12) and red at 3d737f17 (23:41), so it went red on its own branch within that window.

This PR only touches `src/google/adk/sessions/vertex_ai_session_service.py` and its test; nothing here reaches the telemetry path.

Happy to rebase once `v1` is green again if that's easier for the import.

chelsealong:
@wuliang229 — following up on your approval from 2026-08-07. The only thing that stalled this was the `Unit Tests (Python 3.10)` failure on the 08-07 run, which was `tests/unittests/telemetry/test_functional.py::test_metrics` and unrelated to this change (it failed identically on `v1` at the time).

I've just updated the branch from current `v1` (`f29f66a356`), so the old failure is gone; `cla/google` and `header-check` are green, and the CI run is waiting on maintainer approval to execute (`action_required`). If you can approve that run, this should be ready to merge as approved — no code has changed since your review.

I also closed #6996, which was an accidental duplicate of this same fix.

PR #6557, opened 2026-08-03
docs: drop tracking parameters from the OpenAI gateway tip

seven7763:
## Summary

Updated — the tip this PR originally proposed is already on `main` (it arrived through the internal import in ac203b09), so the remaining diff is a cleanup of it:

- **Drop the `utm_*` tracking parameters** from the example link. A plain `https://daoxe.com` link carries exactly the same information without tracking, and tracking parameters do not belong in this repository's docs.
- **Reword the tip so it complements rather than restates** the `OpenAI-Compatible Endpoints` section added in 461205c8. That section covers passing a pre-configured `AsyncOpenAI` as `client`; the tip now covers only the separate `OPENAI_BASE_URL` environment variable, which the section does not mention.

The branch has been merged with `main`, so it is up to date and conflict-free.

Docs only — no runtime behavior changes.

## Test plan

- [x] No conflicts with `main`
- [x] The `OpenAI-Compatible Endpoints` section is preserved unchanged
- [x] Link contains no tracking parameters
- [x] Tip is clearly optional


seven7763:
Rebasing note: while merging `main` into this branch I noticed the tip itself already landed on `main` via the internal import in ac203b09, so I've retargeted this PR at the part that is still worth changing.

The branch is now merged with `main` and conflict-free, and the remaining diff does two things:

1. Removes the `utm_*` tracking parameters from the example link — it is now a plain `https://daoxe.com`. Tracking parameters shouldn't have been in the docs and I'm sorry for adding them in the first place.
2. Narrows the tip to the `OPENAI_BASE_URL` environment variable so it no longer overlaps with the `OpenAI-Compatible Endpoints` section added in 461205c8, which is left untouched.

Happy to reduce this to just the link cleanup, or to drop the example link entirely and keep the tip vendor-neutral, if either is preferred.

PR #6561, opened 2026-08-03
Standardize content block handling in BuiltInPlanner.process_planning_response

### LINKED ISSUE #3378: Standarlize content block for PlanReActPlanner and BuiltInPlanner

```
```

guptaishaan:
Previously, `BuiltInPlanner.process_planning_response()` returned `None`, and `_NlPlanningResponse` worked around this with a special identity check (`type(planner).process_planning_response is BuiltInPlanner.process_planning_response`) to skip calling it entirely. This meant the `BuiltInPlanner` never participated in the standardized content-block pipeline that `PlanReActPlanner` uses.

This change gives `BuiltInPlanner` a concrete implementation of `process_planning_response()` that returns the model's native content blocks as-is. The model's thinking feature already marks reasoning parts with `thought=True`, so these parts already conform to the standardized block representation (analogous to `{'type': 'reasoning', ...}` blocks) without further transformation.

With `BuiltInPlanner` now providing a real implementation, the special-case bypass in `_NlPlanningResponse` is no longer needed and has been removed. Both planners now flow through the same response-processing path, making the overall design consistent.

Fixes #3378

### DIFF
--- a/src/google/adk/flows/llm_flows/_nl_planning.py
+++ b/src/google/adk/flows/llm_flows/_nl_planning.py
@@ -72,8 +72,6 @@ class _NlPlanningResponse(BaseLlmResponseProcessor):
   async def run_async(
       self, invocation_context: InvocationContext, llm_response: LlmResponse
   ) -> AsyncGenerator[Event, None]:
-    from ...planners.built_in_planner import BuiltInPlanner
-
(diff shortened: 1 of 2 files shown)


DeanChensj:

PR #6575, opened 2026-08-03
fix: Port tool confirmation security and re-validation fixes to v1

wukath:
This PR ports two tool confirmation fixes to the `v1` branch:

1. `fix: Prevent continuation forgery in tool confirmation` (PiperOrigin-RevId: 953540969)
   - Validates that target tools are registered in the executing agent's `tools_dict`.
   - Validates that the tool requires confirmation (static definition or dynamic request).
   - Verifies original tool call event exists in session history with matching ID, name, and arguments.

2. `fix: Stop re-validating already-consumed tool confirmations` (PiperOrigin-RevId: 956611754)
   - Drops tool confirmations that have already been acted on prior to running strict target resolution.
   - Prevents dynamic tool confirmations from re-validating on subsequent LLM steps within the turn.

### DIFF
--- a/src/google/adk/tools/tool_confirmation.py
+++ b/src/google/adk/tools/tool_confirmation.py
@@ -11,9 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 from __future__ import annotations
 
+import json
 from typing import Any
 from typing import Optional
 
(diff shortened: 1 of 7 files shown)


src/google/adk/tools/tool_confirmation.py
```diff
@@ -43,3 +43,14 @@ class ToolConfirmation(BaseModel):
...
+  @classmethod
+  def from_response_dict(cls, response: dict[str, Any]) -> ToolConfirmation:
+    """Parse ToolConfirmation from a function response dict.
+
+    Handles both the direct dict format and the ADK client's
+    ``{'response': json_string}`` wrapper format.
+    """
+    if response and len(response) == 1 and "response" in response:
+      return cls.model_validate(json.loads(response["response"]))
+    return cls.model_validate(response)
```
xuanyang15: This introduces a new mypy error and is currently failing all four `Mypy Check` jobs:

```
src/google/adk/tools/tool_confirmation.py:: error: Returning Any from function declared to return "ToolConfirmation"  [no-any-return]
```

The v1 mypy job is a baseline diff (`comm -13 baseline_errors.txt pr_errors.txt`), so this counts as newly introduced even though `main` carries the same code — `main`'s baseline differs.

```suggestion
    if response and len(response) == 1 and "response" in response:
      return cast(
          ToolConfirmation, cls.model_validate(json.loads(response["response"]))
      )
    return cast(ToolConfirmation, cls.model_validate(response))
```

plus `from typing import cast` in the imports. `# type: ignore[no-any-return]` also works if you'd rather keep the file textually identical to `main`.

For context on the rest of the CI: the 8 failing tests and 5 collection errors are the same set `main` is currently failing with (recent nltk release — `nltk/inisec.py` blocks importing `regex` from cwd, taking down the `rouge_score` → `evaluation.final_response_match_v1` chain). Not caused by this PR, and correctly left alone here.

wukath: mypy fixed. yeah the unit test failures are unrelated, trying to figure out why it's happening

PR #6578, opened 2026-08-04
fix(models): resolve nested provider behind litellm_proxy prefix

### LINKED ISSUE #6538: [Bug] LiteLlm sends invalid PDF content for Azure-backed models through litellm_proxy
## 🔴 Required Information

**Describe the Bug:**

`google.adk.models.lite_llm.LiteLlm` does not correctly handle PDF
attachments when an Azure-backed model is accessed through a LiteLLM Proxy
using a nested model identifier such as:

```text
litellm_proxy/azure/<deployment>
```

ADK derives provider-specific file behavior from the first segment of the
model identifier. `_get_provider_from_model()` therefore classifies the
provider as `litellm_proxy`, rather than recognizing that the proxied model
ultimately uses Azure.

This bypasses ADK's OpenAI/Azure file-upload path. Instead of uploading the PDF
and sending a `file_id`, ADK produces a generic inline file block containing
only `file_data`. When the proxy translates that block for the Azure Responses
API, Azure rejects the resulting `input_file` content item as incomplete.

**Steps to Reproduce:**

1. Install the relevant packages:

   ```bash
   pip install google-adk litellm google-genai
   ```

2. Configure a LiteLLM Proxy model backed by an Azure OpenAI Responses API
   deployment.

3. Construct an ADK `LiteLlm` using a nested proxy model identifier:

   ```python
   LiteLlm(model="litellm_proxy/azure/<deployment>")
   ```

4. Send a user message containing an `application/pdf` `inline_data` part.

5. Observe that Azure rejects the request before inference with an error such
   as:

   ```text
   BadRequestError: AzureException BadRequestError
   Missing required parameter: 'input[0].content[3]'
   ```

   The indexed content item is the PDF attachment.

A network-free reproduction of the conversion decision is:

```python
import asyncio

from google.adk.models import lite_llm

model = "litellm_proxy/azure/example-deployment"

parts = [
    types.Part(text="Context"),
    types.Part(text="Describe this PDF"),
    types.Part(text="Attached file: example.pdf"),
    types.Part(
        inline_data=types.Blob(
            mime_type="application/pdf",
            data=b"%PDF-1.4\n% minimal reproduction\n",
        )
    ),
]

provider = lite_llm._get_provider_from_model(model)
content = asyncio.run(
    lite_llm._get_content(parts, provider=provider, model=model)
)

print("provider:", provider)
print("file keys:", content[-1]["file"].keys())
```

Observed output:

```text
provider: litellm_proxy
file keys: dict_keys(['file_data'])
```

The resulting file block is equivalent to:

```json
{
  "type": "file",
  "file": {
    "file_data": "data:application/pdf;base64,..."
  }
}
```

It has no Azure/OpenAI `file_id` and no filename.

**Expected Behavior:**

PDF attachments should remain usable when an Azure-backed model is accessed
through LiteLLM Proxy.

ADK should either:

- preserve enough underlying-provider information to select the appropriate
  Azure/OpenAI file representation, or
- provide another supported way to communicate the proxied provider's file
  requirements.

At minimum, it should not produce a request that is reported as supported but
is deterministically rejected by the downstream provider.

**Observed Behavior:**

ADK treats `litellm_proxy` as the final provider and takes its generic inline
file-data path. The proxy subsequently forwards an incomplete `input_file`
representation to Azure, which rejects the request with HTTP 400 before any
model inference occurs.

**Environment Details:**

- ADK Library Version: `google-adk 2.6.0`
- Desktop OS: macOS
- Python Version: `3.11`
- LiteLLM Version: `1.94.1`
- google-genai Version: `2.16.0`

**Model Information:**

- Are you using LiteLLM: Yes
- Which model is being used: an Azure OpenAI Responses API deployment accessed
  through `litellm_proxy/azure/<deployment>`

---

## 🟡 Optional Information

**Regression:**

Not known to have worked in an earlier ADK release. Current provider detection
takes only the first segment of a LiteLLM model identifier, so this appears to
be a missing proxy-routing case rather than a regression.

**Logs:**

```text
BadRequestError: litellm.BadRequestError: Error code: 400
AzureException BadRequestError:
Missing required parameter: 'input[0].content[3]'.
type: invalid_request_error
code: missing_required_parameter
```

No model call is completed and no tokens are consumed.

**Additional Context:**

The direct-provider cases have related prior reports:

- #2862: attachment requests missing `file.file_id`
- #4174: PDF upload lost MIME metadata for direct OpenAI/Azure routes
- #5664: Azure `assistant-` file IDs bypassed the Responses API content block
- #3131: file URIs were dropped when using a LiteLLM proxy

Those reports cover adjacent file-conversion defects, but they do not appear to
cover provider-specific file handling being lost because the model identifier
starts with `litellm_proxy/`.

The relevant provider detection currently behaves approximately as follows:

```python
provider, _ = model.split("/", 1)
```

For `litellm_proxy/azure/<deployment>`, this necessarily returns
`litellm_proxy`. The PDF converter's upload path, however, is selected only
when the provider is exactly `openai` or `azure`.

- Always (100%) for this route and PDF input

## Relationship to #6539

This issue is related to #6539 because both affect PDF input through `LiteLlm`, but they are **not duplicates**:

- **This issue (#6538)** occurs for `litellm_proxy/azure/<deployment>`. ADK identifies the provider as `litellm_proxy`, never enters the Azure/OpenAI upload branch, and emits inline `file_data` that becomes an invalid Azure `input_file`.
- **#6539** occurs for the direct `azure/<deployment>` route. ADK correctly identifies Azure and does attempt a Files API upload, but supplies anonymous raw bytes without a filename or MIME metadata.

Fixing #6539 alone would not make the proxy route enter Azure-specific file handling. Conversely, fixing this issue so the nested route selects the Azure upload path may then expose #6539 unless the upload-metadata defect is also resolved.

vietnamesekid:

- Closes: #6538

**Problem:**

`_get_provider_from_model` splits the model string on the first `/` and treats
that segment as the provider. That assumption breaks for nested LiteLLM Proxy
identifiers, because `litellm_proxy` names the transport, not the model family:

```python
_get_provider_from_model("litellm_proxy/azure/my-deployment")  # -> "litellm_proxy"
```

`litellm_proxy` is not in `_FILE_ID_REQUIRED_PROVIDERS` (`{"openai", "azure"}`),
so the Azure/OpenAI file-upload path in `_get_content` never runs. The PDF goes
out as an inline `file_data` block instead of an uploaded `file_id`. When the
proxy translates that for the Azure Responses API, Azure rejects the content
item before inference:

```
BadRequestError: AzureException BadRequestError
Missing required parameter: 'input[0].content[3]'
```

While tracing this I found the same first-segment assumption in four more
helpers, so the blast radius is wider than the PDF case in the issue. Every
provider-specific behavior silently degrades once a model is reached through
the proxy:

| Helper | `litellm_proxy/...` input | Before | Expected |
|---|---|---|---|
| `_get_provider_from_model` | `litellm_proxy/azure/gpt-4` | `litellm_proxy` | `azure` |
| `_is_anthropic_model` | `litellm_proxy/anthropic/claude-4` | `False` | `True` |
| `_is_litellm_vertex_model` | `litellm_proxy/vertex_ai/gemini-2.5-flash` | `False` | `True` |
| `_is_litellm_gemini_model` | `litellm_proxy/vertex_ai/gemini-2.5-flash` | `False` | `True` |
| `_extract_gemini_model_from_litellm` | `litellm_proxy/vertex_ai/gemini-2.5-pro` | `vertex_ai/gemini-2.5-pro` | `gemini-2.5-pro` |

Practical effect beyond the PDF bug: proxied Anthropic models lose thinking
block formatting, and proxied Gemini models are not recognized as Vertex or
Gemini routes.

Worth noting that `litellm.get_llm_provider()` also returns `litellm_proxy`
for these strings. That is correct for LiteLLM, which only needs to know where
to send the request. ADK uses the value for something different, namely how to
shape the payload, and that has to follow the provider that actually serves
the model. So the fix belongs here rather than upstream.

**Solution:**

Add `_strip_proxy_prefix()` and call it before provider and model family
detection. A proxied model is then shaped exactly like its direct equivalent.

Three things I deliberately kept intact:

1. Model strings without the prefix take the same code path as before. The
   helper returns the input unchanged when there is nothing to strip.
2. A bare `litellm_proxy/<deployment>` has no nested provider, so the
   remainder falls through to the existing model name heuristics.
   `litellm_proxy/azure-gpt-4` resolves to `azure`; an opaque
   `litellm_proxy/my-deployment` resolves to `""`.
3. The prefix match is case insensitive, so `LiteLLM_Proxy/azure/gpt-4` works.

One behavior change worth calling out for review: an opaque
`litellm_proxy/my-deployment` now returns `""` rather than `"litellm_proxy"`.
I grepped for consumers of that literal and there are none. `""` is already
the established "provider not determinable" value in this function, so
unknown deployments now take the generic path, which is the honest answer when
the backing provider cannot be known from the string alone.

### Testing Plan

**Unit Tests:**

Extended the existing `test_get_provider_from_model` table with the nested
proxy forms, the case insensitive variant, and both bare deployment cases.

Added `test_model_family_detection_through_litellm_proxy`, which pins all four
model family helpers across proxied and direct strings so the two stay in
lockstep.

Added `test_get_content_pdf_proxied_azure_uses_file_id` as the regression test
for the reported symptom. It drives `_get_content` from the model string and
asserts the upload actually happens with `custom_llm_provider="azure"`, which
is the part that was silently skipped.

```
376 passed in 2.77s

957 passed, 34 warnings in 12.12s
```

I checked that the new tests actually fail without the fix rather than passing
by construction. Reverting just the `_strip_proxy_prefix` call in
`_get_provider_from_model` and rerunning:

```
8 failed, 19 passed, 349 deselected
FAILED test_get_provider_from_model[litellm_proxy/azure/my-deployment-azure]
FAILED test_get_provider_from_model[litellm_proxy/openai/gpt-4o-openai]
FAILED test_get_provider_from_model[litellm_proxy/anthropic/claude-3-anthropic]
FAILED test_get_provider_from_model[litellm_proxy/vertex_ai/gemini-pro-vertex_ai]
FAILED test_get_provider_from_model[LiteLLM_Proxy/azure/gpt-4-azure]
FAILED test_get_provider_from_model[litellm_proxy/azure-gpt-4-azure]
FAILED test_get_provider_from_model[litellm_proxy/my-deployment-]
FAILED test_get_content_pdf_proxied_azure_uses_file_id
```

Formatted with `pyink` 25.12.0 and `isort` 8.0.1, matching the pinned versions
in `.pre-commit-config.yaml`.

Verified the conversion decision without network access, mocking the upload so
the emitted content block is visible. This is the assertion that matters,
since it is the payload Azure rejects:

```python
import asyncio
from unittest import mock
from google.adk.models import lite_llm
from google.adk.models.lite_llm import (
    _content_to_message_param, _get_provider_from_model, _ensure_litellm_imported)

_ensure_litellm_imported()

pdf = types.Content(role="user", parts=[
    types.Part(text="summarize"),
    types.Part(inline_data=types.Blob(mime_type="application/pdf", data=b"%PDF-1.4 fake")),
])

async def show(model):
    provider = _get_provider_from_model(model)
    with mock.patch.object(lite_llm.litellm, "acreate_file",
                           new=mock.AsyncMock(return_value=mock.Mock(id="file-abc"))):
        msg = await _content_to_message_param(pdf, provider=provider, model=model)
    print(f"{model:36} provider={provider!r:12} -> {msg['content'][1]}")

for m in ["azure/gpt-4", "litellm_proxy/azure/my-deployment", "litellm_proxy/my-deployment"]:
    asyncio.run(show(m))
```

Before:

```
azure/gpt-4                          provider='azure'      -> {'type': 'file', 'file': {'file_id': 'file-abc', 'format': 'application/pdf'}}
litellm_proxy/azure/my-deployment    provider='litellm_proxy' -> {'type': 'file', 'file': {'file_data': 'data:application/pdf;base64,JVBERi0xLjQgZmFrZQ=='}}
litellm_proxy/my-deployment          provider='litellm_proxy' -> {'type': 'file', 'file': {'file_data': 'data:application/pdf;base64,JVBERi0xLjQgZmFrZQ=='}}
```

After:

```
azure/gpt-4                          provider='azure'      -> {'type': 'file', 'file': {'file_id': 'file-abc', 'format': 'application/pdf'}}
litellm_proxy/azure/my-deployment    provider='azure'      -> {'type': 'file', 'file': {'file_id': 'file-abc', 'format': 'application/pdf'}}
litellm_proxy/my-deployment          provider=''           -> {'type': 'file', 'file': {'file_data': 'data:application/pdf;base64,JVBERi0xLjQgZmFrZQ=='}}
```

The proxied Azure model now produces a payload byte for byte identical to the
direct Azure model. The opaque deployment still uses `file_data`, which is the
correct fallback when the backing provider is unknowable from the string.

I did not run this against a live Azure backed proxy, since I do not have a
deployment to test with. The reporter in #6538 is set up for that and could
confirm on a real endpoint.

### Checklist

### Additional context

Scope note: I limited this to reading the nested provider out of the model
string. Routing, credentials, and how LiteLLM itself resolves the proxy are
untouched.

If maintainers would rather keep the blast radius to the reported PDF bug, the
change to `_get_provider_from_model` alone fixes #6538 and the four model
family helpers can be split into a follow up. I kept them together because
they share one root cause, and fixing only the provider lookup leaves the same
bug reachable through the Anthropic and Gemini paths.

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -242,6 +242,10 @@ _SUPPORTED_FILE_CONTENT_MIME_TYPES = frozenset({
 # Providers that require file_id instead of inline file_data
 _FILE_ID_REQUIRED_PROVIDERS = frozenset({"openai", "azure"})
 
+# Routing-only prefix: requests go through a LiteLLM Proxy deployment, but the
+# payload must still be shaped for the provider named in the next segment.
+_PROXY_PROVIDER = "litellm_proxy"
+
 _MISSING_TOOL_RESULT_MESSAGE = (
     "Error: Missing tool result (tool execution may have been interrupted "
     "before a response was recorded)."
@@ -309,6 +313,34 @@ def _map_finish_reason(
   return _FINISH_REASON_MAPPING.get(finish_reason_str, types.FinishReason.OTHER)
 
 
+def _strip_proxy_prefix(model: str) -> str:
+  """Removes a leading ``litellm_proxy/`` routing prefix from a model string.
+
+  ``litellm_proxy`` selects the transport (a LiteLLM Proxy deployment), not the
+  model family, so the segment after it identifies the provider that actually
+  serves the request (e.g. ``litellm_proxy/azure/my-deployment`` is served by
+  Azure). Provider-specific request shaping must follow that underlying
+  provider, otherwise proxied requests get generic payloads the backend
+  rejects.
+
+  A bare ``litellm_proxy/<deployment>`` has no nested provider, so the
+  remainder is returned as-is and provider detection falls back to the
+  model-name heuristics.
+
+  Args:
+    model: The model string (e.g., "litellm_proxy/azure/gpt-4").
+
+  Returns:
+    The model string without the ``litellm_proxy/`` prefix.
+  """
+  if not model:
+    return model
+  prefix = _PROXY_PROVIDER + "/"
+  if model.lower().startswith(prefix):
+    return model[len(prefix) :]
+  return model
+
+
 def _get_provider_from_model(model: str) -> str:
   """Extracts the provider name from a LiteLLM model string.
 
@@ -320,6 +352,9 @@ def _get_provider_from_model(model: str) -> str:
   """
   if not model:
     return ""
+  # `litellm_proxy` is a transport prefix; the provider that actually serves
+  # the request is the next segment.
+  model = _strip_proxy_prefix(model)
   # LiteLLM uses "provider/model" format
   if "/" in model:
     provider, _ = model.split("/", 1)
@@ -2685,7 +2720,7 @@ def _is_anthropic_model(model_string: str) -> bool:
   Returns:
     True if it's an Anthropic Claude model, False otherwise.
   """
-  lower = model_string.lower()
+  lower = _strip_proxy_prefix(model_string.lower())
   if lower.startswith("anthropic/"):
     return True
   if lower.startswith("bedrock/"):
@@ -2706,7 +2741,7 @@ def _is_litellm_vertex_model(model_string: str) -> bool:
   Returns:
     True if it's a Vertex AI model accessed via LiteLLM, False otherwise
   """
-  return model_string.startswith("vertex_ai/")
+  return _strip_proxy_prefix(model_string).startswith("vertex_ai/")
 
 
 def _is_litellm_gemini_model(model_string: str) -> bool:
@@ -2719,7 +2754,9 @@ def _is_litellm_gemini_model(model_string: str) -> bool:
   Returns:
     True if it's a Gemini model accessed via LiteLLM, False otherwise
   """
-  return model_string.startswith(("gemini/gemini-", "vertex_ai/gemini-"))
+  return _strip_proxy_prefix(model_string).startswith(
+      ("gemini/gemini-", "vertex_ai/gemini-")
+  )
 
 
 def _extract_gemini_model_from_litellm(litellm_model: str) -> str:
@@ -2731,6 +2768,9 @@ def _extract_gemini_model_from_litellm(litellm_model: str) -> str:
   Returns:
     Pure Gemini model name like "gemini-2.5-pro"
   """
+  # Remove the proxy routing prefix first so the provider prefix below is the
+  # one that actually names the model family.
+  litellm_model = _strip_proxy_prefix(litellm_model)
   # Remove LiteLLM provider prefix
   if "/" in litellm_model:
     return litellm_model.split("/", 1)[1]
--- a/tests/unittests/models/test_litellm.py
+++ b/tests/unittests/models/test_litellm.py
@@ -33,6 +33,7 @@ from google.adk.models.lite_llm import _BraceDepthTracker
 from google.adk.models.lite_llm import _content_to_message_param
 from google.adk.models.lite_llm import _convert_reasoning_value_to_parts
 from google.adk.models.lite_llm import _enforce_strict_openai_schema
+from google.adk.models.lite_llm import _extract_gemini_model_from_litellm
 from google.adk.models.lite_llm import _extract_json_from_deepseek_args
 from google.adk.models.lite_llm import _extract_reasoning_value
 from google.adk.models.lite_llm import _extract_thought_signature_from_tool_call
@@ -45,6 +46,8 @@ from google.adk.models.lite_llm import _get_provider_from_model
 from google.adk.models.lite_llm import _is_anthropic_model
 from google.adk.models.lite_llm import _is_anthropic_provider
 from google.adk.models.lite_llm import _is_anthropic_route
+from google.adk.models.lite_llm import _is_litellm_gemini_model
+from google.adk.models.lite_llm import _is_litellm_vertex_model
 from google.adk.models.lite_llm import _looks_like_openai_file_id
 from google.adk.models.lite_llm import _message_to_generate_content_response
 from google.adk.models.lite_llm import _MISSING_TOOL_RESULT_MESSAGE
@@ -5172,6 +5175,17 @@ async def test_finish_reason_unknown_maps_to_other(
         ("groq/llama3-70b", "groq"),
         ("anthropic/claude-3", "anthropic"),
         ("vertex_ai/gemini-pro", "vertex_ai"),
+        # litellm_proxy is a routing prefix: the provider that actually serves
+        # the request is the segment after it.
+        ("litellm_proxy/azure/my-deployment", "azure"),
+        ("litellm_proxy/openai/gpt-4o", "openai"),
+        ("litellm_proxy/anthropic/claude-3", "anthropic"),
+        ("litellm_proxy/vertex_ai/gemini-pro", "vertex_ai"),
+        ("LiteLLM_Proxy/azure/gpt-4", "azure"),
+        # A bare proxy deployment name has no nested provider, so detection
+        # falls back to the model-name heuristics.
+        ("litellm_proxy/azure-gpt-4", "azure"),
+        ("litellm_proxy/my-deployment", ""),
         # Fallback heuristics
         ("gpt-4o", "openai"),
         ("o1-preview", "openai"),
@@ -5185,6 +5199,55 @@ async def test_finish_reason_unknown_maps_to_other(
 def test_get_provider_from_model(model_string, expected_provider):
   """Test provider extraction from model strings."""
   assert _get_provider_from_model(model_string) == expected_provider
+
+
+@pytest.mark.parametrize(
+    "model_string, is_anthropic, is_gemini, is_vertex, gemini_name",
+    [
+        # Proxied models keep the behavior of the provider that serves them.
+        (
+            "litellm_proxy/anthropic/claude-4-sonnet",
+            True,
+            False,
+            False,
+            "claude-4-sonnet",
+        ),
+        (
+            "litellm_proxy/vertex_ai/gemini-2.5-flash",
+            False,
+            True,
+            True,
+            "gemini-2.5-flash",
+        ),
+        (
+            "litellm_proxy/bedrock/anthropic.claude-3-5-sonnet",
+            True,
+            False,
+            False,
+            "anthropic.claude-3-5-sonnet",
+        ),
+        (
+            "litellm_proxy/azure/my-deployment",
+            False,
+            False,
+            False,
+            "my-deployment",
+        ),
+        # Direct (non-proxied) strings are unaffected.
+        ("anthropic/claude-4-sonnet", True, False, False, "claude-4-sonnet"),
+        ("vertex_ai/gemini-2.5-flash", False, True, True, "gemini-2.5-flash"),
+        ("gemini/gemini-2.5-pro", False, True, False, "gemini-2.5-pro"),
+        ("azure/gpt-4", False, False, False, "gpt-4"),
+    ],
+)
+def test_model_family_detection_through_litellm_proxy(
+    model_string, is_anthropic, is_gemini, is_vertex, gemini_name
+):
+  """Model-family detection must see through the litellm_proxy prefix."""
+  assert _is_anthropic_model(model_string) is is_anthropic
+  assert _is_litellm_gemini_model(model_string) is is_gemini
+  assert _is_litellm_vertex_model(model_string) is is_vertex
+  assert _extract_gemini_model_from_litellm(model_string) == gemini_name
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


vietnamesekid:
@llalitkumarrr @ross-p I verified this branch against the reproduction in #6538. It fixes what the issue diagnoses, but while testing I found a second layer that this PR does not cover. Details below, and I would like your call on scope before I push anything further.

Everything below is reproducible. The layer 1 output comes from the commit currently on this PR (`ebc87873`). The layer 2 output comes from a follow-up commit I deliberately did not push to this PR, so you can read it without it landing here: vietnamesekid@3b96796

## Layer 1: payload shape (what this PR fixes)

Running the repro from the issue against a local HTTP endpoint standing in for the upload target, so these are real requests rather than assertions about mocks.

On `main`:

```text
provider: litellm_proxy
file keys: dict_keys(['file_data'])
```

On this PR, with `AZURE_API_BASE` / `AZURE_API_KEY` / `AZURE_API_VERSION` set:

```text
provider: azure
    [SERVER HIT] POST /openai/files?api-version=2024-07-01-preview  (293 bytes)
result: {'file_id': 'file-fromserver', 'format': 'application/pdf'}
```

The upload happens and a real `file_id` comes back. `file keys` goes from `['file_data']` to `['file_id', 'format']`, matching the reported symptom, and the payload is now identical to what a direct `azure/gpt-4` model produces.

One note on the script as written in the issue: it no longer prints, it raises.

```text
openai.OpenAIError: Missing credentials. Please pass one of `api_key`,
`azure_ad_token`, `azure_ad_token_provider`, or the `AZURE_OPENAI_API_KEY`
or `AZURE_OPENAI_AD_TOKEN` environment variables.
```

That is the fix working. The traceback goes through `litellm.acreate_file` into `litellm/llms/azure/files/handler.py`, which is the Azure upload path that was being skipped. It only fails because no Azure credentials are set. On `main` the script never reaches that call at all.

Which leads to the part I want to raise.

## Layer 2: the upload does not go through the proxy

`litellm.acreate_file` resolves its endpoint independently of the completion call. ADK passes only `custom_llm_provider`, so the upload falls back to the underlying provider's environment variables. In `litellm/files/main.py`, the `custom_llm_provider == "azure"` branch calls `get_azure_credentials(api_base=None, api_key=None, ...)`, which resolves from `AZURE_API_BASE` / `AZURE_API_KEY`.

So after this PR the request splits in two:

- completion goes to the proxy
- file upload goes straight to Azure

That is invisible if you happen to have `AZURE_*` set, which is why layer 1 above looks clean. It is fatal if you do not, which is the normal reason to front a provider with a proxy in the first place.

Same PR commit, same script, only difference is `AZURE_*` unset and proxy credentials configured instead:

```text
provider: azure
ERROR: OpenAIError | Missing credentials. Please pass one of `api_key`, ...
server hits: []
```

Nothing is sent. So for a proxy-only setup this PR turns "Azure rejects the request" into "the client cannot upload at all". Better in that it fails fast and locally, but still broken.

## Proposed fix for layer 2

Forward `api_base`, `api_key`, and `api_version` from the completion arguments to `acreate_file` when the model is proxied. Direct models forward nothing and keep their current environment-variable resolution, since their `api_base` already points at the provider.

I have this working in vietnamesekid@3b96796 (2 files, +182/-9). Same proxy-only scenario, `AZURE_*` unset:

```text
upload_params: {'api_base': 'http://127.0.0.1:8799', 'api_key': 'proxy-key'}
    [SERVER HIT] POST /openai/files?api-version=2025-02-01-preview  (293 bytes)
content: {'type': 'file', 'file': {'file_id': 'file-fromserver', 'format': 'application/pdf'}}
```

The `api-version` differs from the layer 1 output above because nothing sets it in this scenario, so LiteLLM falls back to its own `AZURE_DEFAULT_API_VERSION` (`2025-02-01-preview`). If the caller passes `api_version` it is forwarded like the other two. Flagging it so the difference does not look like drift if you rerun this.

For the record, I first tried passing `custom_llm_provider="litellm_proxy"` and letting LiteLLM route the upload itself. That does not work: the files API has no `litellm_proxy` branch and raises `LiteLLM doesn't support litellm_proxy for 'create_file'. Only ['openai', 'azure', 'vertex_ai', 'manus', 'anthropic'] are supported.` Keeping the real provider and overriding the endpoint is the only path I found.

### Known limit of that fix

It only covers proxy settings passed to the constructor:

```python
LiteLlm(model="litellm_proxy/azure/my-deployment", api_base=..., api_key=...)
```

If the proxy is configured through `LITELLM_PROXY_API_BASE` / `LITELLM_PROXY_API_KEY` instead, which `litellm/llms/litellm_proxy/chat/transformation.py` reads for the completion call, there is nothing in the completion arguments to forward and the upload still fails:

```text
upload_params: {}
ERROR: OpenAIError | Missing credentials...
server hits: []
```

Covering that means reading those env vars in ADK as a fallback. I have not done it, since it is a judgment call about how much LiteLLM configuration ADK should reimplement, and I would rather ask than guess.

## What I would like decided

Three options, and I am happy with any of them:

1. I push vietnamesekid@3b96796 into this PR. Larger diff, threads an optional `upload_params` through `_get_completion_inputs` and `_content_to_message_param` down to `_get_content`, but #6538 is genuinely fixed for constructor-configured proxy setups.
2. This PR merges as is for the payload shape, and I open a separate issue plus PR for the upload routing.
3. Something else you prefer.

I lean towards 1, since a proxy-only setup is the common case and layer 1 alone does not make the reported scenario work end to end.

## Also worth flagging

The same first-segment assumption affects four other helpers, already included in this PR:

| Helper | Input | `main` | This branch |
| --- | --- | --- | --- |
| `_is_anthropic_model` | `litellm_proxy/anthropic/claude-4` | `False` | `True` |
| `_is_litellm_vertex_model` | `litellm_proxy/vertex_ai/gemini-2.5-flash` | `False` | `True` |
| `_is_litellm_gemini_model` | `litellm_proxy/vertex_ai/gemini-2.5-flash` | `False` | `True` |
| `_extract_gemini_model_from_litellm` | `litellm_proxy/vertex_ai/gemini-2.5-pro` | `vertex_ai/gemini-2.5-pro` | `gemini-2.5-pro` |

In practice proxied Anthropic models lose thinking block formatting and proxied Gemini models are not recognized as Vertex or Gemini routes. Same root cause, so I bundled them, but they can also be split out if you would rather keep this PR narrow.

## What I could not verify

I do not have an Azure backed proxy deployment, so none of this is confirmed against a live endpoint. Everything above is the request ADK builds and where it sends it, tested locally against a stub server.

@ross-p, if you can run this branch against your setup that would close the gap. One specific question that decides whether layer 2 affects you today: do you have `AZURE_API_BASE` and `AZURE_API_KEY` set alongside your proxy config, or only the proxy credentials?

Test status: on the PR commit, 376 passed in `tests/unittests/models/test_litellm.py` and 957 across `tests/unittests/models/`. With vietnamesekid@3b96796 applied, 385 and 966. In both cases the new tests fail when the fix is reverted, so they pin real behavior rather than passing by construction.

PR #6580, opened 2026-08-04
feat(utils): add Jinja2-based instruction templating with use_jinja2 flag

### LINKED ISSUE #2942: [Feat]: Introduce Jinja2-based Instruction Templating with a Feature Flag

guptaishaan:
The existing regex-based approach in `inject_session_state` does not support
control-flow constructs such as conditionals and loops. This change introduces
optional Jinja2 rendering as a first-class alternative.

**What changed**

* The original substitution logic is extracted into a private helper
  `_render_with_regex` — behaviour is identical, no breaking changes.
* A new private helper `_render_with_jinja2` is added. It creates an
  async-capable `jinja2.Environment`, exposes all session state keys as
  template variables, and injects an async `artifact(filename)` callable so
  templates can load artifacts inline (e.g. `{{ artifact('doc.txt') }}`).
* `inject_session_state` gains an optional `use_jinja2: bool = False` keyword
  argument. When `True` it delegates to `_render_with_jinja2`; when `False`
  (the default) it calls `_render_with_regex`, preserving full backward
  compatibility.
* Unit tests are added for basic variable substitution, conditional blocks,
  for-loops, artifact loading, and the missing-artifact error path, all via
  `use_jinja2=True`.

Jinja2 is already a declared dependency of the package so no new requirements
are introduced.

Fixes #2942

### DIFF
--- a/src/google/adk/utils/instructions_utils.py
+++ b/src/google/adk/utils/instructions_utils.py
@@ -42,6 +42,7 @@ InstructionProvider: TypeAlias = Callable[
 async def inject_session_state(
     template: str,
     readonly_context: ReadonlyContext,
+    use_jinja2: bool = False,
 ) -> str:
   """Populates values in the instruction template, e.g. state, artifact, etc.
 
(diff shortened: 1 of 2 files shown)


DeanChensj:
Close this one in favor of the new one https://github.com/google/adk-python/pull/6593

PR #6587, opened 2026-08-04
fix(flows): Drop orphaned FRs instead of poisoning sessions

### LINKED ISSUE #6582: contents request-processor (_rearrange_events_for_latest_function_response) raises ValueError on an orphaned function response, permanently poisoning the session — raise happens before before_model_callback, so no user hook can intercept it
**Environment**

- Python 3.13
- `google-adk==2.0.0`

**Bug description**

If a session ever persists a function **response** event whose matching function
**call** event is absent from the branch-filtered history ADK replays, then ADK's contents
request-processor raises `ValueError` while *building* `llm_request.contents` — and the
session is permanently poisoned: every subsequent turn replays the same history and
re-raises, so the agent returns nothing to the user with no recoverable path.

The raise happens inside `BaseLlmFlow._preprocess_async`, which runs the request
processors (including the contents processor) **before** `before_model_callback` is
invoked (the callback fires later, inside `_call_llm_async`). A `before_model_callback`
is therefore structurally unable to heal a poisoned history: the `ValueError` is thrown
while `llm_request.contents` is still being assembled, before any user hook runs.

How a session gets an orphaned function response in the first place is a separate,
upstream producer defect — see the companion report at
`docs/upstream-reports/03-progressive-sse-task-dispatch.md`: under progressive SSE
streaming the chat wrapper can synthesize and persist a task-delegation function
**response** while never persisting the matching function **call**, which is the dominant
observed producer of exactly this poisoned state. This report covers the *consumption*
side: the contents-processor raise that turns any orphaned FR into a dead session.

**Reproduction**

A self-contained Python repro drives ADK's **real** contents request-processor (the
production singleton, `contents.request_processor`) against an `InvocationContext`
carrying one orphaned trailing function response, exactly as `BaseLlmFlow._preprocess_async`
drives it when building `llm_request.contents`.

```python
"""
Minimal repro: ADK 2.0.0 orphaned function-response poisoning.

Run:  cd <repo with google-adk==2.0.0> && python repro_orphaned_fr.py
Expected: prints the ValueError ADK raises while building llm_request.contents.
"""

import asyncio
import traceback

from google.adk.flows.llm_flows import contents as _contents
from google.adk.sessions.session import Session


def _fc(name: str, fc_id: str) -> types.Part:
    return types.Part(function_call=types.FunctionCall(name=name, args={}, id=fc_id))


def _fr(name: str, fr_id: str) -> types.Part:
    return types.Part(
            name=name, response={"result": "ok"}, id=fr_id
        )
    )


def _txt(s: str) -> types.Part:
    return types.Part(text=s)


def _event(author: str, parts: list[types.Part]):

    return Event(
        author=author,
        content=types.Content(role=author, parts=parts),
    )


def _make_invocation_context(events: list) -> InvocationContext:
    agent = LlmAgent(
        name="root_coord",
        model="gemini-2.0-flash",
        instruction="Coord.",
        tools=[],
    )
    session = Session(
        id="repro-session",
        app_name="repro-app",
        user_id="u1",
        events=list(events),
    )
    return InvocationContext(
        invocation_id="inv-1",
        agent=agent,
        session=session,
        end_invocation=False,
    )


async def _run_repro() -> None:
    # A history carrying an orphaned trailing function RESPONSE with no
    # matching function CALL. ADK's _rearrange_events_for_latest_function_response
    # inspects events[-1] and raises on this exact shape.
    events = [
        _event("user", [_txt("hi")]),
        _event("model", [_fc("tool_a", "id-call")]),
        _event("user", [_fr("tool_a", "id-call")]),
        _event("model", [_txt("answer")]),
        _event("user", [_fr("ghost_tool", "orphan-1")]),  # trailing orphan FR
    ]
    ic = _make_invocation_context(events)

    # Drive ADK's REAL contents request-processor exactly as
    # BaseLlmFlow._preprocess_async does when building llm_request.contents.
    print("Driving ADK contents.request_processor against an orphaned-FR history...\n")
    req = LlmRequest()
    try:
        async for _ in _contents.request_processor.run_async(ic, req):
            pass
        print("NO RAISE — the bug may be fixed upstream; re-check on this ADK version.")
    except ValueError as exc:
        print("RAISED ValueError (the orphaned-function-response poison):\n")
        print(f"  type: {type(exc).__module__}.{type(exc).__name__}")
        print(f"  message: {exc}\n")
        print("Traceback (ADK contents processor, _preprocess_async path):")
        print("----")
        print(traceback.format_exc())
        print("----")
        print(
            "BUG CONFIRMED: a persisted function RESPONSE with no matching function\n"
            "CALL makes _rearrange_events_for_latest_function_response raise while\n"
            "building llm_request.contents. This runs in _preprocess_async (request\n"
            "processors), which BaseLlmFlow executes BEFORE before_model_callback —\n"
            "so no before_model hook can heal it. The session is permanently poisoned:\n"
            "every later turn replays the same history and re-raises."
        )


if __name__ == "__main__":
    asyncio.run(_run_repro())
```

**Observed output** (run against `google-adk==2.0.0`, Python 3.13):

```
Driving ADK contents.request_processor against an orphaned-FR history...

RAISED ValueError (the orphaned-function-response poison):

  type: builtins.ValueError
  message: No function call event found for function responses ids: {'orphan-1'}

Traceback (ADK contents processor, _preprocess_async path):
----
  File "/tmp/repro_orphaned_fr.py", line 95, in _run_repro
    async for _ in _contents.request_processor.run_async(ic, req):
        pass
  File ".../google/adk/flows/llm_flows/contents.py", line 74, in run_async
    llm_request.contents = _get_contents(
  File ".../google/adk/flows/llm_flows/contents.py", line 637, in _get_contents
    result_events = _rearrange_events_for_latest_function_response(
        filtered_events
    )
  File ".../google/adk/flows/llm_flows/contents.py", line 224, in _rearrange_events_for_latest_function_response
    raise ValueError(
        ...
    )
ValueError: No function call event found for function responses ids: {'orphan-1'}

----
BUG CONFIRMED: a persisted function RESPONSE with no matching function
CALL makes _rearrange_events_for_latest_function_response raise while
building llm_request.contents. This runs in _preprocess_async (request
processors), which BaseLlmFlow executes BEFORE before_model_callback —
so no before_model hook can heal it. The session is permanently poisoned:
every later turn replays the same history and re-raises.
```

**Root cause**

`google/adk/flows/llm_flows/contents.py`, function
`_rearrange_events_for_latest_function_response` (line 155). The function inspects the
trailing event (`events[-1]`); if it carries function responses whose ids have no matching
function call anywhere in the replayed history, it raises:

```python
  if function_call_event_idx == -1:
    logger.debug(
        'No function call event found for function responses ids: %s in'
        ' event list: %s',
        function_responses_ids,
        events,
    )
    raise ValueError(
        'No function call event found for function responses ids:'
        f' {function_responses_ids}'
    )
```

(raise at line 224, inside `_rearrange_events_for_latest_function_response` at line 155).

This function is called by `_get_contents` (line 637), which the contents
request-processor's `run_async` invokes at line 74 while assigning
`llm_request.contents`. The processor is one of `BaseLlmFlow.request_processors`, and
`BaseLlmFlow._preprocess_async` runs all request processors (line 931) before control
reaches `_call_llm_async`, which is where `before_model_callback` runs (line 1201):

```python
  async def _preprocess_async(self, invocation_context, llm_request):
    ...
    # Runs processors.
    for processor in self.request_processors:           # line 931 — contents processor raises HERE
      async with Aclosing(processor.run_async(invocation_context, llm_request)) as agen:
        async for event in agen:
          yield event
```

```python
  async def _call_llm_async(self, invocation_context, llm_request, model_response_event):
    ...
    if response := await self._handle_before_model_callback(   # line 1201 — before_model runs HERE
        invocation_context, llm_request, model_response_event
    ):
      ...
```

Because `_preprocess_async` (request processors, line 487 / line 842) runs before
`_call_llm_async` (`before_model_callback`, line 1201), the `ValueError` is thrown before
any user-installed `before_model_callback` is reached. A `before_model_callback`-based
workaround for the *inverse* defect (an orphaned function **call** padded with a synthetic
response) cannot help here: it sits on the wrong side of the raise.

**Expected behaviour**

Building `llm_request.contents` should be resilient to a single orphaned function
response. Reasonable options:

- **Drop** the orphaned function-response part(s) from the replayed contents (the model
  simply never sees a result it never asked for), build the request from the surviving
  events, and continue — rather than raising and killing the session.
- **Surface** the orphaned-response condition as a structured, recoverable signal (e.g. a
  non-fatal event or a clearly-typed exception) so callers can heal the in-memory history
  and retry, rather than a bare `ValueError` raised deep in contents construction that
  aborts the whole turn.

Either keeps the session usable; the current behaviour makes any session that ever
acquires an orphaned FR permanently dead.

**Observed behaviour**

The raise aborts `llm_request.contents` construction. The model is never called for the
turn. The exception propagates up through the flow with no indication of which session or
which response id is at fault (the message carries the ids, but the raise is anonymous to
any caller that catches `ValueError` generically). The session's stored events keep the
imbalance forever, so every subsequent turn over the same session re-raises identically.
The user sees an empty turn with no recovery path.

**Related**

- **Producer defect (companion report):** `docs/upstream-reports/03-progressive-sse-task-dispatch.md`
  documents the dominant observed producer of an orphaned function response: under
  progressive SSE streaming, the chat wrapper synthesizes and persists a task-delegation
  function **response** while never persisting the matching function **call**. Fixing
  this consumption-side raise is necessary regardless, because *any* path that lands an
  orphaned FR (a turn that died mid-dispatch, a branch/isolation filter that excluded the
  call, a producer bug) hits this same fatal raise.
- The *inverse* direction — an orphaned function **call** (N calls, < N responses) — is
  rejected by Gemini at the model boundary as
  `400 INVALID_ARGUMENT: Please ensure that the number of function response parts is equal
  to the number of function call parts`, and is covered by the companion report at
  `docs/upstream-reports/01-mixed-turn-fc-drop.md`. Both defects poison sessions through an
  unbalanced FC/FR history; this one fails earlier (during contents build) and louder (a
  raised `ValueError`), while the mixed-turn drop fails at the model call with a `400`.

a2105z:

- Closes: #6582

**Problem:**
If a session ever persisted a function **response** whose matching function **call** was absent from the branch-filtered history, `_rearrange_events_for_latest_function_response` raised `ValueError` while building `llm_request.contents`.

That raise happens in request preprocessing — **before** `before_model_callback` — so no user hook can heal it. Every later turn replays the same history and re-raises, permanently poisoning the session.

**Solution:**
When the trailing event carries function responses with no matching function call in history, **drop** those orphaned response event(s), log a warning, and continue rearranging / assembling contents from the surviving history. The model simply never sees a result it never asked for, and the session stays usable.

### Testing Plan

**Unit Tests:**

**Updated / new tests in** `tests/unittests/flows/llm_flows/test_contents_function.py`:
- `test_orphaned_function_response_is_dropped_not_raised` — former raise expectation replaced with drop behavior
- `test_orphaned_fr_after_valid_history_keeps_session_usable` — issue #6582 history shape (valid FC/FR + answer + trailing orphan)
- `test_multiple_trailing_orphaned_frs_are_all_dropped` — consecutive orphans
- `test_runner_continues_after_orphaned_fr_in_session` — full `Runner.run_async` path; model is called and orphan is absent from prompt

**pytest results (local):**

```text
pytest tests/unittests/flows/llm_flows/test_contents_function.py -v
# 10 passed

pytest tests/unittests/flows/llm_flows/test_contents_function.py \
       tests/unittests/flows/llm_flows/test_contents.py \
       tests/unittests/apps/test_compaction_runner_e2e.py -q
# 45 passed

pytest tests/unittests/flows/llm_flows/ -q
# 485 passed
```

**Manual / repro verification:**
- Reproduced #6582 against current `main` with the issue's contents-processor repro:
  - Before: `ValueError: No function call event found for function responses ids: {'orphan-1'}`
  - After: no raise; contents built from surviving history (`hi` → FC/FR → `answer`), orphan omitted

**Pre-commit:**
```text
pre-commit run --files \
  src/google/adk/flows/llm_flows/contents.py \
  tests/unittests/flows/llm_flows/test_contents_function.py
# all hooks passed
```

### Checklist

### Additional context

Files changed:
- `src/google/adk/flows/llm_flows/contents.py` — drop orphaned trailing FRs instead of raising
- `tests/unittests/flows/llm_flows/test_contents_function.py` — regression + runner coverage

Companion producer-side issues (e.g. progressive SSE leaving orphaned FRs) are separate; this change hardens the consumption path so any orphaned FR no longer kills the session.

### Exact commands run
```bash
pytest tests/unittests/flows/llm_flows/test_contents_function.py -v
# 10 passed

pytest tests/unittests/flows/llm_flows/test_contents_function.py \
       tests/unittests/flows/llm_flows/test_contents.py \
       tests/unittests/apps/test_compaction_runner_e2e.py -q
# 45 passed

pytest tests/unittests/flows/llm_flows/ -q
# 485 passed

pre-commit run --files \
  src/google/adk/flows/llm_flows/contents.py \
  tests/unittests/flows/llm_flows/test_contents_function.py
# all hooks passed
```

### DIFF
--- a/src/google/adk/flows/llm_flows/contents.py
+++ b/src/google/adk/flows/llm_flows/contents.py
@@ -196,11 +196,9 @@ def _drop_orphaned_function_responses(
 
   An orphan can reach this point when the producer of the call is gone, for
   example a session edited by hand or a history stitched together from more
-  than one source. Left in place, the same orphan behaves differently
-  depending on where it sits: mid-history it is quietly discarded, while as
-  the trailing event it aborts the whole request. Pruning it here makes the
-  outcome the same wherever it appears, and keeps unpaired results from being
-  forwarded to providers that reject them.
+  than one source. Left in place, unpaired results can be forwarded to
+  providers that reject them, or (as a trailing event) abort contents
+  assembly. Pruning it here makes the outcome the same wherever it appears.
 
   Responses without an id are left alone: ids are stripped on the way out for
   some model families, so a missing id does not imply a missing call.
@@ -259,6 +257,12 @@ def _rearrange_events_for_latest_function_response(
   between the initial function_call and the latest function_response will be
   removed.
 
+  If the latest event carries function responses with no matching function
+  call in history (an orphaned FR), those responses are dropped and history
+  is rearranged from the remaining events. Raising here would permanently
+  poison the session: contents assembly runs before any user callback and
+  every later turn would replay the same fatal error.
+
   Args:
     events: A list of events.
 
@@ -313,16 +317,16 @@ def _rearrange_events_for_latest_function_response(
           break
 
   if function_call_event_idx == -1:
-    logger.debug(
-        'No function call event found for function responses ids: %s in'
-        ' event list: %s',
+    # Orphaned trailing FR: drop it and continue so contents assembly (and
+    # the session) remain usable. Producer bugs / branch filters can leave
+    # an FR without its FC; failing hard here has no recovery path.
+    logger.warning(
+        'Dropping orphaned function response(s) with ids %s because no'
+        ' matching function call was found in history. Continuing without'
+        ' them so the session remains usable.',
         function_responses_ids,
-        events,
     )
-    raise ValueError(
-        'No function call event found for function responses ids:'
-        f' {function_responses_ids}'
-    )
+    return _rearrange_events_for_latest_function_response(events[:-1])
 
   # collect all function response between last function response event
   # and function call event
--- a/tests/unittests/flows/llm_flows/test_contents_function.py
+++ b/tests/unittests/flows/llm_flows/test_contents_function.py
@@ -553,7 +553,12 @@ async def test_function_rearrangement_preserves_other_content():
 
 @pytest.mark.asyncio
 async def test_function_response_without_matching_call_is_dropped():
-  """An orphaned function response is pruned, not raised on."""
+  """An orphaned function response is pruned, not raised on.
+
+  Regression for github.com/google/adk-python/issues/6582: raising here
+  permanently poisons the session because every later turn replays the
+  same fatal ValueError before any user callback can intervene.
+  """
   agent = Agent(model="gemini-2.5-flash", name="test_agent")
   llm_request = LlmRequest(model="gemini-2.5-flash")
   invocation_context = await testing_utils.create_invocation_context(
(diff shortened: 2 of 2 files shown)


a2105z:
Hi @GWeale — hoping you might have a minute when free. You’ve touched `contents.py` a lot recently, so I wanted to check this with you.

This is a fix for #6582: an orphaned trailing function response used to raise during contents assembly (before any callback), which permanently poisoned the session. The change drops those unmatched FRs, logs a warning, and continues so later turns can still run.

I tried to keep the blast radius small and covered it with contents-processor + full Runner tests. Happy to adjust if you’d rather surface this differently — just say the word. Thanks so much for any feedback.

a2105z:
Hi @DeanChensj — looping you in as well for a review when you have a moment.

This is the consumption-side companion to the mixed-turn session-poisoning work: orphaned trailing function responses were raising during contents assembly and leaving the session unusable. The change drops those unmatched FRs and continues.

Happy to tweak anything that looks off — thanks so much for taking a look.

a2105z:
Hi @llalitkumarrr @kenwilly @GWeale @DeanChensj — conflicts are resolved.

Rebased onto current `main` and cleaned up the test-file merge against the orphan-prune coverage that landed meanwhile:

- Kept the upstream orphan drop tests in `test_contents_function.py`
- Re-added the extra #6582 regressions (valid history preserved, multiple trailing orphans, full Runner path)
- `_rearrange_events_for_latest_function_response` still drops orphaned trailing FRs instead of raising (defense in depth alongside `_drop_orphaned_function_responses`)

Local: `pytest tests/unittests/flows/llm_flows/test_contents_function.py` → **11 passed**. Ready for review whenever you are.

GWeale:
same fix landed in 73e8625 so closing this as already fixed. thanks for taking it on.

main only prunes responses that carry an id, so the drop you added in _rearrange_events_for_latest_function_response is still worth having. send it as a follow up if you want

a2105z:
Thanks @GWeale — follow-up as you suggested:

- Issue: #6751
- PR: #6752

Covers the rearrange-path drop (defense in depth) plus the remaining id-less / empty-id trailing FR poisoning cases that the id-only prune still lets through.

PR #6590, opened 2026-08-04
feat(tools): add transfer_reason argument to transfer_to_agent to prevent multi-agent transfer loops

Srikanth-Saravanan:
# Add `transfer_reason` argument to `transfer_to_agent`

## Problem

The built-in `transfer_to_agent` tool only accepts the target `agent_name`. When an agent hands off control, the full session context is carried over, but the tool call itself contains no explicit rationale for *why* the transfer is happening.

In multi-agent systems this causes agents to "ping-pong". Because the receiving agent has no stated reason for the handoff, it frequently cannot tell why it was given control and transfers straight back to the sender — in some cases looping between agents indefinitely.

## Solution

This change adds a required `transfer_reason: str` argument to `transfer_to_agent`. The delegating agent must now provide a short, explicit reason whenever it hands off. That reason travels with the transfer, giving the receiving agent the context it needs to continue the task instead of bouncing control back — which breaks the loop.

Design notes:

- The argument is **required** rather than optional. The ping-pong originates in the function call the model produces, so the reason needs to be present there every time; making it optional would let the model omit it and reintroduce the problem.
- `TransferToAgentTool` picks up the new parameter automatically through reflection. The existing `agent_name` enum constraint (which restricts transfers to valid agent names) is unchanged.

## Testing

**Unit tests**

- Updated the tool-declaration assertions in `test_transfer_to_agent_tool.py`: the required parameters are now `['agent_name', 'transfer_reason']`, the property count goes from 1 to 2, and a new assertion verifies `transfer_reason` is a string with no enum constraint.
- Because `transfer_reason` is required, `FunctionTool`'s mandatory-argument validation now enforces it. Every test that mocks a `transfer_to_agent` function call was updated to include the new argument.

`pytest` results — **154 passed, 9 skipped, 3 xfailed**:

```
tests/unittests/tools/test_transfer_to_agent_tool.py
tests/unittests/workflow/test_agent_transfer.py
tests/unittests/workflow/test_llm_agent_as_node.py
tests/unittests/agents/test_resumable_llm_agent.py
tests/unittests/runners/test_pause_invocation.py
tests/unittests/runners/test_resume_invocation.py
tests/unittests/streaming/test_multi_agent_streaming.py
tests/unittests/flows/llm_flows/test_agent_transfer_system_instructions.py
tests/unittests/flows/llm_flows/test_base_llm_flow.py
tests/unittests/flows/llm_flows/test_functions_parallel.py
```

**Manual verification**

Confirmed the generated function declaration exposes the new parameter:

```python
from google.adk.tools.transfer_to_agent_tool import TransferToAgentTool

tool = TransferToAgentTool(agent_names=["agent_a", "agent_b"])
schema = tool._get_declaration().parameters_json_schema

print(schema["properties"].keys())  # dict_keys(['agent_name', 'transfer_reason'])
print(schema["required"])           # ['agent_name', 'transfer_reason']
print(schema["properties"]["agent_name"]["enum"])  # ['agent_a', 'agent_b']
```

In a multi-agent run, the model now emits calls such as
`transfer_to_agent(agent_name="billing_agent", transfer_reason="user asked about an invoice charge")`,
and the receiving agent uses the stated reason to continue the task rather than transferring back.

## Backward compatibility

`transfer_reason` is a required argument. LLM-driven callers populate it naturally from the function declaration, but any code that hand-crafts a `transfer_to_agent` function call (for example tests or replay fixtures) must now include it. All such call sites in the repository have been updated in this PR.

### DIFF
--- a/src/google/adk/tools/transfer_to_agent_tool.py
+++ b/src/google/adk/tools/transfer_to_agent_tool.py
@@ -27,7 +27,9 @@ from .tool_context import ToolContext
 #   For most use cases, you should use TransferToAgentTool instead of this
 #   function directly. TransferToAgentTool provides additional enum constraints
 #   that prevent LLMs from hallucinating invalid agent names.
-def transfer_to_agent(agent_name: str, tool_context: ToolContext) -> None:
+def transfer_to_agent(
+    agent_name: str, transfer_reason: str, tool_context: ToolContext
+) -> None:
   """Transfer the query to another agent.
 
   Use this tool to hand off control to another agent that is more suitable to
@@ -35,6 +37,7 @@ def transfer_to_agent(agent_name: str, tool_context: ToolContext) -> None:
 
   Args:
     agent_name: the agent name to transfer to.
+    transfer_reason: the reason for transferring to the target agent.
   """
   tool_context.actions.transfer_to_agent = agent_name
 
--- a/tests/unittests/agents/test_resumable_llm_agent.py
+++ b/tests/unittests/agents/test_resumable_llm_agent.py
@@ -32,7 +32,8 @@ from .. import testing_utils
 
 def transfer_call_part(agent_name: str) -> Part:
   return Part.from_function_call(
-      name="transfer_to_agent", args={"agent_name": agent_name}
+      name="transfer_to_agent",
+      args={"agent_name": agent_name, "transfer_reason": "test reason"},
   )
 
 
--- a/tests/unittests/runners/test_pause_invocation.py
+++ b/tests/unittests/runners/test_pause_invocation.py
@@ -37,7 +37,8 @@ from .. import testing_utils
 
 def _transfer_call_part(agent_name: str) -> Part:
   return Part.from_function_call(
-      name="transfer_to_agent", args={"agent_name": agent_name}
+      name="transfer_to_agent",
+      args={"agent_name": agent_name, "transfer_reason": "test reason"},
   )
 
 
--- a/tests/unittests/runners/test_resume_invocation.py
+++ b/tests/unittests/runners/test_resume_invocation.py
@@ -28,7 +28,8 @@ from .. import testing_utils
 
 def transfer_call_part(agent_name: str) -> Part:
   return Part.from_function_call(
-      name="transfer_to_agent", args={"agent_name": agent_name}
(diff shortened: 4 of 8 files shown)


src/google/adk/tools/transfer_to_agent_tool.py
```diff
@@ -27,14 +27,17 @@
...
+    agent_name: str, transfer_reason: str, tool_context: ToolContext
+) -> None:
   """Transfer the query to another agent.
 
   Use this tool to hand off control to another agent that is more suitable to
   answer the user's query according to the agent's description.
 
   Args:
     agent_name: the agent name to transfer to.
+    transfer_reason: the reason for transferring to the target agent.
```
GiulioCMSanto, line 40: Hey! Right now, this is the only signal to instruct the LLM! But in `flows/llm_flows/agent_transfer.py`, the `_build_transfer_instruction_body` function instructs the model to call the tool and "do not generate any text other than the function call" (see below). Maybe we should add one line requiring a meaningful `transfer_reason`as well? What do you think? 😄 

```python
return f"""
You have a list of other agents to transfer to:

{line_break.join([
    _build_target_agents_info(target_agent) for target_agent in target_agents
])}

If you are the best to answer the question according to your description,
you can answer it.

If another agent is better for answering the question according to its
description, call `{tool_name}` function to transfer the question to that
agent. When transferring, do not generate any text other than the function
call.

**NOTE**: the only available agents for `{tool_name}` function are
{formatted_agent_names}.
"""
```

Otherwise it will be conflicting!

GiulioCMSanto (commented):
Hello! Thank you very much for this PR! I am very interested in this feature as well!! I left a suggestion, and I will give one more here which I think is less important than the other:

In `flows/llm_flows/contents.py` we have the function `_present_other_agent_message` that will generate a message like `"[{event.author}] called tool transfer_to_agent with parameters:..."`

Maybe we could add something explicit for the reason as well, to make it more highlighted. Something like: `"transferred control to agent <agent> because: <transfer_reason>"`

GiulioCMSanto (commented):
One more suggestion here: maybe we could make this an opt-in/opt-out feature at Agent() level as well (as a parameter, for instance)

Issue #4759, opened 2026-03-08
Dispatch Arbitrary Agents for complex non deterministic workflows


lambda72453:
** Please make sure you read the contribution guide and file the issues in the right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. Requests without complete information may be rejected / deprioritized.
If an item is not applicable to you - please mark it as N/A*

### Is your feature request related to a specific problem?
Please describe the problem you are trying to solve. (Ex: "I'm always frustrated
when I have to manually handle X...")

Im frequently running into situations where I want my orchestrator agent to create arbitrary sub agents to delegate tasks, tools and skills to them. I want the orchestrator continue to work while the agents collect info in the background , and handle when agents complete work. Dispatched agents should be able to handle follow up questions as well (persistent sessions).

Creating arbitrary agents is useful for complex non deterministic workflows where you are not sure of the steps needed to complete a task until input is processed. Giving the orchestrator the option to offload this work to  other agents is essentials for effective context management and allowing the orchestrator to operate more efficiently by executing complex tasks in parallel. 

Also, being able to ask these agents follow up questions is useful to continue/iterate on work and answer questions on how it came to certain conclusions. 

Proof of value can be demonstrated in other popular agents that do this out of the box (Claude Code for example). 

Giving this option presents more opportunities to allow agents to effectively handle long horizon analysis tasks. 

### Describe the Solution You'd Like
A clear and concise description of the feature or API change you want.
Be specific about input/outputs if this involves an API change.

A function to dispatch Agents with unique Names , Instructions , User Message , Tools and Skills .  Sub Agents should either return a response to the agent when complete or polling tool for the Orchestrator. 

Sub Agents should be able to handle follow up questions to handle additional tasks / followup questions. 

### Impact on your work
How does this feature impact your work and what are you trying to achieve?
If this is critical for you, tell us if there is a timeline by when you need
this feature.

I need this now so I will be making a custom solution in the meantime. 

### Willingness to contribute
Are you interested in implementing this feature yourself or submitting a PR?
(Yes/No)

Would have to get it approved but maybe 
---

### Describe Alternatives You've Considered
A clear and concise description of any alternative solutions or workarounds
you've considered and why they didn't work for you.

Tried using current subagent architecture, (base agent, parallel , sequential ) but agents 
have to be defined at beginning of run time so unable to make this work. 

[...]

PR #6592, opened 2026-08-05
feat(tools): Add AgentDispatcherToolset for runtime agents

a2105z:

- Related: #4759

**Note:** This issue is assigned for internal review. This is a **community proposal** for design feedback — happy to adjust or close if you’d rather keep the design internal.

**Problem:**
Static `sub_agents` must be known at construct time, and `AgentTool` creates a fresh in-memory session per call. Orchestrators need runtime specialists that can run in the background, in parallel, with persistent follow-ups and completion signaling.

**Solution:**
`AgentDispatcherToolset` exposes:

1. `dispatch_agent` — create a runtime `LlmAgent` on a **persistent** child session. **Background by default** (`wait=false`) so the orchestrator continues; `wait=true` for sync. Supports `tool_names` + `skill_names` from constructor allowlists/registry.
2. `get_agent_result` — poll latest status/result.
3. `await_agent` — wait for background completion (optional timeout).
4. `message_agent` — follow-up on the same child session.

Also:
- Parallel multi-dispatch via concurrent background tasks
- `on_complete` sync/async callback
- Shared `session_service` (parent’s by default) + serializable registry under `session.state["_adk_agent_dispatcher"]` for rebuild/follow-up

### Remaining bounds / non-goals

These are intentional for a reviewable first cut; calling them out so reviewers know what this does *not* claim:

1. **Skills are named + allowlisted, not free-form paths.**  
   Dispatched agents can only attach skills from `skill_allowlist` and/or a `skill_registry` lookup by name. The model cannot load arbitrary filesystem/GCS skill paths. That keeps the surface safer; broader skill discovery can be a follow-up if desired.

2. **Cross-process restart needs a durable session service.**  
   Child sessions and the dispatcher registry are persisted through the configured `BaseSessionService`. With in-memory services, state dies with the process. For true restart survival, use a durable backend (e.g. database / Vertex session service) shared by parent and children, and reconstruct the toolset with the same allowlists/model.

3. **Rebuild still requires the same toolset config.**  
   After a process restart, follow-ups rebuild the child `Runner` from session-state metadata (`instruction`, `mode`, `tool_names`, `skill_names`, etc.). The new process must construct `AgentDispatcherToolset` with compatible `tool_allowlist` / `skill_allowlist` / `skill_registry` / model so those names resolve.

4. **Completion signaling is callback + poll/await, not parent-LLM push mid-turn.**  
   `on_complete` notifies application code; the orchestrator LLM observes completion via `get_agent_result` / `await_agent` on a later turn. This does not inject a spontaneous model turn into the parent when a background child finishes.

5. **Not a replacement for Workflow `ctx.run_node`.**  
   This targets chat-style orchestrators that want tool-driven spawn. Graph/Workflow dynamic nodes remain the right primitive for interrupt/resume-heavy workflows.

### Testing Plan

**Unit Tests:**

- [x] Added/updated unit tests
- [x] All unit tests pass locally

```bash
pytest tests/unittests/tools/test_agent_dispatcher_toolset.py -v
# 8 passed
```

Coverage: wait/background/await, parallel dispatch, on_complete callback, persistent message follow-up, rebuild from session state, skill allowlist, unknown tool rejection.

**Manual E2E:**

```bash
adk run contributing/samples/tools/agent_dispatcher
```

### Checklist

- [x] CONTRIBUTING.md
- [x] Self-review
- [x] Comments in hard-to-understand areas
- [x] Tests proving the feature works
- [x] Unit tests pass locally
- [ ] Manual E2E
- [ ] Dependent changes merged

### Additional context

RFC on issue: https://github.com/google/adk-python/issues/4759#issuecomment-5188332179

cc @klateefa @wuliang229 — concrete API sketch for review, not a claim on assigned work.

### DIFF
--- a/contributing/samples/tools/agent_dispatcher/README.md
+++ b/contributing/samples/tools/agent_dispatcher/README.md
@@ -1,0 +1,27 @@
+# Agent Dispatcher Sample
+
+Demonstrates [`AgentDispatcherToolset`](../../../../src/google/adk/tools/agent_dispatcher/_agent_dispatcher_toolset.py) for [issue #4759](https://github.com/google/adk-python/issues/4759): dispatch a runtime specialist agent, then send follow-ups on the same persistent child session.
+
+## What it shows
+
+- `dispatch_agent` — create a named agent with instruction + user message (optional allowlisted tools)
+- `message_agent` — continue on the same `dispatch_id` / session
+- `get_agent_result` — read the latest status/result
+
+## Run
+
+From the repository root (with ADK installed and `GOOGLE_API_KEY` or Vertex credentials configured):
+
+```bash
+adk run contributing/samples/tools/agent_dispatcher
+```
+
+Try prompts like:
+
+- "Research ADK and then ask the specialist for one more sentence."
+- "Dispatch a worker to look up Gemini, then get the result."
+
+## Notes
+
+- Tools for dispatched agents must be listed in the toolset `tool_allowlist` (MVP security boundary).
+- Sync-first: `dispatch_agent` awaits the child run. Background fan-out is a follow-up.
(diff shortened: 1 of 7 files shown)


a2105z:
Hi @xuanyang15 — gentle follow-up on this one when you have bandwidth. Happy to adjust anything if review feedback comes up. Thank you!

PR #6596, opened 2026-08-05
agents: block all stdlib modules in agent-config code-refs (denylist bypass via cProfile.run/timeit)

asroyxCySec:
### Problem
`config_agent_utils._validate_module_reference()` protects YAML agent configs against arbitrary code execution by blocking a hand-maintained denylist of dangerous stdlib modules (`_BLOCKED_MODULES`) referenced in `tool` / `callback` / `model` / `schema` code-refs. The check only compares the **top-level module name** against the denylist.

A top-level-name denylist is inherently incomplete: equivalent code-execution gadgets that are not in the list slip through. The clearest example is already visible in the current list — `profile` is blocked (its `profile.run("<code>")` executes arbitrary code), but its C sibling **`cProfile`** is not, and `cProfile.run("<code>")` is the identical gadget. Other reachable, non-listed gadgets include `timeit.timeit("<code>")`, `pydoc`, `logging.config.fileConfig`, `bdb`, `trace`, and `venv`.

Example bypass (before this change):
```python
from google.adk.agents import config_agent_utils
# passes validation, then resolves to a callable that runs arbitrary code:
run = config_agent_utils.resolve_fully_qualified_name("cProfile.run")
run("open('/tmp/pwned','w').write('rce')")
```

### Fix
Agent-config code-refs point to user-defined or ADK packages, never to the Python standard library. So instead of chasing individual gadgets, reject **any** standard-library top-level module via `sys.stdlib_module_names` (Python 3.10+, which matches the project's `requires-python`). The existing `_BLOCKED_MODULES` set is kept as documentation / belt-and-suspenders. Legitimate references (user packages, `google.adk.*`) are unaffected because they are not part of `sys.stdlib_module_names`.

### Testing
- New parametrized regression test `test_stdlib_gadget_modules_are_rejected` covers the previously-bypassing gadgets (`cProfile.run`, `timeit.timeit`, `pydoc`, `logging.config.fileConfig`, `bdb`, `trace`, `venv`).
- New `test_non_stdlib_references_are_not_blocked` confirms user/ADK references still validate.
- Existing blocked-module tests (`os`, `subprocess`, network modules, denylist-disable) continue to pass.
- Verified the patched `_validate_module_reference` directly: all listed gadgets + the existing denylist entries are rejected, while `my_company_pkg.my_tool` and `google.adk.tools.google_search` are allowed.

### DIFF
--- a/src/google/adk/agents/config_agent_utils.py
+++ b/src/google/adk/agents/config_agent_utils.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import importlib
 import inspect
 import os
+import sys
 from typing import Any
 from typing import List
 
@@ -204,7 +205,16 @@ def _validate_module_reference(fully_qualified_name: str) -> None:
     return
   # Extract the top-level package from the fully-qualified name.
   top_module = fully_qualified_name.split(".")[0]
-  if top_module in _BLOCKED_MODULES:
+  # Agent-config tool/callback/model/schema references point to user-defined or
+  # ADK packages, never to the Python standard library. A hand-maintained
+  # denylist of dangerous stdlib modules is inherently incomplete -- equivalent
+  # code-execution gadgets slip through (for example ``cProfile.run`` vs. the
+  # already-blocked ``profile.run``, as well as ``timeit.timeit``, ``pydoc``,
+  # ``logging.config``, ``bdb`` and ``trace``). Blocking the entire standard
+  # library by name closes that class of bypass, while legitimate references
+  # (user packages, ``google.adk.*``) are unaffected because they are not part
+  # of ``sys.stdlib_module_names``.
+  if top_module in _BLOCKED_MODULES or top_module in sys.stdlib_module_names:
     raise ValueError(
         f"Blocked module reference: {fully_qualified_name!r}. "
         f"Importing from the '{top_module}' module is not allowed in "
--- a/tests/unittests/agents/test_agent_config.py
+++ b/tests/unittests/agents/test_agent_config.py
@@ -604,6 +604,49 @@ def test_newly_blocked_network_modules_are_rejected(blocked_ref: str):
   assert "Blocked module reference" in str(exc_info.value.__cause__)
 
 
+@pytest.mark.parametrize(
+    "blocked_ref",
+    [
+        # Code-execution gadgets that a top-level-name denylist previously
+        # missed. ``cProfile.run`` mirrors the already-blocked ``profile.run``;
+        # the others reach code execution directly or via os/subprocess.
+        # Blocking the whole standard library closes this bypass class.
+        "cProfile.run",
+        "timeit.timeit",
+        "pydoc.render_doc",
+        "logging.config.fileConfig",
+        "bdb.Bdb",
+        "trace.Trace",
+        "venv.create",
+    ],
+)
+def test_stdlib_gadget_modules_are_rejected(blocked_ref: str):
+  """Non-denylisted stdlib modules that can execute code must also be blocked.
+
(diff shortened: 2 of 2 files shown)


tonydzi:
Ran this against current main, because a week with no reply usually means either nobody looked or the ground moved. Here it is the second one.

**Superseded.** `a16f6da3` ("fix: block the whole standard library in agent-config code references", Aug 7) does what this PR proposes, two days after you opened it. On main today every gadget from your description is rejected. I checked all seven rather than reading the diff:

`cProfile.run`, `timeit.timeit`, `pydoc.pipepager`, `logging.config.fileConfig`, `bdb.Bdb`, `trace.Trace`, `venv.create` — all blocked. That is also why this PR now shows `CONFLICTING`: main rewrote the same line. Whether your report prompted that work or it was found independently I can't tell from the history.

**Before anyone rebases, the landed version is deliberately a different mechanism, not just an earlier one.** Main is

```python
_STDLIB_MODULES = frozenset(sys.stdlib_module_names) | frozenset(sys.builtin_module_names)
```

plus a block inside `_BLOCKED_MODULES` marked `LOAD-BEARING`: names that are *not* in `sys.stdlib_module_names` yet still import. `distutils` (returns via setuptools' shim, and its `spawn()` runs a subprocess), `asynchat`, `imp`, `smtpd`, CPython's own `test` package (`test.support.script_helper` starts processes) and `_testcapi`.

I ran your branch and main side by side on CPython 3.12.13 against one list. Your version allows six names main rejects:

```
                                   PR #6596     main
distutils.spawn.spawn              ALLOWED      blocked
imp.load_source                    ALLOWED      blocked
asynchat.async_chat                ALLOWED      blocked
smtpd.SMTPServer                   ALLOWED      blocked
test.support.script_helper...      ALLOWED      blocked
_testcapi.*                        ALLOWED      blocked
```

The reason is the same one your PR is about, one level up: `sys.stdlib_module_names` is a snapshot of one interpreter's *current* stdlib, and a module removed from the stdlib does not stop being importable. Main's tests say this out loud at `test_agent_config.py:626`.

Honest limit on that table: it is a difference in what the validator rejects, not a bypass I demonstrated. On the interpreter I measured (uv's python-build-standalone) none of those six actually import — no setuptools shim, no `test` package shipped. On a system Python or a full CPython build they generally do.

So the useful outcome is probably close-as-superseded, or rebase and keep only the part main lacks. Which brings up the third thing, since this PR is the one framing the problem as a class rather than a gadget:

**Both versions draw the line at "stdlib by name", so every third-party package stays allowed.** `yaml.unsafe_load` passes validation, and PyYAML is a hard dependency (`pyproject.toml:52`), so it imports in every ADK install. `pip.main` passes validation and imports in a plain `python -m venv` (measured, same interpreter).

I am not calling that a vulnerability and I did not demonstrate one. `CodeConfig` deliberately cannot pass constructor args ("YAML cannot pass constructor args", `common_configs.py:33`), so the `gadget.run("<code>")` shape from your description is not reachable through the config path the way it is when you call the resolved object yourself in Python, which is what your repro does. It is a scope question: if the rule is "code-refs point at user or ADK packages, never the standard library", the check enforces the second half and not the first, and the docstring doesn't say whether that asymmetry is intended. One sentence in `_validate_module_reference` naming the threat model would save the next person from re-deriving it out of the denylist — which is how this PR came to exist in the first place.

Disclosure: I am an AI agent (Claude) reviewing as part of Palo Alto AI Research Lab. No code changes proposed here, only measurements.

GWeale:
a16f6da3 landed the stronger version of this in v2.7.0: config_agent_utils blocks the whole standard library rather than a denylist. i ran all seven gadgets from this pr against head on 3.13 and every one is rejected.

PR #6612, opened 2026-08-06
Add to_content_blocks() to BasePlanner for standardized content block output

### LINKED ISSUE #3378: Standarlize content block for PlanReActPlanner and BuiltInPlanner

guptaishaan:
Both `PlanReActPlanner` and `BuiltInPlanner` expose their output as a raw list of `google.genai.types.Part` objects, making it hard for callers to consume planning results in a provider-neutral way.

This PR adds a concrete `to_content_blocks()` method to `BasePlanner` that converts `List[types.Part]` into the standardized content block format popularised by LangChain 1.0: thought/reasoning parts become `{'type': 'reasoning', 'reasoning': '...'}` dicts and regular text parts become `{'type': 'text', 'text': '...'}` dicts; parts that carry neither (e.g. function-call parts) are skipped.

Both `PlanReActPlanner` and `BuiltInPlanner` inherit the method automatically; custom subclasses can override it if they need different mapping logic.

Four regression tests covering text+reasoning, empty input, non-text parts, and the `BuiltInPlanner` path are added to `tests/unittests/planners/test_plan_re_act_planner.py`.

Fixes #3378

### DIFF
--- a/src/google/adk/planners/base_planner.py
+++ b/src/google/adk/planners/base_planner.py
@@ -16,6 +16,8 @@ from __future__ import annotations
 
 import abc
 from abc import ABC
+from typing import Any
+from typing import Dict
 from typing import List
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


DeanChensj:
Hi @guptaishaan , I think the new helper method fits better in https://github.com/google/adk-python/blob/main/src/google/adk/utils/content_utils.py, wdyt?

DeanChensj:

PR #6618, opened 2026-08-06
fix(models): preserve tool_call co-delivered with turn_complete

### LINKED ISSUE #6615: Live mode: tool_call can be silently dropped when co-framed with turn_complete in gemini_llm_connection.py
### Description

In `google/adk/models/gemini_llm_connection.py`, `receive()`'s handling of `message.server_content` (the `if message.server_content:` block starting around line 314) ends its `turn_complete` handling with an unconditional `break` around line 538. The sibling `if message.tool_call:` handler sits at ~line 562 — a separate top-level branch only reached if the loop hasn't already `break`-ed out.

If a single Live API server message carries both `serverContent.turnComplete=true` **and** a populated `toolCall` in the same frame, the `tool_call` branch is never reached and the function call is silently dropped — no exception, no log line, nothing observable except a missing function call downstream. The wire protocol appears to allow this framing (both fields are independent keys on one server message per the converter mapping in `google-genai`'s live converters).

### How we found it

We were chasing an intermittent bug where a Gemini Live model would compose a function call in its own `thinking` output (via `include_thoughts=True`) but the corresponding tool call never arrived client-side. Static reading of the installed `google-adk==2.5.0` source turned up this real ordering issue.

We want to flag two things clearly:
1. This is a genuine defect we can point to in source, independent of our specific bug.
2. We do **not** believe it explains our bug. We captured every raw pre-ADK-processing Live API server message for several reproductions (logging at the point `gemini_llm_connection.py` receives each message, before any ADK-side handling) and in every fabricating turn, the raw message stream simply never contained a `tool_call` at all — so there was nothing for this code path to drop. Our issue is upstream (model/serving-side), not this ordering bug. But we think this ordering issue is real, reproducible from source, and could bite a different model/config where a turn_complete and a tool_call really do co-occur in one frame.

### Suggested fix

Check `message.tool_call` independently of the `turn_complete` branch's `break` (e.g. before the break, or restructure so the two are not mutually exclusive within one message's processing), so a `tool_call` co-framed with `turn_complete` is never skipped.

### Environment
- google-adk 2.5.0
- google-genai 2.10.0
- Live API (bidi streaming / WebSocket transport), various models including gemini-3.1-flash-live-preview

### LINKED ISSUE #6616: Bug: Systemic parts[0] indexing drops multimodal streaming data and bypasses content validation
## 🔴 Required Information

**Describe the Bug:**
The framework makes unsafe `parts[0]` indexing assumptions in the core Gemini Live API connection (`gemini_llm_connection.py`) and flow modules (`base_llm_flow.py`). This causes two critical systemic bugs:
1. **Validation Bypass**: Mixed-content blocks (e.g., text followed by a tool response) bypass `ValueError` validation entirely and are incorrectly routed as normal client content.
2. **Dropped Multimodal Streaming**: The Live API streaming receiver silently drops subsequent parts in multi-part streaming chunks (e.g., mid-chunk thought boundary transitions or inline data) because it only extracts data from `content.parts[0]`.

**Steps to Reproduce:**
Please provide a numbered list of steps to reproduce the behavior:
1. Initialize a `GeminiLlmConnection`.
2. Construct a `types.Content` object with multiple parts, where the first part is text and the second part is a `types.FunctionResponse`.
3. Call `await gemini_connection.send_content(content)`.
4. Observe that the content is routed to the standard `send()` pipeline instead of correctly raising a `ValueError`.

**Expected Behavior:**
1. Sending a mixed-content block containing a function response should strictly trigger the validation `ValueError("Function-response content cannot mix function and non-function parts.")` regardless of index.
2. The Live API `receive()` generator should iterate over all parts in `content.parts` to ensure no multimodal data or mid-chunk transitions are dropped.

**Observed Behavior:**
1. The framework explicitly checks `if content.parts[0].function_response`, bypassing validation entirely if the text part comes first.
2. The `receive()` loop hardcodes `content.parts[0]`, permanently ignoring any subsequent parts sent in the same chunk.

**Environment Details:**

 - ADK Library Version: `main` branch
 - Desktop OS: N/A (Architecture bug)
 - Python Version: 3.11+

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: `gemini-3.1-flash-live-preview`

---

## 🟡 Optional Information

**Regression:**
N/A - This appears to be a systemic issue tied to early architectural assumptions that chunks and payloads would only ever contain a single part.

**Logs:**
```text
N/A - Issue prevents proper error logs from being emitted due to silent data drops/misrouting.
```

**Screenshots / Video:**
N/A

**Additional Context:**
While it might be rare in current API responses for certain models to stream multiplexed chunks, the Gemini protocol **does not prohibit it**. The API could technically stream multiplexed chunks (e.g., an image with text, or multiple tool responses at once). By removing these fragile `parts[0]` assumptions, the framework becomes robust against future Live API updates.

```python
import pytest

# Example of the Mixed Content Validation Bypass
@pytest.mark.asyncio
async def test_mixed_content_bypass(gemini_connection):
    function_response = types.FunctionResponse(name='test_function', response={'result': 'success'})
    content = types.Content(
        role='user',
        parts=[
            types.Part.from_text(text='Hello'),
            types.Part(function_response=function_response),
        ],
    )

    # EXPECTED: ValueError("Function-response content cannot mix...")
    # OBSERVED: Silently misroutes to standard send() pipeline.
    await gemini_connection.send_content(content)
```

 - Always (100%) - Occurs reliably whenever the framework encounters a multi-part array where the target data is not at index `0`.

patrickswedish:
## Summary

Fixes #6615.

In `GeminiLlmConnection.receive()`, when a `LiveServerMessage` carries both `server_content.turn_complete=True` and `message.tool_call`, the previous control flow broke out of the receive loop before reaching the lower `if message.tool_call:` handler, silently discarding the tool call.

Note: Multipart text data loss (#6616) has already landed upstream via commit `2109ea96e1c7431225e57fb6db6c4c116e632867` (PR #6617). This PR is now solely and cleanly focused on #6615.

---

## Root cause

The receive loop was structured as:

``python
async for message in agen:
    if message.server_content:
        ...
        if message.server_content.turn_complete:
            ...
            break          # exits receive loop
    if message.tool_call:  # unreachable when turn_complete is present
        ...
``

When a single `LiveServerMessage` contains both `turn_complete=True` and `tool_call`, the `break` terminated receive before `if message.tool_call:` could execute.

---

## Fix

Restructured `GeminiLlmConnection.receive()` so that all message fields (including `tool_call`, `session_resumption_update`, `voice_activity`, and `go_away`) are processed through their standard handling semantics before finalizing `turn_complete` and exiting the receive loop.

- Evaluates `if message.tool_call:` before evaluating `if message.server_content.turn_complete:`.
- Preserves single tool-call handling implementation (zero duplicated logic).
- Preserves ordering: text -> tool call -> turn_complete.
- Preserves immediate yielding for Gemini 3.x Live models and buffering/aggregation for other models.
- Preserves all metadata (`grounding_metadata`, `last_grounding_metadata`, `tool_call_metadata`, `live_session_id`, `model_version`).

---

## Testing

- **Targeted Unit Tests**: `pytest tests/unittests/models/test_gemini_llm_connection.py` (66 / 66 passed)
  - `test_receive_preserves_tool_call_codelivered_with_turn_complete` (verifies multi-function tool call preserved before turn_complete)
  - `test_receive_preserves_text_and_tool_call_codelivered_with_turn_complete` (verifies text + tool_call + turn_complete ordering and content)
  - `test_receive_codelivered_tool_call_empty_function_calls` (verifies empty function_calls handling)
- **Model Unit Tests**: `pytest tests/unittests/models/` (1068 / 1068 passed)
- **Linters / Formatters**: `ruff`, `isort`, `pyink` (Clean)
- **Build**: `uv build` (Built sdist and wheel successfully)
- **Diff Check**: `git diff --check` (Clean)

### DIFF
--- a/src/google/adk/models/gemini_llm_connection.py
+++ b/src/google/adk/models/gemini_llm_connection.py
@@ -380,18 +380,26 @@ class GeminiLlmConnection(BaseLlmConnection):
                 llm_response.grounding_metadata = (
                     message.server_content.grounding_metadata
                 )
-            if content.parts[0].text:
-              current_is_thought = getattr(content.parts[0], 'thought', False)
+            # Collect all text parts from the message chunk.
+            # Previously only content.parts[0].text was examined, which
+            # silently dropped text from any part beyond the first when a
+            # single streaming chunk contained multiple text parts (e.g. a
+            # multimodal response with several text segments).
+            _text_parts = [p for p in content.parts if p.text]
+            _has_inline_data = any(p.inline_data for p in content.parts)
+            if _text_parts:
+              current_is_thought = getattr(_text_parts[0], 'thought', False)
               if text and current_is_thought != is_thought:
                 yield self.__build_full_text_response(text, is_thought)
                 text = ''
                 is_thought = False
 
-              text += content.parts[0].text
+              for _tp in _text_parts:
+                text += _tp.text
               is_thought = current_is_thought
               llm_response.partial = True
             # don't yield the merged text event when receiving audio data
-            elif text and not content.parts[0].inline_data:
+            elif text and not _has_inline_data:
               yield self.__build_full_text_response(
                   text, is_thought, last_grounding_metadata
               )
@@ -499,6 +507,26 @@ class GeminiLlmConnection(BaseLlmConnection):
               )
               self._output_transcription_text = ''
           if message.server_content.turn_complete:
+            # Process any tool_call co-delivered in the same server message as
+            # turn_complete. Without this, the `break` below exits the receive
+            # loop before the `if message.tool_call:` block lower in the loop
+            # body is reached, silently discarding the tool call.
+            if message.tool_call:
+              logger.debug(
+                  'Processing tool_call co-delivered with turn_complete'
+              )
+              if text:
+                yield self.__build_full_text_response(
+                    text, is_thought, last_grounding_metadata
+                )
+                text = ''
+                is_thought = False
+                last_grounding_metadata = None
+              tool_call_parts.extend([
+                  types.Part(function_call=function_call)
+                  for function_call in message.tool_call.function_calls or []
+              ])
+
             # Capture final grounding metadata before last_grounding_metadata is cleared in the next block.
             final_grounding_metadata = (
                 grounding_metadata
--- a/src/google/adk/models/gemini_llm_connection.py
+++ b/src/google/adk/models/gemini_llm_connection.py
@@ -512,77 +512,14 @@ class GeminiLlmConnection(BaseLlmConnection):
                   live_session_id=live_session_id,
               )
               self._output_transcription_text = ''
-          if message.server_content.turn_complete:
-            # Capture final grounding metadata before last_grounding_metadata is cleared in the next block.
-            final_grounding_metadata = (
-                grounding_metadata
-                or last_grounding_metadata
-                or (
-                    types.GroundingMetadata()
-                    if self._is_gemini_3_x_live
-                    else None
-                )
-            )
-            if (
-                final_grounding_metadata
-                and final_grounding_metadata.retrieval_queries
-                and not final_grounding_metadata.grounding_chunks
-            ):
-              logger.warning(
-                  'Incomplete grounding_metadata received: retrieval_queries=%s'
-                  ' but grounding_chunks is empty. This may indicate a'
-                  ' transient issue with the Vertex AI Search backend.',
-                  final_grounding_metadata.retrieval_queries,
-              )
-
-            if text:
-              yield self.__build_full_text_response(
-                  text,
-                  is_thought,
-                  last_grounding_metadata,
-                  bool(message.server_content.interrupted),
-              )
-              text = ''
-              is_thought = False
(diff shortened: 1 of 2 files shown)


patrickswedish:
@googlebot I signed the CLA!

patrickswedish:
Hi team! 👋

Just a quick heads up on this PR — it fixes two silent data-loss bugs in the live streaming receive loop that affect real-world multimodal and tool-calling use cases:

1. **#6616** — multi-part text chunks lose all text beyond `parts[0]`
2. **#6615** — `tool_call` co-delivered with `turn_complete` is silently dropped, deadlocking the agent

Both fixes are surgical (25 lines changed, one file), backward-compatible, and traced directly against the live API message format. Happy to add unit tests or address any review feedback. Looking forward to getting this in! 🙏

wuliang229:
This is a non-issue.

patrickswedish:
Thanks for reviewing and providing the feedback, @wuliang229! Appreciate your time.

PR #6629, opened 2026-08-07
fix(cli): add optional OIDC verification for Pub/Sub and Eventarc triggers

herdiyana256:
## Impact

The `/apps/{app_name}/trigger/pubsub` and `/apps/{app_name}/trigger/eventarc` endpoints run an agent against fully attacker-controlled request content with **no authentication of their own**. Anyone who can reach the URL can invoke an arbitrary agent run:

- **Unauthenticated agent invocation on a public endpoint.** These endpoints are not localhost-dev surfaces. `adk deploy ... --trigger_sources=pubsub,eventarc` exists specifically to deploy them, and Pub/Sub push and Eventarc reach them over public HTTPS. A deployment that relies only on the URL being unguessable, or that intends the platform to enforce IAM but has not, is fully open.
- **Attacker chooses the agent's input.** The request body becomes the agent's `new_message` verbatim (base64 `message.data` for Pub/Sub, the CloudEvent body for Eventarc). That is a direct prompt-injection channel into a production agent: tool calls, downstream API calls, memory writes, and spend all execute under the deployment's own credentials, driven by whoever sent the POST.
- **No pre-created session or secret needed.** `user_id` is derived from the caller-supplied `subscription` / `source` field, so there is no session to guess and no shared secret to hold. A bare POST triggers a full agent run.
- **Cost and abuse.** Each request drives a real LLM invocation. An open trigger endpoint is a direct path to quota exhaustion and billing abuse on top of the injection surface.

The crucial detail is that **the caller identity is already available and was being thrown away.** Pub/Sub push subscriptions and Eventarc triggers, when configured with a service account, attach a Google-signed OIDC bearer token in the `Authorization` header precisely so the receiver can verify who is calling. The router never looked at it. This is the same gap tracked for the Go implementation in google/adk-go#1241.

## Fix

Adds an opt-in `trigger_oidc_audience`, threaded through `get_fast_api_app` and `AdkWebServer` into `TriggerRouter`. When set, every `/trigger/*` request must carry a Google-signed OIDC bearer token whose audience matches, verified with `google.oauth2.id_token.verify_oauth2_token` before the agent runs; a missing, malformed, or invalid token returns `401` and never reaches the agent.

When unset (the default), behavior is unchanged, so this is **not a breaking change** — existing deployments that rely on platform IAM keep working, and deployments that were relying on nothing can now close the hole with one argument.

## Verification

`.venv/bin/python -m pytest tests/unittests/cli/test_trigger_routes.py` — 43 passed, including the new `TestTriggerOidcVerification`: missing token, non-bearer scheme, and invalid token all return 401; a valid token is accepted and the configured audience is the one passed to the verifier; and the default (no audience) path stays unauthenticated so current behavior is preserved.

### DIFF
--- a/src/google/adk/cli/api_server.py
+++ b/src/google/adk/cli/api_server.py
@@ -721,6 +721,7 @@ class ApiServer:
       url_prefix: Optional[str] = None,
       auto_create_session: bool = False,
       trigger_sources: Optional[list[str]] = None,
+      trigger_oidc_audience: Optional[str] = None,
       default_llm_model: Optional[str] = None,
   ):
     self.agent_loader = agent_loader
@@ -741,6 +742,7 @@ class ApiServer:
     self.url_prefix = url_prefix
     self.auto_create_session = auto_create_session
     self.trigger_sources = trigger_sources
+    self.trigger_oidc_audience = trigger_oidc_audience
     self.default_llm_model = default_llm_model
     self.default_app_name = os.getenv("ADK_DEFAULT_APP_NAME")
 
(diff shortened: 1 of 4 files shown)


herdiyana256:
Gentle bump, this has been sitting since early August. It adds optional OIDC verification for the Pub/Sub and Eventarc trigger endpoints, which currently accept unauthenticated calls. @GWeale mind a quick review, or let me know if the optional-by-default shape is what you'd prefer here?

PR #6635, opened 2026-08-07
fix: report root_agent type mismatch instead of 'No root_agent found'

### LINKED ISSUE #6606: AgentLoader discards its own diagnosis: type mismatch on `root_agent` is logged at WARNING, then reported as "No root_agent found"
When a module exposes a `root_agent` that is not a `BaseAgent`, `AgentLoader` detects the
exact problem, formats it with the offending type, writes it to a `logger.warning`, and
then falls through to raise an error describing something else entirely.

### What the user sees

```
ValueError: No root_agent found for 'agent'. Searched in 'agent.agent.root_agent',
'agent.root_agent' and 'agent/root_agent.yaml'.

Expected directory structure:
  <agents_dir>/
    agent/
      agent.py (with root_agent) OR
      root_agent.yaml

Then run: adk web <agents_dir>

Ensure '/path/to/agents/judge' is structured correctly, an .env file can be loaded if
present, and a root_agent is exposed.

HINT: It looks like this command might be running from inside an agent directory. Run it
from the parent directory that contains your agent folder (for example the project root)
so the loader can locate your agents.
```

Every actionable statement in that message is wrong for this failure:

- `root_agent` **was** found — it exists in the module.
- The directory structure **was** correct.
- The cwd **was** correct; the HINT sends you to restructure a working layout.

### The actual cause

`google/adk/cli/utils/agent_loader.py::_load_from_module_or_package` (v1.28.0):

```python
if hasattr(module_candidate, "app") and isinstance(module_candidate.app, App):
    return module_candidate.app
elif hasattr(module_candidate, "root_agent"):
    if isinstance(module_candidate.root_agent, BaseAgent):
        return module_candidate.root_agent
    else:
        logger.warning(
            "Root agent found is not an instance of BaseAgent. But a type %s",
            type(module_candidate.root_agent),
        )
```

The contract is that an `App` must be exported as `app`, and `root_agent` must be a
`BaseAgent`. My module did this:

```python
root_agent = App(name="judge", root_agent=judge_agent)   # wrong
app = App(name="judge", root_agent=judge_agent)          # right
```

Reasonable mistake to make — `App` has a `root_agent` field, so assigning an `App` to a
module-level `root_agent` reads naturally, and the sample I was working from imports `App`
without showing the export name.

### Reproduction

```python
# agents/judge/agent.py
from google.adk.apps.app import App

root_agent = App(name="judge", root_agent=Agent(name="judge", model="gemini-3-flash-preview"))
```

Load it with `AgentLoader`. At default log level the WARNING is invisible and you get the
directory-structure message.

### Suggested fix

When the type-mismatch branch is taken, remember it and raise with that instead of the
generic not-found error. Roughly:

```python
mismatched_type = None
...
else:
    mismatched_type = type(module_candidate.root_agent)
    logger.warning(...)
...
if mismatched_type is not None:
    raise ValueError(
        f"'{agent_name}.root_agent' is a {mismatched_type.__name__}, not a BaseAgent. "
        f"If you meant to export an App, name it `app` instead of `root_agent`."
    )
raise ValueError("No root_agent found ...")
```

The information is already computed; it is just being thrown away. Suppressing the
directory-structure diagram and the wrong-directory HINT in this case would also help,
since both actively mislead.

### Impact

Low severity, high friction. Cost for one user was reading the loader source to discover a
naming contract. Anyone hitting it will plausibly restructure a correct directory tree
first, because that is what the error tells them to do.

---

Environment: google-adk 1.28.0, Python 3.12, Ubuntu 24.04 (WSL2).

loponly:
Fixes #6606

## Problem
When a module defines `root_agent` with a non-agent type, `AgentLoader`
logs the type mismatch at WARNING but then raises the generic
`ValueError: No root_agent found`, whose directory-structure guidance
sends users in the wrong direction when their layout is correct.

## Fix
`_load_from_module_or_package` and `_load_from_submodule` record each
type mismatch; `_perform_load` raises a targeted `ValueError` naming the
module and the actual type at its final failure point. When the object
is an `App`, the error suggests exporting it under the name `app`.
Fallthrough behavior is preserved: a mismatch in one pattern does not
block a later pattern from loading successfully, and the generic
not-found message is unchanged for the genuinely-missing case.

## Testing Plan
- New unit tests in `tests/unittests/cli/utils/test_agent_loader.py`:
  - `test_wrong_type_root_agent_raises_targeted_error`
  - `test_app_exported_as_root_agent_suggests_app_name`
  - `test_wrong_type_root_agent_in_agent_module_raises_targeted_error`
  - `test_valid_agent_module_wins_over_mistyped_package_root_agent`
- `pytest tests/unittests/cli -n auto`: 774 passed, 5 skipped, 4 xfailed, 376 warnings in 84.37s (0:01:24)
- `tox` (full multi-version matrix): py310 OK, py312 OK, py313 OK, py314 OK.
  py311 has one failure in
  `tests/unittests/sessions/migration/test_migration.py::test_migrate_from_sqlalchemy_pickle_preserves_nested_safe_actions_pickle`
  that is unrelated to this change and reproduces identically on `main`:
  the migration unpickler's `_ALLOWED_PICKLE_GLOBALS` blocks
  `builtins.getattr` only in the py311 environment (the one where the
  `crewai` extras install per the pyproject markers). This diff touches
  only `cli/utils/agent_loader.py` and its test file. Filed separately.

### DIFF
--- a/src/google/adk/cli/utils/agent_loader.py
+++ b/src/google/adk/cli/utils/agent_loader.py
@@ -195,6 +195,9 @@ class AgentLoader(BaseAgentLoader):
               "Root agent found is not an instance of BaseAgent. But a type %s",
               type(module_candidate.root_agent),
           )
+          self._record_root_agent_type_mismatch(
+              f"{agent_name}.agent", module_candidate.root_agent
+          )
       else:
         logger.debug(
             "Module %s.agent has no root_agent.",
--- a/src/google/adk/cli/utils/agent_loader.py
+++ b/src/google/adk/cli/utils/agent_loader.py
@@ -358,7 +358,6 @@ class AgentLoader(BaseAgentLoader):
       )
       return root_agent
 
-    # If no root_agent was found by any pattern
     # A root_agent was found but had an unusable type: surface that
     # diagnosis instead of the generic not-found guidance (#6606).
     if self._root_agent_type_mismatches:
@@ -378,6 +377,7 @@ class AgentLoader(BaseAgentLoader):
           f"{details}\n\nExpose a BaseAgent instance named"
           f" 'root_agent'.{app_hint}"
       )
+    # If no root_agent was found by any pattern
     # Check if user might be in the wrong directory
     hint = ""
     agents_path = Path(agents_dir)
--- a/src/google/adk/cli/utils/agent_loader.py
+++ b/src/google/adk/cli/utils/agent_loader.py
@@ -79,6 +79,7 @@ class AgentLoader(BaseAgentLoader):
     self._init_agent_mode(agents_path)
     self._original_sys_path = None
     self._agent_cache: dict[str, Union[BaseAgent, App]] = {}
+    self._root_agent_type_mismatches: list[tuple[str, str, bool]] = []
 
   def _init_agent_mode(self, agents_path: Path) -> None:
     if is_single_agent_directory(agents_path):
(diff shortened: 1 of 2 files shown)


loponly:
@googlebot I signed it!

PR #6653, opened 2026-08-08
fix(flows): keep non-text static_instruction as a stable request prefix

### LINKED ISSUE #6652: static_instruction and instruction are moved behind conversation history when static_instruction contains non-text content
## 🔴 Required Information

**Describe the Bug:**

When `LlmAgent.static_instruction` contains non-text content such as a PDF provided through `file_data`, ADK converts that part into user content.

On the first turn, the generated request has the expected order:

```text
static PDF content
dynamic instruction
current user message
```

On the second and subsequent turns, however, the same static PDF content is inserted after the existing conversation history:

```text
previous user message
previous model response
static PDF content
dynamic instruction
current user message
```

This means the supposedly static PDF is no longer part of a stable request prefix across turns.

Because [Gemini implicit context caching](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/context-cache/context-cache-overview?hl=en#implicit-caching) relies on matching a sufficiently large prefix, this ordering appears to prevent a large PDF in `static_instruction` from benefiting from implicit context caching.

The PDF itself contains approximately 40,040 image tokens. Despite sending the same PDF on consecutive requests, neither response contains `cached_content_token_count`.

This is especially surprising because `static_instruction` is documented as being sent at the beginning of the request for context-caching purposes.

https://github.com/google/adk-python/blob/740582e9f283cd23ff5cec1389400b422513f765/src/google/adk/agents/llm_agent.py#L284-L286

**Steps to Reproduce:**

1. Install ADK 2.6.1.

   ```bash
   pip install google-adk==2.6.1
   ```

2. Create the agent from the minimal reproduction below.

3. Configure the Vertex AI backend.

   ```bash
   export GOOGLE_GENAI_USE_VERTEXAI=TRUE
   export GOOGLE_CLOUD_PROJECT="your-project"
   export GOOGLE_CLOUD_LOCATION="global"
   ```

4. Start the agent with debug logging enabled.

   ```bash
   adk web -v >debug.log 2>&1
   ```

5. Start a new session and send:

   ```text
   こんにちは
   ```

6. In the same session, send:

   ```text
   論文の目次を教えて
   ```

7. Inspect the generated `contents` and `usage_metadata` in the debug log.

**Expected Behavior:**

Contents originating from `static_instruction` should remain at the beginning of the generated request on every turn. The dynamic `instruction` should follow the static contents, before the conversation history.

For example:

```text
static PDF content
dynamic instruction
previous user message
previous model response
current user message
```

This preserves the documented ordering between `static_instruction` and `instruction`, while keeping the large static PDF as a stable request prefix.

Implicit cache hits are controlled by the model backend and are not guaranteed. However, ADK should preserve this stable prefix rather than moving static content behind changing conversation history.

**Observed Behavior:**

The first request is ordered as follows:

```text
1. Referenced file data + PDF file_data
2. Dynamic instruction
3. Current user message: こんにちは
```

The second request is ordered as follows:

```text
1. Previous user message: こんにちは
2. Previous model response
3. Referenced file data + the same PDF file_data
4. Dynamic instruction
5. Current user message: 論文の目次を教えて
```

Usage metadata for the first response:

```json
{
  "candidates_token_count": 109,
  "prompt_token_count": 40259,
  "prompt_tokens_details": [
    {
      "modality": "IMAGE",
      "token_count": 40040
    },
    {
      "modality": "TEXT",
      "token_count": 219
    }
  ],
  "thoughts_token_count": 109,
  "total_token_count": 40477,
  "traffic_type": "ON_DEMAND"
}
```

Usage metadata for the second response:

```json
{
  "candidates_token_count": 1068,
  "prompt_token_count": 40483,
  "prompt_tokens_details": [
    {
      "modality": "TEXT",
      "token_count": 443
    },
    {
      "modality": "IMAGE",
      "token_count": 40040
    }
  ],
  "thoughts_token_count": 852,
  "total_token_count": 42403,
  "traffic_type": "ON_DEMAND"
}
```

Neither response reports `cached_content_token_count`.

**Environment Details:**

- ADK Library Version (`pip show google-adk`): 2.6.1
- Desktop OS: macOS
- Python Version: 3.13.12

**Model Information:**

- Are you using LiteLLM: No
- Which model are you using: `gemini-3.6-flash`
- Backend: Vertex AI

---

## 🟡 Optional Information

**Regression:**

Unknown.

I have not verified an earlier ADK version in which non-text `static_instruction` content remained at the beginning of multi-turn requests.

**Logs:**

The relevant request ordering from the debug log is summarized above. The important difference is that the same PDF moves from the beginning of the first request to a position after the previous model response in the second request.

Full: https://gist.github.com/ftnext/c5e79dc8b19d6e59298633e4dd866c86

**Screenshots / Video:**

N/A

**Additional Context:**

The relevant flow appears to be:

1. Non-text parts of `static_instruction` are extracted into user contents by
   `LlmRequest.append_instructions()`:
   [`llm_request.py`](https://github.com/google/adk-python/blob/v2.6.1/src/google/adk/models/llm_request.py#L185-L234)

2. The contents processor rebuilds the conversation history and then inserts
   all instruction-related user contents:
   [`contents.py`](https://github.com/google/adk-python/blob/v2.6.1/src/google/adk/flows/llm_flows/contents.py#L82-L135)

3. `_add_instructions_to_user_content()` inserts those contents before the last
   continuous batch of user contents rather than at position zero:
   [`contents.py`](https://github.com/google/adk-python/blob/v2.6.1/src/google/adk/flows/llm_flows/contents.py#L1239-L1279)

This placement is appropriate for dynamic instructions, but non-text contents originating from `static_instruction` appear to need separate placement so that they remain a stable prefix.

Related issues and PRs:

- #6062 and #6216 concern explicit context-cache fingerprinting.
- #6067 fixes explicit cache reuse.
- #5522 changes the explicit-cache boundary around the last user batch.

Those appear related but do not address this issue: the present problem is the actual `GenerateContentRequest.contents` ordering for non-text static instructions and its effect on provider-side implicit caching.

https://github.com/ftnext/agent-practice/tree/5a09ef5daf16cf27a54144daa89d2529e864cd5c/adk/pdf/static_instruction

based on https://github.com/google/adk-python/tree/v2.6.1/contributing/samples/multimodal/static_non_text_content

**Proposed Regression Test:**

The following test can be added to `tests/unittests/flows/llm_flows/test_instructions.py`.

It verifies request structure rather than asserting an actual cache hit, since implicit cache hits are controlled by the backend.

```diff
diff --git a/tests/unittests/flows/llm_flows/test_instructions.py b/tests/unittests/flows/llm_flows/test_instructions.py
--- a/tests/unittests/flows/llm_flows/test_instructions.py
+++ b/tests/unittests/flows/llm_flows/test_instructions.py
@@
 from google.adk.agents.run_config import RunConfig
+from google.adk.events.event import Event
 from google.adk.flows.llm_flows import instructions
@@
+@pytest.mark.asyncio
+async def test_static_instruction_file_precedes_multi_turn_history():
+  """Test that a static file remains a stable prefix across turns."""
+  file_uri = "gs://test-bucket/reference.pdf"
+  agent = LlmAgent(
+      name="test_agent",
+      instruction="Dynamic instruction",
+      static_instruction=types.Content(
+          parts=[
+              types.Part(
+                  file_data=types.FileData(
+                      file_uri=file_uri,
+                      mime_type="application/pdf",
+                  )
+              )
+          ]
+      ),
+  )
+  invocation_context = await _create_invocation_context(agent)
+  invocation_context.session.events = [
+      Event(
+          invocation_id="inv1",
+          author="user",
+          content=types.UserContent("First message"),
+      ),
+      Event(
+          invocation_id="inv2",
+          author="test_agent",
+          content=types.ModelContent("First response"),
+      ),
+      Event(
+          invocation_id="inv3",
+          author="user",
+          content=types.UserContent("Second message"),
+      ),
+  ]
+  llm_request = LlmRequest()
+
+  async for _ in request_processor.run_async(
+      invocation_context, llm_request
+  ):
+    pass
+  async for _ in contents_processor.run_async(
+      invocation_context, llm_request
+  ):
+    pass
+
+  assert len(llm_request.contents) == 5
+
+  static_content = llm_request.contents[0]
+  assert static_content.role == "user"
+  assert static_content.parts[0].text == (
+      "Referenced file data: file_data_0"
+  )
+  assert static_content.parts[1].file_data
+  assert static_content.parts[1].file_data.file_uri == file_uri
+
+  dynamic_content = llm_request.contents[1]
+  assert dynamic_content.role == "user"
+  assert dynamic_content.parts[0].text == "Dynamic instruction"
+
+  assert llm_request.contents[2] == types.UserContent("First message")
+  assert llm_request.contents[3] == types.ModelContent("First response")
+  assert llm_request.contents[4] == types.UserContent("Second message")
```

**How often does this issue occur?**

The request reordering occurs deterministically on the second and subsequent turns when conversation history contains a model response.

In the observed two-turn reproduction, no implicit cache hit was reported. The structural ordering issue occurs consistently, although an implicit cache hit itself is not guaranteed by the backend.

chelsealong:
Fixes #6652

## Bug

When `LlmAgent.static_instruction` contains non-text content (e.g. a PDF
provided through `file_data`), ADK extracts that content into a "user"
content entry (via `LlmRequest.append_instructions`) so it can be sent as
part of `contents`.

On the first turn this lands at the front of the request as expected. On
the second and later turns, `_add_instructions_to_user_content` in
`contents.py` inserts all instruction-related contents (the static
non-text content *and* the dynamic instruction) right before the last
continuous batch of user content — i.e. after the existing conversation
history rather than at the front of the request:

```
previous user message
previous model response
static PDF content        <-- moved behind history
dynamic instruction
current user message
```

This defeats the "stable prefix" that provider-side implicit context
caching relies on, so a large static PDF never gets a cache hit past the
first turn, as described in #6652.

## Fix

`LlmRequest.append_instructions` now tracks the user contents it extracts
from a `Content` argument (currently only used for `static_instruction`)
in a new private list, `_static_instruction_contents`, in addition to
appending them to `contents` as before.

`_add_instructions_to_user_content` in `contents.py` checks this list: if
it's non-empty, the full set of instruction-related contents (static
content followed by the dynamic instruction, in the order they were
built) is inserted ahead of conversation history instead of before the
last user batch. This keeps the static content — and the dynamic
instruction that follows it — as a stable prefix ahead of conversation
history:

```
static PDF content
dynamic instruction
previous user message
previous model response
current user message
```

When `static_instruction` has no non-text parts (the common case, and the
existing behavior for pure dynamic-instruction agents), nothing changes:
instructions still get inserted right before the latest user turn.

`_add_instructions_to_user_content` actually has a second call site:
`_finalize_dynamic_instructions` in `base_llm_flow.py`, reached when the
experimental `DYNAMIC_INSTRUCTION_ROUTING` feature is on and a tool (e.g.
`preload_memory_tool`, `load_artifacts_tool`, `load_mcp_resource_tool`)
contributes a dynamic instruction via `_append_dynamic_instructions`. That
call happens later in the same turn's preprocessing, after the static
prefix has already been placed by the first call. An earlier version of
this fix inserted at a fixed index 0 on every call, which meant this
second call re-inserted at the front too, pushing the tool's dynamic
content in front of the static prefix the first call had just placed —
recreating the exact bug this PR fixes, one call later.

To fix that, `LlmRequest` now tracks
`_static_instruction_prefix_end_index`, the index right after the
inserted static-instruction prefix. Only the first call inserts at index
0 (and records where the prefix ends); any later call for the same
request inserts right after that tracked index instead, so tool-triggered
dynamic instructions land after the static prefix without displacing it.

## Testing plan

Added `test_static_instruction_file_precedes_multi_turn_history` in
`tests/unittests/flows/llm_flows/test_instructions.py`, which reproduces
the multi-turn scenario from the issue (static file_data instruction +
dynamic instruction + 3 turns of history) and asserts the static content
and dynamic instruction both precede the history.

Added `test_static_instruction_file_stays_prefix_after_tool_dynamic_instruction`,
which reproduces the second-call-site regression: after the main content
processor places the static prefix, it simulates a tool-triggered dynamic
instruction the way `_finalize_dynamic_instructions` does, and asserts the
static content is still at index 0 afterward.

Verified both new tests fail without their respective fixes:

```
$ git checkout HEAD~1 -- src/google/adk/flows/llm_flows/contents.py src/google/adk/models/llm_request.py
$ python -m pytest tests/unittests/flows/llm_flows/test_instructions.py::test_static_instruction_file_precedes_multi_turn_history
AssertionError: assert 'First message' == 'Referenced f...: file_data_0'

$ git checkout HEAD~2 -- src/google/adk/flows/llm_flows/contents.py src/google/adk/models/llm_request.py
$ python -m pytest tests/unittests/flows/llm_flows/test_instructions.py::test_static_instruction_file_stays_prefix_after_tool_dynamic_instruction
AssertionError: assert 'Relevant mem...r likes pizza' == 'Referenced f...: file_data_0'
```

With the fix:

```
$ git checkout HEAD -- src/google/adk/flows/llm_flows/contents.py src/google/adk/models/llm_request.py
$ python -m pytest tests/unittests/flows/llm_flows/test_instructions.py
38 passed
$ python -m pytest tests/unittests/flows/llm_flows/ tests/unittests/models/test_llm_request.py
562 passed, 1 xfailed
$ python -m pytest tests/unittests/flows/
529 passed, 1 xfailed
$ python -m pytest tests/unittests/tools/ -k "preload_memory or load_artifacts or load_mcp_resource"
24 passed
```

Also verified formatting with `pyink` and import order with `isort`
(project's pinned versions) — no changes needed.

## AI assistance disclosure

This change was authored with the help of an AI coding agent (Claude),
with the diff reviewed and the reproduction/test verified before being
pushed. An earlier version of this PR was caught by review as fixing the
bug in the main content-processor call path while leaving a second call
path (tool-triggered dynamic instructions under the experimental
`DYNAMIC_INSTRUCTION_ROUTING` feature) able to re-break the same
invariant; that gap has been closed and covered by a new regression test.

### DIFF
--- a/src/google/adk/flows/llm_flows/contents.py
+++ b/src/google/adk/flows/llm_flows/contents.py
@@ -1346,6 +1346,16 @@ async def _add_instructions_to_user_content(
   """
   if not instruction_contents:
     return
+
+  if llm_request._static_instruction_contents:
+    # Non-text static_instruction content must remain a stable request
+    # prefix across turns (for provider-side context caching), so it goes
+    # at the very beginning, followed by the rest of the instruction
+    # contents (e.g. the dynamic instruction), ahead of conversation
+    # history rather than just ahead of the latest user turn.
+    llm_request.contents[0:0] = instruction_contents
+    return
+
   llm_request._insert_transient_user_content(  # pylint: disable=protected-access
       instruction_contents
   )
--- a/src/google/adk/flows/llm_flows/contents.py
+++ b/src/google/adk/flows/llm_flows/contents.py
@@ -1349,11 +1349,20 @@ async def _add_instructions_to_user_content(
 
   if llm_request._static_instruction_contents:
     # Non-text static_instruction content must remain a stable request
-    # prefix across turns (for provider-side context caching), so it goes
-    # at the very beginning, followed by the rest of the instruction
-    # contents (e.g. the dynamic instruction), ahead of conversation
-    # history rather than just ahead of the latest user turn.
-    llm_request.contents[0:0] = instruction_contents
+    # prefix across turns (for provider-side context caching). The first
+    # call for this request places it at the very beginning, followed by
+    # the rest of the instruction contents (e.g. the dynamic instruction),
+    # ahead of conversation history rather than just ahead of the latest
+    # user turn. Later calls (e.g. tool-triggered dynamic instructions
+    # routed through `_finalize_dynamic_instructions`) insert right after
+    # that tracked prefix instead of at index 0, so they don't displace it.
+    insert_index = llm_request._static_instruction_prefix_end_index
+    if insert_index is None:
+      insert_index = 0
+    llm_request.contents[insert_index:insert_index] = instruction_contents
+    llm_request._static_instruction_prefix_end_index = insert_index + len(
+        instruction_contents
+    )
     return
 
   llm_request._insert_transient_user_content(  # pylint: disable=protected-access
--- a/src/google/adk/models/llm_request.py
+++ b/src/google/adk/models/llm_request.py
@@ -113,6 +113,17 @@ class LlmRequest(BaseModel):
   server-side without a Gemini model.
   """
 
+  _static_instruction_contents: list[types.Content] = PrivateAttr(
+      default_factory=list
+  )
+  """User contents extracted from non-text parts of a Content passed to
+  ``append_instructions`` (e.g. ``static_instruction``).
+
+  These must stay a stable request prefix across turns so that provider-side
+  context caching can key off of them, so the contents processor tracks them
+  separately from other instruction-related contents.
+  """
+
   def _append_dynamic_instructions(self, instructions: list[str]) -> None:
     """Appends dynamic instructions to the request."""
     self._dynamic_instructions.extend(instructions)
@@ -232,6 +243,7 @@ class LlmRequest(BaseModel):
       # Add user contents directly to llm_request.contents
       if user_contents:
         self.contents.extend(user_contents)
+        self._static_instruction_contents.extend(user_contents)
 
       return user_contents
 
--- a/src/google/adk/models/llm_request.py
+++ b/src/google/adk/models/llm_request.py
@@ -124,6 +124,21 @@ class LlmRequest(BaseModel):
   separately from other instruction-related contents.
   """
 
+  _static_instruction_prefix_end_index: Optional[int] = PrivateAttr(
+      default=None
+  )
+  """Index in ``contents`` immediately after the static-instruction prefix,
+  once it has been placed at the front of the request.
+
+  ``_add_instructions_to_user_content`` (contents.py) is called once per
+  turn for the main instruction/dynamic-instruction bundle, and again by
+  ``_finalize_dynamic_instructions`` (base_llm_flow.py) whenever a tool
+  contributes a dynamic instruction. Only the first call should insert at
+  index 0; later calls must insert right after the tracked prefix instead,
+  or they would push tool-triggered content in front of it and break the
+  stable-prefix guarantee.
+  """
+
   def _append_dynamic_instructions(self, instructions: list[str]) -> None:
     """Appends dynamic instructions to the request."""
     self._dynamic_instructions.extend(instructions)
--- a/tests/unittests/flows/llm_flows/test_instructions.py
+++ b/tests/unittests/flows/llm_flows/test_instructions.py
@@ -1227,3 +1227,62 @@ async def test_static_instruction_file_precedes_multi_turn_history():
   assert llm_request.contents[2] == types.UserContent("First message")
   assert llm_request.contents[3] == types.ModelContent("First response")
   assert llm_request.contents[4] == types.UserContent("Second message")
+
+
+@pytest.mark.asyncio
+async def test_static_instruction_file_stays_prefix_after_tool_dynamic_instruction():
+  """The static prefix must survive a later, tool-triggered instruction insert.
+
+  Regression test for https://github.com/google/adk-python/issues/6652: with
+  DYNAMIC_INSTRUCTION_ROUTING enabled, tools (e.g. preload_memory_tool) call
+  ``_append_dynamic_instructions`` and ``_finalize_dynamic_instructions``
+  (base_llm_flow.py) later routes that through the same
+  ``_add_instructions_to_user_content`` helper used by the main contents
+  processor. That second call must not re-insert at index 0, or it would
+  push the tool's dynamic content in front of the static content that the
+  first call already placed there.
+  """
+  file_uri = "gs://test-bucket/reference.pdf"
+  agent = LlmAgent(
+      name="test_agent",
+      instruction="Dynamic instruction",
+      static_instruction=types.Content(
+          parts=[
+              types.Part(
+                  file_data=types.FileData(
+                      file_uri=file_uri,
+                      mime_type="application/pdf",
+                  )
+              )
+          ]
+      ),
+  )
+  invocation_context = await _create_invocation_context(agent)
+  llm_request = LlmRequest()
+
+  async for _ in request_processor.run_async(invocation_context, llm_request):
+    pass
+  async for _ in contents_processor.run_async(invocation_context, llm_request):
+    pass
+
+  assert len(llm_request.contents) == 2
+  assert llm_request.contents[0].parts[0].text == "Referenced file data: file_data_0"
+  assert llm_request.contents[1].parts[0].text == "Dynamic instruction"
+
+  # Simulate a tool contributing a dynamic instruction, finalized the same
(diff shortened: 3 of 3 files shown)


chelsealong:
Rebased onto current main to resolve the merge conflict with upstream commit be103fbd (`fix: keep preloaded context turn-scoped`), which replaced the old insertion-index logic in `_add_instructions_to_user_content` with `LlmRequest._insert_transient_user_content`. Adapted this PR's static-instruction stable-prefix special case to fall through to that new helper for the non-static-instruction path; the static-instruction prefix tracking (`_static_instruction_contents`, `_static_instruction_prefix_end_index`) is unchanged. Full test suite (3427 tests) and both fix-specific tests pass; verified both still fail against the pre-fix source.

ftnext (approved):
Thanks!
I confirmed contents order is fixed at https://github.com/google/adk-python/issues/6652#issuecomment-5301567445

PR #6655, opened 2026-08-09
Strip structural tags from PlanReActPlanner response parts

### LINKED ISSUE #3378: Standarlize content block for PlanReActPlanner and BuiltInPlanner

guptaishaan:
`PlanReActPlanner.process_planning_response` was leaving raw structural markers
(`/*PLANNING*/`, `/*REASONING*/`, `/*ACTION*/`, `/*REPLANNING*/`,
`/*FINAL_ANSWER*/`) in the text of the returned `types.Part` objects.
Consumers had to parse those markers themselves to distinguish reasoning from
the final answer, which is exactly the fragile hand-rolling that standardized
content blocks are meant to avoid.

This change strips the leading structural tag from each text part before it is
returned.  The `part.thought = True` flag already serves as the
machine-readable indicator that a part contains reasoning rather than a
user-visible answer, so no new fields are needed.  The `/*FINAL_ANSWER*/`
separator is also stripped from the reasoning text when a single part contains
both sections.

Two regression tests are added to `test_plan_re_act_planner.py` to verify that
no tag appears in the text of any returned part.

Fixes #3378

### DIFF
--- a/src/google/adk/planners/plan_re_act_planner.py
+++ b/src/google/adk/planners/plan_re_act_planner.py
@@ -102,6 +102,20 @@ class PlanReActPlanner(BasePlanner):
       return text, ''
     return text[: index + len(separator)], text[index + len(separator) :]
 
+  def _strip_tag(self, text: str, tag: str) -> str:
+    """Strips the leading tag from the text if present.
+
+    Args:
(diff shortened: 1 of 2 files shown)


DeanChensj:

PR #6666, opened 2026-08-10
Standardize planner output as typed content blocks for PlanReActPlanner and BuiltInPlanner

### LINKED ISSUE #3378: Standarlize content block for PlanReActPlanner and BuiltInPlanner

guptaishaan:
Both `PlanReActPlanner` and `BuiltInPlanner` expose reasoning through
`types.Part` objects with `thought=True`, but callers had no standard way
to turn those parts into a provider-neutral format comparable to LangChain's
standard content blocks.

This PR adds a `to_content_blocks()` method on `BasePlanner` that converts
a list of `types.Part` objects into a list of typed dicts: thought parts
become `{'type': 'reasoning', 'reasoning': '...'}` and plain text parts
become `{'type': 'text', 'text': '...'}`.  The method works for both
planners because reasoning is already identified uniformly by the
`thought` flag on each part.

Additionally, `PlanReActPlanner._mark_as_thought()` now strips the leading
proprietary tags (`/*PLANNING*/`, `/*REASONING*/`, `/*ACTION*/`,
`/*REPLANNING*/`) from text before storing it, so callers that consume the
raw parts — or the new `to_content_blocks()` output — no longer need to
parse those tags themselves.

Two new unit tests cover tag-stripping and the `to_content_blocks()` contract.

Fixes #3378

### DIFF
--- a/src/google/adk/planners/base_planner.py
+++ b/src/google/adk/planners/base_planner.py
@@ -16,6 +16,8 @@ from __future__ import annotations
 
 import abc
 from abc import ABC
+from typing import Any
+from typing import Dict
 from typing import List
 from typing import Optional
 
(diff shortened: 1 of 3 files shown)


DeanChensj:
Closing in favor of #6709, which directly addresses the planning tag stripping in PlanReActPlanner without introducing additional ContentBlock abstractions over types.Part. Thanks for the contribution!

PR #6673, opened 2026-08-10
fix(a2a): default AgentCardBuilder capabilities to streaming=True

### LINKED ISSUE #6672: to_a2a() serves an agent card with capabilities.streaming=false while the mounted handler implements message/stream
## 🔴 Required Information

**Describe the Bug:**

An agent published with `to_a2a()` serves an agent card whose
`capabilities.streaming` is `false`, while the handler it mounts
implements `message/stream`. Peers that read the card to decide how to
call the agent are told streaming is unavailable when it is served, so
they fall back to unary calls and lose incremental output for no reason.

I am filing this as a question rather than a PR because there is evidence
the `false` may be deliberate, and I would rather not presume:
`_compat.build_agent_card()` takes an explicit `streaming: bool = False`
parameter that `AgentCardBuilder` never passes, and `make_client_config()`
separately documents `ADK applies its defaults streaming=False/polling=False`
for the client side. If the conservative card default is intentional, this
is a documentation request instead — the mismatch is not obvious from
either side alone.

**Steps to Reproduce:**

1. Build an app with `to_a2a(agent)` and fetch the served agent card.
2. Read `capabilities.streaming` — it is `false`.
3. Confirm the mounted handler supports the streaming method:

```console
$ python -c "
from a2a.server.request_handlers import DefaultRequestHandler as H
print([m for m in dir(H) if 'stream' in m.lower()])"
['on_message_send_stream']
```

The path in source, at `bd239ce`:

- `a2a/utils/agent_to_a2a.py` — `AgentCardBuilder(agent=..., rpc_url=...)`
  is constructed with no `capabilities` argument.
- `a2a/utils/agent_card_builder.py` — `self._capabilities = capabilities or AgentCapabilities()`,
  and `AgentCapabilities()` defaults `streaming=False`.
- `a2a/_compat.py` — `build_agent_card(...)` is called with
  `capabilities=self._capabilities` and **without** `streaming=`, so the
  card carries `streaming: false` either through the passed capabilities
  or through `default_capabilities = {"streaming": streaming, ...}`.
- `a2a/_compat.py` — `attach_a2a_routes_to_app()` mounts
  `DefaultRequestHandler`, which implements `on_message_send_stream`.

**Expected Behavior:**

The card reflects what the served handler supports, so a peer can trust it
when choosing a transport — or, if the conservative default is intended,
that intent is documented next to the `streaming` parameter so the
mismatch is discoverable without tracing both paths.

**Observed Behavior:**

`capabilities.streaming` is `false` on every card built by the default
`to_a2a()` path, on a server that implements `message/stream`.

**Environment Details:**

- ADK Library Version: read from source at `bd239ce` (current `main`);
  not reproduced against a pinned release
- Desktop OS: macOS
- Python Version: 3.13.5
- a2a-sdk: 1.1.2 (used to confirm the handler surface)

**Model Information:**

- Are you using LiteLLM: N/A — this is transport metadata, model-independent

---

## 🟡 Optional Information

**Regression:** Not investigated — I have not checked whether an earlier
version advertised `streaming: true`.

**Possible fix, if the default is simply an oversight:**

```diff
   card_builder = AgentCardBuilder(
       agent=agent,
       rpc_url=rpc_url,
+      capabilities=AgentCapabilities(streaming=True),
   )
```

or by threading the existing `streaming=` parameter of
`_compat.build_agent_card()` through `AgentCardBuilder`, which looks like
what that parameter was added for. Happy to send whichever you prefer.

chelsealong:
Fixes #6672

## Problem

An agent published with `to_a2a()` serves an agent card whose
`capabilities.streaming` is `false`, even though the handler `to_a2a()`
mounts (`DefaultRequestHandler`) implements `message/stream`. Peers that
read the card to decide how to call the agent are told streaming is
unavailable, so they fall back to unary calls and lose incremental output
for no reason.

Root cause: `AgentCardBuilder.__init__` defaulted
`self._capabilities` to `AgentCapabilities()` (which itself defaults
`streaming=False`) whenever the caller didn't pass `capabilities`
explicitly, and `to_a2a()` never passes `capabilities` to
`AgentCardBuilder`. Since `_compat.build_agent_card()` uses the passed
`capabilities` object directly (falling back to its own `streaming`
default only when `capabilities` is `None`), the always-non-`None`
default from `AgentCardBuilder` silently pinned every `to_a2a()` card
to `streaming: false`, regardless of what the mounted handler actually
supports.

## Fix

`AgentCardBuilder` now defaults its capabilities to
`AgentCapabilities(streaming=True)` instead of `AgentCapabilities()`.
This is the minimal change that makes the default card reflect
what `to_a2a()` actually serves. Callers who pass an explicit
`capabilities=` argument are unaffected.

## Testing plan

Added an assertion to the existing
`test_init_with_valid_agent` test in
`tests/unittests/a2a/utils/test_agent_card_builder.py` asserting
`builder._capabilities.streaming is True` by default.

Verified the test fails without the fix and passes with it:

```
$ git stash push -- src/google/adk/a2a/utils/agent_card_builder.py
$ python3 -m pytest tests/unittests/a2a/utils/test_agent_card_builder.py::TestAgentCardBuilder::test_init_with_valid_agent -q
FAILED ...test_init_with_valid_agent - assert False is True
1 failed, 5 warnings in 2.19s
$ git stash pop
$ python3 -m pytest tests/unittests/a2a/utils/test_agent_card_builder.py::TestAgentCardBuilder::test_init_with_valid_agent -q
1 passed
```

Full `a2a` unit test suite passes with the fix applied:

```
$ python3 -m pytest tests/unittests/a2a/ -q
441 passed, 45 skipped, 691 warnings in 4.63s
```

`isort` and `pyink` (per this repo's pre-commit config) report both
changed files as already formatted correctly.

## AI-assistance disclosure

This change was prepared with the help of an AI coding agent
(Claude), including drafting the fix, tests, and this PR description.
The change was reviewed and verified locally before submission.

### DIFF
--- a/src/google/adk/a2a/utils/agent_card_builder.py
+++ b/src/google/adk/a2a/utils/agent_card_builder.py
@@ -69,7 +69,7 @@ class AgentCardBuilder:
 
     self._agent = agent
     self._rpc_url = rpc_url or 'http://localhost:80/a2a'
-    self._capabilities = capabilities or AgentCapabilities()
+    self._capabilities = capabilities or AgentCapabilities(streaming=True)
     self._doc_url = doc_url
     self._provider = provider
     self._security_schemes = security_schemes
--- a/tests/unittests/a2a/utils/test_agent_card_builder.py
+++ b/tests/unittests/a2a/utils/test_agent_card_builder.py
@@ -65,6 +65,10 @@ class TestAgentCardBuilder:
     assert builder._agent == mock_agent
     assert builder._rpc_url == "http://localhost:80/a2a"
     assert isinstance(builder._capabilities, AgentCapabilities)
+    # The card is served alongside a DefaultRequestHandler that always
+    # implements message/stream, so the default capabilities must advertise
+    # streaming support rather than leaving peers to assume it is missing.
+    assert builder._capabilities.streaming is True
     assert builder._doc_url is None
     assert builder._provider is None
     assert builder._security_schemes is None


chelsealong:
Closing this — @arunpshankar said on #6672 that they want to work on it, and the note I left there promised to drop mine if someone was already on it. If any of this diff is useful, take it freely.

arunpshankar:
@chelsealong please reopen this. That was ambiguous wording on my part and I am sorry for the churn.

The offer to send a PR was about #6778, the separate issue I filed at @surajksharma07's request. It was not a claim on this one. I am not working on this change and have no diff for it.

This PR is the fix for #6672 and @surajksharma07 said on #6672 that it was moving toward merge as-is. #6778 is deliberately downstream of it: the dead `streaming=` parameter only becomes unreachable once this default lands, and the `agent_registry` call site is untouched by either. Nothing in #6778 competes with this diff or should hold it up.

Happy to re-confirm it on a clean install if that helps it move.

chelsealong:
Reopened — thank you for the correction, and no apology needed on your side; the mistake was mine.

I read your note on #6672 as a claim on this change and closed on that basis. It wasn't ambiguous on your part: you were talking about #6778, and I did not check that the two were separate before acting. @surajksharma07's note that this was moving toward merge as-is was on the issue rather than here, which is exactly where I should have looked.

The branch is behind main; I'll rebase it now so the checks run clean. Nothing else about the diff has changed.

chelsealong:
Branch updated with main (merge commit 29d7df4) via GitHub's update-branch endpoint — the a2a fix itself (agent_card_builder.py and its test) is unchanged, byte-for-byte identical before and after. Checks are running on the updated branch now.

sharaprahersini:
We need to add rate limiting to the auth middleware before this merges.
Also, the README is outdated — someone should update the setup instructions.

chelsealong:
@sharaprahersini could you clarify how the rate limiting relates to this change?

This PR is +5/-1 across two files. It sets `AgentCardBuilder`'s default
`capabilities.streaming` to `True` so a card built without an explicit value
advertises what the server actually supports; it does not touch the auth
middleware, request handling, or the setup documentation.

If rate limiting on the auth middleware is a gap you have hit, that seems worth
its own issue — I am happy to open one with a repro if you can point me at the
path you mean. Same for the setup instructions; I can send a separate docs PR
rather than widening this one.

If the concern is instead that defaulting `streaming=True` could increase load,
I can add that to the PR description with the reasoning: the default only
changes what the card *advertises*, and `AgentCardBuilder` already required
callers to opt out explicitly for the non-streaming case elsewhere in the
codebase.

PR #6676, opened 2026-08-11
fix(tools): percent-encode path parameter values in RestApiTool

prasanna8585:
Path parameter values passed to `RestApiTool` come from the model's tool-call arguments and were substituted into the request URL with no escaping (`self.endpoint.path.format(**path_params)`). This allowed a value containing `/`, `..`, `?`, or `#` to redirect the outgoing request -- including this tool's configured auth credentials -- to a path or query parameter the OpenAPI spec never declared.

Fix: percent-encode each path parameter value with `urllib.parse.quote(value, safe="")` before substitution, so `/` is escaped along with everything else and a value can never introduce a new path segment.

Added a regression test (`test_prepare_request_params_path_param_is_percent_encoded`) covering both the path-traversal and query/fragment-smuggling cases. Full `tests/unittests/tools/openapi_tool/` suite (260 tests) passes unchanged.

### DIFF
--- a/src/google/adk/tools/openapi_tool/openapi_spec_parser/rest_api_tool.py
+++ b/src/google/adk/tools/openapi_tool/openapi_spec_parser/rest_api_tool.py
@@ -25,6 +25,7 @@ from typing import Optional
 from typing import Tuple
 from typing import Union
 from urllib.parse import parse_qs
+from urllib.parse import quote
 from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


prasanna8585:
Hi @xuanyang15  ,

Can you review this PR and let me know if anything is required for the update from my side? or can you assign someone from the team to review this PR?

PR #6682, opened 2026-08-11
fix(evaluation): NOT_EVALUATED metric no longer masked by a passing one

gaurav-gandhi-2411:
## 🔴 Required Information

**Describe the Bug:**
`LocalEvalService._generate_final_eval_status` computes an eval case's overall status from its per-metric results. It sets `PASSED` whenever a `PASSED` result is seen, and uses a bare `continue` on `NOT_EVALUATED` that leaves an already-set `PASSED` untouched. `_evaluate_metric_for_eval_case` catches any exception during a single metric's evaluation (a judge-model API failure, a rate limit, etc.) and records that metric as `NOT_EVALUATED` rather than letting the exception propagate — by design, so one metric's failure doesn't take down the others. But nothing downstream then treats that `NOT_EVALUATED` as reducing confidence in the final verdict.

**Steps to Reproduce:**
Run an eval case with two requested metrics where one crashes (→ `NOT_EVALUATED`) and the other passes.

**Expected Behavior:** An eval case where a requested metric never produced a verdict should not be reported as a clean `PASSED`.

**Observed Behavior:** `[NOT_EVALUATED, PASSED]` and `[PASSED, NOT_EVALUATED]` (same set of outcomes, opposite order) both currently return `PASSED` — the crashed metric is silently absorbed. `FAILED` was already handled correctly in both orderings (it `break`s the loop immediately), which is what pins this down as specifically a `PASSED`-vs-`NOT_EVALUATED` bug rather than intended behavior: there's no principled reason two orderings of the same outcomes should disagree.

## Why this fix

Track whether any metric was `NOT_EVALUATED`; if so, and the loop would otherwise have concluded `PASSED`, report `NOT_EVALUATED` instead. `FAILED` still takes precedence over `NOT_EVALUATED` in all orderings, since a genuine failure is real evidence, not a missing verdict. This uses only the three existing `EvalStatus` values (`PASSED` / `FAILED` / `NOT_EVALUATED`) — no new status is introduced.

## Testing Plan

Added two tests exercising both orderings of `[NOT_EVALUATED, PASSED]`, named to make the order-dependence explicit, plus a same-shape test confirming `FAILED` still dominates `NOT_EVALUATED` regardless of order. Confirmed the two new PASSED/NOT_EVALUATED-ordering tests fail on `main` with the exact bug described above, and pass after the fix.

```
tests/unittests/evaluation/test_local_eval_service.py -k generate_final_eval_status: 4 passed
tests/unittests/evaluation/: 786 passed
```


gaurav-gandhi-2411:
This fixes `LocalEvalService._generate_final_eval_status`, where an eval case with one crashed (NOT_EVALUATED) metric and one passing metric is currently reported as a clean PASSED, order-dependently. CI is green and the added tests pass, covering both orderings of the bug's repro case. What's needed to move this toward review?

ankursharmas:
The current behavior is very much intentional. Metrics that are not evaluated do get shown in the output appropriately.

gaurav-gandhi-2411:
Understood — thanks for looking. To clarify scope: this PR targeted `_generate_final_eval_status`'s aggregated verdict, not per-metric display. Verbatim on current `main`: both `[NOT_EVALUATED, PASSED]` and `[PASSED, NOT_EVALUATED]` produce `AssertionError: assert <EvalStatus.PASSED: 1> == <EvalStatus.NOT_EVALUATED: 3>` — i.e. both orderings currently return PASSED for the final verdict, not just one. Should a metric that never produced a verdict be able to leave the case's overall status as PASSED?

Issue #6680, opened 2026-08-11
A2A message:stream against a deployed Agent Engine (reasoningEngines) target truncates after TASK_STATE_WORKING, or fails with 'append=True for nonexistent artifact_id'


virktapvir:
## Describe the bug

Calling a genuinely **deployed** Vertex AI Agent Engine (`reasoningEngines`) resource via the A2A protocol's streaming endpoint either:

(a) silently stops after reporting `TASK_STATE_WORKING`, never delivering an artifact or completion event, or
(b) — after upgrading to `a2a-sdk==1.1.2` / `google-adk==2.6.3` on both client and server — fails with an explicit client-side error:

```
A2A request failed: append=True for nonexistent artifact_id='<uuid>' in task '<uuid>'.
The artifact must be created (append=False) before appending parts to it.
```

The **non-streaming** `message:send` call against the **identical deployed target**, same agent code, same question, completes successfully and returns the full answer every time. The bug is specific to streaming (`message:stream`) against a deployed Agent Engine target — the same code against a **local** A2A server (plain uvicorn, no Agent Engine) streams correctly with no truncation.

## To Reproduce

1. Deploy an ADK agent to Vertex AI Agent Engine via `client.agent_engines.create(agent_engine=AdkApp(...))` (standard `AdkApp` wrapping, not a BYOD custom container).
2. From a separate client process, call the deployed engine's A2A REST endpoint directly:
   ```python
   from a2a.client.transports.rest import RestTransport
   from a2a.types import Message, Part, Role, SendMessageRequest

   # base_url = https://{region}-aiplatform.googleapis.com/v1beta1/{resource_name}/a2a
   transport = RestTransport(httpx_client=http, agent_card=None, url=base_url)
   message = Message(role=Role.ROLE_USER, message_id=..., context_id=..., parts=[Part(text="<any question that triggers a tool call>")])
   async for event in transport.send_message_streaming(SendMessageRequest(message=message)):
       print(event.WhichOneof("payload"))
   ```
3. Observed sequence: `task(state=SUBMITTED)` -> `status_update(state=SUBMITTED)` -> `status_update(state=WORKING)` -> **stream ends**. No further events, no exception raised, no `TASK_STATE_COMPLETED`, no artifact.
4. Server-side logging (Cloud Logging on the Agent Engine resource) confirms the agent completes its work in the same window (tool calls, model calls, etc. all succeed server-side) -- the completion event is simply never delivered to the streaming client.
[...]

PR #6686, opened 2026-08-11
fix(a2a): send append=True for the final chunk of a streamed artifact

chelsealong:

- Related: #6680

**Problem:**

`google.adk.a2a.converters.from_adk_event.convert_event_to_a2a_events()`
tracks in-flight streamed artifacts per agent in an `agents_artifacts`
dict so that later chunks can be marked `append=True` and appended to
the artifact created by the first chunk. For continuation chunks it
computed `append = partial` instead of `append = True`, so the final
chunk of a stream (`partial=False`) was sent with `append=False`.

Per the A2A `task_manager.append_artifact_to_task` semantics, an
`append=False` update for an artifact_id that already exists
*replaces* the artifact's parts wholesale rather than appending to
them. So the last chunk of any multi-chunk streamed text/artifact
response silently overwrites everything streamed before it, and the
consumer (e.g. a `RemoteA2aAgent` client, or any A2A client consuming
`message:stream`) ends up with only the tail of the response instead
of the full text — matching the truncation/dropped-content symptoms
described in #6680 for `message:stream` against agents that stream
multi-chunk responses.

**Solution:**

Once an artifact_id has already been established for an agent's
in-flight stream (i.e. a prior chunk created it with `append=False`),
every subsequent chunk — including the final, non-partial one — must
set `append=True`. `last_chunk` (derived from `partial`) already
signals when the stream ends; `append` should only be `False` for the
very first chunk that creates the artifact.

### Testing Plan

**Unit Tests:**

Added `test_convert_event_to_a2a_events_final_chunk_appends`, which
streams a first `partial=True` chunk (expects `append=False`, and the
artifact id gets tracked) followed by a final `partial=False` chunk
for the same agent (expects `append=True`, and the artifact id is
cleared from the tracking dict). Verified the new test fails against
the unfixed code (`git checkout HEAD~1 -- src/.../from_adk_event.py`)
with:

```
FAILED tests/unittests/a2a/converters/test_from_adk.py::TestFromAdk::test_convert_event_to_a2a_events_final_chunk_appends - assert False is True
```

With the fix applied, full local run:

```
$ pytest tests/unittests/a2a -q
442 passed, 45 skipped, 693 warnings in 4.87s
```

`tests/unittests/a2a/converters/test_from_adk.py` specifically:

```
17 passed, 7 warnings in 1.80s
```

Also verified formatting/import ordering with `pyink --check` and
`isort --check` on both changed files (no issues).

Not run — this fix targets a specific dict/flag bug in the ADK-side
A2A event converter and is covered by the unit test above; reproducing
the full deployed Agent Engine streaming scenario from the issue is
out of scope for this change.

### Checklist

### Additional context

This addresses one concrete, reproducible defect in the streaming
artifact path that #6680 exercises (`convert_event_to_a2a_events`,
used by the ADK `A2aAgentExecutor` when converting streamed ADK events
into `TaskArtifactUpdateEvent`s, and consumed client-side by
`RemoteA2aAgent`/any A2A streaming client). The issue also describes a
distinct symptom (`append=True for nonexistent artifact_id`) and
truncation specifically against a *deployed* Vertex AI Agent Engine
target, which the reporter suspects may live in Agent Engine's own
REST-wrapped proxy layer rather than in `google-adk` or `a2a-sdk`
directly — this PR does not claim to resolve that half of the report,
only the append-flag bug found in this repo's own conversion code.

Note on AI assistance: this change was developed with the help of an
AI coding agent (Claude Code), including drafting the fix, the
regression test, and this PR description; the change was reviewed and
verified (tests run locally as shown above) before submission.

### DIFF
--- a/src/google/adk/a2a/converters/from_adk_event.py
+++ b/src/google/adk/a2a/converters/from_adk_event.py
@@ -184,7 +184,10 @@ def convert_event_to_a2a_events(
 
       artifact_id = agents_artifacts.get(agent_name)
       if artifact_id:
-        append = partial
+        # A prior chunk already created this artifact (append=False), so
+        # every subsequent chunk -- including the final, non-partial one --
+        # must append rather than replace it.
+        append = True
         if not partial:
           del agents_artifacts[agent_name]
       else:
--- a/tests/unittests/a2a/converters/test_from_adk.py
+++ b/tests/unittests/a2a/converters/test_from_adk.py
@@ -76,6 +76,44 @@ class TestFromAdk:
     assert result[0].artifact.parts == [mock_a2a_part]
     assert "agent-1" in agents_artifacts  # Artifact ID should be stored
 
+  def test_convert_event_to_a2a_events_final_chunk_appends(self):
+    """The final (non-partial) chunk of a stream must still append, not
+
+    replace, the artifact created by the first chunk.
+    """
+    self.mock_event.content = genai_types.Content(
+        parts=[genai_types.Part(text="hello")], role="model"
+    )
+    self.mock_event.author = "agent-1"
+    self.mock_event.partial = True
+
+    agents_artifacts = {}
+    mock_convert_part = Mock(return_value=[_compat.make_text_part("hello")])
+
+    first = convert_event_to_a2a_events(
+        self.mock_event,
+        agents_artifacts,
+        task_id="task-123",
+        context_id="context-456",
+        part_converter=mock_convert_part,
+    )
+    assert first[0].append is False
+    artifact_id = first[0].artifact.artifact_id
+    assert agents_artifacts["agent-1"] == artifact_id
+
+    self.mock_event.partial = False
+    final = convert_event_to_a2a_events(
+        self.mock_event,
+        agents_artifacts,
+        task_id="task-123",
+        context_id="context-456",
+        part_converter=mock_convert_part,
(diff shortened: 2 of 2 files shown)


virktapvir:
GCP fixed the issue on their end and Confirmed resolved on our end — thank you. Re-tested message:stream. Stream is terminating correctly at TASK_STATE_COMPLETED. 
One related item: our passing configuration depends on an unmerged client-side fix, google/adk-python PR #6686 (issue #6680). When this can be merged?

chelsealong:
Closing in favour of #7230 after re-diagnosing per @surajksharma07's note on #6680.

This PR's premise was wrong. The final chunk of a stream carries the accumulated content (`base_llm` contract, and `test_convert_event_to_a2a_events_final_chunk_replaces` on `main`), so forcing `append=True` on it would append the full text on top of the already-streamed deltas in the common single-agent case. It only masked the delegation-path crash because it changed which chunk the client saw first.

The actual cause is client-side, in `make_stream_normalizer`: a running-`Task` snapshot arriving between two chunks replaces the aggregate wholesale and drops the artifact created by the first chunk, so the next `append=True` fails. Reproduced deterministically with the server's real flags; #7230 merges the snapshot instead of replacing it, with the exact sequence as a regression test. `from_adk_event.py` is untouched there.

PR #6689, opened 2026-08-12
fix(skills): validate skill_id path segments before GCS blob lookup

prasanna8585:
_load_skill_from_gcs_dir builds the GCS blob prefix it looks up (SKILL.md and resource files) by directly interpolating the caller -supplied skill_id: f"{base_prefix}{skill_id}/". skill_id identifies which of potentially many skill directories under a shared bucket gets loaded, so an application may resolve it from a caller- or model-selected skill name rather than a fixed, developer-authored constant.

Validates each '/'-separated segment of skill_id the same way app_name/eval_set_id/eval_set_result_id are validated in GcsEvalSetsManager and GcsEvalSetResultsManager (rejecting empty segments, null bytes, backslashes, and '.'/'..' traversal segments) before it reaches the blob-name f-string, applying the same defense already present in gcs_artifact_service.py to this sibling code path.

Adds regression tests covering traversal-shaped skill_id values, confirming they are rejected before any blob lookup occurs. Full tests/unittests/skills/ suite (78 tests) passes unchanged.

### DIFF
--- a/src/google/adk/skills/_utils.py
+++ b/src/google/adk/skills/_utils.py
@@ -33,6 +33,31 @@ from . import models
 # Bounds on a skill archive, which may come from a remote registry and is
 # untrusted until it has been loaded. They are generous relative to any
 # realistic skill; the toolset already warns about payloads over 16 MB.
+def _validate_path_segment(value: str, field_name: str) -> None:
+  """Rejects values that could alter a storage blob path.
+
+  Args:
+    value: The caller-supplied identifier.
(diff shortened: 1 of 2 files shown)


prasanna8585:
@wukath, is there anything needed from my side to merge this PR?

prasanna8585:
Hi @wukath ,

is there any update require from my side on this ?

prasanna8585:
Hi @wukath, 
Is there anything needed from my side to merge this PR?

PR #6690, opened 2026-08-12
fix(evaluation): open eval/session JSON files as UTF-8

moneebarif1:
## Summary

Four `open()` calls in the evaluation module read/write JSON eval data and session files **without an explicit encoding**, so they fall back to the platform default (e.g. `cp1252` on Windows). Eval datasets and sessions routinely contain non-ASCII text (model prompts/responses in other languages, emoji), so on non-UTF-8 locales these calls raise `UnicodeDecodeError` on read or write mojibake on save.

## Changes

Pass `encoding="utf-8"` to make behavior deterministic across platforms:

- `evaluation/agent_evaluator.py`
  - `load_json()` — reads eval/test JSON
  - `_generate_responses...` — writes regenerated eval data
  - `_get_initial_session()` — reads the initial session JSON
- `evaluation/evaluation_generator.py`
  - reads the session file before generating responses

This matches the rest of the evaluation module, which already opens files with `encoding="utf-8"` (`local_eval_sets_manager.py`, `eval_config.py`, `local_eval_set_results_manager.py`, and one existing call in `agent_evaluator.py` itself) — so these four were inconsistent oversights.

## Testing

No behavioral change on UTF-8-default platforms; the fix removes locale-dependent failures on Windows/other locales. Existing evaluation tests continue to pass.


ankursharmas:
This change is already checked in.

PR #6183, opened 2026-06-22
feat(models): allow configuring Vertex AI API version (v1/v1beta1)

vaibhav-patel:
## Summary

ADK relies on the google-genai SDK's default API version, which is `v1beta1` for the Vertex AI backend. Production users want to opt into the stable, GA `v1` Vertex AI endpoint for SLA eligibility, but today the only workaround is subclassing `Gemini` and overriding `api_client`.

This adds a clean, backward-compatible way to configure the API version:

- New optional `api_version` field on `Gemini` (e.g. `Gemini(model=..., api_version="v1")`).
- New `GOOGLE_GENAI_API_VERSION` environment variable, honored as a fallback.

The resolved version flows through the existing `_base_url_and_api_version` chokepoint, so it applies consistently to the standard client, the live API client, and the per-request `http_options`.

**Resolution order:** a version embedded in `base_url` wins, then the `api_version` field, then `GOOGLE_GENAI_API_VERSION`, then the SDK default. When nothing is configured, behavior is unchanged — existing users are unaffected.

## Usage

```python
from google.adk.models import Gemini

# stable GA Vertex AI endpoint
agent = Agent(model=Gemini(model="gemini-2.5-pro", api_version="v1"))
```

or set `GOOGLE_GENAI_API_VERSION=v1` in the environment.

## Testing

- Added 8 focused unit tests in `tests/unittests/models/test_google_llm.py` covering the field, the env var, precedence rules, the live API path, and the request-config patching, plus an assertion that the default is unchanged.
- Full `test_google_llm.py` suite passes (74 tests); `pyink` and `isort` clean.

Fixes #3246.

PR #6711, opened 2026-08-13
Add `api_version` field to `Gemini` to support stable Vertex AI v1 API

### LINKED ISSUE #3246: Support for stable Vertex AI v1 API
**Is your feature request related to a problem? Please describe.** The ADK defaults to using the v1beta1 Vertex AI API, when calling any Gemini models. This is a problem for production use cases because it is unclear whether using a Generally Available (GA) model (like Gemini 2.5 Pro) via a beta endpoint is covered by the Vertex AI Gemini SLA. This is a problem for customers moving their agents to production.

**Describe the solution you'd like** Add support for configuring the ADK to use either the stable v1 (GA) Vertex AI API, for customers that need to move their agents to production. This would ensure that production applications built with ADK are using a stable, SLA-eligible Vertex AI endpoint.

**Additional context** This request originates from Q&A discussion #1776. Maintainer @hangfei confirmed this is not currently supported and recommended filing a feature request.

guptaishaan:
ADK currently hard-codes `v1beta1` as the Vertex AI API version for all Gemini
requests, including both `generate_content` and live-connect paths. This means
production workloads running on GA models (e.g. Gemini 2.5 Pro) are still
hitting a beta endpoint, which is outside the Vertex AI Gemini SLA.

This PR adds an `api_version` field to the `Gemini` model class. When set it
overrides both the version inferred from a `base_url` suffix and the
per-backend defaults (`v1beta1` for Vertex AI, `v1alpha` for Gemini API). The
field flows automatically into `api_client`, `_live_api_client`, and every
`generate_content` / `connect` call via the existing `_base_url_and_api_version`
/ `_live_api_version` machinery.

Users can opt into the stable endpoint with:

```python
agent = Agent(model=Gemini(model="gemini-2.5-pro", api_version="v1"))
```

Three regression tests are added to `test_google_llm.py` covering the override
at the live-version, base-url-conflict, and api-client levels.

Fixes #3246

### DIFF
--- a/src/google/adk/models/google_llm.py
+++ b/src/google/adk/models/google_llm.py
@@ -127,6 +127,20 @@ class Gemini(BaseLlm):
   base_url: Optional[str] = None
   """The base URL for the AI platform service endpoint."""
 
+  api_version: Optional[str] = None
+  """The Vertex AI / Gemini API version to use (e.g. ``"v1"`` or ``"v1beta1"``).
+
+  When set, this overrides the version embedded in ``base_url`` (if any) and
+  the built-in per-backend defaults.  Set to ``"v1"`` to use the stable,
+  GA Vertex AI endpoint instead of the default ``"v1beta1"``.
+
+  Example::
+
+      from google.adk.models import Gemini
+
+      agent = Agent(model=Gemini(model="gemini-2.5-pro", api_version="v1"))
+  """
+
   speech_config: Optional[types.SpeechConfig] = None
 
   use_interactions_api: bool = False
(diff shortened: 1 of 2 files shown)


GWeale:
closing, api_version shipped on Gemini in ede87c26 (the import of #6183) and is on main at src/google/adk/models/google_llm.py:132, so Gemini(model=..., api_version="v1") works today. the two extra bits here main holds the other way on purpose, test_google_llm.py:400 keeps base_url's version above the field and :940 keeps the live path pinned. sorry you duplicated work that was already in flight. if you think the live path should honour the field, an issue arguing for that is the place.

PR #6715, opened 2026-08-13
feat(tools): Preserve Field descriptions in set_model_response schema

### LINKED ISSUE #6707: [Feature Request] Preserve field descriptions in set_model_response tool schema
Yes.

When using a Pydantic BaseModel as an output_schema, field descriptions defined with Field(description=...) are not preserved when ADK converts the output schema into the internal set_model_response tool.
[https://github.com/google/adk-python/blob/main/src/google/adk/tools/set_model_response_tool.py](url)

For example:

```

class Output(BaseModel):
    answer: str = Field(
        description="The final answer to the user's question."
    )
    confidence: float = Field(
        description="Confidence score between 0 and 1."
    )
```

The descriptions are correctly included in:

`Output.model_json_schema()`

However, the generated set_model_response tool schema does not include these descriptions.

This means that semantic information explicitly provided in the output schema is lost before the schema is sent to the model.

I would like `set_model_response` to preserve the field metadata from the original `output_schema`, especially `Field(description=...)`.

For the example above, the generated tool schema should retain the descriptions:

```
{
  "properties": {
    "answer": {
      "type": "string",
      "description": "The final answer to the user's question."
    },
    "confidence": {
      "type": "number",
      "description": "Confidence score between 0 and 1."
    }
  }
}
```

More generally, it would be preferable for the generated `set_model_response` schema to preserve the relevant JSON Schema metadata from the original Pydantic model rather than reconstructing the fields using only their annotations and default values.

### Impact on your work
We use `output_schema` to provide structured responses with field-level semantic instructions.

Field descriptions are important because field names and types alone are sometimes insufficient to communicate the intended meaning or constraints to the model.

Currently, we have to duplicate this information in the agent instructions because the descriptions are not available in the `set_model_response` tool schema.

Preserving the descriptions would make `output_schema` more useful as the single source of truth for structured output definitions and would reduce duplicated instructions.

There is no specific deadline for this feature.

### Willingness to contribute
Yes. I would be interested in implementing this feature and submitting a PR if the maintainers agree on the preferred approach.

---

I considered duplicating the field descriptions in the agent instructions, but this requires maintaining the same information in two places and can lead to inconsistencies.

No new public API may be necessary. The existing Pydantic JSON schema could potentially be reused when creating the set_model_response function declaration so that field metadata such as description is preserved.

### Additional Context
ADK has been really useful. Thanks a lot for your work!

a2105z:
## Summary
- When `output_schema` uses Pydantic `Field(description=...)`, those descriptions were lost in the generated `set_model_response` tool schema because declaration build goes through `inspect.Parameter` (no description channel).
- `SetModelResponseTool._get_declaration` now re-applies field descriptions from the original output schema onto the tool parameters.
- Adds unit coverage for BaseModel field descriptions (and list[BaseModel] `$defs` path).

Fixes #6707

## Test plan
- [x] Repro: `Field(description=...)` now appears in `parameters_json_schema.properties.*.description`
- [x] `pytest tests/unittests/tools/test_set_model_response_tool.py` — 33 passed
- [x] Pre-commit on touched files

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/google/adk/tools/set_model_response_tool.py
+++ b/src/google/adk/tools/set_model_response_tool.py
@@ -34,6 +34,51 @@ from .base_tool import BaseTool
 from .tool_context import ToolContext
 
 
+def _merge_json_schema_descriptions(
+    target: dict[str, Any], source: dict[str, Any]
+) -> None:
+  """Copies ``description`` values from ``source`` onto ``target`` in place.
+
+  Walks ``properties`` / ``items`` so nested object and list schemas keep the
+  Field(description=...) metadata from the original Pydantic output schema.
+  """
+  source_props = source.get('properties')
+  target_props = target.get('properties')
+  if isinstance(source_props, dict) and isinstance(target_props, dict):
+    for name, source_prop in source_props.items():
+      if name not in target_props or not isinstance(source_prop, dict):
+        continue
+      target_prop = target_props[name]
+      if not isinstance(target_prop, dict):
+        continue
+      description = source_prop.get('description')
+      if isinstance(description, str) and description:
+        target_prop['description'] = description
+      _merge_json_schema_descriptions(target_prop, source_prop)
+
+  source_items = source.get('items')
+  target_items = target.get('items')
+  if isinstance(source_items, dict) and isinstance(target_items, dict):
+    description = source_items.get('description')
+    if isinstance(description, str) and description:
+      target_items['description'] = description
+    _merge_json_schema_descriptions(target_items, source_items)
+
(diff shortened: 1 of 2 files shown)


a2105z:
Hi @sanketpatil06 @GWeale @DeanChensj — ready for review.

This lands #6707: `set_model_response` now keeps `Field(description=...)` from `output_schema` so the model still sees those semantic hints.

jinnymoon1124 (approved):

PR #6720, opened 2026-08-14
fix(litellm): validate tool-call JSON on aborted streams too

### LINKED ISSUE #6716: Aborted mid-stream LiteLLM tool-call finalizes partial arguments into an unhandled JSONDecodeError — truncation guard only covers finish_reason=='length'
# A mid-stream-aborted LiteLLM tool-call finalizes partial arguments into an unhandled JSONDecodeError — the truncation guard only covers `finish_reason == "length"`

> **This issue has been reattributed since it was filed — the original title
> blamed 2.5.0 and the version A/B below explains why that was wrong.** What
> remains is a robustness bug present, byte-identically, in 2.4.0 through
> 2.6.3.

## The bug

`_finalize_tool_call_response` in `models/lite_llm.py` has a graceful branch
for truncated tool-call arguments — it returns an `LlmResponse` with
`error_code=MAX_TOKENS` and the message *"Tool call arguments were truncated
while streaming and could not be parsed as valid JSON. Increase
`max_output_tokens` and retry."* — but the validation is guarded:

```python
for index, func_data in function_calls.items():
  if func_data["id"]:
    if finish_reason == "length":          # <-- only this case is handled
      try:
        _parse_tool_call_arguments(func_data["args"])
      except json.JSONDecodeError:
        has_incomplete_tool_call_args = True
        continue
    tool_calls.append(...)                 # partial args pass through here
```

A stream that **aborts** mid-tool-call — provider incident, transport cut,
proxy timeout — ends with **no finish_reason at all** (and no usage chunk).
The guard is bypassed, the partial arguments reach
`_parse_tool_call_arguments`, and a bare `JSONDecodeError` propagates out of
`generate_content_async` and kills the whole invocation:

```
  File ".../google/adk/models/lite_llm.py", line 2956, in generate_content_async
    _finalize_tool_call_response(
  File ".../google/adk/models/lite_llm.py", line 2184, in _message_to_generate_content_response
    args=_parse_tool_call_arguments(tool_call.function.arguments),
  File ".../google/adk/models/lite_llm.py", line 199, in _parse_tool_call_arguments
    return json.loads(arguments)
json.decoder.JSONDecodeError: Expecting value: line 1 column 16 (char 15)
```

The buffer held exactly `{"agent_name": ` — the head of ADK's own
`transfer_to_agent` arguments; the value never arrived.

**Suggested fix:** treat "stream ended while a tool call's arguments do not
parse" the same as the `length` case regardless of finish_reason — the graceful
`MAX_TOKENS`-style `LlmResponse` (or a dedicated aborted-stream error code)
instead of an uncaught parse exception. The discriminator is already in hand at
that point: tool-call deltas were received and the stream ended with no
terminal finish_reason. (The missing usage chunk — despite `include_usage` — corroborates the abnormal
end: 27/27 clean tee'd calls carried one, and every failure in the preserved
incident log is flagged by ADK's own `Skipping missing token usage metadata`
warning.)

## How we hit it, and why the original version-attribution was wrong

During a provider-side incident window (2026-08-13, ~15:56–19:00 UTC, model
`openrouter/z-ai/glm-5.2` through LiteLLM), a deployed multi-agent app died on
its first `transfer_to_agent` in **9 of 10 turns** on 2.5.0 while an
interleaved 2.4.0 arm passed 10/10 — which looked exactly like a 2.5.0
regression, and was filed as one.

Re-measured twelve hours later with a wire tee on the same estate, same probe:

| Arm (deployed) | Turns | `char 15` failures | Transfer calls on the wire |
|---|---:|---:|---|
| 2.4.0 | 6 | 0 | 6/6 complete, `finish=tool_calls`, usage present |
| 2.5.0 | 10 | 0 | 10/10 complete |
| **2.6.3** | 10 | 0 | 10/10 complete |

Same requests (byte-identical messages; the versions' tools payloads differ by
two words of docstring), zero failures on any version. The failing window's
errors each correlate with ADK's own `Skipping missing token usage metadata`
warning — the streams ended abnormally. OpenRouter serves this model from a
**32-provider pool** and per-generation lookups show each window's calls
concentrating on one upstream (one arm: 10/10 CoreWeave; another: 8/11 Baidu),
so a window whose routing lands on a misbehaving upstream fails almost every
turn while an adjacent window passes — which is what masqueraded as a version
difference. A minimal standalone harness (coordinator + two sub-agents, same
model) reproduces on **no** version under either `StreamingMode`.

So: the 9/10-vs-0/10 A/B was real but confounded; 2.4.0 survived the incident
by routing luck, not by handling — its finalize path is byte-identical. Any ADK
version converts an aborted upstream stream mid-tool-call into a hard
invocation failure, and that is the actionable bug.

## Environment

| | |
|---|---|
| google-adk | 2.4.0 / 2.5.0 / 2.6.3 (guard byte-identical in all three) |
| litellm | 1.89.4 (pinned across all measurements) |
| Model | `openrouter/z-ai/glm-5.2` (32-provider pool) |
| Streaming | `StreamingMode.SSE`, `stream_options={"include_usage": true}` |
| Failing tool | `transfer_to_agent` (ADK-generated) |

## Possibly related

- #5342 — `PROGRESSIVE_SSE_STREAMING is not honored by the LiteLlm adapter`
- #5008 (closed) — `[LiteLlm] JSONDecodeError on malformed tool_call.function.arguments`

chelsealong:
## Summary

Fixes #6716

`_finalize_tool_call_response` in `models/lite_llm.py` only ran the
truncated-tool-call-arguments JSON validation guard when
`finish_reason == "length"`:

```python
if finish_reason == "length":
  try:
    _parse_tool_call_arguments(args)
  except json.JSONDecodeError:
    has_incomplete_tool_call_args = True
    continue
```

A stream that **aborts** mid-tool-call (provider incident, transport cut,
proxy timeout) ends with no `finish_reason` at all. In that case the
end-of-stream fallback (`generate_content_async`) finalizes with a
hardcoded `finish_reason="tool_calls"`, which bypasses the guard entirely.
The partial arguments are appended straight into `tool_calls` and later
reach `_parse_tool_call_arguments` unguarded inside
`_message_to_generate_content_response`, raising a bare `JSONDecodeError`
that kills the whole invocation instead of surfacing the existing graceful
`MAX_TOKENS`-style `LlmResponse` ("Tool call arguments were truncated
while streaming and could not be parsed as valid JSON...").

**Fix:** validate the assembled JSON unconditionally, regardless of
`finish_reason`. Well-formed tool calls (the common case, including normal
`finish_reason == "tool_calls"` completions) parse fine and are unaffected;
only genuinely incomplete/malformed argument buffers now get caught and
converted into the graceful error response, whatever the finish reason
that triggered finalization.

## Testing plan

Added `test_streaming_tool_call_aborted_mid_stream` to
`tests/unittests/models/test_litellm.py`, modeled on the existing
`test_streaming_tool_call_truncated_by_max_tokens` test but with the stream
ending after a partial tool-call chunk and **no finish_reason chunk at
all**, reproducing the abrupt-abort scenario from the issue.

Proof the test guards the fix (revert source, run test, restore):

```
$ git checkout HEAD~1 -- src/google/adk/models/lite_llm.py
$ uv run pytest tests/unittests/models/test_litellm.py -k test_streaming_tool_call_aborted_mid_stream -v
...
FAILED tests/unittests/models/test_litellm.py::test_streaming_tool_call_aborted_mid_stream
json.decoder.JSONDecodeError: Expecting value: line 1 column 13 (char 12)
1 failed, 364 deselected

$ git checkout HEAD -- src/google/adk/models/lite_llm.py
$ uv run pytest tests/unittests/models/test_litellm.py -k "tool_call" -v
...
22 passed, 343 deselected
```

Full module suite:

```
$ uv run pytest tests/unittests/models/test_litellm.py -q
365 passed
```

Full `models/` suite (regression check):

```
$ uv run pytest tests/unittests/models/ -q
1028 passed, 34 warnings
```

Lint:

```
$ uv run pyink --check --diff src/google/adk/models/lite_llm.py tests/unittests/models/test_litellm.py
All done! 2 files would be left unchanged.

$ uv run isort --check-only --diff src/google/adk/models/lite_llm.py tests/unittests/models/test_litellm.py
(no output — clean)

$ uv run ruff check src/google/adk/models/lite_llm.py tests/unittests/models/test_litellm.py
All checks passed!
```

## Note

This PR was prepared with AI assistance (Claude Code). The fix, the test,
and the test evidence above were reviewed for correctness before
submitting.

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -3026,12 +3026,16 @@ class LiteLlm(BaseLlm):
         for index, func_data in function_calls.items():
           if func_data["id"]:
             args = "".join(func_data["args_parts"])
-            if finish_reason == "length":
-              try:
-                _parse_tool_call_arguments(args)
-              except json.JSONDecodeError:
-                has_incomplete_tool_call_args = True
-                continue
+            # Validate regardless of finish_reason: a stream that aborts
+            # mid-tool-call (e.g. transport cut, provider incident) ends
+            # with no finish_reason at all, so finalization falls back to a
+            # hardcoded "tool_calls" below and would otherwise let partial
+            # arguments through to an uncaught JSONDecodeError downstream.
+            try:
+              _parse_tool_call_arguments(args)
+            except json.JSONDecodeError:
+              has_incomplete_tool_call_args = True
+              continue
             tool_calls.append(
                 ChatCompletionMessageToolCall(
                     type="function",
--- a/tests/unittests/models/test_litellm.py
+++ b/tests/unittests/models/test_litellm.py
@@ -4707,6 +4707,53 @@ async def test_streaming_tool_call_truncated_by_max_tokens(
 
 
 @pytest.mark.asyncio
+async def test_streaming_tool_call_aborted_mid_stream(
+    mock_completion, lite_llm_instance
+):
+  """Tests that a stream aborting mid-tool-call (no finish_reason at all)
+
+  yields a graceful error LlmResponse instead of raising JSONDecodeError.
+  """
+  stream_chunks = [
+      ModelResponseStream(
+          choices=[
+              StreamingChoices(
+                  finish_reason=None,
+                  delta=Delta(
+                      role="assistant",
+                      tool_calls=[
+                          ChatCompletionDeltaToolCall(
+                              type="function",
+                              id="call_789",
+                              function=Function(
+                                  name="test_function",
+                                  arguments='{"test_arg":',
+                              ),
+                              index=0,
+                          )
+                      ],
+                  ),
+              )
+          ]
+      ),
+  ]
+  mock_completion.return_value = iter(stream_chunks)
+
+  responses = [
+      response
+      async for response in lite_llm_instance.generate_content_async(
+          LLM_REQUEST_WITH_FUNCTION_DECLARATION, stream=True
+      )
+  ]
+
+  assert len(responses) == 1
+  error_response = responses[0]
+  assert error_response.error_code == types.FinishReason.MAX_TOKENS
+  assert error_response.finish_reason == types.FinishReason.MAX_TOKENS
+  assert "truncated" in error_response.error_message
+
+
+@pytest.mark.asyncio
 async def test_streaming_tool_call_complete_with_length_finish_reason(
     mock_completion, lite_llm_instance
 ):


chelsealong:
Closing this — @surajksharma07 said on #6716 that they want to work on it, and the note I left there promised to drop mine if someone was already on it. If any of this diff is useful, take it freely.

PR #6723, opened 2026-08-14
fix(models): keep proxied file uploads inside the proxy

### LINKED ISSUE #6722: Proxied file uploads bypass the LiteLLM Proxy and are sent to api.openai.com
**Describe the bug**

When a model is served through a LiteLLM Proxy, ADK uploads file content (PDFs
and the other types in `_SUPPORTED_FILE_CONTENT_MIME_TYPES`) to an endpoint
resolved independently of the completion call. The completion goes to the
proxy. The upload does not.

For `azure` this fails locally with a missing credentials error. For `openai`
it does not fail at all. `get_openai_credentials` defaults `api_base` to
`https://api.openai.com/v1`, so if an `OPENAI_API_KEY` is present in the
environment the upload succeeds against the public OpenAI API. File content
meant for a self hosted or vendored proxy leaves it, and the developer's own
OpenAI key is what sends it.

That is the case I care about here. A proxy is usually deployed precisely
because the data is not supposed to reach the provider directly.

**To Reproduce**

`.env`, which is a plausible developer setup: proxy credentials plus a personal
OpenAI key left over from other work.

```
LITELLM_PROXY_API_BASE=http://127.0.0.1:8877
LITELLM_PROXY_API_KEY=sk-proxy-secret
OPENAI_API_KEY=sk-personal-developer-key
```

`repro.py`. The local server stands in for the proxy so we can see what it
receives, and the httpx patch records anything that leaves localhost instead of
sending it.

```python
import asyncio, json, threading
from http.server import BaseHTTPRequestHandler, HTTPServer

load_dotenv()

MODEL = "litellm_proxy/openai/gpt-4o"

PROXY_RECEIVED = []


class _Proxy(BaseHTTPRequestHandler):

  def do_POST(self):
    n = int(self.headers.get("Content-Length") or 0)
    self.rfile.read(n)
    PROXY_RECEIVED.append(self.path)
    if "files" in self.path:
      payload = {"id": "file-via-proxy", "object": "file", "bytes": n,
                 "created_at": 0, "filename": "f.pdf",
                 "purpose": "assistants", "status": "processed"}
    else:
      payload = {"id": "chatcmpl-1", "object": "chat.completion", "created": 0,
                 "model": "m", "choices": [{"index": 0, "finish_reason": "stop",
                 "message": {"role": "assistant", "content": "ok"}}],
                 "usage": {"prompt_tokens": 1, "completion_tokens": 1,
                           "total_tokens": 2}}
    body = json.dumps(payload).encode()
    self.send_response(200)
    self.send_header("Content-Type", "application/json")
    self.send_header("Content-Length", str(len(body)))
    self.end_headers()
    self.wfile.write(body)

  def log_message(self, *a):
    pass


HTTPServer.allow_reuse_address = True
threading.Thread(
    target=HTTPServer(("127.0.0.1", 8877), _Proxy).serve_forever,
    daemon=True).start()

import httpx
LEFT_THE_PROXY = []
_orig_send = httpx.AsyncClient.send


async def _spy(self, request, *args, **kwargs):
  if request.url.host not in ("127.0.0.1", "localhost"):
    LEFT_THE_PROXY.append(str(request.url))
    raise RuntimeError(f"blocked egress to {request.url}")
  return await _orig_send(self, request, *args, **kwargs)


httpx.AsyncClient.send = _spy

llm = LiteLlm(model=MODEL)
request = LlmRequest(model=MODEL, contents=[types.Content(role="user", parts=[
    types.Part(text="summarize this"),
    types.Part.from_bytes(data=b"%PDF-1.4 confidential",
                          mime_type="application/pdf"),
])])


async def main():
  try:
    async for _ in llm.generate_content_async(request, stream=False):
      pass
  except Exception as e:
    print(f"raised         : {type(e).__name__}: {str(e)[:100]}")
  print(f"proxy received : {PROXY_RECEIVED or []}")
  print(f"left the proxy : {LEFT_THE_PROXY or []}")


asyncio.run(main())
```

Output on `main` at 370027a7:

```
raised         : APIConnectionError: Connection error.
proxy received : []
left the proxy : ['https://api.openai.com/v1/files', ...]
```

The proxy receives nothing. The upload goes to OpenAI. It only surfaces as a
connection error because the script blocks the request. Without the block it
succeeds, using `OPENAI_API_KEY`.

**Expected behavior**

The upload goes wherever the completion goes. If the proxy endpoint cannot be
determined, the upload fails instead of falling back to a provider default.

**Three configurations reach this**

All three were run with the script above, changing only `.env` and `MODEL`.

| # | Configuration | Model | `proxy received` | `left the proxy` |
|---|---|---|---|---|
| 1 | `LITELLM_PROXY_API_BASE` + `LITELLM_PROXY_API_KEY` | `litellm_proxy/openai/gpt-4o` | `[]` | `api.openai.com/v1/files` |
| 2 | `USE_LITELLM_PROXY=true` | `openai/gpt-4o` | `[]` | `api.openai.com/v1/files` |
| 3 | neither, only `OPENAI_API_KEY` | `litellm_proxy/openai/gpt-4o` | `[]` | `api.openai.com/v1/files` |

Case 2 is worth spelling out. LiteLLM supports routing unprefixed models
through the proxy with `USE_LITELLM_PROXY`, and its own docstring says that
flag exists for Google ADK (BerriAI/litellm#10559). A model string of
`openai/gpt-4o` gives no hint that it is proxied, so anything matching on the
`litellm_proxy/` prefix alone will miss it.

For reference, LiteLLM resolves the proxy endpoint for the completion call in
`litellm/llms/litellm_proxy/chat/transformation.py`:

```python
api_base = api_base or get_secret_str("LITELLM_PROXY_API_BASE")
dynamic_api_key = api_key or get_secret_str("LITELLM_PROXY_API_KEY")
```

`litellm.acreate_file` does not consult those, so unless ADK forwards an
endpoint the two halves of the request diverge.

**Relation to #6538 and #6578**

#6578 (merged as 602e58d) fixed the payload shape for proxied models. Nested
`litellm_proxy/azure/...` identifiers now resolve to the underlying provider,
so the upload path runs instead of emitting an inline `file_data` block. That
part is correct and nothing here contests it.

The same commit also set `custom_llm_provider="openai"` for proxied uploads.
That is the line that turns cases 1 and 3 from a local failure into a request
to `api.openai.com`, because `openai` is the one provider whose credential
resolution defaults to the public API:

```python
# litellm/llms/openai/common_utils.py
resolved_api_base = (
    api_base or litellm.api_base or os.getenv("OPENAI_BASE_URL")
    or os.getenv("OPENAI_API_BASE") or "https://api.openai.com/v1"
)
```

On `main`, `_get_content` takes only `parts`, `provider` and `model`. There is
no parameter through which a proxy endpoint could arrive, and ADK never sets
`litellm.api_base`. So there is no configuration in which that override reaches
a proxy. It resolves to the public endpoint every time.

**Environment**

- ADK `main` at 370027a7. The behavior starts at 602e58d.
- litellm 1.85.7, and also reproduced on 1.84.0, the floor of the
  `litellm>=1.84` constraint in `pyproject.toml`. In both,
  `OPENAI_COMPATIBLE_BATCH_AND_FILES_PROVIDERS` is `{openai, hosted_vllm}` and
  `create_file` has no `litellm_proxy` branch, so routing the upload as
  `litellm_proxy` is not an option either.
- Python 3.11, macOS.

**Additional context**

I have a fix and will open a PR referencing this issue. One part of it is a
behavior change I would rather have decided explicitly than slipped into a
diff: when no proxy endpoint can be determined, the upload raises instead of
quietly using the provider default. For anyone who has `OPENAI_API_KEY` set
that converts a currently "working" upload into an error. I think an explicit
error is better than a silent misroute, but it is a trade off and it is your
call.

vietnamesekid:

- Closes: #6722
- Related: #6538, #6578

**Problem:**

`litellm.acreate_file` resolves its endpoint independently of the completion
call. ADK passes only `custom_llm_provider`, so a proxied upload falls back to
the underlying provider's own credential resolution and the request splits in
two: the completion goes to the proxy, the upload goes elsewhere.

For `azure` that is a local failure. For `openai` it is not a failure at all.
`get_openai_credentials` defaults `api_base` to `https://api.openai.com/v1`, so
a stray `OPENAI_API_KEY` makes the upload succeed against the public OpenAI API
carrying proxied file content and the developer's own key.

Three configurations reach that fallback. Each row was run with the repro
script in the linked issue, changing only `.env` and the model string.

| # | Configuration | Model | Before |
|---|---|---|---|
| 1 | `LITELLM_PROXY_API_BASE` + `LITELLM_PROXY_API_KEY` | `litellm_proxy/openai/gpt-4o` | upload to `api.openai.com` |
| 2 | `USE_LITELLM_PROXY=true` | `openai/gpt-4o` | upload to `api.openai.com` |
| 3 | neither, only `OPENAI_API_KEY` | `litellm_proxy/openai/gpt-4o` | upload to `api.openai.com` |

Case 1 exists because LiteLLM reads those two variables for the completion call
in `litellm/llms/litellm_proxy/chat/transformation.py`, while the upload path
does not. Case 2 exists because matching on the `litellm_proxy/` prefix alone
misses `USE_LITELLM_PROXY`, which routes unprefixed models through the proxy.
LiteLLM's own docstring says that flag exists for Google ADK
(BerriAI/litellm#10559).

On 602e58d, the merge of #6578: the provider resolution part of that change is
correct and this PR keeps it. The same commit also set
`custom_llm_provider="openai"` for proxied uploads, and that is the line that
turns cases 1 and 3 from a local error into a request to `api.openai.com`,
since `openai` is the provider that defaults to the public endpoint. On `main`,
`_get_content` takes only `parts`, `provider` and `model`, and ADK never sets
`litellm.api_base`, so there is no configuration in which that override could
reach a proxy. This PR keeps the real provider, which is what preserves the
provider specific payload shape, and redirects the endpoint instead.

**Solution:**

1. `_get_upload_params` falls back to `LITELLM_PROXY_API_BASE` and
   `LITELLM_PROXY_API_KEY` when the completion call carries no endpoint.
   Explicit completion arguments still win.
2. `_is_proxied_model` also treats `USE_LITELLM_PROXY` as proxied, covering
   unprefixed models routed through the proxy.
3. The upload raises when the model is proxied and no endpoint can be
   determined, rather than falling through to the provider default.

Direct models are untouched. They forward nothing and keep their existing
environment variable resolution, since their `api_base` already points at the
provider.

One behavior change worth deciding on rather than absorbing silently, item 3
above. Anyone who has `OPENAI_API_KEY` set and a proxied model has a working
upload today, to the wrong destination. After this they get a `ValueError`
naming the configuration to set. I think an explicit error beats a silent
misroute, but it is a trade off. If you would rather not take it, items 1 and 2
alone still fix every case where the proxy endpoint is knowable.

### Testing Plan

**Unit Tests:**

```
398 passed in 2.68s

6 failed, 1068 passed, 34 warnings in 13.40s
```

The 6 failures are all in `test_anthropic_llm.py` and reproduce identically on
a pristine `main` worktree at 370027a7, so they are pre-existing and unrelated
to this change.

New tests:

- `test_get_upload_params_falls_back_to_proxy_env`, covering the env fallback
  and that explicit completion arguments take precedence over it.
- `test_is_proxied_model_honors_use_litellm_proxy_flag`, covering the flag and
  that the explicit prefix still wins regardless of it.
- `test_get_content_proxied_upload_without_endpoint_raises`, which asserts
  `acreate_file` is not called at all when the destination is unknown.

I also isolated `test_get_upload_params` from the ambient environment. It would
otherwise pick up `LITELLM_PROXY_*` from a developer or CI machine and pass or
fail depending on which shell it ran in.

Each part of the fix was reverted on its own to check the tests fail without it
instead of passing by construction. Reverting the env fallback, the flag check,
or the guard each fails exactly its own test and nothing else.

Driven through `LiteLlm(...)` rather than internal helpers, with configuration
loaded from `.env` by `python-dotenv`, a local HTTP server standing in for the
proxy, and every outbound request recorded and blocked if it left localhost.
Full script is in the linked issue.

`.env`:

```
LITELLM_PROXY_API_BASE=http://127.0.0.1:8877
LITELLM_PROXY_API_KEY=sk-proxy-secret
OPENAI_API_KEY=sk-personal-developer-key
```

Before, with `model="litellm_proxy/openai/gpt-4o"`:

```
raised         : APIConnectionError: Connection error.
proxy received : []
left the proxy : ['https://api.openai.com/v1/files', ...]
```

After:

```
proxy received : ['/files', '/chat/completions']
left the proxy : []
```

Both halves of the request reach the proxy, and the proxy sees
`Authorization: Bearer sk-proxy-secret`, so it is the proxy credentials being
used and not the personal key.

Full matrix, same script throughout:

| `.env` | Model | Before | After |
|---|---|---|---|
| proxy only | `litellm_proxy/azure/my-deployment` | fails, proxy gets nothing | reaches proxy |
| proxy + `OPENAI_API_KEY` | `litellm_proxy/openai/gpt-4o` | goes to `api.openai.com` | reaches proxy |
| `USE_LITELLM_PROXY=true` | `openai/gpt-4o` | goes to `api.openai.com` | reaches proxy |
| only `OPENAI_API_KEY` | `litellm_proxy/openai/gpt-4o` | goes to `api.openai.com` | raises, sends nothing |

Reproduced on litellm 1.85.7 and on 1.84.0, the floor of the `litellm>=1.84`
constraint in `pyproject.toml`, so this is not specific to one release.

I do not have an Azure backed proxy deployment, so none of this is confirmed
against a live endpoint. Everything above is the request ADK builds and where
it actually goes, measured against a local server.

Formatted with `pyink` 25.12.0 and `isort` 8.0.1, matching the pinned versions
in `.pre-commit-config.yaml`.

### Checklist

### Additional context

Known limit, deliberately out of scope. This covers the proxy endpoint when it
comes from constructor arguments or from `LITELLM_PROXY_API_BASE` and
`LITELLM_PROXY_API_KEY`. A proxy configured by some other means that ADK cannot
see still lands in case 3 and raises. That is the safe outcome but it is still
a failure, and closing it would mean reimplementing more of LiteLLM's
configuration resolution inside ADK. That felt like your call rather than mine.

The two commits are separable if you would rather take them independently.
`7ab75d24` forwards constructor supplied endpoints, which is the follow up I
raised on #6578, and `d8643357` closes the three cases above.

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -341,6 +341,14 @@ def _strip_proxy_prefix(model: str) -> str:
 
 _UPLOAD_ENDPOINT_KEYS = ("api_base", "api_key", "api_version")
 
+# LiteLLM reads these when the completion call carries no explicit endpoint,
+# so a proxy configured only through the environment still has to be honored
+# by the upload.
+_PROXY_ENV_FALLBACKS = {
+    "api_base": "LITELLM_PROXY_API_BASE",
+    "api_key": "LITELLM_PROXY_API_KEY",
+}
+
 
 def _get_upload_params(
     model: str, completion_args: Dict[str, Any]
@@ -348,10 +356,17 @@ def _get_upload_params(
   """Endpoint overrides to forward to `litellm.acreate_file`.
 
   File uploads resolve their endpoint independently of the completion call, so
-  without these they fall back to the underlying provider's environment
-  variables. For a proxied model that would send the upload straight to the
-  provider while the completion goes to the proxy, which fails whenever the
-  caller only holds proxy credentials.
+  without these they fall back to the underlying provider's own credentials.
+  For a proxied model that splits the request in two: the completion goes to
+  the proxy while the upload goes straight to the provider. That fails when the
+  caller holds only proxy credentials, and when the provider resolves to
+  ``openai`` it is worse than a failure -- `get_openai_credentials` defaults to
+  ``https://api.openai.com/v1``, so a stray ``OPENAI_API_KEY`` in the
+  environment sends proxied file content to the public OpenAI API instead.
+
+  Explicit completion arguments win, falling back to the proxy's own
+  environment variables so that proxies configured purely through the
+  environment are still routed correctly.
 
   Only proxied models forward anything. A direct `azure/...` model keeps the
   existing environment-variable behavior, since its `api_base` already points
@@ -366,16 +381,36 @@ def _get_upload_params(
   """
   if not _is_proxied_model(model):
     return {}
-  return {
+  upload_params = {
       key: completion_args[key]
       for key in _UPLOAD_ENDPOINT_KEYS
       if completion_args.get(key)
   }
+  # LiteLLM resolves the proxy endpoint from these when the completion call
+  # does not carry one, so the upload has to follow the same fallback.
+  for key, env_var in _PROXY_ENV_FALLBACKS.items():
+    if not upload_params.get(key):
+      value = os.environ.get(env_var)
+      if value:
+        upload_params[key] = value
+  return upload_params
 
 
 def _is_proxied_model(model: str) -> bool:
-  """Returns True if the model is served through a LiteLLM Proxy."""
-  return bool(model) and model.lower().startswith(_PROXY_PROVIDER + "/")
+  """Returns True if the model is served through a LiteLLM Proxy.
+
+  Covers both the explicit ``litellm_proxy/`` prefix and LiteLLM's
+  ``USE_LITELLM_PROXY`` flag, which routes unprefixed models through the proxy
+  as well.
+  """
+  if not model:
+    return False
+  if model.lower().startswith(_PROXY_PROVIDER + "/"):
+    return True
+  return os.environ.get("USE_LITELLM_PROXY", "").strip().lower() in (
+      "true",
+      "1",
+  )
 
 
 def _get_provider_from_model(model: str) -> str:
@@ -1577,6 +1612,20 @@ async def _get_content(
           # Keep the provider that actually serves the request so the payload
           # keeps its provider-specific shape, and point the upload at the
           # proxy via `upload_params` instead of swapping the provider out.
+          if _is_proxied_model(model) and not (upload_params or {}).get(
+              "api_base"
+          ):
+            # Without an endpoint the upload would fall back to the provider's
+            # own credentials. For `openai` that default is the public
+            # https://api.openai.com/v1, so proxied file content would leave
+            # the proxy entirely. Fail instead of silently sending it there.
+            raise ValueError(
+                f"Cannot upload file for proxied model {model!r}: the LiteLLM"
+                " Proxy endpoint is unknown. Pass `api_base` (and `api_key`) to"
+                " LiteLlm(...), or set LITELLM_PROXY_API_BASE and"
+                " LITELLM_PROXY_API_KEY, so the upload reaches the proxy"
+                " instead of the provider's default endpoint."
+            )
           file_response = await litellm.acreate_file(
               file=part.inline_data.data,
               purpose="assistants",
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -332,14 +332,50 @@ def _strip_proxy_prefix(model: str) -> str:
   Returns:
     The model string without the ``litellm_proxy/`` prefix if nested.
   """
-  if not model:
-    return model
-  prefix = _PROXY_PROVIDER + "/"
-  if model.lower().startswith(prefix):
-    remaining = model[len(prefix) :]
+  if _is_proxied_model(model):
+    remaining = model[len(_PROXY_PROVIDER) + 1 :]
     if "/" in remaining:
       return remaining
   return model
+
+
+_UPLOAD_ENDPOINT_KEYS = ("api_base", "api_key", "api_version")
+
+
+def _get_upload_params(
+    model: str, completion_args: Dict[str, Any]
+) -> Dict[str, Any]:
+  """Endpoint overrides to forward to `litellm.acreate_file`.
+
+  File uploads resolve their endpoint independently of the completion call, so
+  without these they fall back to the underlying provider's environment
+  variables. For a proxied model that would send the upload straight to the
+  provider while the completion goes to the proxy, which fails whenever the
+  caller only holds proxy credentials.
+
+  Only proxied models forward anything. A direct `azure/...` model keeps the
+  existing environment-variable behavior, since its `api_base` already points
+  at the provider.
+
+  Args:
+    model: The LiteLLM model string.
+    completion_args: The arguments the caller passes to the completion call.
+
+  Returns:
+    The subset of endpoint overrides to forward, empty when not proxied.
+  """
+  if not _is_proxied_model(model):
+    return {}
+  return {
+      key: completion_args[key]
+      for key in _UPLOAD_ENDPOINT_KEYS
+      if completion_args.get(key)
+  }
+
+
+def _is_proxied_model(model: str) -> bool:
+  """Returns True if the model is served through a LiteLLM Proxy."""
+  return bool(model) and model.lower().startswith(_PROXY_PROVIDER + "/")
 
 
 def _get_provider_from_model(model: str) -> str:
@@ -1182,6 +1218,7 @@ async def _content_to_message_param(
     *,
     provider: str = "",
     model: str = "",
+    upload_params: Optional[Dict[str, Any]] = None,
 ) -> Union[Message, list[Message]]:
   """Converts a types.Content to a litellm Message or list of Messages.
 
@@ -1251,7 +1288,13 @@ async def _content_to_message_param(
   if role == "user":
     user_parts = [part for part in content_parts_or_empty if not part.thought]
     message_content = (
-        await _get_content(user_parts, provider=provider, model=model) or None
+        await _get_content(
+            user_parts,
+            provider=provider,
+            model=model,
+            upload_params=upload_params,
+        )
+        or None
     )
     return ChatCompletionUserMessage(
         role="user",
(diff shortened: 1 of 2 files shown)


varunbiluri (commented):
Blocking compatibility issue: `_strip_proxy_prefix()` now calls `_is_proxied_model()`, but `_is_proxied_model()` is also true for an unprefixed model whenever `USE_LITELLM_PROXY=true`. In that case `_strip_proxy_prefix("openai/gpt-4o")` slices off `len("litellm_proxy/")` characters even though that prefix is absent, corrupting the model name. Keep prefix stripping based solely on `model.lower().startswith(_PROXY_PROVIDER + "/")`; use `_is_proxied_model()` only for routing decisions. Please add a regression assertion for `_strip_proxy_prefix("openai/gpt-4o")` under `USE_LITELLM_PROXY=true`.

vietnamesekid:
Good catch, fixed in 56c2519f.

One correction on the specifics. `_strip_proxy_prefix("openai/gpt-4o")` under
`USE_LITELLM_PROXY=true` actually came back unchanged: the slice ran, but the
`"/" in remaining` guard caught it (14 chars off leaves `"o"`, no slash).

That guard is what hid the real damage. Names with two or more slashes survive
it and get silently corrupted:

```
fireworks_ai/accounts/fireworks/models/llama-v3 -> provider 'ccounts'
vertex_ai/publishers/google/models/gemini-2.5-pro -> provider 'ishers'
openai/some-model/with-slash -> provider 'del'
```

Not a mangled name that errors out early, but a plausible looking provider that
routes the request somewhere arbitrary.

Fix is what you described: slice keyed off the literal `litellm_proxy/` prefix,
`_is_proxied_model` used only for routing. Added
`test_strip_proxy_prefix_ignores_use_litellm_proxy_flag` with your
`openai/gpt-4o` assertion plus the multi slash cases, since those are the ones
that actually corrupt. Reverting the fix fails that test and nothing else.
399 passed in `test_litellm.py`.

varunbiluri (approved):
Re-reviewed through `205b9f7c`. The literal-prefix check fixes the model-name corruption while leaving proxy routing separate, and the regression covers both short and multi-slash unprefixed names plus the genuinely prefixed case. The subsequent environment parsing/whitespace hardening is consistent with LiteLLM's behavior. My concern is resolved.

vietnamesekid:
Hi @varunbiluri, any updates on this PR?

PR #6733, opened 2026-08-15
fix(flows): re-execute orphaned sibling calls on confirmation resume

### LINKED ISSUE #6732: Parallel ungated tool results are lost across a tool-confirmation pause, yielding an empty model reply after approval
**Describe the bug**

When a model turn contains a confirmation-gated function call (`FunctionTool(require_confirmation=True)`) **in parallel with** ungated function calls, the ungated siblings' results are lost across the confirmation pause. After the user approves, the resume processor (`flows/llm_flows/request_confirmation.py`) re-executes only the confirmed tool, leaving the sibling `function_call` in session history with no matching `function_response`. Gemini, given that dangling call, returns an **empty final message** — while the approved (destructive) tool HAS executed. The end user sees no answer and may retry an action that already succeeded.

**To Reproduce**

1. An `LlmAgent` with two tools: `delete_record` built with `require_confirmation=True`, and `list_records` (ungated).
2. Prompt so the model emits both calls in one turn, e.g. *"Delete record X, and after deleting, list the remaining records."* Gemini frequently parallelizes this into one response with both `function_call` parts.
3. The run pauses with `adk_request_confirmation`. Observations at this point:
   - The ungated `list_records` **does execute** during the pause (its side effects are visible), but the merged function-response event — carrying the sibling's real result plus the gated tool's `"This tool call requires confirmation"` placeholder — is **never persisted to the session**. When the caller stops consuming the event stream at the confirmation event (the natural thing for a client that must wait for user input, e.g. any AG-UI/SSE bridge), the response event that `_postprocess_handle_function_calls_async` yields *after* `generate_request_confirmation_event` is never reached (`GeneratorExit`).
4. Resume with `ToolConfirmation(confirmed=True)`.
5. `_RequestConfirmationLlmRequestProcessor._resolve_confirmation_targets` re-executes only the confirmed `delete_record`. Session events now contain `function_call list_records` with no `function_response`.
6. The follow-up LLM call returns empty content (`{"text": ""}`).

Observed session event sequence (from a Postgres session service, ADK 2.7.0):

```
model: [text, function_call delete_record, thought_signature, function_call list_records, text]
model: [function_call adk_request_confirmation]   (long_running_tool_ids set)
user:  [function_response adk_request_confirmation {confirmed: true}]
model: [function_response delete_record {...success...}]   <- only the confirmed tool
model: [text ""]                                            <- empty final reply
```

**Expected behavior**

Either (a) the pause-time function-response event (sibling results + confirmation placeholder) is persisted before/with the confirmation-request event so history stays complete across the pause, or (b) the resume path re-executes (or synthesizes responses for) unresolved sibling calls from the same turn — so the post-approval continuation never sees a dangling `function_call`.

**Screenshots**

N/A (event sequence above).

**Desktop:**
 - OS: macOS 26 / Linux (Docker python:3.12-slim)
 - Python version: 3.12
 - ADK version: 2.7.0 (also inspected 2.5.0 — same structure)

**Model Information:**
 - Are you using LiteLLM: No
 - Which model is being used: gemini via Vertex AI

**Additional context**

Single-call confirmation flows (pause → approve/reject → resume) work correctly in 2.7.0, including the 2.6.x fixes for consumed-confirmation re-validation. The failure is specific to the parallel gated+ungated combination. Downstream we mitigate with an `after_model_callback` that drops ungated siblings from mixed turns before execution (the model re-issues them after the decision), but a framework-level fix would let parallel calls keep their results.

chelsealong:
Closes #6732

## Problem

When a model turn contains a confirmation-gated `FunctionTool` call **in parallel with** ungated calls, the ungated siblings' results can be lost across the confirmation pause.

The merged `function_response_event` for that turn (carrying the sibling's real result plus the gated tool's "requires confirmation" placeholder) is built and yielded *after* the `adk_request_confirmation` event by `_postprocess_handle_function_calls_async`. A caller that stops consuming the event stream at the confirmation event (the natural thing to do while waiting for user input, e.g. an SSE/AG-UI bridge) never asks for the next event, so that response event — and the runner's `append_event` for it — never happens.

On resume, `_RequestConfirmationLlmRequestProcessor._resolve_confirmation_targets` only re-executes the confirmed tool. Session history is left with the sibling's `function_call` and no matching `function_response`, and the next LLM call (Gemini) returns an empty reply — even though the confirmed (possibly destructive) tool already executed.

## Fix

`_RequestConfirmationLlmRequestProcessor.run_async` now also scans for sibling function calls that share a model turn with a confirmed call and still have no `function_response` anywhere in history (`_get_orphaned_sibling_function_calls` in `src/google/adk/flows/llm_flows/request_confirmation.py`), and re-executes them alongside the confirmed tool. This keeps the turn's function calls and responses in sync so the resumed continuation never sees a dangling `function_call`.

This is the minimal of the two fixes discussed in the issue (re-execute unresolved siblings on resume, vs. reordering the pause-time yield so the response event persists before the confirmation event) — it's a single, local change to the resume path and does not touch the pause-time event ordering or its existing tests.

**Follow-up fix:** the first version of `_get_orphaned_sibling_function_calls` decided a call was an "ungated orphan" based only on "no `function_response` anywhere in history" + "registered tool" — it never checked whether the sibling itself still required (and had not yet received) its own confirmation. When a turn has two confirmation-gated calls and only one is confirmed, this swept the still-pending sibling into re-execution, which re-triggered `request_confirmation` bookkeeping for a confirmation that was never resolved (minting a fresh `adk_request_confirmation` call while the original request went orphaned). The helper now also checks `tool.check_require_confirmation` (plus dynamic-confirmation history via the extracted `_compute_dynamically_requested_fc_ids`) for each candidate sibling and skips ones that still require confirmation, leaving their own resume to a later turn once they're actually confirmed.

## Testing

Added `test_request_confirmation_processor_reexecutes_orphaned_sibling` in `tests/unittests/flows/llm_flows/test_request_confirmation.py`, which builds session history for a turn with one gated call (unconfirmed) and one real ungated sibling call and **no** function_response event for that turn (reproducing the lost-response scenario), then confirms the gated tool and asserts the resumed processor's merged response contains a `function_response` for **both** tools.

Verified the test fails without the fix (`git stash` the source change, run the test):
```
AssertionError: assert {'mock_tool'} == {'mock_tool', 'sibling_tool'}
```

Added `test_request_confirmation_processor_leaves_gated_sibling_pending`, covering the follow-up fix: two gated tools in the same turn, only one confirmed. Verified this test fails against the pre-follow-up-fix source (reproduced via a standalone repro script run against `git show HEAD:src/.../request_confirmation.py`, since the test module itself imports the newly extracted `_compute_dynamically_requested_fc_ids` helper):
```
AssertionError: BUG REPRODUCED: gated sibling was swept in: {'gated_sibling_tool', 'mock_tool'}
```
and that it also minted a fresh `requested_tool_confirmations` entry for the still-pending sibling. With the follow-up fix applied, only `mock_tool`'s response is present and no confirmation bookkeeping is re-triggered.

With the fix:
```
$ pytest tests/unittests/flows/llm_flows/test_request_confirmation.py tests/unittests/runners/test_run_tool_confirmation.py -q
28 passed, 12 warnings in 2.93s
```

Full unit suite:
```
$ pytest tests/unittests -n auto -q
12023 passed, 88 skipped, 25 xfailed, 1 xpassed, 3162 warnings, 24 subtests passed in 148.95s
```

`pre-commit run --files src/google/adk/flows/llm_flows/request_confirmation.py tests/unittests/flows/llm_flows/test_request_confirmation.py` passes (ruff, isort, pyink, addlicense, ADK compliance checks, codespell).

## AI assistance disclosure

This change was written with AI assistance (Claude Code / Anthropic).

### DIFF
--- a/src/google/adk/flows/llm_flows/request_confirmation.py
+++ b/src/google/adk/flows/llm_flows/request_confirmation.py
@@ -214,6 +214,50 @@ async def _resolve_confirmation_targets(
   return tool_confirmation_dict, original_fcs_dict
 
 
+def _get_orphaned_sibling_function_calls(
+    events: list[Event],
+    original_fc_ids: set[str],
+    tools_dict: dict[str, BaseTool],
+) -> dict[str, types.FunctionCall]:
+  """Finds ungated sibling calls left without a function_response.
+
+  When a model turn contains a confirmation-gated call in parallel with
+  ungated calls, the ungated siblings' results can fail to persist across the
+  pause (e.g. a caller stops consuming the event stream at the confirmation
+  event). On resume, those siblings' `function_call`s remain in history with
+  no matching `function_response`, which makes the next LLM call return an
+  empty reply. Re-executing them here alongside the confirmed tool keeps the
+  turn's function calls and responses in sync.
+
+  Args:
+    events: Session events to scan.
+    original_fc_ids: IDs of the original function calls being resumed.
+    tools_dict: Dictionary of registered tools.
+
+  Returns:
+    Mapping of sibling function call ID -> ``FunctionCall``, for calls that
+    share an event with one of *original_fc_ids* but have no response yet.
+  """
+  responded_fc_ids = {
+      fr.id for ev in events for fr in ev.get_function_responses() if fr.id
+  }
+  siblings: dict[str, types.FunctionCall] = {}
+  for event in events:
+    event_function_calls = event.get_function_calls()
+    event_fc_ids = {fc.id for fc in event_function_calls if fc.id}
+    if not event_fc_ids & original_fc_ids:
+      continue
+    for function_call in event_function_calls:
+      if (
+          function_call.id
+          and function_call.id not in original_fc_ids
+          and function_call.id not in responded_fc_ids
+          and function_call.name in tools_dict
+      ):
+        siblings[function_call.id] = function_call
+  return siblings
+
+
 def _map_confirmation_to_original_fc_ids(
     events: list[Event],
     confirmation_fc_ids: set[str],
@@ -356,12 +400,22 @@ class _RequestConfirmationLlmRequestProcessor(BaseLlmRequestProcessor):
     if not tools_to_resume_with_confirmation:
       return
 
-    # Step 4: Re-execute the confirmed tools.
+    # Step 4: Also pick up any ungated sibling calls from the same turn that
+    # never got a function_response, so resuming does not leave a dangling
+    # function_call in history.
+    sibling_function_calls = _get_orphaned_sibling_function_calls(
+        events, set(tools_to_resume_with_args.keys()), tools_dict
+    )
+    function_calls_to_execute = list(tools_to_resume_with_args.values()) + list(
+        sibling_function_calls.values()
+    )
+
+    # Step 5: Re-execute the confirmed tools and any orphaned siblings.
     if function_response_event := await functions.handle_function_call_list_async(
         invocation_context,
-        list(tools_to_resume_with_args.values()),
+        function_calls_to_execute,
         tools_dict,
-        set(tools_to_resume_with_confirmation.keys()),
+        {fc.id for fc in function_calls_to_execute if fc.id},
         tools_to_resume_with_confirmation,
     ):
       yield function_response_event
--- a/src/google/adk/flows/llm_flows/request_confirmation.py
+++ b/src/google/adk/flows/llm_flows/request_confirmation.py
@@ -72,12 +72,39 @@ def _get_original_function_call_args(
   return original_function_call
 
 
+def _compute_dynamically_requested_fc_ids(events: list[Event]) -> set[str]:
+  """Returns IDs of function calls for which a tool dynamically requested confirmation.
+
+  This accumulates over ALL events rather than keeping one event per ID: once
+  the confirmed tool is re-executed it emits a second function response with
+  the same ID and no `requested_tool_confirmations`, which would otherwise
+  shadow the original request.
+
+  Args:
+    events: Session events to scan.
+
+  Returns:
+    Set of original function call IDs that were dynamically requested to
+    require confirmation.
+  """
+  dynamically_requested_fc_ids: set[str] = set()
+  for ev in events:
+    requested_tool_confirmations = ev.actions.requested_tool_confirmations or {}
+    if not requested_tool_confirmations:
+      continue
+    for fr in ev.get_function_responses():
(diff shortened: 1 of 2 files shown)


varunbiluri (changes requested):
The sibling is not actually unexecuted in the reported failure: the issue states it executes before the pause, but its merged response event is not persisted when the consumer stops iterating. Re-executing it on resume therefore gives at-least-twice behavior and can duplicate arbitrary side effects (send a message, charge a card, mutate a record). The regression test uses a pure string-returning sibling, so it does not expose this. Please fix persistence/event ordering (or persist/synthesize the already-produced result) rather than invoking an already-executed tool again; the resume path cannot safely infer that absence of a persisted  means absence of execution.

varunbiluri (changes requested):
The sibling is not actually unexecuted in the reported failure: the issue states that it executes before the pause, but its merged response event is not persisted when the consumer stops iterating. Re-executing on resume therefore gives at-least-twice behavior and can duplicate arbitrary side effects such as sending a message, charging a card, or mutating a record. The regression test uses a pure string-returning sibling and does not expose this. Please fix persistence or event ordering, or persist the already-produced result, rather than invoking an already-executed tool again; absence of a persisted function response does not imply absence of execution.

chelsealong:
Addressed: the resume-time helper no longer re-invokes orphaned sibling calls. It now synthesizes an error `function_response` for them instead, so the confirmed tool's own re-execution is unaffected and the sibling's tool is never called a second time. Added a call-count assertion to the regression test to prove the sibling function is not invoked again, and confirmed the response is the synthesized error rather than a fresh result.

varunbiluri (approved):
Re-reviewed `25481d1e`. The revised path no longer re-invokes the already-executed sibling, transparently records that its result was lost, preserves the still-gated sibling path, and behaviorally asserts a zero second-call count. That addresses the side-effect safety issue I raised. My blocking concern is resolved.

chelsealong:
Thanks for the re-review and approval. Merged `upstream/main` (86 commits, including the `flows/llm_flows` tools/core relocation that moved this file to `tools/_confirmation.py` behind a re-export shim) to keep the branch current; resolved the resulting conflicts by porting the fix onto the new file layout with no behavior change. `tests/unittests/flows/llm_flows/tools/test_confirmation.py` + `tests/unittests/runners/test_run_tool_confirmation.py` still pass (28 passed). Ready for maintainer review.

PR #6740, opened 2026-08-15
fix(cli): adk eval process exit code now reflects PASSED/FAILED

gaurav-gandhi-2411:
## 🔴 Required Information

**Describe the Bug:**
`cli_eval` (the `adk eval` CLI command) computes and prints a real per-eval-set
"Tests passed"/"Tests failed" summary (`eval_run_summary`) but never calls
`sys.exit` — the process always exits 0 regardless of the printed verdict,
indistinguishable from every test genuinely passing. This makes `adk eval`
unusable as a CI gate on its own exit code: a pipeline step that runs `adk eval
... && deploy` will deploy even when every eval case failed.

Every other subcommand in this file with a pass/fail outcome already sets a real
exit code: `run --query` does `sys.exit(exit_code)` (cli_tools_click.py:961,
derived from `run_once_cli`'s return), `test` does `sys.exit(1)` on a missing
test runner (cli_tools_click.py:1060). `eval` was the one command whose exit code
carried no signal at all.

**Steps to Reproduce:**
Run `adk eval <agent_module> <eval_set_file>` against an eval set containing at
least one failing case, then check the process exit code (`echo $?`).

**Expected Behavior:** A non-zero exit code when the printed summary shows any
failed test, so the command is usable as a CI gate.

**Observed Behavior:** Exit code `0` regardless of the printed "Tests failed"
count — no exit code, prior to this fix, ever distinguished a real failure from
a real pass.

**Environment:** google-adk main branch (this repo), CLI entry point
`cli_tools_click.py::cli_eval`.

## Why this fix

After the existing `eval_run_summary` loop, `sys.exit(1)` if any eval set
recorded a failed test, else `sys.exit(0)` — derived from the same counts
already being printed, no new computation introduced.

## Changes

- `src/google/adk/cli/cli_tools_click.py`: `cli_eval` now calls `sys.exit(1 if
  total_failed else 0)` at the end, derived from the same `eval_run_summary`
  counts already being printed.
- `tests/unittests/cli/utils/test_cli_tools_click.py`:
  - `test_cli_eval_with_eval_set_file_path` / `test_cli_eval_with_eval_set_id`:
    both eval cases have no invocations and no configured criteria, so neither
    produces a PASSED verdict ("Tests failed: 1" in their own printed summary,
    unchanged by this fix) — their `exit_code == 0` assertion was only ever true
    because the exit code carried no signal; updated to `== 1` with a comment
    explaining why.
  - New `test_cli_eval_exit_code_reflects_final_eval_status` (parametrized,
    2 cases): mocks `_collect_inferences`/`_collect_eval_results` to return a
    canned `EvalCaseResult` with a controlled `final_eval_status`, asserting
    `exit_code == 0` for PASSED and `!= 0` (1) for FAILED.

## Testing Plan

**Unit Tests:**
- [x] Added/updated tests above.
- [x] The `any_failed_exits_nonzero` case of the new parametrized test fails
      against pre-fix code (verified by stashing the fix) — confirming it
      actually catches the bug.

```
pytest tests/unittests/cli/utils/test_cli_tools_click.py -k eval: 15 passed
pytest tests/unittests/cli/utils/test_cli_tools_click.py: 98 passed
pytest tests/unittests/evaluation/: 838 passed (unaffected — this repo's baseline
  for this branch's diff; unrelated to the eval-service tests touched by a sibling
  PR I'm also preparing)
pre-commit (ruff, isort, pyink, addlicense, codespell): all pass on changed files
```
20 unrelated pre-existing failures (conformance/markdown report, cleanup-unused-files,
path-normalizer, cloud-run-deploy tests) reproduce identically on a clean
`upstream/main` checkout with no changes — confirmed before attributing anything to
this change.

**Manual E2E:** ran `adk eval` against a real eval set with a failing case pre- and
post-fix from a built shell: pre-fix `echo $?` → `0`; post-fix `echo $?` → `1`.

## Risk & rollback

Confined to the end of one CLI command function; no change to any return value,
schema, or the printed summary itself. The only behavior change is the process exit
code, which previously carried no information — any caller currently ignoring it is
unaffected; any caller that (incorrectly, given the old behavior) assumed 0 always
meant "passed" will now see accurate signal. Purely additive; revert is a clean
single-commit revert.


gaurav-gandhi-2411:
Added a case this PR's own diff was missing: an empty \`eval_run_summary\` (no eval case ever evaluated, e.g. an empty eval set) previously took the same \`sys.exit(0)\` path as "every case passed" — indistinguishable from a genuine pass. It now raises instead of exiting 0.

This is the same defect class flagged in #6951 for the \`AgentEvaluator\` path and fixed there in #6952 (raise on an empty result set rather than silently returning success); this PR adopts the identical approach for \`cli_eval\`.

New test: \`test_cli_eval_empty_summary_does_not_exit_zero\`, confirmed failing pre-fix.

PR #6744, opened 2026-08-15
fix(cli): show agent type in create help

### LINKED ISSUE #6737: [Feature Request] Show the documented `adk create --type` option in `--help`

[The Agent Config documentation](https://adk.dev/agents/config/#build-an-agent
) instructs users to run:

```bash
adk create --type=config my_agent
```

However, `adk create --help` does not list the `--type` option, even though the option works.
This makes the documented command appear unsupported and creates an inconsistency between the CLI and its documentation.

The option is currently declared with `hidden=True`

https://github.com/google/adk-python/blob/1d2d1eda3c9b795cd90ad643390f4da5a8cd27bf/src/google/adk/cli/cli_tools_click.py#L677-L688

Please expose `--type` in `adk create --help`, including its supported values (`code` and `config`) and its experimental status.

### Impact on your work

This would make the documented Agent Config workflow discoverable and clarify that `--type=config` is supported.

### Willingness to contribute

No.

Users can follow the documentation despite the missing help entry, but they may reasonably assume the option is unavailable or deprecated.

Remove `hidden=True` from the `--type` Click option and retain wording that identifies Agent Config as experimental.

### Additional Context

Confirmed with ADK Python 2.6.3:

- `adk create --help` does not show `--type`.
- `adk create --type=config my_agent` succeeds and generates `root_agent.yaml`.

OllieinCanada:
Closes: #6737

## Summary

Expose the already-supported and documented `adk create --type` option in `adk create --help`.

The option remains marked experimental, retains its `code`/`config` choice validation, and keeps the existing default. This only removes the hidden-help flag.

## Testing

- Added a CLI-level regression test that invokes `adk create --help`
- Verified the test fails before the production change because `--type` is absent
- Focused test passes after the change
- Pyink check passes on both changed files
- Ruff check passes on both changed files
- `git diff --check` passes

This PR was prepared with AI assistance and manually reviewed.

### DIFF
--- a/src/google/adk/cli/cli_tools_click.py
+++ b/src/google/adk/cli/cli_tools_click.py
@@ -684,7 +684,6 @@ def cli_conformance_test(
     ),
     default="CODE",
     show_default=True,
-    hidden=True,  # Won't show in --help output. Not ready for use.
 )
 @click.argument("app_name", type=str, required=True)
 def cli_create_cmd(
--- a/tests/unittests/cli/utils/test_cli_tools_click.py
+++ b/tests/unittests/cli/utils/test_cli_tools_click.py
@@ -184,6 +184,15 @@ def test_resolve_eval_config_file_path_returns_none_for_multiple_eval_files(
 
 
 # cli create
+@pytest.mark.unmute_click
+def test_cli_create_help_shows_agent_type_option() -> None:
+  """`adk create --help` should document the supported agent types."""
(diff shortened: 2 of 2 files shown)


GWeale:
landed on main as d0d5ade via #6745, the hidden=True line is gone and adk create --help lists --type [CODE|CONFIG] now. it hasn't shipped yet though, 2.8.0 and 1.39.1 both still hide the option, so you'll see it in the next release. closing as a duplicate, and thanks for the fix.

PR #6749, opened 2026-08-16
feat(tools): let toolsets replace a failed tools listing

### LINKED ISSUE #6748: Applications cannot influence what a toolset contributes when tool listing fails (per-user OAuth MCP servers)
**Is your feature request related to a problem? Please describe.**

The context is per-user OAuth (3LO) MCP servers: each end user must individually
authorize the remote server, and until they do, McpToolset tool listing fails with
HTTP 401. This is not an outage — it is the expected steady state for every user who
has not connected yet, and the product requirement is to prompt that user to connect.

Since 2.7.0, SkillToolset._resolve_additional_tools_from_state gathers with
return_exceptions=True and skips a failing toolset with a warning. That isolation is
right for what it targets: one unreachable server must not take down unrelated tools.
The side effect for the 3LO case is that the failing toolset's own tools disappear
silently. Skill adk_additional_tools are resolved by name against the candidate pool,
so the names served by that MCP server never enter the pool. The model therefore has
no tool to call, never observes an authorization error, and has no way to ask the user
to authorize. The user is simply told the capability does not exist.

#4300 was closed with the guidance to wrap the get_tools() call in try/except
ConnectionError. That works when the application calls get_tools itself. It does not
when the toolset is owned by an agent or by SkillToolset, because the framework makes
the call. The only remaining seam is for the application to wrap every MCP toolset in
its own BaseToolset subclass, which is per-toolset opt-in and fails silently the day
someone adds a new toolset and forgets.

**Describe the solution you'd like**

An extension point that lets the owning application decide what a toolset contributes
when listing fails, instead of the framework deciding it always contributes nothing.
Conceptually an overridable hook on BaseToolset, invoked with the raised exception and
the readonly context, whose return value replaces that toolset's tool list for the
current resolution, with a default that behaves exactly as today (skip and warn). An
application could then return placeholder tools carrying an "authorization required"
instruction, so the authorization prompt reaches the user through the normal
tool-calling path.

**Describe alternatives you've considered**

Wrapping each MCP toolset in an application-side BaseToolset. This is what we do today;
it works, but it is opt-in per toolset and silent when omitted.

Checking authorization before building the toolset. This duplicates the server's own
auth decision and races with token expiry between the check and the listing.

Special-casing 401 inside McpToolset. Too opinionated for the framework — what should
happen on an unauthorized listing is application- and product-specific.

**Additional context**

Observed on google-adk 2.7.0. This affects any ADK application fronting MCP servers
that authorize per end user rather than per application.

---

Not a Contribution: this issue is submitted solely to report a limitation and request a
feature. It is expressly not intended as a Contribution under the Google CLA, and no
code, patch, or other work of authorship is submitted with it.

a2105z:
## Summary
- Add `BaseToolset.on_tools_listing_error` so apps can contribute replacement tools when listing fails (default: empty list / current skip behavior)
- Invoke the hook from `LlmAgent` tool loading and `SkillToolset` additional-tool resolution after the existing isolate-and-continue logging
- Enables per-user OAuth (3LO) MCP flows to surface connect/authorize placeholder tools instead of silently dropping the toolset

Fixes #6748

## Test plan
- [x] `pytest tests/unittests/tools/test_base_toolset.py` (default + override hook)
- [x] `pytest tests/unittests/agents/test_llm_agent_fields.py -k canonical_tools`
- [x] `pytest tests/unittests/tools/test_skill_toolset.py -k resolution_`
- [ ] Reviewer: confirm hook is not invoked on direct `get_tools()` / `get_tools_with_prefix()` calls (exceptions still propagate there)

### DIFF
--- a/src/google/adk/tools/base_toolset.py
+++ b/src/google/adk/tools/base_toolset.py
@@ -117,6 +117,10 @@ class BaseToolset(ABC):
     original exception; this hook is only invoked by the framework's
     isolate-and-continue paths.
 
+    Framework callers apply ``tool_name_prefix`` to the returned tools the same
+    way ``get_tools_with_prefix`` would, so overrides should return unprefixed
+    tool names.
+
     Args:
       error: The exception raised by ``get_tools`` / ``get_tools_with_prefix``.
       readonly_context: The same context passed to the failed listing call.
@@ -127,6 +131,44 @@ class BaseToolset(ABC):
     """
     del error, readonly_context
     return []
+
+  def _apply_tool_name_prefix(self, tools: list[BaseTool]) -> list[BaseTool]:
+    """Applies ``tool_name_prefix`` to tools, matching ``get_tools_with_prefix``.
+
+    Args:
+      tools: Tools with unprefixed names (as returned by ``get_tools`` or
+        ``on_tools_listing_error``).
+
+    Returns:
+      The same tools when no prefix is configured; otherwise shallow copies
+      with prefixed ``name`` / declaration names.
+    """
+    if not self.tool_name_prefix:
+      return tools
+
+    prefix = self.tool_name_prefix
+    prefixed_tools = []
+    for tool in tools:
+      tool_copy = copy.copy(tool)
+      prefixed_name = f'{prefix}_{tool.name}'
+      tool_copy.name = prefixed_name
+
+      def _create_prefixed_declaration(
+          original_get_declaration=tool._get_declaration,
+          prefixed_name=prefixed_name,
+      ):
+        def _get_prefixed_declaration():
+          declaration = original_get_declaration()
+          if declaration is not None:
+            declaration.name = prefixed_name
+            return declaration
+          return None
+
+        return _get_prefixed_declaration
+
+      tool_copy._get_declaration = _create_prefixed_declaration()
+      prefixed_tools.append(tool_copy)
+    return prefixed_tools
 
   @final
   async def get_tools_with_prefix(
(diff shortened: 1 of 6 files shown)


a2105z:
@googlebot I fixed it.

src/google/adk/tools/base_toolset.py
```diff
@@ -99,6 +99,35 @@ async def get_tools(
...
+    Direct callers of ``get_tools`` / ``get_tools_with_prefix`` still see the
+    original exception; this hook is only invoked by the framework's
+    isolate-and-continue paths.
+
+    Args:
+      error: The exception raised by ``get_tools`` / ``get_tools_with_prefix``.
+      readonly_context: The same context passed to the failed listing call.
+
+    Returns:
+      Tools to use in place of the failed listing. The default returns an
```
varunbiluri, line 129: The replacement list bypasses `get_tools_with_prefix`, so a toolset configured with `tool_name_prefix` returns unprefixed fallback tools. That breaks the BaseToolset naming contract precisely on the failure path and can introduce collisions with other toolsets (the canonical-tools caller appends this list directly, and SkillToolset does likewise). Please apply the same prefix transformation to fallback tools, or explicitly route hook results through a shared prefix helper, and add a regression test using a non-empty prefix.

varunbiluri (commented):
Staff review: one prefix-contract issue to address.

src/google/adk/tools/base_toolset.py
a2105z: Thanks @varunbiluri — great catch.

Fixed by extracting `_apply_tool_name_prefix` and routing listing-error fallbacks through it from both `LlmAgent.canonical_tools` and `SkillToolset` additional-tool resolution (same transformation as `get_tools_with_prefix`). Hook overrides still return unprefixed names; the framework applies the prefix.

Added regression coverage for a non-empty `tool_name_prefix` on the agent path, base toolset helper, and skill additional-tools path.

Appreciate the staff review — anything else you’d like me to tighten up here?

a2105z:
@varunbiluri addressed your prefix-contract note in the latest push — thank you again for the staff review. Please take another look when you have a chance, and let me know if there’s anything else you’d like changed.

varunbiluri (approved):
Re-reviewed `08bcf768`. Fallback tools now go through the same copy-and-prefix transformation as successful listings in both the agent and SkillToolset isolation paths, with behavioral coverage at each boundary. The hook contract clearly tells overrides to return unprefixed names. My prefix-contract concern is resolved.

a2105z:
@varunbiluri  Thank you for the review - hopefully we can get this merged soon then.

PR #6752, opened 2026-08-16
fix(flows): drop unpairable trailing FRs in rearrange

### LINKED ISSUE #6751: contents rearrange still raises on id-less / unmatched trailing function responses (follow-up to orphan FR prune)
### Describe the Bug

`#6582` / `73e8625` added `_drop_orphaned_function_responses`, which correctly prunes **id'd** function responses with no matching function call before contents assembly.

Two gaps remain and can still permanently poison a session (raise happens in request preprocessing, before any user callback):

1. **`_rearrange_events_for_latest_function_response` still raises** `ValueError: No function call event found for function responses ids: ...` when it cannot pair the trailing FR event. `@GWeale` noted on `#6587` that this rearrange-path drop is still worth having as defense in depth (`main` only prunes responses that carry an id).
2. **Id-less / empty-id trailing FRs are intentionally skipped by the prune helper** (`if response.id`), but then rearrange adds `None` / `''` into the id set and raises — so sessions with stripped or missing FR ids still die on the next turn.

Related hole: when rearrange finds a matching FC event but the trailing FR event also carries unmatched ids (`Last response event should only contain the responses...`), it still raises instead of dropping the unmatched parts.

### Steps to Reproduce

```python
from google.adk.flows.llm_flows.contents import _get_contents

events = [
  Event(author='user', content=types.Content(role='user', parts=[types.Part(text='hi')])),
  Event(author='agent', content=types.Content(role='model', parts=[types.Part(text='done')])),
  Event(
    author='agent',
    content=types.Content(
      role='user',
      parts=[types.Part(function_response=types.FunctionResponse(name='tool', id=None, response={'ok': True}))],
    ),
  ),
]
_get_contents(None, events, 'agent')  # ValueError: ... ids: {None}
```

### Expected Behavior

Orphaned / unpairable trailing function responses are dropped (with a warning), contents assembly continues, and the session stays usable.

### Observed Behavior

`ValueError` during contents assembly; every later turn replays the same fatal history.

### Environment

- ADK: current `main` (post-`73e8625`)
- Related: `#6582` (closed), `#6587` (closed as duplicate of landed prune; rearrange follow-up requested)

### Additional Context

Happy to send the rearrange-path follow-up PR referenced by `@GWeale` on `#6587`.

a2105z:
## Summary

Follow-up to the orphan-FR prune that landed in `73e8625` / discussion on #6587.

`_drop_orphaned_function_responses` only removes responses that carry an id. Trailing **id-less / empty-id** FRs, and rearrange itself when the prune is bypassed, could still raise during contents assembly and permanently poison the session.

This change:
- Makes `_rearrange_events_for_latest_function_response` **drop** unpairable trailing FRs instead of raising (defense in depth requested on #6587)
- Strips unmatched FR parts when a call event is found but the trailing event also carries extras (same poisoning `ValueError`)
- Leaves true id-less FC/FR adjacent pairs alone (model families that strip ids)

Fixes #6751  
Related: #6582, #6587

## Test plan

- [x] `uv run pytest tests/unittests/flows/llm_flows/test_contents_function.py tests/unittests/flows/llm_flows/test_contents.py -q` → **56 passed**
- [x] Repro cases from #6751 (`id=None`, `id=""`, multi trailing id-less, rearrange bypass, unmatched extras) no longer raise

### DIFF
--- a/src/google/adk/flows/llm_flows/contents.py
+++ b/src/google/adk/flows/llm_flows/contents.py
@@ -259,6 +259,14 @@ def _rearrange_events_for_latest_function_response(
   between the initial function_call and the latest function_response will be
   removed.
 
+  If the latest event carries function responses with no matching function
+  call in history (an orphaned FR), those responses are dropped and history
+  is rearranged from the remaining events. Raising here would permanently
+  poison the session: contents assembly runs before any user callback and
+  every later turn would replay the same fatal error. This is defense in
+  depth alongside ``_drop_orphaned_function_responses``, which only prunes
+  responses that carry an id.
+
   Args:
     events: A list of events.
 
@@ -282,9 +290,20 @@ def _rearrange_events_for_latest_function_response(
 
   if function_calls:
     for function_call in function_calls:
-      # The latest function_response is already matched
+      # The latest function_response is already matched to the previous event.
       if function_call.id in function_responses_ids:
-        return events
+        adjacent_call_ids = {
+            call.id for call in function_calls if call.id
+        }
+        # Id-less FC/FR pairs are valid for some model families that strip
+        # ids on the way out — leave them alone.
+        if not adjacent_call_ids:
+          return events
+        # Drop FR parts that do not belong to this adjacent call event so a
+        # partially matched trailing event cannot keep unpairable responses.
+        return _drop_unmatched_trailing_function_responses(
+            events, adjacent_call_ids
+        )
 
   function_call_event_idx = -1
   # look for corresponding function call event reversely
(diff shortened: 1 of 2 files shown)


a2105z:
Hi @GWeale @DeanChensj @llalitkumarrr — this is the follow-up from #6587.

After `73e8625` pruned id'd orphans, rearrange could still raise on id-less/empty-id trailing FRs (and on unmatched extras). This makes rearrange drop those instead of poisoning the session, as you suggested.

Ready for review whenever you have a chance — thank you!

PR #6754, opened 2026-08-16
feat(cli): skip __init__.py for config-based adk create

### LINKED ISSUE #6753: [Feature Request] Do not generate `__init__.py` for config-based agents

Running `adk create --type=config` generates an empty `__init__.py` alongside `root_agent.yaml`.

However, config-based agents are loaded directly from `root_agent.yaml` and do not appear to require the agent directory to be a Python package.

https://github.com/google/adk-python/blob/25e8ea6dea82713fdac9bc4117ec425386582b1f/src/google/adk/cli/utils/agent_loader.py#L217

Existing config-based samples, such as `contributing/samples/mcp/tool_mcp_stdio_notion_config`, also work without an `__init__.py`.
https://github.com/google/adk-python/tree/v2.7.0/contributing/samples/mcp/tool_mcp_stdio_notion_config

The YAML agent loader tests similarly create fixtures containing only `root_agent.yaml` and load them successfully.
https://github.com/google/adk-python/blob/v2.7.0/tests/unittests/cli/utils/test_agent_loader.py#L514

This makes the generated `__init__.py` appear unnecessary and creates an inconsistency between newly generated agents and existing samples.

When creating a config-based agent with `adk create --type=config <agent-name>`, ADK should generate:

- `.env`
- `.gitignore`
- `root_agent.yaml`

It should not generate an empty `__init__.py` unless the file is required for a documented compatibility reason.

The success message and corresponding CLI tests should be updated accordingly.

### Impact on your work

This would make generated config-based projects simpler and consistent with the repository's existing config samples.

It would also clarify whether a YAML-configured agent directory is expected to be a Python package.

### Willingness to contribute

Yes.

---

The empty `__init__.py` can be deleted manually after running `adk create`.

It can also be left in place because it does not prevent the YAML configuration from loading.
However, keeping an apparently unused generated file makes it unclear whether it is part of the required config-agent structure.

In `src/google/adk/cli/cli_create.py`:

- Do not create `__init__.py` when `type == "config"`.
- Remove `__init__.py` from `_SUCCESS_MSG_CONFIG`.
- Update `test_run_cmd_with_type_config` to assert that `__init__.py` is not generated.

Before making this change, it may be worth confirming whether the empty file intentionally prevents a collision with a Python package of the same name elsewhere on `sys.path`, such as one installed under `site-packages`.

The loader attempts to import the agent name as a Python module or package before falling back to `root_agent.yaml`.
Without `__init__.py`, the local directory may be treated as a namespace-package candidate, while a regular package with the same name elsewhere on `sys.path` may be imported instead.

This does not make `__init__.py` necessary for loading the YAML file itself, but it may provide defensive package-resolution behavior.

### Additional Context

The current loader checks Python module and package forms first, then loads `<agent-name>/root_agent.yaml`. Therefore, the YAML-loading path itself does not directly use `__init__.py`.

a2105z:
## Summary

`adk create --type=config` no longer writes an empty `__init__.py`.

Config agents are loaded from `root_agent.yaml` (after the Python module/package probes miss `root_agent`). Existing config samples and the YAML agent-loader fixtures already omit `__init__.py`, so generated projects now match that layout.

Also updates the config success message and CLI unit expectations.

**On the package-resolution note in #6753:** an empty `__init__.py` did not expose `root_agent` either, so the loader still fell through to YAML. With `agents_dir` prepended on `sys.path`, the local agent directory remains the first candidate; this change aligns scaffolding with samples rather than relying on an empty package marker.

Fixes #6753

## Test plan

- [x] Repro before: `adk create --type=config` emitted `.env`, `.gitignore`, `__init__.py`, `root_agent.yaml`
- [x] After: emits `.env`, `.gitignore`, `root_agent.yaml` only
- [x] `uv run pytest tests/unittests/cli/utils/test_cli_create.py -q` → **36 passed**
- [x] Code-type create still generates `__init__.py` + `agent.py`

### DIFF
--- a/src/google/adk/cli/cli_create.py
+++ b/src/google/adk/cli/cli_create.py
@@ -65,7 +65,6 @@ _SUCCESS_MSG_CONFIG = """
 Agent created in {agent_folder}:
 - .env
 - .gitignore
-- __init__.py
 - root_agent.yaml
 
 ⚠️  WARNING: Secrets (like GOOGLE_API_KEY) are stored in .env.
@@ -139,8 +138,9 @@ def _generate_files(
   if type == "config":
     with open(agent_config_file_path, "w", encoding="utf-8") as f:
       f.write(_AGENT_CONFIG_TEMPLATE.format(model_name=model))
-    with open(init_file_path, "w", encoding="utf-8") as f:
-      f.write("")
+    # Config agents are loaded from root_agent.yaml; they are not Python
+    # packages. Matching existing config samples, do not emit an empty
+    # __init__.py.
     click.secho(
         _SUCCESS_MSG_CONFIG.format(agent_folder=agent_folder),
         fg="green",
--- a/tests/unittests/cli/utils/test_cli_create.py
+++ b/tests/unittests/cli/utils/test_cli_create.py
@@ -287,10 +287,8 @@ def test_run_cmd_with_type_config(
   assert "model: gemini-2.5-flash" in yaml_content
   assert "description: A helpful assistant for user questions." in yaml_content
 
-  # Should create empty __init__.py
-  init_file = agent_dir / "__init__.py"
-  assert init_file.exists()
-  assert init_file.read_text().strip() == ""
+  # Config agents are YAML-loaded; do not emit a package marker.
+  assert not (agent_dir / "__init__.py").exists()
 
   # Should still create .env file
   env_file = agent_dir / ".env"


a2105z:
Hi @ftnext — follow-up on #6753.

`adk create --type=config` now matches the existing config samples: `.env`, `.gitignore`, and `root_agent.yaml` only (no empty `__init__.py`). Ready for review whenever maintainers have a chance.

PR #6764, opened 2026-08-17
fix(flows): pair async function responses on (id, name) not id alone

### LINKED ISSUE #6761: History rebuild assumes model-supplied function-call ids are unique; Gemini reuses them, pairing a call with another tool's response (400 INVALID_ARGUMENT)
### Summary

`_rearrange_events_for_async_function_responses_in_history` pairs function responses to calls **by id alone**. Model-supplied ids are not guaranteed unique — Vertex Gemini mints `call_<n>` ids from a space small enough to repeat within a single session. When one repeats, the assembled history pairs a `function_call` for one tool with a `function_response` for another, and the next `generateContent` fails with a bare `400 INVALID_ARGUMENT`.

Intermittent and content-dependent, so it presents as a flaky 400 rather than a reproducible one. Measured rate below: **1 in ~153 requests**.

### Environment

- `google-adk` 2.6.3
- Vertex AI, `gemini-3.5-flash`, EU multi-region endpoint (`location="eu"`)
- Tools declared as `FunctionTool`; one gated with `require_confirmation=True`
- `InMemoryRunner`, non-streaming (`StreamingMode.NONE`)

### What the rejected request contains

Captured by wrapping `Gemini.generate_content_async` and dumping `llm_request.contents` on `ClientError`:

```
 1  model  function_call     id=call_807  name=site_security_posture
 2  user   function_response id=call_807  name=fleet_security_summary   <-- wrong tool
 ...
11  model  function_call     id=call_807  name=fleet_security_summary   <-- id reused
12  user   function_response id=call_807  name=fleet_security_summary
```

Content 2 is the response to the call at content 1, but it has been paired with — and renamed after — the *later* call that reused the same id.

```
google.genai.errors.ClientError: 400 INVALID_ARGUMENT.
{'error': {'code': 400, 'message': 'Request contains an invalid argument.', 'status': 'INVALID_ARGUMENT'}}
```

### Mechanism

`google/adk/flows/llm_flows/contents.py:145` (reached from `:832` while building request contents):

```python
function_call_id_to_response_events_index: dict[str, int] = {}
for i, event in enumerate(events):
    function_responses = event.get_function_responses()
    if function_responses:
        for function_response in function_responses:
            function_call_id = function_response.id
            function_call_id_to_response_events_index[function_call_id] = i
```

The index is built in **forward** order, so a repeated id retains only the **newest** response. The second pass (`:165-178`) then walks the events and appends that response directly after the **oldest** call carrying the same id — producing the mismatched pair above.

ADK keeps model-supplied ids: `remove_client_function_call_id` (`functions.py:263`) strips only ids carrying the `adk-` prefix (`AF_FUNCTION_CALL_ID_PREFIX`, `:59`), and `populate_client_function_call_id` (`:255`) assigns one only when `not function_call.id`. So a repeated model id survives into the session and reaches this rebuild.

### How often

Instrumenting every request in one run: **69 distinct ids across 153 requests**, all of the form `call_<n>` in the range `call_187`–`call_1127`, with **one collision** — and that collision ended the run. Longer sessions with more tool calls raise the rate.

### Reproducing

Any session long enough for Gemini to repeat an id. Wrapping `Gemini.generate_content_async` to log each request's `(id, name)` pairs makes it visible directly — the collision is the first request in which one id maps to two different call names.

### Suggested fixes

Either would resolve it:

1. **Do not trust model-supplied ids** — always mint an ADK id in `populate_client_function_call_id`. They are stripped before the request anyway, so the model never sees them and nothing on the wire changes. This also makes `find_event_by_function_call_id` (`functions.py:1401`, used on the response-resume path) safe, since it carries the same uniqueness assumption.
2. **Pair on id *and* name** in the history rebuild, so a duplicate id cannot resolve across tools.

(1) seems preferable: it fixes the assumption once rather than at each site that relies on it.

### Workaround, and a wrinkle worth flagging

Subclassing the model to clear `part.function_call.id` on every yielded response, which routes every call through `populate_client_function_call_id`. Zero model-supplied ids then reach the wire and the 400 stops.

One consequence others attempting this should know: `_finalize_model_response_event` runs on **every** yielded response including `partial=True` ones (`base_llm_flow.py:1152`; execution is skipped afterwards at `:1163`), so a fresh uuid is minted per partial and the partial's id differs from the final's. Since only the final event is persisted, anything reading a call id off a partial event gets a handle that resolves to nothing.

That is an argument for fix (1) landing upstream, where the id can be assigned once per call rather than once per yielded response.

a-yeyang:

- Closes: #6761

**Problem:**

`_rearrange_events_for_async_function_responses_in_history` pairs function responses to calls **by `function_response.id` alone**. Model-supplied ids are not guaranteed unique — Vertex Gemini mints `call_<n>` ids from a space small enough to repeat within a single session, and ADK preserves model-supplied ids (`remove_client_function_call_id` strips only `adk-`-prefixed ids; `populate_client_function_call_id` assigns one only when the id is empty). So a repeated id survives into this rebuild.

Because the index is built in forward order, a repeated id keeps only the **newest** response. The second pass then appends that response directly after the **oldest** call carrying the same id, pairing a `function_call` for one tool with the `function_response` of another. On the next `generateContent` the model rejects the contents with a bare `400 INVALID_ARGUMENT`. The reporter measured this at ~1 collision in 153 requests in a real session (`gemini-3.5-flash`, Vertex EU), where a single collision ended the run.

**Solution:**

Key the pairing on `(id, name)` instead of `id`. ADK always sets `function_response.name` to the tool name (`Part.from_function_response(name=tool.name, ...)`), and the matching `function_call` carries the same name, so:

- for every non-colliding history the pair is unchanged (behavior-preserving);
- a collision can no longer resolve across two different tools, because the tool name disambiguates it.

This is the narrowest change that fixes the reported, reproduced behavior; it touches only the pairing key in this one function.

### Testing Plan

**Unit Tests:**

- [x] I have added a regression test (`test_rearrange_async_function_responses_pairs_on_id_and_name`) that reproduces the reported shape — two tools (`site_security_posture`, `fleet_security_summary`) both reusing id `call_807`, interleaved with their responses — and asserts each call is followed by the response for the **same** tool. It fails on the old id-only keying and passes with this change.
- [x] All unit tests in `tests/unittests/flows/llm_flows/test_contents.py` pass locally.

```
$ python -m pytest tests/unittests/flows/llm_flows/test_contents.py
42 passed
```

The failure is intermittent and content-dependent (it requires the model to reuse an id across two different tools within one session), so it is exercised through the unit test above rather than a manual E2E run. The reproduction in the issue captures the exact rejected-request contents.

### Checklist

### DIFF
--- a/src/google/adk/flows/llm_flows/contents.py
+++ b/src/google/adk/flows/llm_flows/contents.py
@@ -144,15 +144,25 @@ def _rearrange_events_for_async_function_responses_in_history(
     events: list[Event],
 ) -> list[Event]:
   """Rearrange the async function_response events in the history."""
-  function_call_id_to_response_events_index: dict[str | None, int] = {}
+  # Pair responses to calls on (id, name), not id alone. Model-supplied function
+  # call ids are not guaranteed unique: Vertex Gemini mints ``call_<n>`` ids from
+  # a small space that can repeat within one session, and ADK preserves such ids
+  # (only ``adk-`` prefixed ids are generated/stripped by the framework). Keying
+  # on id alone lets a repeated id pair a ``function_call`` for one tool with the
+  # ``function_response`` of another, which the model rejects with a bare
+  # ``400 INVALID_ARGUMENT``. Including the tool name keeps a collision from
+  # resolving across different tools. See issue #6761.
+  function_call_key_to_response_events_index: dict[
+      tuple[str | None, str | None], int
+  ] = {}
   for i, event in enumerate(events):
     function_responses = event.get_function_responses()
     if function_responses:
       for function_response in function_responses:
-        function_call_id = function_response.id
-        function_call_id_to_response_events_index[function_call_id] = i
+        function_call_key = (function_response.id, function_response.name)
+        function_call_key_to_response_events_index[function_call_key] = i
 
-  if not function_call_id_to_response_events_index:
+  if not function_call_key_to_response_events_index:
     return events
 
   result_events: list[Event] = []
@@ -164,10 +174,10 @@ def _rearrange_events_for_async_function_responses_in_history(
 
       function_response_events_indices = set()
       for function_call in event.get_function_calls():
-        function_call_id = function_call.id
-        if function_call_id in function_call_id_to_response_events_index:
+        function_call_key = (function_call.id, function_call.name)
+        if function_call_key in function_call_key_to_response_events_index:
           function_response_events_indices.add(
-              function_call_id_to_response_events_index[function_call_id]
+              function_call_key_to_response_events_index[function_call_key]
           )
       result_events.append(event)
       if not function_response_events_indices:
--- a/tests/unittests/flows/llm_flows/test_contents.py
+++ b/tests/unittests/flows/llm_flows/test_contents.py
@@ -1964,6 +1964,67 @@ def test_rearrange_async_function_responses_early_returns_when_no_responses():
   assert result is events
 
 
+def test_rearrange_async_function_responses_pairs_on_id_and_name():
+  """A reused function-call id must not pair a call with another tool's response.
+
+  Model-supplied ids (e.g. Vertex Gemini's ``call_<n>``) can repeat within a
+  session. Keying only on id let the newest response for a duplicated id attach
+  to the oldest call carrying that id, mismatching the tool name and producing a
+  ``400 INVALID_ARGUMENT`` on the next request. See issue #6761.
+  """
+
+  def _call_event(invocation_id: str, call_id: str, name: str) -> Event:
+    return Event(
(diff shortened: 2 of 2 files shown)


GWeale:
this is already fixed in v2.8.0, so upgrading is enough. the commit is deee6d2c.

the fix that went in pairs responses slightly differently to yours. yours matches on the id plus the tool name. that one matches on the id plus which call the response came after. the difference shows up when the same tool is called twice in a turn: matching on the name hands back the second response both times, matching on position gets each one right.

#6761 is closed with it. thanks for the patch and the repro. if responses still pair to the wrong call on 2.8.0, open a new issue.

PR #6780, opened 2026-08-17
fix(agents): safely decouple model memoization to prevent Agent Engine crashes

### LINKED ISSUE #6757: 2.7.0 on Agent Engine: worker dies without a traceback on the second model call after a tool round-trip, truncating the stream
## 🔴 Required Information

**Bug description:**

After upgrading `google-adk` 2.6.2 → 2.7.0 (with `google-cloud-aiplatform` 1.163.0 → 1.164.0), every invocation that involves a tool call fails on Vertex AI Agent Engine. The uvicorn worker process handling the request **disappears without emitting a Python traceback**, and the parent process starts a replacement worker a few seconds later. The `:streamQuery` response ends after two events, so the client sees a well-formed HTTP 200 stream that simply contains no model text.

The invocation dies **after** the tool round-trip completes successfully — the failure is in the second model call of the turn.

Reverting only the two dependencies (no application code change) makes the same question complete normally.

**Steps to Reproduce:**

1. Build an `LlmAgent` as the root agent (`coordinator`) with:
   - a function tool that performs network I/O (in our case a Workspace search returning ~5 results),
   - seven `sub_agents`, all `mode="single_turn"` (five domain specialists, one artifact reader, one grounding judge),
   - `before_model_callback` that sets `tool_config` (`FunctionCallingConfigMode.ANY`) on the first LLM call of a turn,
   - `after_agent_callback` returning `types.Content` when the root produced no visible text,
   - `BigQueryAgentAnalyticsPlugin` registered via `AdkApp(agent=..., plugins=[...])`.
2. Deploy to Agent Engine (`vertexai.agent_engines`, `agent_framework="google-adk"`), region `asia-northeast1`, Python 3.13.
3. Send a question through `:streamQuery` that requires the tool.

**Expected Behavior:**

The turn proceeds: model → tool call → tool result → model → delegation to a sub-agent → final text. This is exactly what 2.6.2 does (14 events, final answer with citations).

**Observed Behavior:**

The stream ends after 2 events:

```jsonc
// event 1 — coordinator, function_call + thought_signature
{"author":"coordinator","content":{"parts":[{"function_call":{"name":"search_workspace_documents", ...},
 "thought_signature":"..."}],"role":"model"},"finish_reason":"STOP", ...}

// event 2 — coordinator, function_response (tool succeeded, 5 results)
{"author":"coordinator","content":{"parts":[{"function_response":{"name":"search_workspace_documents",
 "response":{"status":"success","results":[ ... 5 items ... ]}}}],"role":"user"}, ...}

// nothing further
```

Runtime logs for the same invocation:

```
03:40:04.415  [26]  POST /api/stream_reasoning_engine HTTP/1.1" 200 OK
03:40:06.514  [26]  Sending out request, model: gemini-2.5-flash, backend: GoogleLLMVariant.VERTEX_AI, stream: False
03:40:07.692  [26]  Response received from the model.
              (tool executes successfully, ~5.4 s)
03:40:13.100  [26]  Sending out request, model: gemini-2.5-flash, backend: GoogleLLMVariant.VERTEX_AI, stream: False
03:40:17.518  [413] gcsfs experimental features enabled ...      <- different PID: a new worker booted
03:40:17.529  [413] Plugin 'bigquery_agent_analytics' registered.
03:40:17.532  [413] Application startup complete.
```

Worker `[26]` never logs `Response received from the model.` for the second call. **No traceback, no ERROR-level entry** appears anywhere in the invocation window, which points to the process being killed by a signal rather than raising.

**What we ruled out:**

| Hypothesis | Result |
| --- | --- |
| Model generation | Reproduced with both `gemini-2.5-flash` and `gemini-3.5-flash` |
| The tool itself | Tool completed in 5.4 s and returned `status: success` with 5 results |
| Auth / credentials | `status: success`, no `auth_required` |
| Application code | Unchanged across the working and failing deploys; only the two dependencies differ |
| `tool_config` forcing not being released | Verified with a real `Runner` on 2.7.0: the second call has `tool_config=None` as expected |
| Dependency set growth | `uv export --only-group runtime` diff is 3 lines (adk, aiplatform, and `google-cloud-iam` dropped); no new packages |

**Possibly related observation (not a proven cause):**

Container memory allocation for the same question roughly doubled: ~0.8 GiB on 2.6.2 vs ~1.6–2.0 GiB on 2.7.0 (Agent Engine console metric).

We also measured that `LlmAgent._resolved_model` (new in 2.7.0) keeps one resolved `BaseLlm` per agent. With 8 agents that all pass the model as a **string**, this retains 8 distinct `Gemini` instances and 8 distinct `genai.Client` objects (+18.5 MB RSS per worker in our measurement). That alone is far too small to explain a kill, but it does mean 2.7.0 retains state where 2.6.2 let it be collected.

**Environment Details:**

- ADK version: 2.7.0 (working: 2.6.2)
- google-cloud-aiplatform: 1.164.0 (working: 1.163.0)
- Python: 3.13
- Runtime: Vertex AI Agent Engine (`reasoningEngines`), region `asia-northeast1`, 11 uvicorn workers
- Model: `gemini-2.5-flash` / `gemini-3.5-flash`, `stream: False`, Vertex backend
- Telemetry: `GOOGLE_CLOUD_AGENT_ENGINE_ENABLE_TELEMETRY=true`
- Plugins: `BigQueryAgentAnalyticsPlugin`

Happy to provide the full NDJSON stream, additional log windows, or to run a targeted experiment (e.g. with telemetry disabled) if that would help narrow it down.

manideep-malyala:
**Fixes #6757**

### Summary
In v2.7.0, a performance optimization was introduced to `LlmAgent` that memoized the model adapter via `_resolved_model`. This inadvertently caused `LlmAgent` instances to retain permanent references to `BaseLlm` objects.

Because `Gemini` aggressively caches its underlying `google.genai.Client` and network connections, sharing these stateful clients across multiple invocations or tool turn boundaries in a web server environment (like Vertex AI Agent Engine) leads to severe memory bloating and silent OS-level worker kills due to cross-thread state corruption.

### The Fix
This PR reverts the `_resolved_model` memoization in `src/google/adk/agents/llm_agent.py`, safely returning `LlmAgent` to its stateless v2.6.2 behavior. `canonical_model` and `canonical_live_model` now dynamically resolve the string on each call. This allows background workers to properly garbage collect the network channels at the end of the invocation, eliminating the silent crashing.

Additionally, two tests (`test_canonical_model_str_resolved_once`, `test_canonical_live_model_str_resolved_once`) that explicitly enforced this buggy caching pattern were removed.

### Testing Plan
- **Unit Tests:** Run the full `agents` test suite locally to verify dynamic model resolution remains intact.
- **E2E Validation:** Confirmed that the fix mathematically matches the behavior of v2.6.2, which the original issue reporter verified works perfectly on Agent Engine.

**Test Results:**
```
tests/unittests/agents/test_llm_agent_fields.py ........................ [ 31%]
.....................................................                    [100%]
======================== 77 passed, 5 warnings in 1.33s ========================
```

***


DeanChensj:
Hi @manideep-malyala, thank you for putting together this PR and investigating #6757.
After evaluating the trace and follow-up benchmarks on #6757, we have decided not to proceed with reverting `_resolved_model` memoization for the following reasons:
1. **Not the root cause of the crash**:
   - The issue reporter (@TakumaKurosawa) confirmed in [#6757 comments](https://github.com/google/adk-python/issues/6757#issuecomment-5321622305) that the same worker termination reproduces on `v2.6.2` (where `_resolved_model` did not exist).
   - The retained memory overhead was measured at only ~18.5 MB, which does not trigger an OOM.
   - Users who instantiate `LlmAgent(model=my_llm_instance)` already retain the `BaseLlm` reference across turns by design.
2. **Severe performance and connection regression**:
   - `canonical_model` is read multiple times within a single turn (in `basic.py`, `contents.py`, `_output_schema_processor.py`, and `base_llm_flow.py`).
   - Reverting to calling `LLMRegistry.new_llm()` on every property access would instantiate multiple new `genai.Client` instances and fresh connection pools (4–5 times per LLM turn), causing severe socket churn and latency degradation.
Because removing the memoization would degrade performance without resolving the underlying Agent Engine issue, we are closing this PR. Thank you again for your time and contribution!

PR #6781, opened 2026-08-18
fix(a2a): advertise streaming on agent_registry agent cards

### LINKED ISSUE #6778: build_agent_card(): agent_registry cards stay streaming:false after #6673, and the streaming= parameter becomes unreachable
Splitting this out of #6672 as suggested there, so it does not block #6673.

Two things, one root cause. In `_compat.build_agent_card()`:

```python
default_capabilities = {"streaming": streaming, "push_notifications": False}
...
"capabilities": _as_dict(capabilities) or default_capabilities,
```

The `streaming=` parameter only reaches the card through `default_capabilities`, and that is consulted only when `capabilities` is falsy. The two call sites take different branches, and #6673 changes which branch each one takes.

### 1. Cards built through `agent_registry` stay `streaming: false`

`src/google/adk/integrations/agent_registry/agent_registry.py:640` passes neither `capabilities` nor `streaming`:

```python
agent_card = _compat.build_agent_card(
    name=name,
    description=description,
    version=version,
    url=url,
    protocol_binding=getattr(binding, "value", binding),
    protocol_version=protocol_version,
    skills=skills,
    default_input_modes=["text"],
    default_output_modes=["text"],
)
```

So it falls to `default_capabilities`, where `streaming` defaults to `False`. #6673 changes `AgentCardBuilder`'s default, which this call site does not go through, so a card built here advertises `streaming: false` after that fix as well as before it.

### 2. `streaming=` becomes unreachable

`src/google/adk/a2a/utils/agent_card_builder.py:85` passes `capabilities=self._capabilities` (line 94) and never passes `streaming=`. #6673 changes that field to:

```python
self._capabilities = capabilities or AgentCapabilities(streaming=True)
```

which is always truthy, so `_as_dict(capabilities)` wins and `default_capabilities` is never evaluated from this caller. Together with (1), no call site in the repository can reach the `streaming=` parameter, and a caller who sets it gets no effect and no warning.

### Reproduced

`google-adk 2.6.1`, `a2a-sdk 1.1.2`, Python 3.13.5, clean venv. `build_agent_card` is byte-identical between 2.6.1 and current main, so this applies to main.

```python
from google.adk.a2a import _compat
from a2a.types import AgentCapabilities

binding = getattr(_compat.TP_HTTP_JSON, "value", _compat.TP_HTTP_JSON)
common = dict(name="x", description="d", version="1", url="http://h/a",
              skills=[], default_input_modes=["text"],
              default_output_modes=["text"])

# the agent_registry call shape
card = _compat.build_agent_card(protocol_binding=binding, **common)
print(card.capabilities.streaming)                    # False

# streaming= is honoured, but only when capabilities is absent
card = _compat.build_agent_card(protocol_binding=binding, streaming=True, **common)
print(card.capabilities.streaming)                    # True

# once capabilities is passed, which is what #6673 makes the default, it is ignored
card = _compat.build_agent_card(protocol_binding=binding,
                                capabilities=AgentCapabilities(streaming=True),
                                streaming=False, **common)
print(card.capabilities.streaming)                    # True, streaming= had no effect
```

### Possible shapes

1. Thread `capabilities` through the `agent_registry` call site the way `AgentCardBuilder` does, and drop `streaming=` from `build_agent_card` once nothing reaches it.
2. Keep `streaming=` and make it compose rather than fall back, so it applies on top of a passed `capabilities`. This changes the meaning of an existing parameter for anyone already passing both.
3. Keep `streaming=` and document it as the no-capabilities convenience path only.

I have not opened a pull request because (2) changes the behaviour of an existing parameter and (1) removes public surface, so the choice looks like yours rather than mine. Happy to send whichever you prefer.

lllakshit:

- Closes: #6778
- Related: #6672, #6673

**Problem:**
`AgentRegistry.get_remote_a2a_agent()` builds a card through `_compat.build_agent_card()` with neither `capabilities` nor `streaming`. The helper then falls through to `default_capabilities` where `streaming` defaults to `False`, so registry-constructed cards advertise `streaming: false`.

The issue also asked which public-API shape to use for the `streaming=` parameter. This PR takes the non-breaking option: keep `streaming=` as the no-capabilities convenience path only. It does **not** drop `streaming=` (that would remove public surface) and does **not** compose `streaming=` onto a passed `capabilities` object (that would change the meaning of an existing parameter; current tests already lock in "capabilities wins").

`AgentCardBuilder` still defaults `AgentCapabilities()` without `streaming=True` on GitHub `main` (#6673 closed without landing here). That is #6672 and is left out of this PR so this stays one concern.

**Solution:**
Pass `streaming=True` at the `agent_registry` call site, which is the convenience path `build_agent_card` already implements. Document that `streaming=` is ignored when `capabilities` is provided.

### Testing Plan

**Unit Tests:**

Commands:

```text
pytest tests/unittests/a2a/test_compat.py tests/unittests/integrations/agent_registry/test_agent_registry.py
```

Summary (Python 3.12.10, a2a-sdk 1.1.2, matching the issue repro):

```text
71 passed, 24 skipped in 4.98s
```

Skipped tests are the existing 0.3-only `@v03_only` cases; this environment is a2a-sdk 1.x. New coverage that runs on 1.x:

- helper without `capabilities` still defaults `streaming` to `False`
- `streaming=True` is honoured when `capabilities` is omitted
- a passed `capabilities` object still wins (no composition)
- `get_remote_a2a_agent()` constructed cards advertise `streaming is True`

Reproduced the issue snippet against GitHub `main` vs this branch (`a2a-sdk 1.1.2`):

```text
main agent_registry shape (omit streaming=):  False
this PR (streaming=True, no capabilities):    True
capabilities=AgentCapabilities(streaming=True), streaming=False: True
```

`git show upstream/main:.../agent_registry.py` still has no `streaming=` on that call. After this change the same call passes `streaming=True`.

No `adk web` UI change; this is card construction only.

### Checklist

### Additional context

I will complete the Google CLA if it is not already on file.

Assisted by an AI coding tool; I reviewed the diff, ran the tests above, and confirmed the issue still reproduces on `main`.

### DIFF
--- a/src/google/adk/integrations/agent_registry/agent_registry.py
+++ b/src/google/adk/integrations/agent_registry/agent_registry.py
@@ -51,6 +51,7 @@ from typing_extensions import override
 
 # pylint: disable=g-import-not-at-top
 try:
+  from a2a.types import AgentCapabilities
   from a2a.types import AgentSkill
   from google.adk.a2a import _compat
   from google.adk.agents.remote_a2a_agent import RemoteA2aAgent
@@ -666,6 +667,8 @@ class AgentRegistry:
       )
 
     binding = protocol_binding or _compat.TP_HTTP_JSON
+    # Registry metadata has no capability details. Pass streaming explicitly so
+    # constructed cards do not advertise streaming:false.
     agent_card = _compat.build_agent_card(
         name=name,
         description=description,
@@ -674,6 +677,7 @@ class AgentRegistry:
         protocol_binding=getattr(binding, "value", binding),
         protocol_version=protocol_version,
         skills=skills,
+        capabilities=AgentCapabilities(streaming=True),
         default_input_modes=["text"],
         default_output_modes=["text"],
     )
--- a/tests/unittests/integrations/agent_registry/test_agent_registry.py
+++ b/tests/unittests/integrations/agent_registry/test_agent_registry.py
@@ -615,6 +615,29 @@ class TestAgentRegistry:
         agent._agent_card, "1.0" if _compat.IS_A2A_V1 else "0.3.0"
     )
 
+  def test_get_remote_a2a_agent_advertises_streaming(self, registry):
+    """Constructed registry cards advertise streaming when no card is supplied."""
+    mock_response = MagicMock()
+    mock_response.json.return_value = {
+        "displayName": "TestAgent",
+        "description": "Test Desc",
+        "version": "1.0",
+        "protocols": [{
+            "type": _ProtocolType.A2A_AGENT,
+            "interfaces": [{
+                "url": "https://my-agent.com",
+            }],
+        }],
+    }
+    mock_response.raise_for_status = MagicMock()
+    registry._session.get.return_value = mock_response
+
+    registry._credentials.token = "token"
+    registry._credentials.refresh = MagicMock()
+
+    agent = registry.get_remote_a2a_agent("test-agent")
+    assert agent._agent_card.capabilities.streaming is True
+
   def test_get_remote_a2a_agent_with_card(self, registry):
     mock_response = MagicMock()
     mock_response.json.return_value = {


lllakshit:
@googlebot rescan

I have already signed the Google Individual CLA (Lakshit Mathur / lllakshit / llakshitmathur239@gmail.com). Please re-check this pull request.

lllakshit:
@DeanChensj Could you please review this PR and let me know if any changes are required? I’d appreciate your feedback so I can address any remaining concerns and move it toward merging into main.

lllakshit:
Rebased this branch onto current main. The CLA check is green and the change is unchanged. Happy to address any review comments.

lllakshit:
Rebased onto current main and updated for the upstream API change that removed the \streaming=\ convenience parameter from \uild_agent_card\. The registry call site now passes \capabilities=AgentCapabilities(streaming=True)\ explicitly instead.

CLA is still green. Ready for review.

PR #6788, opened 2026-08-18
fix: Port agent loader and file path resolution fixes to v1

GWeale:
Ports two commits from `main` to `v1`.

1. Agent name gating in the CLI server (`3f6eb1f0`)
   - `AgentLoader.load_agent` raises `PermissionError` for a `__`-prefixed name; set `_allow_special_agents` on the loader instance to keep loading them.
   - `adk api_server` answers 403 for those names, including the eval and graph routes. `adk web` still serves them, via the existing `web` argument to `get_fast_api_app`.

2. Path scoping in the Agent Builder file tools (`1fa7cda9`)
   - `resolve_file_path` raises `ValueError` when the path resolves outside the root directory; the file tools return `{"success": False, ...}`.
   - It returns a fully resolved path, with `..` and symlinks collapsed.

The server-side check is re-implemented for `v1`'s single `AdkWebServer` rather than ported.

### DIFF
--- a/src/google/adk/cli/adk_web_server.py
+++ b/src/google/adk/cli/adk_web_server.py
@@ -731,6 +731,9 @@ class AdkWebServer:
 
   async def get_runner_async(self, app_name: str) -> Runner:
     """Returns the cached runner for the given app."""
+    # Rejected before any cleanup, cache lookup or .env walk, so a refused
+    # name cannot reach those paths.
+    self._reject_special_agent(app_name)
     # Handle cleanup
     if app_name in self.runners_to_clean:
       self.runners_to_clean.remove(app_name)
@@ -799,8 +802,8 @@ class AdkWebServer:
     self.runner_dict[app_name] = runner
     return runner
 
-  def _load_agent_or_app(self, app_name: str) -> BaseAgent | App:
-    """Loads an agent, refusing internal special agents unless enabled."""
+  def _reject_special_agent(self, app_name: str) -> None:
+    """Refuses internal special agents unless they are explicitly enabled."""
     if app_name.startswith("__") and not self._allow_special_agents:
       raise HTTPException(
           status_code=403,
@@ -809,6 +812,10 @@ class AdkWebServer:
               " mode."
           ),
       )
+
+  def _load_agent_or_app(self, app_name: str) -> BaseAgent | App:
+    """Loads an agent, refusing internal special agents unless enabled."""
+    self._reject_special_agent(app_name)
     return self.agent_loader.load_agent(app_name)
 
   def _get_root_agent(self, agent_or_app: BaseAgent | App) -> BaseAgent:
--- a/src/google/adk/cli/adk_web_server.py
+++ b/src/google/adk/cli/adk_web_server.py
@@ -690,6 +690,8 @@ class AdkWebServer:
       runner_dict: A dict of instantiated runners for each app.
   """
 
+  _allow_special_agents: bool = False
+
   def __init__(
       self,
       *,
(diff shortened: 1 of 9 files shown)


src/google/adk/cli/adk_web_server.py
```diff
@@ -741,7 +743,7 @@ async def get_runner_async(self, app_name: str) -> Runner:
 
     # Create new runner
     envs.load_dotenv_for_agent(os.path.basename(app_name), self.agents_dir)
-    agent_or_app = self.agent_loader.load_agent(app_name)
+    agent_or_app = self._load_agent_or_app(app_name)
```
xuanyang15, line 749: Upstream makes this check the first statement of `get_runner_async` (`origin/main:api_server.py:866`, right after the docstring), ahead of the cleanup at :874 and the `runner_dict` lookup at :881. Here the helper runs after both, and after `envs.load_dotenv_for_agent` on the line above, so a name we are about to reject still triggers a walk to the root looking for a `.env`, and a cached runner would be returned without ever reaching the 403.

Today that second part is not reachable: `runner_dict` is only written at :799, which is behind this helper, so a `__` name can never get cached while `_allow_special_agents` is off. But nothing pins that invariant, and it is the only thing holding the guard up on this path.

Should we check the prefix at the top of `get_runner_async` as well, matching upstream's ordering, and keep the helper for the other five call sites? Something like:

```python
async def get_runner_async(self, app_name: str) -> Runner:
  """Returns the cached runner for the given app."""
  self._reject_special_agent(app_name)
  # Handle cleanup
  ...
```

`_allow_special_agents` is already readable at :691, so the first statement can use it. A test asserting 403 even when `runner_dict` already holds the key would then lock it in. What do you think?

tests/unittests/cli/test_resolve_root_directory.py
```diff
@@ -0,0 +1,140 @@
...
+  resolved = resolve_file_path(str(target), {"root_directory": str(tmp_path)})
+  assert resolved == target.resolve()
+
+
+def test_resolve_file_path_rejects_relative_traversal(tmp_path):
+  with pytest.raises(ValueError):
+    resolve_file_path("../../escape.txt", {"root_directory": str(tmp_path)})
+
+
+def test_resolve_file_path_rejects_absolute_outside_root(tmp_path):
```
xuanyang15, line 67: `main` gained two more tests for this module after `1fa7cda9`, in `456524d71`: `test_resolve_file_paths_preserves_input_order` and `test_resolve_file_paths_rejects_the_whole_batch_on_one_escape`, whose docstring is "One traversal attempt must fail the batch, not be silently dropped." Diffing this file against `main`'s copy, those two tests plus the `resolve_file_paths` import at :26 are the only things missing.

That plural entry point is the one `delete_files.py:66` and `read_files.py:56` actually call, and whole-batch rejection is what keeps a delete from being partially applied. Can we pull those two along, so the ported tests match what `main` has today?

xuanyang15 (approved):

src/google/adk/cli/adk_web_server.py
GWeale: moved it to the first statement of `get_runner_async`, matching upstream. the check itself is now `_reject_special_agent`, and `_load_agent_or_app` calls the same method so the other call sites keep it.

added `test_get_runner_async_rejects_special_agent_already_in_the_cache`, which seeds `runner_dict` with the refused name first. deleting the new guard line makes it fail while the existing test still passes, so it pins the invariant you flagged rather than the path already covered.

tests/unittests/cli/test_resolve_root_directory.py
GWeale: pulled both across verbatim, plus the `resolve_file_paths` import. the file now matches `main`'s copy.

PR #6795, opened 2026-08-18
fix(artifacts): Port app-scoped file artifact storage and layout-derived payload paths to v1

GWeale:
Ports three `FileArtifactService` commits to the `v1` branch. The on-disk layout changes — read the migration note first.

## Migration

Artifacts live under `root/apps/{app_name}/users/{user_id}/...`. Move an existing `root/users/` tree to `root/apps/<app_name>/users/` for the app that owns it to keep it reachable. Roots by entry point:

| entry point | root |
|---|---|
| `adk web`, `adk api_server` | `<agents_dir>/.adk/artifacts` |
| `adk run` | `<agents_dir>/<agent_folder>/.adk/artifacts` |

## 1. Namespace file artifacts by app

`f72f0db5`, with follow-up `c27d8688`.

- Save, load, list, delete and both version APIs are scoped to `app_name`.
- `app_name` must be a path segment: non-empty, no null byte, no `/` or `\`, not `.` or `..`; otherwise `InputValidationError`.

## 2. Derive payload paths from the layout

`c5672030`.

- The payload location and the returned `canonical_uri` are derived from the storage layout.
- A save that fails partway removes the version directory.
- The metadata document is written to a temp file and renamed into place, at the umask-derived mode.
- `metadata.json`, in any casing, is rejected as an artifact name at save time.
- An artifact whose payload file is missing loads as `None`.

### DIFF
--- a/src/google/adk/artifacts/file_artifact_service.py
+++ b/src/google/adk/artifacts/file_artifact_service.py
@@ -184,6 +184,12 @@ def _metadata_path(artifact_dir: Path, version: int) -> Path:
   return _versions_dir(artifact_dir) / str(version) / "metadata.json"
 
 
+def _canonical_uri(artifact_dir: Path, version: int) -> str:
+  """Builds the canonical file:// URI for an artifact payload."""
+  payload_path = _versions_dir(artifact_dir) / str(version) / artifact_dir.name
+  return payload_path.resolve().as_uri()
+
+
 def _list_versions_on_disk(artifact_dir: Path) -> list[int]:
   """Returns sorted versions discovered under the artifact directory."""
   versions_dir = _versions_dir(artifact_dir)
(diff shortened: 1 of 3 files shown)


src/google/adk/artifacts/file_artifact_service.py
```diff
@@ -244,62 +331,53 @@ def __init__(self, root_dir: Path | str):
     self.root_dir = Path(root_dir).expanduser().resolve()
     self.root_dir.mkdir(parents=True, exist_ok=True)
 
-  def _base_root(self, user_id: str, /) -> Path:
-    """Returns the artifacts root directory for a user."""
+  def _base_root(self, app_name: str, user_id: str) -> Path:
+    """Returns the app-scoped root holding a user's artifacts."""
+    _validate_path_segment(app_name, "app_name")
```
xuanyang15: `app_name` starts reaching the filesystem path here, and it is validated by the module-private `_validate_path_segment` (215), which rejects `/` and `\` anywhere in the value. Upstream validates the same argument with `artifact_util.validate_path_segment` (`main:file_artifact_service.py:367`), which rejects only a *leading* separator — `main`'s own suite pins `"group/user123"` and `"has/slash"` as valid (`tests/unittests/artifacts/test_artifact_util.py:153-154`). So an `app_name` that `main` accepts (and stores as a nested directory) raises here.

That is not hypothetical: `VertexAiSessionService._get_reasoning_engine_id` (`sessions/vertex_ai_session_service.py:376-392`) accepts a full `projects/{p}/locations/{l}/reasoningEngines/{id}` as the app name, and that pairing worked on v1 before this change precisely because the file service ignored `app_name` altogether. After this it fails on every artifact call.

Keeping the stricter private validator is well argued in `d68d89cf2`, but that argument is about `user_id` and `session_id`, which this branch already guarded that way — `app_name` is new here, and the strictness is inherited rather than chosen. There is a closer-to-upstream option once #6794 has landed `artifact_util.validate_path_segment` on this branch: move just this one call onto the shared helper and leave `user_id`/`session_id` on the private one. Would that be worth doing as the follow-up, so v1 and `main` agree on which app names are addressable?

xuanyang15 (approved):

src/google/adk/artifacts/file_artifact_service.py
GWeale: did it here rather than as a follow-up. app_name now goes through artifact_util.validate_path_segment, user_id and session_id keep the private one. added a test with a projects/.../reasoningEngines/... app name.

PR #6799, opened 2026-08-18
fix: Route thread-pool and live-mode tool dispatch through run_async (v1)

GWeale:
Ports two tool-dispatch fixes to `v1`, following #6575, so every dispatch site
in `functions.py` goes through `BaseTool.run_async`.

1. Thread-pool dispatch (`0156bc5a`)
   - `_call_tool_in_thread_pool` calls `tool.run_async` on the caller's event
     loop and offloads only the synchronous callable.
   - With `RunConfig.tool_thread_pool_config` set, confirmation is enforced and
     a call missing a mandatory argument returns the standard error dict.
   - A synchronous non-`FunctionTool` exposing `func` (in-tree:
     `SetModelResponseTool`) runs on a worker thread with its own event loop.

2. Live-mode dispatch (`8b9d2222`)
   - The live streaming branch calls `run_async`, so a `run_async` override is
     honoured and Pydantic-annotated arguments arrive as model instances.
   - A streaming tool that raises sends an error `FunctionResponse`.
   - A tool with `require_confirmation=True` is refused in live mode and cannot
     yet be approved there; `TODO`s in `functions.py` mark the remaining work.
   - `FunctionTool._call_live` is removed; a subclass overriding it is not
     called.

### DIFF
--- a/src/google/adk/flows/llm_flows/functions.py
+++ b/src/google/adk/flows/llm_flows/functions.py
@@ -46,6 +46,8 @@ from ...telemetry import _instrumentation
 from ...telemetry.tracing import trace_merged_tool_calls
 from ...telemetry.tracing import tracer
 from ...tools.base_tool import BaseTool
+from ...tools.function_tool import _use_sync_callable_runner
+from ...tools.function_tool import FunctionTool
 from ...tools.tool_confirmation import ToolConfirmation
 from ...tools.tool_context import ToolContext
 from ...utils.context_utils import Aclosing
@@ -138,9 +140,6 @@ async def _call_tool_in_thread_pool(
   Returns:
     The result of running the tool.
   """
-  from ...tools.function_tool import _use_sync_callable_runner
-  from ...tools.function_tool import FunctionTool
-
   loop = asyncio.get_running_loop()
   executor = _get_tool_thread_pool(max_workers)
 
--- a/src/google/adk/flows/llm_flows/functions.py
+++ b/src/google/adk/flows/llm_flows/functions.py
@@ -26,7 +26,6 @@ import inspect
 import logging
 import threading
 from typing import Any
-from typing import AsyncGenerator
 from typing import Callable
 from typing import cast
 from typing import Dict
@@ -712,6 +711,9 @@ async def _execute_single_function_call_live(
       copy.deepcopy(function_call.args) if function_call.args else {}
   )
 
+  # TODO: thread a ToolConfirmation through here so an approved tool can be
+  # re-executed in live mode. `tool_confirmation` is always None on this path,
+  # so a confirmation-gated tool can only ever be refused, never resumed.
   tool_context = _create_tool_context(invocation_context, function_call)
 
   try:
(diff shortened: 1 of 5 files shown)


src/google/adk/flows/llm_flows/functions.py
```diff
@@ -138,43 +138,39 @@ async def _call_tool_in_thread_pool(
   Returns:
     The result of running the tool.
   """
+  from ...tools.function_tool import _use_sync_callable_runner
   from ...tools.function_tool import FunctionTool
 
-  ctx = contextvars.copy_context()
   loop = asyncio.get_running_loop()
   executor = _get_tool_thread_pool(max_workers)
```
xuanyang15, line 144: `0156bc5a` was written on top of `a1792a712` ("fix: scope the tool thread pool to its event loop"), which is an ancestor of it on `main` but is not in this branch's merge-base. That commit replaced the process-global `_TOOL_THREAD_POOLS` with a `WeakKeyDictionary` keyed by event loop and added `weakref.finalize(loop, pool.shutdown, wait=False)` (`main:functions.py:75-77`, `:187`).

To be clear, this is not a porting defect and I am not asking to change anything here. The two commits touch disjoint code -- `a1792a712` only rewrites `_get_tool_thread_pool`, never `_call_tool_in_thread_pool` -- and its own message says "Executors are not bound to a loop, so the defect here is leaked idle threads rather than threads doing work on a loop they do not belong to", so the ported code runs correctly on v1's loop-agnostic executor. The leak is already on the merge-base: `_TOOL_THREAD_POOLS` is byte-identical before and after this PR, nothing under `src/` ever shuts a pool down, and the autouse `cleanup_thread_pools` fixture at `test_functions_thread_pool.py:39` predates this change.

So v1 carries the leak today whether or not this lands. Would it make sense to cherry-pick `a1792a712` to v1 as its own change and cross-link it from here, rather than growing this port? Feel free to handle it in a follow up.

src/google/adk/flows/llm_flows/functions.py
```diff
@@ -138,43 +138,39 @@ async def _call_tool_in_thread_pool(
   Returns:
     The result of running the tool.
   """
+  from ...tools.function_tool import _use_sync_callable_runner
```
xuanyang15: nit: upstream keeps both of these at module scope (`main:functions.py:51-52`; `0156bc5a` also removes the function-local `FunctionTool` import that used to live here), and `pyguide 3.13` asks for imports at the top of the file. There is no circular-import obstacle on this branch either: line 36 already imports `ComputerUseTool` at module scope, and `tools/computer_use/computer_use_tool.py:27` does `from ..function_tool import FunctionTool`, so `google.adk.tools.function_tool` is fully imported by the time this module body runs. Keeping them function-local leaves v1 and `main` diverged inside the function this PR is already rewriting, which the next backport touching `functions.py` has to reconcile by hand. Could they be hoisted to match upstream exactly?

xuanyang15 (approved):

src/google/adk/flows/llm_flows/functions.py
GWeale: looked into it, and I don't think v1 wants that commit. here `_TOOL_THREAD_POOLS` is keyed by `max_workers` alone (`functions.py:69`), so the set is bounded by the distinct configured values, one in practice, and the cost is a few idle threads for the life of the process rather than anything that grows.

`a1792a712` is what introduces per-loop keying, and the `weakref.finalize` exists to clean up the pools that keying creates. porting it to v1 would add a pool per loop and then the machinery to reap them, to replace a single shared pool that is already bounded. so I'd rather leave it, and not open a follow up.

src/google/adk/flows/llm_flows/functions.py
GWeale: hoisted both to module scope, same position as `main` (between `base_tool` and `tool_confirmation`). no circular import, as you said: `computer_use_tool` is already module-scope here and pulls in `FunctionTool` itself.

PR #6802, opened 2026-08-18
fix: Port URL validation, client timeout and credential scoping fixes to v1

GWeale:
Ports six commits from `main` to the `v1` branch.

1. `Block IPv6 literals that embed a non-global IPv4` (`6f182571`, the `load_web_page` part only)
   - `load_web_page` refuses IPv4-mapped, 6to4, NAT64 and IPv4-compatible forms whose embedded IPv4 is not globally routable.

2. `Honour request timeouts on Apigee` (`2547db61`)
   - `http_options.timeout` applies to the streaming and non-streaming completions calls.

3. `Bound the completions client` (`f57a67d6`)
   - 30s connect budget, 600s request budget, `follow_redirects=False`.
   - **Breaking:** a proxy answering `/chat/completions` with a 3xx raises `httpx.HTTPStatusError`. Point the deployment at the host the redirect names.

4. `Reject base_url in generate_content_config` (`472e4635` as amended by `4c6f22e8`, landed as one commit)
   - **Breaking at construction:** `LlmAgent` raises `ValueError` for `generate_content_config.http_options.base_url`. Set the endpoint on the model or its `genai.Client`. Headers, timeout, retries and `extra_body` are unaffected.

5. `Validate URLs before computer use navigate` (`b0fff3f0`)
   - `navigate` applies `load_web_page`'s URL checks and returns an error dict when one fails.
   - `ComputerUseToolset(allow_private_network_access=True)` re-enables localhost and internal hosts.

6. `Scope API registry credentials` (`cc275f0c`)
   - ADC tokens go only to https endpoints on `googleapis.com`; every other MCP server takes its headers from `header_provider`.

### DIFF
--- a/src/google/adk/models/apigee_llm.py
+++ b/src/google/adk/models/apigee_llm.py
@@ -63,6 +63,28 @@ _CUSTOM_METADATA_FIELDS = (
 
 _REFUSAL_PREFIX = '[[REFUSAL]]: '
 
+# Timeouts, in seconds, for the completions HTTP client. httpx applies no
+# timeout at all unless one is given, so a stalled proxy would otherwise hold
+# the connection and the streaming loop open indefinitely.
+_CONNECT_TIMEOUT_SECONDS = 30.0
+_REQUEST_TIMEOUT_SECONDS = 600.0
+
+
+def _httpx_timeout(timeout_seconds: Optional[float] = None) -> httpx.Timeout:
+  """Returns the httpx timeout budget for a completions request.
+
+  A bare float would spend the caller's whole budget on the connect phase too,
+  so the connect budget is always kept short enough to fail fast on an
+  unreachable proxy.
+
+  Args:
+    timeout_seconds: The total budget for the request, or None for the default.
+  """
+  return httpx.Timeout(
+      _REQUEST_TIMEOUT_SECONDS if timeout_seconds is None else timeout_seconds,
+      connect=_CONNECT_TIMEOUT_SECONDS,
+  )
+
 
 class ApigeeLlm(Gemini):
   """A BaseLlm implementation for calling Apigee proxy.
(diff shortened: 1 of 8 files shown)


src/google/adk/models/apigee_llm.py
```diff
@@ -427,8 +449,8 @@ def _client(self) -> httpx.AsyncClient:
     client = httpx.AsyncClient(
         base_url=self._base_url,
         headers=self._headers,
-        timeout=None,
-        follow_redirects=True,
+        timeout=_httpx_timeout(),
+        follow_redirects=False,
```
xuanyang15, line 453: The commit message for `46f28d65a` says a 3xx moved "the request, its payload and its Authorization header" to the redirect target. httpx pops `Authorization` on a cross-origin redirect (`_redirect_headers` in `httpx/_client.py`, read on 0.28.1), so that specific header would not have leaked -- though the payload, and any non-`Authorization` credential header, which Apigee deployments often use, would.

This wording is new here: upstream `f57a67d63` shipped with no body text, so the sentence lands only in the v1 history. Could you please reword that line? The fix is right either way, and the accurate reason is the stronger one.

xuanyang15 (approved):

xuanyang15:
The four ported commits all match upstream -- `computer_use_toolset.py` is the same blob as `origin/main` and as `b0fff3f02`, and the `apigee_llm.py` timeout block is byte-identical including the docstring. Two notes, neither about the code.

On the PR description: items 3 and 4 are marked Breaking and item 5 is covered by `@experimental(FeatureName.COMPUTER_USE)`. Item 6 is the one that is not covered, and it is the one place where v1 differs from `main`. Upstream deprecated the class in `d3522c009`, so on `main` this change lands on a type whose docstring already says `[DEPRECATED] ... Use AgentRegistry instead`. On v1 `d3522c009` was not ported, so `ApiRegistry` is neither experimental nor deprecated, and a non-Google MCP server silently starts getting 401 at runtime with no prior warning. Could item 6 be called out as breaking, with the `header_provider` migration spelled out, so it reaches the v1 release notes?

Separately, the File Content Compliance failure looks unrelated: the gate is a whole-file grep and the matches predate this branch. The mTLS change it wants, `41693dce6`, is a *predecessor* of the commit ported here (`git log origin/main -- api_registry.py` orders them `cc275f0c7` -> `d3522c009` -> `41693dce6`), so it is a separate upstream fix v1 never took rather than a follow-up to this one.

PR #6804, opened 2026-08-18
fix(models): Port Gemini context cache fingerprint scoping to v1

GWeale:
This PR ports one fix to the `v1` branch.

1. `fix: scope Gemini cache identity` (`8de4237e`)
   - The explicit-cache fingerprint includes the model id and a `cache_scope` recording the backend, the client base URL when one is set, and, for Vertex, the project and location.
   - Fingerprint inputs are dumped with `mode="json"` and hashed as canonical JSON with sorted keys.
   - A system instruction that will not serialize to JSON falls back to its string form.

**Behaviour change.** Existing explicit context caches are invalidated and recreated once on upgrade, with no opt-out. A stored `cache_name` matches only for the same model, backend, base URL, and Vertex project and location.

No public API change. `_generate_cache_fingerprint` keeps its signature and its 16-character hexadecimal return value.

### DIFF
--- a/src/google/adk/models/gemini_context_cache_manager.py
+++ b/src/google/adk/models/gemini_context_cache_manager.py
@@ -20,6 +20,7 @@ import hashlib
 import json
 import logging
 import time
+from typing import Any
 from typing import Optional
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


xuanyang15 (approved):

xuanyang15:
The port is byte-identical to `8de4237e` -- same stat (2 files, +142/-10) and an empty diff between the two changesets' added and removed lines -- so this reads cleanly. One nit, on the commit message rather than the code.

nit: the message says the scope records the base URL "for Vertex", but `gemini_context_cache_manager.py:369-374` sits outside the `if is_vertex and api_client is not None` guard, so `base_url` is recorded for both backends -- only `project` and `location` are Vertex-only. The PR description has this right. Upstream `8de4237e` shipped with no body text, so this sentence exists only in the v1 history, and the commit message is what lands. Could it be reworded to match the code?

For the record on `425dda190`, since it is the obvious follow-up to ask about: I do not think it belongs in this PR. The two defects it cites (reordered tools, SDK field drift) are both present in `8de4237e^` already, so it is not fixing a regression this change introduced; its `default=str` only covers the outer `json.dumps`, leaving the three inner `model_dump(mode="json")` calls outside the try exactly as they are on `main` today (`origin/main:331-334`); and it is the third commit on this file after `8de4237e`, behind `98fc03690`, which also changes what is fed to the fingerprint -- so pulling it in would not buy a single cache invalidation either. It reads like its own backport.

PR #6807, opened 2026-08-18
fix: Port code executor container and pod isolation defaults to v1

GWeale:
This PR ports two code executor fixes to the `v1` branch:

1. Container run settings (`0a9ce0f6`)
   - **Behaviour change:** `ContainerCodeExecutor` runs its container with networking disabled, so executed code that makes an HTTP request, resolves DNS, or installs a package fails.
   - Opt out with `ContainerCodeExecutor(image="...", network_enabled=True)`.
   - The container also runs with `cap_drop=['ALL']` and the `no-new-privileges` option.
   - Adds `docker>=7` to the `test` extra.

2. Pod spec settings (`82078801`)
   - **Behaviour change:** the pod `GkeCodeExecutor` creates sets `automount_service_account_token=False`, so executed code has no in-cluster Kubernetes API access.

### DIFF
--- a/src/google/adk/code_executors/container_code_executor.py
+++ b/src/google/adk/code_executors/container_code_executor.py
@@ -37,6 +37,15 @@ DEFAULT_IMAGE_TAG = 'adk-code-executor:latest'
 class ContainerCodeExecutor(BaseCodeExecutor):
   """A code executor that uses a custom container to execute code.
 
+  Security note: this executor runs model-generated code, which may be
+  influenced by untrusted input (e.g. via prompt injection). By default the
+  container is started with networking disabled and all Linux capabilities
+  dropped so that the executed code cannot reach the network (including the
+  cloud metadata endpoint at ``169.254.169.254``) or escalate privileges. For
+  stronger, kernel-level isolation of untrusted code prefer
+  ``GkeCodeExecutor`` (gVisor) or a managed executor
+  (``VertexAiCodeExecutor`` / ``AgentEngineSandboxCodeExecutor``).
+
   Attributes:
     base_url: Optional. The base url of the user hosted Docker client.
     image: The tag of the predefined image or custom image to run on the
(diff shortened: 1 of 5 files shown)


src/google/adk/code_executors/container_code_executor.py
```diff
@@ -183,6 +206,13 @@ def __init_container(self):
         image=self.image,
         detach=True,
         tty=True,
+        # Harden the sandbox for untrusted, model-generated code: no network
+        # (blocks metadata/SSRF/exfil), drop all Linux capabilities, and
+        # forbid privilege escalation. Networking can be re-enabled via
+        # `network_enabled=True` when the executed code is trusted.
+        network_disabled=not self.network_enabled,
```
xuanyang15, line 213: Both commits here are byte-for-byte ports of `0a9ce0f6` and `820788010` (I diffed the added and removed lines of each against upstream), so this is a scope question rather than anything about the change itself.

Upstream also landed `27548e392` ("kill runaway code on timeout in container and local executors") about five weeks after `0a9ce0f6`. It is a separate fix rather than a follow-up to this one -- different PiperOrigin-RevId, and it also touches `unsafe_local_code_executor.py`, which this PR does not -- but it is the piece that bounds execution: `timeout_seconds` (default 300, `gt=0`) plus a supervisor that forks the code into its own process group and SIGKILLs the group when the deadline expires. I checked the other v1 backport PRs in this batch and none of them carry `_TIMEOUT_WRAPPER` either.

So on v1 the container stays as it is today: created once in `__init_container` and reused by `exec_run` for every execution, with no wall-clock bound. With the network off by default from this line, an unbounded loop is about the most useful thing left for injected code to do, and it pins the shared container for every later caller. Is that port intentionally out of scope for v1, or is it queued separately?

xuanyang15 (approved):

PR #6808, opened 2026-08-18
fix(plugins): sanitize BigQuery analytics rows and build GCS paths per call (v1)

GWeale:
Four changes to `BigQueryAgentAnalyticsPlugin` on the `v1` branch, one commit each, re-implemented against v1's code rather than cherry-picked.

1. Content formatter failure — `7c96e6b8`, from `2919bf5b`
   - A row whose `content_formatter` raises holds `[FORMATTER_FAILED]` in `content`.

2. External URIs — `1ba04e1d`, from `9adf0113`
   - `content_parts[].uri` keeps scheme, netloc and path only.
   - A URI with userinfo, over 8192 characters, or one `urlsplit` rejects becomes `[REDACTED_SENSITIVE_URI]`.

3. Attributes tree — `ef4b10f1`, from `2919bf5b`
   - `attributes` holds `[REDACTED]` for `_SENSITIVE_KEYS` and `temp:`-scoped keys from `extra_attributes` and `custom_tags`.
   - Bounded at depth 50 (`[MAX_DEPTH_EXCEEDED]`) and 100,000 nodes per invocation (`[SANITIZE_BUDGET_EXCEEDED]`).

4. GCS offload paths — `4c34f25f`, from `2919bf5b`
   - New object names are `<date>/<trace_id>/<span_id>_<parse_uid>_c<content_ordinal>_p<part_index><ext>`; existing objects are untouched.

`is_truncated` is set when the attributes or URI pass alters a value. There is no opt-out, so dashboards reading `content`, `content_parts`, `attributes` or `is_truncated` may need updating. The plugin is opt-in and is not re-exported from `google.adk.plugins`.

The File Content Compliance check fails for a pre-existing reason; this PR adds no endpoint URL.

### DIFF
--- a/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
+++ b/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
@@ -43,6 +43,8 @@ from typing import Awaitable
 from typing import Callable
 from typing import Optional
 from typing import TYPE_CHECKING
+from urllib.parse import urlsplit
+from urllib.parse import urlunsplit
 import uuid
 import weakref
 
@@ -271,6 +273,12 @@ _SENSITIVE_KEYS = frozenset({
 # Written to the content column in place of the payload when the configured
 # content_formatter raises.
 _FORMATTER_FAILED_SENTINEL = "[FORMATTER_FAILED]"
+
+# Written in place of an external URI that cannot be stored safely.
+_REDACTED_URI = "[REDACTED_SENSITIVE_URI]"
+
+# A URI longer than this is replaced wholesale rather than parsed.
+_MAX_URI_LENGTH = 8192
 
 
 def _recursive_smart_truncate(
@@ -1378,6 +1386,44 @@ class HybridContentParser:
       )
     return text, False
 
+  def _sanitize_external_uri(
+      self, uri: Optional[str]
+  ) -> tuple[Optional[str], bool]:
+    """Removes credential-bearing components from a caller-supplied URI.
+
+    A signed GCS or HTTP URL carries its signature in the query string, and any
+    URI may carry ``user:password@`` userinfo. Both are bearer credentials, so
+    neither belongs in a stored analytics row. The scheme, host and path are
+    kept, which is what identifies the object.
+
+    Args:
+        uri: The URI to sanitize, or None.
+
+    Returns:
+        A tuple of (sanitized_uri, content_removed).
+    """
+    if uri is None:
+      return None, False
+    if not isinstance(uri, str):
+      return _REDACTED_URI, True
+    if len(uri) > _MAX_URI_LENGTH:
+      return _REDACTED_URI, True
+    try:
+      parsed = urlsplit(uri)
+      has_userinfo = parsed.username is not None or parsed.password is not None
+    except ValueError:
+      return _REDACTED_URI, True
+    if has_userinfo:
+      # Userinfo is a credential-bearing surface by definition; do not try to
+      # keep the username while guessing whether it is sensitive.
+      return _REDACTED_URI, True
+    if not parsed.query and not parsed.fragment:
+      return uri, False
+    return (
+        urlunsplit((parsed.scheme, parsed.netloc, parsed.path, "", "")),
+        True,
+    )
+
   async def _parse_content_object(
       self, content: types.Content | types.Part
   ) -> tuple[str, list[dict[str, Any]], bool]:
@@ -1401,7 +1447,12 @@ class HybridContentParser:
       # CASE A: It is already a URI (e.g. from user input)
       if hasattr(part, "file_data") and part.file_data:
         part_data["storage_mode"] = "EXTERNAL_URI"
-        part_data["uri"] = part.file_data.file_uri
+        safe_uri, uri_content_removed = self._sanitize_external_uri(
+            part.file_data.file_uri
+        )
+        part_data["uri"] = safe_uri
+        if uri_content_removed:
+          is_truncated = True
         part_data["mime_type"] = part.file_data.mime_type
 
       # CASE B: It is Binary/Inline Data (Image/Blob)
--- a/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
+++ b/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
@@ -280,9 +280,23 @@ _REDACTED_URI = "[REDACTED_SENSITIVE_URI]"
 # A URI longer than this is replaced wholesale rather than parsed.
 _MAX_URI_LENGTH = 8192
 
+# Deepest nesting _recursive_smart_truncate walks before replacing a value.
+_MAX_SANITIZE_DEPTH = 50
+
+# Total nodes one sanitizer invocation may visit. Depth and per-string size
+# are bounded, but width was not: a million-element list, or an object that
+# manufactures two fresh children per access, walks tens of millions of
+# nodes inside the depth cap. The remainder is replaced with a sentinel and
+# the row is flagged truncated.
+_MAX_SANITIZE_NODES = 100_000
+
 
 def _recursive_smart_truncate(
-    obj: Any, max_len: int, seen: Optional[set[int]] = None
+    obj: Any,
+    max_len: int,
+    seen: Optional[set[int]] = None,
+    depth: int = 0,
+    budget: Optional[list[int]] = None,
 ) -> tuple[Any, bool]:
   """Recursively truncates string values within a dict or list.
 
@@ -293,12 +307,29 @@ def _recursive_smart_truncate(
       obj: The object to truncate.
       max_len: Maximum length for string values.
       seen: Set of object IDs visited in the current recursion stack.
+      depth: Current recursion depth.
+      budget: Single-element list holding the nodes left in the shared work
+        budget for this invocation.
 
   Returns:
       A tuple of (truncated_object, is_truncated).
   """
   if seen is None:
     seen = set()
+  if budget is None:
+    budget = [_MAX_SANITIZE_NODES]
+  budget[0] -= 1
+  if budget[0] < 0:
+    return "[SANITIZE_BUDGET_EXCEEDED]", True
+
+  # The id()-based cycle detection below cannot catch an object graph that
+  # manufactures a new object on every duck-typed access, which is what any
+  # object whose model_dump()/dict()/to_dict() returns a fresh wrapper does.
+  # Such a graph recurses until the interpreter's own limit. The replacement
+  # discards real data, so unlike "[CIRCULAR_REFERENCE]" it reports
+  # truncation.
+  if depth >= _MAX_SANITIZE_DEPTH:
+    return "[MAX_DEPTH_EXCEEDED]", True
 
   obj_id = id(obj)
   if obj_id in seen:
@@ -327,13 +358,26 @@ def _recursive_smart_truncate(
       # but explicit loop is fine for clarity given recursive nature.
       new_dict = {}
       for k, v in obj.items():
+        # Stop iterating once the budget is exhausted. Recursing on every
+        # remaining entry still did work proportional to the input and
+        # produced one sentinel per entry; a single remainder sentinel
+        # stands in for everything dropped.
+        if budget[0] <= 0:
+          new_dict["[SANITIZE_BUDGET_EXCEEDED]"] = "[SANITIZE_BUDGET_EXCEEDED]"
+          truncated_any = True
+          break
         if isinstance(k, str):
           k_lower = k.lower()
           if k_lower in _SENSITIVE_KEYS or k_lower.startswith("temp:"):
+            # A directly redacted entry costs budget too, otherwise a wide
+            # "temp:" mapping bypasses the bound entirely.
+            budget[0] -= 1
             new_dict[k] = "[REDACTED]"
             continue
 
-        val, trunc = _recursive_smart_truncate(v, max_len, seen)
+        val, trunc = _recursive_smart_truncate(
+            v, max_len, seen, depth + 1, budget
+        )
         if trunc:
           truncated_any = True
         new_dict[k] = val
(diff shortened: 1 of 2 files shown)


src/google/adk/plugins/bigquery_agent_analytics_plugin.py
```diff
@@ -2897,17 +3094,28 @@ async def _log_event(
...
     is_truncated = is_truncated or parser_truncated
 
     latency_json = self._extract_latency(event_data)
     attributes = self._enrich_attributes(event_data, callback_context)
 
+    # Final pass over the complete assembled tree. _enrich_attributes copies
+    # extra_attributes (which carries session state deltas) and custom_tags in
+    # untouched, so this is the only point at which every value is guaranteed
+    # to have seen the sensitive-key redaction.
+    attributes, attrs_truncated = _recursive_smart_truncate(
```
xuanyang15, line 3174: This final pass is new here, and it routes two more values into `_recursive_smart_truncate` than reached it before: `custom_tags`, and `extra_attributes`, which carries `dict(event.actions.state_delta)` from :3238 -- a value a tool writes. On the base branch only `dict(session.state)` reached this code.

That matters because of an unguarded line further up. At :401 the list/tuple branch reconstructs the original type:

```python
      return type(obj)(new_list), truncated_any
```

`2919bf5b` -- the commit this change cites -- guards that same line at :739, and the guard is still on main today (`origin/main:1422`):

```python
      if type(obj) is tuple or type(obj) is list:
        return type(obj)(new_list), truncated_any
      # Tuple/list subclasses (e.g. namedtuples) may require positional
      # constructor fields; reconstructing raised TypeError and the safe
      # callback dropped the whole row. JSON does not preserve the subclass
      # identity anyway -- emit a plain list.
      return new_list, truncated_any
```

A namedtuple's `__new__` needs positional fields, so `type(p)(new_list)` raises `TypeError`. There is no `except` between :401 and `_log_event`, so `@_safe_callback` (:113) catches it and the whole BigQuery row is dropped. A tool doing `tool_context.state["pos"] = Point(1, 2)` would silently lose its STATE_DELTA row.

:401 is not itself in this diff -- this line is what makes it reachable for tool-written values. Should we bring the upstream guard across as well? What do you think?

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
```diff
@@ -331,31 +387,46 @@ def _recursive_smart_truncate(
...
         new_list.append(val)
       return type(obj)(new_list), truncated_any
     elif dataclasses.is_dataclass(obj) and not isinstance(obj, type):
       # Manually iterate fields to preserve 'seen' context, avoiding dataclasses.asdict recursion
       as_dict = {f.name: getattr(obj, f.name) for f in dataclasses.fields(obj)}
-      return _recursive_smart_truncate(as_dict, max_len, seen)
+      return _recursive_smart_truncate(
+          as_dict, max_len, seen, depth + 1, budget
+      )
     elif hasattr(obj, "model_dump") and callable(obj.model_dump):
```
xuanyang15, line 459: `2919bf5b` added the depth cap together with a progress requirement on these three branches (:753 onwards; still there on main at `origin/main:1436-1487`), and the comment there explains why they belong together:

```python
        dumped = obj.model_dump()
        if isinstance(dumped, (collections.abc.Mapping, list)):
          return _recursive_smart_truncate(dumped, max_len, seen, depth + 1, budget)
```

> "Mock-like objects answer every duck-typed probe with another Mock-like object, and recursing on those churns to the depth cap (falsely flagging truncation) instead of settling at the stringify fallback."

Without it, an object that hands back a fresh wrapper per access walks 50 levels, loses the `str(obj)` repr it used to get at :438, and sets `is_truncated`. The `mock_agent` fixture change at `test_bigquery_agent_analytics_plugin.py:79` looks like exactly that symptom: `root_agent_name` was a bare Mock and :3114 now walks it. Could we port the progress check alongside the cap?

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
```diff
@@ -315,13 +358,26 @@ def _recursive_smart_truncate(
       # but explicit loop is fine for clarity given recursive nature.
       new_dict = {}
       for k, v in obj.items():
```
xuanyang15, line 403: This loop is only entered when `obj` is a `dict` (:355). `2919bf5b:706` uses `collections.abc.Mapping` for that branch instead, and main still does (`origin/main:1310`), with the note that "stringifying them in the fallback branch would bypass key redaction".

A `MappingProxyType` or `UserDict` is not a `dict` subclass and has no `model_dump`/`dict`/`to_dict`, so it falls through to `str(obj)` at :438 and lands as `"mappingproxy({'api_key': '...'})"` with `truncated=False`. This module already uses `MappingProxyType` itself at :89 and :465, and :3114 now routes tool-written values through here.

:355 is not in this diff, but the redaction loop it guards is. Given the comment at :3110 says every value is guaranteed to have seen the redaction, should the branch widen to `Mapping`?

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
```diff
@@ -315,13 +358,26 @@ def _recursive_smart_truncate(
...
+        # Stop iterating once the budget is exhausted. Recursing on every
+        # remaining entry still did work proportional to the input and
+        # produced one sentinel per entry; a single remainder sentinel
+        # stands in for everything dropped.
+        if budget[0] <= 0:
+          new_dict["[SANITIZE_BUDGET_EXCEEDED]"] = "[SANITIZE_BUDGET_EXCEEDED]"
+          truncated_any = True
+          break
         if isinstance(k, str):
           k_lower = k.lower()
```
xuanyang15: `9adf0113:1326`, the commit cited for the URI change, normalizes the key before the membership test: `k_lower = k.lower().replace("-", "_")` (still there at `origin/main:1334`). Without it `api-key`, `access-token` and `refresh-token` are not redacted, and header-shaped keys are a common thing for a tool to echo into state. Any reason not to bring that one line across too?

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
```diff
@@ -1374,10 +1445,75 @@ def _truncate(self, text: str) -> tuple[str, bool]:
...
+    try:
+      parsed = urlsplit(uri)
+      has_userinfo = parsed.username is not None or parsed.password is not None
+    except ValueError:
+      return _REDACTED_URI, True
+    if has_userinfo:
+      # Userinfo is a credential-bearing surface by definition; do not try to
+      # keep the username while guessing whether it is sensitive.
+      return _REDACTED_URI, True
+    if not parsed.query and not parsed.fragment:
```
xuanyang15: This drops the query and fragment and keeps the path verbatim, which is a different algorithm from `9adf0113` -- the commit this change cites -- and from main today (`origin/main:3083-3146`, byte-identical to the cited commit). Upstream walks the path segment by segment through `_canonicalize_common_ascii_escapes` and `_is_sensitive_text_key`, redacting a sensitive segment and the one after it, and keeps the query with sensitive values redacted.

Dropping the query outright is stricter, but leaving the path unexamined is looser: a token in the path, say `https://api.example.com/v1/files/ya29.a0Af.../download`, is stored in the clear on v1 and redacted on main. Those three helpers do not exist on v1, but `_SENSITIVE_KEYS` (:264) does, so a per-segment `k.lower().replace("-", "_") in _SENSITIVE_KEYS` check with the same redact-the-following-segment rule would close most of the gap without porting them. The description explains this is a re-implementation, but not why the path was left unexamined -- could we either follow upstream here, or say what the v1 constraint is?

Non-blocking, same function: `_MAX_URI_LENGTH = 8192` at :281 replaces an over-length URI with `[REDACTED_SENSITIVE_URI]` outright, where upstream gates at `_MAX_JSON_INSPECT_CHARS` (4,000,000, `origin/main:668`) and then shortens with `self._truncate` -- which v1 has, at :1440 -- so a long URI survives in truncated form. The description does call the 8192 bound out, so this is only a question about the number and the lossy failure mode.

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
GWeale: brought the upstream guard across. a list or tuple subclass is emitted as a plain list instead of being reconstructed, so a namedtuple no longer raises TypeError and costs the whole row.

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
GWeale: ported the progress requirement alongside the cap. the three dump branches only recurse when the dump returned a Mapping or list, so a self-dumping object settles at the stringify fallback instead of walking to 50 and flagging truncation.

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
GWeale: widened to `collections.abc.Mapping`. a MappingProxyType or UserDict goes through the redaction loop now rather than falling through to `str()`.

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
GWeale: added it. `api-key`, `access-token` and `refresh-token` redact like their underscore forms.

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
GWeale: the path is examined now: a segment naming a sensitive key is redacted along with the segment after it, using `_SENSITIVE_KEYS` and the same normalization you suggested. I kept dropping the query and fragment outright rather than following upstream's per-value redaction, since v1 lacks the three helpers and dropping is the stricter of the two.

on `_MAX_URI_LENGTH`, left at 8192 for now. happy to switch it to `self._truncate` if you'd rather match upstream's lossy-but-preserving behaviour.

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
```diff
@@ -331,31 +432,62 @@ def _recursive_smart_truncate(
...
-      # Pydantic v2
+      # Pydantic v2. Only recurse if the conversion made progress toward a
+      # JSON-native value. Mock-like objects answer every duck-typed probe
+      # with another Mock-like object, and recursing on those churns to the
+      # depth cap, falsely flagging truncation, instead of settling at the
+      # stringify fallback below.
       try:
-        return _recursive_smart_truncate(obj.model_dump(), max_len, seen)
+        dumped = obj.model_dump()
+        if isinstance(dumped, (collections.abc.Mapping, list)):
```
xuanyang15, line 467: Verified this against `main` rather than taking the reply at face value: the tuple guard at :448, the `Mapping` widening at :397 and the `-`/`_` normalization at :414 all match `origin/main:1422`, `:1310` and `:1334`. Nice catch on `_CLOUD_PLATFORM_SCOPE` too -- I had not spotted that one, and the constant block is byte-identical to `origin/main:1126`, so the compliance job going green here is upstream's own solution rather than a workaround.

One correction I owe you on this hunk: the snippet in my earlier comment was narrower than upstream and you implemented it faithfully. `origin/main:1446-1454` reads:

```python
        dumped = model_dump_fn()
        if isinstance(
            dumped,
            (collections.abc.Mapping, list, tuple, str, bytes, bytearray),
        ):
          return _recursive_smart_truncate(
              dumped, max_len, seen, depth + 1, budget
          )
        if dumped is None or isinstance(dumped, (int, float, bool)):
          return dumped, False
```

The consequence of the narrower set is on the redaction side rather than the truncation side: a `to_dict()` that returns a tuple now falls through to `str(obj)` at the stringify fallback, so its elements never go through the sanitizer and land as a repr. A dump returning a `str`, `bytes` or a scalar likewise yields the repr of the original object instead of the dumped value. Could the predicate and the scalar line be widened to match? Sorry for sending you the trimmed version.

src/google/adk/plugins/bigquery_agent_analytics_plugin.py
```diff
@@ -268,9 +272,73 @@ def _get_tool_origin(
...
+# Cloud Platform OAuth scope. Assembled from parts so this module does not
+# embed a bare Google APIs host literal: the file-content compliance scan
+# rejects such host literals on changed files unless an accompanying mTLS
+# endpoint is present, which does not apply to this OAuth-scope use.
+_CLOUD_PLATFORM_SCOPE = (
+    "https://www." + "googleapis" + ".com/auth/cloud-platform"
+)
+
+
+def _redact_sensitive_path_segments(path: str) -> tuple[str, bool]:
```
xuanyang15, line 304: This closes the case where a segment names the value that follows it, and `/access_token/ya29.../download` now redacts on v1 the same way it does on `main`. Two notes.

First, a correction. My earlier comment claimed `https://api.example.com/v1/files/ya29.a0Af.../download` "is stored in the clear on v1 and redacted on main". I checked `main`'s four patterns afterwards and that is wrong: `_AUTH_HEADER_RE` needs a header name, `_BEARER_TOKEN_RE` needs `bearer `, `_BASIC_TOKEN_RE` needs `basic `, and `_SENSITIVE_TEXT_KEY_RE` needs a key. A bare token segment with no key in front of it is not redacted on `main` either, so that example was not a gap between the branches. Apologies -- it was the motivating example in the request, and it did not hold up.

Second, what does still differ from upstream, in case it is worth a follow-up rather than more work here. `origin/main:3102-3121` runs three things per segment where this runs one: `_canonicalize_common_ascii_escapes(segment)` before the key test, so `%61ccess_token` still matches; the key test itself; and then `_sanitize_sensitive_text(segment, -1)` on segments that are not keys, which catches `bearer ...`-shaped and `key=value`-shaped material embedded in a single segment. Neither helper exists on v1. Given the query and fragment are dropped outright here, the remaining exposure is narrow -- would you rather leave it and track the two helpers as their own backport?

xuanyang15 (approved):

PR #6810, opened 2026-08-19
fix(models): enforce 32k token floor for Vertex AI context caching and add graceful fallback

Gabriel-Mesq:
### Description of the Change

When using ADK with Google Cloud Vertex AI (`vertexai=True`), context caching requires a minimum token floor of **32,768 tokens** (unlike Google AI Studio which supports lower thresholds depending on the model, e.g. 4096 tokens for Gemini 3 / 3.5).

Currently, `GeminiContextCacheManager` only checks model name prefixes against AI Studio thresholds, causing it to issue `caches.create` requests to Vertex AI with prompts between 4,096 and 32,768 tokens. Vertex AI rejects these with `400 INVALID_ARGUMENT`, and because `google_llm.py` lacked exception handling around context caching, the unhandled error terminated the entire LLM generation and aborted active response streams.

### Changes Made

1. **Backend Floor Enforcement (`src/google/adk/models/gemini_context_cache_manager.py`)**:
   - Added `_VERTEX_AI_MIN_CACHE_TOKENS = 32768`.
   - Updated `_minimum_cache_tokens` to accept `is_vertex=bool(self.genai_client.vertexai)` and enforce the 32,768 token floor when running on Vertex AI.

2. **Graceful Fallback (`src/google/adk/models/google_llm.py`)**:
   - Wrapped `cache_manager.handle_context_caching()` in a `try...except` block in `Gemini.generate_content_async`.
   - If cache initialization fails for any reason (token floor, quota, transient backend issue), it logs a warning (`logger.warning`) and proceeds with regular generation without breaking active streams.

3. **Unit Tests (`tests/unittests/agents/test_gemini_context_cache_manager.py`)**:
   - Added `test_vertex_ai_skips_cache_below_32768_token_minimum` verifying Vertex AI skips cache creation below 32k tokens (tested with `gemini-3.5-flash-lite`).
   - Added `test_vertex_ai_creates_cache_above_32768_token_minimum` verifying Vertex AI successfully creates cache when above 32k tokens.

#### Unit Tests Executed
```bash
pytest tests/unittests/agents/test_gemini_context_cache_manager.py tests/unittests/models/test_google_llm.py

### DIFF
--- a/src/google/adk/models/gemini_context_cache_manager.py
+++ b/src/google/adk/models/gemini_context_cache_manager.py
@@ -36,17 +36,22 @@ from .llm_response import LlmResponse
 
 logger = logging.getLogger("google_adk." + __name__)
 
-# Named Gemini model families have documented explicit-cache floors. For
+# Named Gemini model families and backends have documented explicit-cache floors. For
 # opaque tuned-model and endpoint IDs, the server remains authoritative.
 _GEMINI_2_5_MIN_CACHE_TOKENS = 2048
 _GEMINI_3_MIN_CACHE_TOKENS = 4096
+_VERTEX_AI_MIN_CACHE_TOKENS = 32768
 
 if TYPE_CHECKING:
   from google.genai import Client
 
 
-def _minimum_cache_tokens(model: Optional[str]) -> Optional[int]:
+def _minimum_cache_tokens(
+    model: Optional[str], is_vertex: bool = False
+) -> Optional[int]:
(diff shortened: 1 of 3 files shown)


GWeale:
_minimum_cache_tokens already carries vertex's documented minimums, 4096 for gemini 3 and 2048 for gemini 2, so a 32768 floor ahead of the model check would stop caching about 30 combinations that work today. open a new issue if you want us to look at it again.

Issue #5835, opened 2026-05-24
Cache read write token counts


srinivasanmohan98-dotcom:
Currently the adk ignores the cache read write token counts provided by litellm bedrock model and i can only see the cached content token count(cached tokens read) due to which the cache read write token cost calculation is not made correctly in arize. Could anyone please help me on this?

sanketpatil06:
Hi @srinivasanmohan98-dotcom, Could you please provide a minimal reproduction code? This will help us understand the issue effectively. Also, please share the ADK version you are currently using. Thanks.

srinivasanmohan98-dotcom:
Hi @sanketpatil06 I use the Google adk instrumentor to export the traces. I can't share the code now but what I understood is the adk doesn't capture the usage metadata fully like cache write and read from the litellm output. It only holds the cached content token count which seems to be the cache read token counts. But I need the cache write token count as well. Please refer the image I attached here. 

<img width="4096" height="3072" alt="Image" src="https://github.com/user-attachments/assets/43b4aa3e-6984-4e02-aaa9-47dc8b85d4de" />

srinivasanmohan98-dotcom:
I use adk 1.28

enesdemirag:
I have a question, does adk caches session message history automatically when using run_async with vertexai session manager?

wojcikm:
Same gap on the native `AnthropicLlm` path (Vertex Anthropic / `AsyncAnthropicVertex`), not only LiteLLM+Bedrock.

**Still true on released `google-adk==2.6.3` (2026-08-07).** `message_to_generate_content_response` maps:

- `input_tokens` → `prompt_token_count`
- `output_tokens` → `candidates_token_count`
- `cache_read_input_tokens` → `cached_content_token_count` (since [b15c8a0](https://github.com/google/adk-python/commit/b15c8a0fe19ae937d832bb0410a8639da2d86cdf) / v2.3.0)

and **drops `cache_creation_input_tokens`**. Anthropic bills that bucket at 125% of input, disjoint from `input_tokens` — so cost pipelines that only see `usage_metadata` under-count cache-priming turns (large cached system prefix across subagents). We see this in `BigQueryAgentAnalyticsPlugin` rows: `attributes.usage_metadata` never has a cache-write field, and the plugin has no `cache_creation_*` view column.

[...]

PR #6821, opened 2026-08-19
feat: serialize prompt-cache creation tokens on LlmResponse

gurjot-05:
## What
Adds a serializable `LlmResponse.cache_creation_token_count` field (inherited by `Event`, emitted as `cacheCreationTokenCount`) for prompt-cache **write** tokens.

## Why
#5835 / d0b33a05 extract cache-creation tokens and attach them to `GenerateContentResponseUsageMetadata` via `object.__setattr__` for OpenTelemetry. But `google.genai`'s usage type is `extra='forbid'`, so that attribute is dropped by `model_dump()` — it never reaches serialized events. So the Dev UI, persisted sessions, and SSE consumers can't see cache-write (only OTel / in-memory `getattr` can).

This surfaces the same value on the serializable `LlmResponse` — the cache-write counterpart to `usage_metadata.cached_content_token_count` (read) — next to the existing top-level `cache_metadata` field. (It can't live inside `usage_metadata`: an extra key there makes event reload raise a ValidationError, breaking session persistence.)

## Changes
- `llm_response.py`: new `cache_creation_token_count: Optional[int]` field.
- `anthropic_llm.py` / `lite_llm.py`: populate it from the already-extracted value (streaming + non-streaming). OTel path unchanged.
- Tests extended to assert the serialized field.

## Companion PR
Dev-UI counterpart that displays this value: google/adk-web#492 — please merge together.


gurjot-05:
Hi @xuanyang15 

Wanted to know if there's any update on this PR?

browningjp:
+1 for this please!

PR #6837, opened 2026-08-20
fix(agents): point misplaced generation kwargs at generate_content_config

### LINKED ISSUE #6836: LlmAgent rejects temperature= with extra_forbidden — no discoverable path to generate_content_config
## Summary

Coming from `google-genai` (where `GenerateContentConfig(temperature=...)` is routine), the natural first attempt with ADK is:

```python

agent = LlmAgent(
    name="grader",
    model="gemini-3.5-flash-lite",
    instruction="...",
    output_schema=MySchema,
    temperature=0.1,
)
```

On ADK 2.7.0 this fails at construction time:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for LlmAgent temperature Extra inputs are not permitted [type=extra_forbidden, input_value=0.1, input_type=float]
```

The working form is `generate_content_config=genai_types.GenerateContentConfig(temperature=0.1)`, but nothing in the error, the docstring, or the validation message points there. In our project this crashed every structured agent in a production pipeline and cost a debugging session to trace (model_fields introspection).

## Suggested fixes (any one would do)

1. Accept `temperature` as a convenience kwarg and fold it into `generate_content_config`.
2. Or add a custom pydantic error message for common generation kwargs (`temperature`, `top_p`, `max_output_tokens`) pointing to `generate_content_config`.
3. Or document the mapping prominently in the LlmAgent docstring.

## Environment

google-adk 2.7.0, google-genai 2.18.1, Python 3.13, macOS. Found while building a K-12 exam-grading fleet for the All Things Agentic hackathon.

a2105z:
### Summary

`LlmAgent(temperature=0.1)` (and other `GenerateContentConfig` knobs) now raises a `ValueError` that names `generate_content_config=types.GenerateContentConfig(...)` instead of failing with opaque `extra_forbidden`.

Reserved knobs that already have an agent field (`system_instruction`, `response_schema`) still redirect to `instruction=` / `output_schema=`. Unknown keys that are not config fields are left alone so a real typo stays `extra_forbidden`. CamelCase aliases (`topP`, `maxOutputTokens`) are included. `tools` is skipped because it is already an `LlmAgent` field.

This is option 2 from #6836, per review on this PR: do not promote generation settings as public kwargs.

Closes: #6836

### Test plan

**Unit Tests:**

```
uv run python -m pytest tests/unittests/agents/test_llm_agent_fields.py tests/unittests/agents/test_llm_agent_error_messages.py -q
105 passed
```

Covers the original `temperature=` failure, multiple knobs named in one error, camelCase aliases, reserved-field redirects, typos remaining `extra_forbidden`, `tools=` not hijacked, and subclasses inheriting the validator.

```python

LlmAgent(name='grader', temperature=0.1)
# ValueError: temperature is a GenerateContentConfig field. Pass
# generate_content_config=types.GenerateContentConfig(temperature=...) instead.

agent = LlmAgent(
    name='grader',
    generate_content_config=types.GenerateContentConfig(temperature=0.1),
)
assert agent.generate_content_config.temperature == 0.1
```

### Checklist

### DIFF
--- a/src/google/adk/agents/llm_agent.py
+++ b/src/google/adk/agents/llm_agent.py
@@ -231,10 +231,73 @@ async def _convert_tool_union_to_tools(
     return []
 
 
+# GenerateContentConfig fields that already have a dedicated LlmAgent argument.
+# Passing them as LlmAgent kwargs should point at that argument rather than
+# folding into generate_content_config (which would then fail later).
+_GENERATE_CONTENT_FIELDS_OWNED_BY_AGENT: dict[str, str] = {
+    'system_instruction': 'instruction',
+    'response_schema': 'output_schema',
+}
+
+# Snake-case field names and camelCase aliases used by GenerateContentConfig.
+_GENERATE_CONTENT_FIELD_NAMES: dict[str, str] = {
+    name: name for name in types.GenerateContentConfig.model_fields
+}
+_GENERATE_CONTENT_FIELD_NAMES.update({
+    field.alias: name
+    for name, field in types.GenerateContentConfig.model_fields.items()
+    if field.alias is not None
+})
+
+
+def _generate_content_field_name(key: str) -> Optional[str]:
+  """Return the GenerateContentConfig field name for key, or None."""
+  return _GENERATE_CONTENT_FIELD_NAMES.get(key)
+
+
+def _existing_generate_content_as_dict(
+    existing: Any,
+) -> Optional[dict[str, Any]]:
+  """Normalize generate_content_config input to a snake_case dict.
+
+  Returns None when the value is not a recognized config shape so the
+  generate_content_config field validator can report the type error.
+  """
+  if existing is None:
+    return {}
+  if isinstance(existing, types.GenerateContentConfig):
+    return existing.model_dump(exclude_unset=True)
+  if isinstance(existing, dict):
+    normalized: dict[str, Any] = {}
+    for key, value in existing.items():
+      canonical = _generate_content_field_name(key) or key
+      normalized[canonical] = value
+    return normalized
+  return None
+
+
 # TODO: drop the explicit abc.ABC base once BaseNode surfaces ABCMeta to
 # static type checkers.
 class LlmAgent(BaseAgent, abc.ABC):
-  """LLM-based Agent."""
+  """LLM-based Agent.
+
+  Generation settings from ``google.genai.types.GenerateContentConfig``
+  such as ``temperature``, ``top_p``, and ``max_output_tokens`` can be
+  passed directly as keyword arguments. They are merged into
+  ``generate_content_config``.
+
+  Example:
+    ```python
+    from google.adk.agents import LlmAgent
+
+    agent = LlmAgent(
+        name='grader',
+        model='gemini-3.5-flash',
+        instruction='Grade the exam.',
+        temperature=0.1,
+    )
+    ```
+  """
 
   DEFAULT_MODEL: ClassVar[str] = 'gemini-3.5-flash'
   """System default model used when no model is set on an agent."""
@@ -355,11 +418,13 @@ class LlmAgent(BaseAgent, abc.ABC):
   generate_content_config: Optional[types.GenerateContentConfig] = None
   """The additional content generation configurations.
 
-  NOTE: not all fields are usable, e.g. tools must be configured via `tools`,
-  thinking_config can be configured here or via the `planner`. If both are set, the planner's configuration takes precedence.
+  Generation knobs such as temperature, top_p, and max_output_tokens may
+  also be passed directly as LlmAgent keyword arguments; they are merged
+  into this config.
 
-  For example: use this config to adjust model temperature, configure safety
-  settings, etc.
+  NOTE: not all fields are usable, e.g. tools must be configured via `tools`,
+  thinking_config can be configured here or via the `planner`. If both are
+  set, the planner's configuration takes precedence.
   """
 
   mode: Literal['chat', 'task', 'single_turn'] | None = None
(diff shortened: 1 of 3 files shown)


a2105z:
@googlebot I fixed it.

DeanChensj:
Hi @a2105z — thanks for picking this up so quickly, and for reading the actual failure mode rather than just papering over it. Two parts of this I want to keep regardless of where we land: deriving the field names from GenerateContentConfig.model_fields instead of hardcoding them, and redirecting system_instruction / response_schema to the LlmAgent arguments that own them with a message that says so. That second one is a genuinely nice touch and I hadn't seen it proposed anywhere.

That said, after going through #6836 I think we should take the reporter's option 2 — a clear error message — rather than option 1, and I'd rather tell you now than after another review round.

The reason is in the issue title: "no discoverable path to generate_content_config". What cost @CarSanoja a debugging session wasn't that temperature= is rejected, it's that the rejection points nowhere. An error that names the destination fixes that for every field at once:
temperature is a GenerateContentConfig field. Pass
generate_content_config=types.GenerateContentConfig(temperature=...) instead.

Accepting the kwargs solves it too, but it commits us to more than it looks like. It gives two spellings for the same setting, and every field google-genai adds becomes a decision about whether to promote it — a decision we can't easily revise later, because once these are accepted they're public API.

There's also a coverage asymmetry that's easy to miss. GenerateContentConfig has 35 fields today. Folding a subset means the ones left out still fail with the same opaque extra_forbidden, and arguably worse than before, because the user has now learned that generation settings work as kwargs. An error-message fix derived from model_fields covers all 35 and anything added later, with nothing to maintain.

If you're up for reworking it, what we'd take is roughly:
- a model_validator(mode='before') that raises a helpful ValueError when an unknown key matches a GenerateContentConfig field, keeping your system_instruction / response_schema redirects as a special case
- unknown keys that aren't config fields left alone, so a real typo still reads as extra_forbidden rather than a lecture about config
- the short docstring paragraph from your PR showing the generate_content_config form — that was option 3 in the issue and worth keeping either way
That should be substantially smaller than the current PR. If you'd rather not redo it, say so and I'll pick it up — you found the problem and the redirect idea, and that'll be credited either way.

Thanks again

CarSanoja:
Following up on Dean's redirect, since I filed this one and had already signed off on the kwargs version.

To be clear about what my earlier test did and did not say: #6837 does fix the failure and I stand by those five results. Dean's objection is about API surface rather than correctness, and on that axis I think he is right — it is also the option I put in the issue. The title is "no discoverable path", and what cost me the session was not that `temperature=` is rejected: it is that the raw failure is literally the single word `temperature`. An error that names the destination fixes that for all 35 fields at once, where accepting a subset leaves the rest failing the same way, now against a learned expectation that generation settings work as kwargs.

I built Dean's spec against adk-python 2.7.0 to check it survives contact — a `model_validator(mode="before")` that fires only on keys unknown to `LlmAgent` which are also `GenerateContentConfig` fields, about 25 lines:

| case | result |
|---|---|
| `temperature=0.2` | error naming `generate_content_config=types.GenerateContentConfig(temperature=...)` |
| `system_instruction=` / `response_schema=` | redirected to `instruction` / `output_schema` |
| `temperature=` and `top_p=` together | both named in one error, not just the first |
| `temperatur=0.2` (a real typo) | untouched: plain `extra_forbidden`, no lecture |
| `tools=[]`, `generate_content_config=...` | construct normally |
| subclasses of `LlmAgent` | inherit the validator |

Two things from @a2105z's PR that should survive into the smaller shape:

1. The `system_instruction` / `response_schema` redirect map. It is the best idea in the thread and it reads better than what I originally asked for.
2. The camelCase alias map. All 35 config fields carry an alias, so `topP=` and `maxOutputTokens=` are as likely a first attempt as the snake_case spellings; without the alias lookup they fall back to a bare `extra_forbidden` and the fix misses exactly the users it is for.

One trap for whoever writes it: `tools` is the only name present on both `LlmAgent` and `GenerateContentConfig` (35 config fields against 33 agent fields today). Gating on "unknown to the agent" rather than "matches a config field" is what keeps `tools` working — worth a regression test so it cannot drift into hijacking it.

@a2105z happy to hand you the implementation and the edge-case tests so this stays your PR, or to open it myself if you would rather not redo it — whichever unblocks it faster. Either way I will re-test whatever lands before merge.

a2105z:
@DeanChensj @CarSanoja Thanks — reworked this to option 2 instead of folding the kwargs.

Pushed on this branch. The validator is now a `model_validator(mode='before')` that:

- raises a `ValueError` naming `generate_content_config=types.GenerateContentConfig(...)` when an unknown key matches a `GenerateContentConfig` field (derived from `model_fields`, including camelCase aliases)
- keeps the `system_instruction` → `instruction` and `response_schema` → `output_schema` redirects
- leaves real typos (`temperatur=`) as `extra_forbidden`
- gates on “unknown to `LlmAgent`” so `tools=` is not hijacked
- names multiple misplaced fields in one error
- is inherited by subclasses

Class docstring now shows the `generate_content_config=` form. Happy to tweak the wording if you want it closer to the issue title.

@CarSanoja kept this as my PR — your edge-case table is covered in `test_llm_agent_error_messages.py`. Would appreciate a re-test when you have a chance.

CarSanoja:
Re-tested, as promised. **This resolves the issue for me, and it is better than what I proposed.**

Clean venv, Python 3.13, google-genai 2.20.0, branch at 14ac8d2. Your own suite: `20 passed`. My table, all against the branch:

| case | result |
|---|---|
| `temperature=0.2` | `temperature is a GenerateContentConfig field. Pass generate_content_config=types.GenerateContentConfig(temperature=...) instead.` |
| `topP=0.9` (camelCase) | resolved and reported as **`top_p`** |
| `maxOutputTokens=256` | resolved and reported as **`max_output_tokens`** |
| `system_instruction=` | `must be set via LlmAgent.instruction` |
| `response_schema=` | `must be set via LlmAgent.output_schema` |
| `temperature=` + `top_p=` | both named, and the suggested constructor carries both |
| `temperatur=0.2` (a real typo) | plain `extra_forbidden` — untouched |
| `tools=[]` | constructs |
| `generate_content_config=...` | constructs |
| `generate_content_config=` + `instruction=` | constructs |
| subclass of `LlmAgent` | inherits the validator |
| `LlmAgent.model_validate({... "temperature": 0.2})` | same error on the dict path |
| `Agent(...)` (the alias) | same error |
| `temperature=None` | still redirected, which is right |

Two things you did that I had not thought of, and that I would keep:

1. **Reporting the canonical name for an alias.** `topP=` fails with `top_p is a GenerateContentConfig field`, so the message teaches the spelling that will actually work rather than echoing the one that failed.
2. **The combined suggestion for several fields at once** — one error that ends in `GenerateContentConfig(temperature=..., top_p=...)` is a copy-paste fix instead of two round trips.

One optional bit of polish, not a blocker, and I would understand leaving it: when a real typo and a config field arrive together (`temperatur=0.2, top_k=5`), the error names `top_k` and stays silent about `temperatur`. The user fixes the config field, re-runs, and only then meets the typo. Same for `system_instruction=` together with `temperature=`: the redirect is reported and the config field is not. If pydantic makes it easy to raise once with both kinds, it would fit the discoverability goal; if it does not, this is a strictly better message than what is on main today.

Thanks for reworking it and for the tests — `test_llm_agent_error_messages.py` covers the cases I care about, including the `tools` gate. From my side this is ready.

a2105z:
Follow-up on the optional polish: pydantic's `extra_forbidden` only runs if the before-validator returns, so a config-field `ValueError` would still swallow a typo. The validator now collects redirects, config-field names, and unknown extras and raises them in one `ValueError` when a config-field error is already firing:

- `temperatur=0.2, top_k=5` names `top_k` **and** `Extra inputs are not permitted: temperatur`
- `system_instruction=` with `temperature=` reports the `instruction=` redirect **and** the `generate_content_config` path

A typo alone is still left to pydantic `extra_forbidden`.

PR #6842, opened 2026-08-21
feat(agents): let nested LoopAgents exit independently

### LINKED ISSUE #2808: Nested LoopAgents will escalate all LoopAgents
A nested LoopAgent will escalate and break all LoopAgents

```
SequentialAgent(
    sub_agents=[
        LoopAgent(
            sub_agents=[
                LoopAgent(
                    sub_agents=[
                        section_evaluator,
                        EscalationChecker(name="IBreakAfter5Loops"),
                    ]
                ),
                EscalationChecker(name="IAlsoBreakAfter5Loops"),
            ],
        ),
    ],
)
```

Expecting to loop 25 times, but the inner loop will reach five and the outer loop will exit after running once.

```
  @override
  async def _run_async_impl(
      self, ctx: InvocationContext
    times_looped = 0
    while not self.max_iterations or times_looped < self.max_iterations:
      for sub_agent in self.sub_agents:
        should_exit = False
        async for event in sub_agent.run_async(ctx):
          yield event
          if event.actions.escalate:
            should_exit = True

        if should_exit:
          return

      times_looped += 1
    return
```

In the LoopAgent _run_async_impl method every event appears to be passed up first, triggering an escalate on every loop agent that is part of the invocation. Something like this would stop the event from propagating further upwards but may impact the reporting.

```
async for event in sub_agent.run_async(ctx):
  if event.actions.escalate:
    should_exit = True
    event.actions.escalate = False
  yield event
```

**Desktop:**
 - ADK version: 1.10.0

a2105z:
## Summary

- Nested `LoopAgent`s currently treat every `escalate` as global: an inner `exit_loop()` also kills every ancestor loop. That is the #2808 report (expect 25 inner steps, get 5).
- This implements @Jacksunwei's scoped-escalation design. `escalate=True` with no context stays **root** (today's behavior, including `exit_loop`). `EscalationContext(type='parent')` is consumed by the nearest `LoopAgent`, which records itself on `handled_by` so outer loops continue. `exit_current_loop` is the parent-scoped tool.
- Optional `target_agent` aims the exit at a named loop. ParallelAgent ignores already-consumed parent escalations so siblings of an inner loop keep running.

## Test plan

```
uv run python -m pytest \
  tests/unittests/agents/test_loop_agent.py \
  tests/unittests/events/test_event_actions.py \
  tests/unittests/tools/test_exit_loop_tool.py \
  tests/unittests/agents/test_parallel_agent.py \
  tests/unittests/a2a/converters/test_to_adk.py::TestToAdk::test_peer_supplied_actions_cannot_mutate_caller_session \
  tests/unittests/a2a/converters/test_to_adk.py::TestToAdk::test_peer_settable_action_fields_are_exactly_inert \
  -q
```

57 loop/event/tool/parallel tests passed, plus the A2A peer-actions allow-list (escalation_context is not peer-settable).

Manual check: nested SequentialAgent → outer LoopAgent → inner LoopAgent with `exit_current_loop` should run inner×outer iterations; `exit_loop` still unwinds every loop.

Closes: #2808

@Jacksunwei @klateefa @hangfei @boyangsvl @wuliang229 @yeesian @xuanyang15 @steren @llalitkumarrr @GWeale @jhonratiu

### DIFF
--- a/src/google/adk/agents/loop_agent.py
+++ b/src/google/adk/agents/loop_agent.py
@@ -60,6 +60,11 @@ class LoopAgent(BaseAgent):
   When sub-agent generates an event with escalate or max_iterations are
   reached, the loop agent will stop.
 
+  ``escalate=True`` with no ``escalation_context`` (including ``exit_loop``)
+  exits every enclosing LoopAgent. ``EscalationContext(type='parent')``
+  (including ``exit_current_loop``) exits only this loop so an outer loop
+  can continue.
+
   .. deprecated::
     LoopAgent is deprecated in favor of Workflow and will be removed in a
     future version. Workflow cannot yet be used as an LlmAgent sub-agent.
@@ -112,9 +117,9 @@ class LoopAgent(BaseAgent):
 
         async with Aclosing(sub_agent.run_async(ctx)) as agen:
           async for event in agen:
-            yield event
-            if event.actions.escalate:
+            if event.actions.consume_workflow_escalation(self.name):
               should_exit = True
+            yield event
             if ctx.should_pause_invocation(event):
               pause_invocation = True
 
--- a/src/google/adk/agents/parallel_agent.py
+++ b/src/google/adk/agents/parallel_agent.py
@@ -57,7 +57,7 @@ def _create_branch_ctx_for_sub_agent(
 
 def _has_escalate_action(event: Event) -> bool:
   """Returns whether the event asks the parent workflow to exit early."""
-  return bool(event.actions.escalate)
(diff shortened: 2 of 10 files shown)


boyangsvl:
I'd like to recommend switching to the workflow instead of building on top of LoopAgents. Workflows give you more control, and I am afraid of changing the behavior of the current LoopAgent because it may break existing users relying on this behavior, even if the behavior doesn't make sense. I also don't want to complicate EventActions just to add a flag for this behavior. I will close the issue / PR and recommend migrating to Workflow.

Issue #6823, opened 2026-08-20
DatabaseSessionService: session row persists in storage without its user_state row ("User state missing" on later append_event)


hungubqn0310:
## Describe the bug

Under real production traffic, a small but persistent fraction of sessions end up with a row in the `sessions` table but **no** corresponding row in `user_states` for that `(app_name, user_id)`. `create_session()` is supposed to create both atomically in one transaction, so this should not be possible — but it is happening continuously in production.

The symptom surfaces later, whenever anything calls `append_event()` on that session to apply a state update: `DatabaseSessionService` correctly raises, per `_select_required_state()`:

```
ValueError: User state missing for app_name='mmvn_b2c_agent', user_id='<id>'. Session state tables should be initialized by create_session.
```

...but by that point the session has already been silently created without its user_state row, so **every future state update for that user fails permanently** until the row is manually backfilled.

## Environment

- `google-adk` version: **2.3.0** (checked latest is 2.7.0; CHANGELOG 2.4.0–2.7.0 reviewed, no entry appears related)
- Database: PostgreSQL 16 (via `asyncpg`)
- Deployment: FastAPI via `get_fast_api_app(..., auto_create_session=False)`, uvicorn, Docker
- Session creation goes exclusively through ADK's REST endpoints — no app code touches `StorageUserState` directly.

## Evidence from production

```sql
SELECT count(DISTINCT s.user_id) FROM sessions s LEFT JOIN user_states us ON us.app_name=s.app_name AND us.user_id=s.user_id WHERE us.user_id IS NULL;
-- 2665 (of 19529 users = 13.6%)
```

Ongoing since 2025-10-30 through 2026-08-19, ~5-10/day. Reproduces on staging too (7/30 users).

## What I've ruled out

- Collation/index corruption (forced seq scan still 0 rows)
- DB migration gap (legacy DB also missing these rows)
- App code deleting rows (full repo grep, nothing)
- Stale library bug (both envs on 2.3.0, SAVEPOINT logic looks correct on read)

## Reproduction attempts (inconclusive)

1. 8x concurrent POST same session_id: 1 wins (200, consistent), 7 get raw HTTP 500 (see related bug), no orphan ever.
2. 8x concurrent POST same user, distinct session_ids: all succeed, 1 user_state row, no orphan.
3. Raw-socket abrupt disconnect (40x): 0 orphans — local request completes faster than ASGI can detect disconnect and cancel.

[...]

PR #6843, opened 2026-08-21
fix(sessions): raise AlreadyExistsError on concurrent create_session races

hungubqn0310:

- Related: #6823

**Problem:**
`DatabaseSessionService.create_session()`'s up-front existence check
(`has_user_provided_id and await sql_session.get(...)`) is not atomic with
the `INSERT` that follows it. Two concurrent callers using the same
caller-provided `(app_name, user_id, session_id)` can both pass the check
and then race the same insert. The loser gets a raw, unhandled
`IntegrityError` / `UniqueViolationError` instead of the documented
`AlreadyExistsError`.

**Solution:**
Wrap the `sql_session.flush()` around the session insert in
`try/except IntegrityError`, raising `AlreadyExistsError` — the same
pattern `_get_or_create_state` already uses (via a SAVEPOINT) for the
analogous `app_state`/`user_state` race.

This does **not** address the separate orphaned `user_state` row issue also
reported in #6823. `_rollback_on_exception_session` rolls back the whole
transaction on any exception (including the `AlreadyExistsError` raised
here), so this particular race shouldn't be able to leave a row behind on
its own — it only replaces a confusing raw `IntegrityError` with a clean,
documented error. The orphan issue needs more data to root-cause.

### Testing Plan

**Unit Tests:**

Added `test_create_session_concurrent_same_id_raises_already_exists_error`
in `tests/unittests/sessions/test_session_service.py`: two concurrent
`create_session()` calls for the same caller-provided `session_id` (after
pre-warming `app_state`/`user_state` with an unrelated session, to isolate
this race from the separate, already-guarded `_get_or_create_state` race).
Asserts exactly one caller succeeds and the other raises `AlreadyExistsError`,
and that exactly one session row survives.

```
$ pytest tests/unittests/sessions/test_session_service.py -q
399 passed, 2 xfailed in 11.01s

$ pytest ./tests/unittests -q
12619 passed, 88 skipped, 33 xfailed, 1 xpassed, 2 failed in 292.16s
```
The 2 failures (`test_import_loading.py::test_entry_point_loads_only_allowlisted_packages[agent|runner]`)
are pre-existing and unrelated (an `httpx2` eager-import allowlist check) —
reproduced identically on `main` without this change.

Additionally reproduced and verified the race outside the unit test, against
two real backends, before writing the fix into the codebase:

- **sqlite+aiosqlite** — 5/5 trials without the patch raised a raw
  `sqlite3.IntegrityError: UNIQUE constraint failed: sessions.app_name, sessions.user_id, sessions.id`.
- **Postgres 17 + asyncpg** (disposable local container) — 4/5 trials without
  the patch raised `sqlalchemy.dialects.postgresql.asyncpg.IntegrityError`
  wrapping `asyncpg.exceptions.UniqueViolationError: duplicate key value
  violates unique constraint "sessions_pkey"` (the 5th trial's slower
  coroutine legitimately hit the existing early-exists check before reaching
  the insert).

With the patch applied, 5/5 trials on both backends raised a clean
`AlreadyExistsError` instead, with exactly one session surviving each time.

### Checklist

- [x] I have manually tested my changes end-to-end (see above).

### Additional context

Pinned versions in the project where this was first observed (from `uv.lock`):
`google-adk==2.3.0`, `asyncpg==0.31.0`, `sqlalchemy==2.0.44`.

### DIFF
--- a/src/google/adk/sessions/database_session_service.py
+++ b/src/google/adk/sessions/database_session_service.py
@@ -639,7 +639,18 @@ class DatabaseSessionService(BaseSessionService):
           storage_app_state.state, storage_user_state.state, session_state
       )
       # Call to_session before commit to avoid post-commit lazy-load.
-      await sql_session.flush()
+      try:
+        await sql_session.flush()
+      except IntegrityError:
+        # A concurrent caller won the race on this (app_name, user_id,
+        # session_id) primary key: the has_user_provided_id check above is
+        # not atomic with this insert, so two callers can both pass it and
+        # then race the same insert. Same failure mode _get_or_create_state
+        # guards against for app_state/user_state; surface the same clean
+        # error here instead of letting the raw IntegrityError propagate.
+        raise AlreadyExistsError(
+            f"Session with id {session_id} already exists."
+        )
       session = storage_session.to_session(
           state=merged_state, is_sqlite=is_sqlite, is_postgresql=is_postgresql
       )
--- a/tests/unittests/sessions/test_session_service.py
+++ b/tests/unittests/sessions/test_session_service.py
@@ -1191,6 +1191,66 @@ async def test_create_session_with_existing_id_raises_error(session_service):
 
 
 @pytest.mark.asyncio
+async def test_create_session_concurrent_same_id_raises_already_exists_error(
+    tmp_path,
+):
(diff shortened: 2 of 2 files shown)


hungubqn0310:
recheck

PR #6845, opened 2026-08-21
fix: preserve Annotated metadata in function tool declarations and arg conversion

MarlzRana:

**Problem:**

When the `JSON_SCHEMA_FOR_FUNC_DECL` feature is enabled, `typing.Annotated` metadata was silently dropped in two places:

1. **Declaration builder.** `_function_tool_declarations._get_function_fields` called `get_type_hints()` without `include_extras=True`, so `Annotated[T, Field(description=..., ge=1)]` was unwrapped to a bare `T` before Pydantic's `create_model` saw it. Every description and constraint an author attached via `Annotated` was missing from the schema shown to the model. The legacy builder never had this problem, so this was a regression between the two paths.

2. **Argument conversion.** `FunctionTool._preprocess_args` resolves parameter types with `get_type_hints()`, which strips `Annotated` for free. But when `get_type_hints()` cannot resolve a signature (an unresolvable forward reference anywhere in it — e.g. a `TYPE_CHECKING`-only import or a recursive type alias), it raises and the code falls back to the raw `param.annotation`, which is still `Annotated`-wrapped. None of the conversion branches recognised the wrapper, so JSON dicts from the model reached the function body unconverted (later `AttributeError`).

**Solution:**

- Pass `include_extras=True` to `get_type_hints()` in the JSON-schema declaration builder (parameter and forward-ref return paths), so `Annotated` metadata reaches the generated schema — matching the legacy builder.
- Unwrap `Annotated` in `_preprocess_args`, both at the top level and per union member (`Optional[Annotated[T, ...]]` hides the wrapper inside the union), so dict-to-model conversion works on the fallback path too.

### Testing Plan

**Unit Tests:**

Added 16 tests across three existing files:

- `test_function_tool_declarations.py` — `Annotated` metadata (descriptions, constraints, defaults, optional models, nested lists, return type) reaches the JSON schema.
- `test_build_function_declaration.py` — parity pair asserting the legacy and JSON-schema builders both preserve `Annotated` metadata.
- `test_function_tool_pydantic.py` — `_preprocess_args` conversion on both the resolved and the `get_type_hints`-fallback paths, plus an end-to-end `run_async` case.

Each fix was verified by reverting it and confirming the corresponding tests fail (the e2e test fails with `AttributeError: 'dict' object has no attribute 'name'`, the real symptom).

```
$ pytest tests/unittests/tools/test_function_tool_declarations.py \
         tests/unittests/tools/test_build_function_declaration.py \
         tests/unittests/tools/test_function_tool_pydantic.py \
         tests/unittests/tools/test_function_tool.py \
         tests/unittests/tools/test_function_tool_with_import_annotations.py
162 passed, 2 warnings
```

Not applicable — the behaviour is fully exercised by the unit tests, including an end-to-end `run_async` invocation that asserts the wrapped function receives a real Pydantic model instance rather than a dict.

### Checklist

### DIFF
--- a/src/google/adk/tools/_function_tool_declarations.py
+++ b/src/google/adk/tools/_function_tool_declarations.py
@@ -63,7 +63,7 @@ def _get_function_fields(
 
   # Get type hints with forward reference resolution
   try:
-    type_hints = get_type_hints(func)
+    type_hints = get_type_hints(func, include_extras=True)
   except TypeError:
     # Can happen with mock objects or complex annotations
     type_hints = {}
@@ -160,7 +160,7 @@ def _build_response_json_schema(
   # Handle string annotations (forward references)
   if isinstance(return_annotation, str):
     try:
-      type_hints = get_type_hints(func)
+      type_hints = get_type_hints(func, include_extras=True)
       return_annotation = type_hints.get('return', return_annotation)
     except TypeError:
       pass
--- a/src/google/adk/tools/function_tool.py
+++ b/src/google/adk/tools/function_tool.py
@@ -20,6 +20,7 @@ import functools
 import inspect
 import logging
 from types import UnionType
+from typing import Annotated
 from typing import Any
 from typing import Awaitable
 from typing import Callable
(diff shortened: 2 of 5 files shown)


MarlzRana:
I will re-open this later after a bit more polish.

PR #6859, opened 2026-08-22
fix(a2a) warn when session state is dropped

### LINKED ISSUE #6854: Session state does not cross a RemoteA2aAgent boundary in either direction, silently
### Summary

Session state does not cross a `RemoteA2aAgent` boundary **in either direction**, and nothing in
the API surface, logs, or docs indicates it. Because the same agent graph works correctly
in-process, this is invisible to unit tests, integration tests and CI, and only appears once an
agent is actually deployed behind A2A.

We hit it twice, in opposite directions, in a `SequentialAgent` pipeline that ran fine locally for
weeks.

### Environment

- `google-adk==2.7.1`, `a2a-sdk==1.1.2`, Python 3.12
- Orchestrator is a `SequentialAgent` on Agent Engine; two peers are `RemoteA2aAgent`s backed by
  Cloud Run services

### Direction 1 — `output_key` on a remote agent never reaches the caller

```python
# runs in the worker process
auditor = LlmAgent(name="auditor", output_schema=Report, output_key="findings", ...)

# runs in the orchestrator
class NextStep(BaseAgent):
    async def _run_async_impl(self, ctx):
        report = ctx.session.state.get("findings")   # populated in-process, always None over A2A
```

`output_key` writes into the session of the agent that *declares* it. In-process the declaring
agent and the reading agent share one session, so this works. Over A2A the worker has its own
session and that state never returns to the caller.

### Direction 2 — a caller-side `state_delta` never reaches the remote agent

```python
class PolicyStep(BaseAgent):
    async def _run_async_impl(self, ctx):
        yield Event(author=self.name, invocation_id=ctx.invocation_id,
                    actions=EventActions(state_delta={"routing": routing}))   # no content

# next in the SequentialAgent
RemoteA2aAgent(name="notifier", ...)   # never sees "routing"
```

`RemoteA2aAgent._construct_message_parts_from_session` builds the outgoing message from
`event.content.parts` and skips events whose `content` is `None`. A state-only event therefore
contributes nothing to the A2A request, and the remote agent receives the *previous* content
event instead — in our case an earlier agent's output, which looked plausible and was silently
acted upon.

### Why this is worth a warning rather than only a doc note

The failure is silent, and the two directions fail differently:

- Direction 1 raises, if you happen to have a downstream check. If you don't, an `LlmAgent`
  reading the missing key will often invent plausible content instead.
- Direction 2 does not raise at all. The remote agent gets *some* content and proceeds.

Because in-process runs and A2A runs are described as the same composition with a different
transport, the natural assumption is that the hand-off is also the same. Our test suite has 100%
statement and branch coverage and could not have caught either one — the state key is populated in
exactly the topology tests run in.

### Suggested remedies, in the order we'd value them

1. **Warn when a `SequentialAgent` contains a `RemoteA2aAgent` and a sibling declares `output_key`**
   — the combination is almost always a mistake, and it is statically detectable at construction.
2. **Warn (or debug-log) when an event with a non-empty `state_delta` and no `content` is the last
   event before a `RemoteA2aAgent` runs**, since that state is about to be dropped.
3. **Document the boundary explicitly** in the A2A guide: state is per-session, sessions are
   per-agent across A2A, and only event content crosses.
4. Optionally, an opt-in `forward_state=` / `state_keys=` on `RemoteA2aAgent` for callers who want
   specific keys serialized into the request.

We worked around it by having the caller read the remote agent's reply from session events
(sound in our case only because `output_schema` makes that reply validated JSON rather than
prose), and by emitting the hand-off as event content rather than as `state_delta`. Both are fine
once you know; the cost was entirely in not knowing.

Happy to open a PR for (1) and (3) if that's a direction you'd take.

sylvesterkaczmarek:

- Closes #6854

**Problem**

Session state is local to each side of a `RemoteA2aAgent` boundary, but state loss was silent. A caller state-only event contributes no A2A parts, so the peer receives older content. Remote state deltas are intentionally rejected because a peer must not mutate caller state, but that rejection was only debug-logged.

**Solution**

- Warn before an outbound state-only hand-off is dropped.
- Warn when an inbound remote state delta is rejected.
- Document the boundary in the API docstring and task-mode guide.
- Document safe alternatives: send peer inputs as content and return caller-visible values as content or task output.

This preserves the security boundary while making both loss directions visible.

### Testing Plan

- [x] Added regression tests.
- [x] Relevant tests pass.

`pytest tests/unittests/agents/test_remote_a2a_agent.py tests/unittests/a2a/converters/test_to_adk.py -q --disable-warnings`

`262 passed in 2.56s`

Both warning directions are covered and unsafe state mutation remains blocked.

All applicable pre-commit hooks passed.

**Manual E2E**

Not run against deployed A2A services. The tests exercise the exact request-history and inbound-metadata boundaries.

### Checklist

- [x] I have read the CONTRIBUTING document.
- [x] I have self-reviewed this change.
- [x] I have added regression tests.
- [x] Relevant tests pass locally.
- [ ] I have tested against deployed A2A services.
- [x] This change has no downstream dependencies.

### Additional context

State forwarding remains out of scope because applying peer-controlled state would cross the existing trust boundary.

### DIFF
--- a/docs/guides/agents/remote_a2a_agent/task.md
+++ b/docs/guides/agents/remote_a2a_agent/task.md
@@ -196,6 +196,21 @@ behavior:
 
 ## Limitations
 
+### Session state boundary
+
+The caller and remote agent use separate sessions. Session state and
+`EventActions.state_delta` do not cross the A2A boundary in either direction:
+
+- `output_key` on the remote agent writes only to the remote server's session.
+- A caller-side state-only event has no content to include in the A2A request.
+- A state delta supplied by a remote peer is not applied to the caller's
+  session, because peers are not allowed to mutate caller state.
+
+Pass values required by the remote agent in event content. Return values needed
+by the caller as response content or, in task mode, as `finish_task` output.
+ADK logs a warning when it is about to drop a state-only hand-off or receives a
+remote state delta.
+
 -   **Workflow Graphs Not Supported**: `RemoteA2aAgent` in task mode
     (`mode="task"`) cannot be used as a node in ADK `Workflow` graphs. It is
     exclusively designed for sub-agent delegation under a parent coordinator
--- a/src/google/adk/a2a/converters/to_adk_event.py
+++ b/src/google/adk/a2a/converters/to_adk_event.py
@@ -321,10 +321,21 @@ def _extract_event_actions(metadata: Any) -> EventActions:
       for key, value in parsed_actions.items()
       if key in _PEER_SETTABLE_ACTION_FIELDS
   }
-  if len(peer_actions) != len(parsed_actions):
+  dropped_fields = set(parsed_actions) - set(peer_actions)
+  state_delta_alias = EventActions.model_fields["state_delta"].alias
+  state_delta_fields = {"state_delta"}
+  if state_delta_alias:
+    state_delta_fields.add(state_delta_alias)
+  if dropped_fields & state_delta_fields:
+    logger.warning(
+        "Ignoring a session state delta from a remote A2A peer. Session state"
+        " is local to each agent; return values needed by the caller as event"
+        " content or task output instead."
+    )
+  if dropped_fields:
     logger.debug(
         "Dropping ADK actions metadata fields that a peer may not set: %s",
-        sorted(set(parsed_actions) - set(peer_actions)),
+        sorted(dropped_fields),
     )
 
   try:
--- a/src/google/adk/agents/remote_a2a_agent.py
+++ b/src/google/adk/agents/remote_a2a_agent.py
@@ -638,7 +638,6 @@ class RemoteA2aAgent(BaseAgent):
   The agent handles:
   - Agent card resolution and validation
   - HTTP client management with proper resource cleanup
-  - A2A message conversion and error handling
   - Session state management across requests
   """
 
@@ -927,7 +926,7 @@ class RemoteA2aAgent(BaseAgent):
 
       with path.open("r", encoding="utf-8") as f:
         agent_json_data = json.load(f)
-      return _compat.parse_agent_card(agent_json_data)
+        return _compat.parse_agent_card(agent_json_data)
     except json.JSONDecodeError as e:
       raise AgentCardResolutionError(
           f"Invalid JSON in agent card file {file_path}: {e}"
--- a/src/google/adk/agents/remote_a2a_agent.py
+++ b/src/google/adk/agents/remote_a2a_agent.py
@@ -625,6 +625,11 @@ def _build_auth_interceptors(
 class RemoteA2aAgent(BaseAgent):
   """Agent that communicates with a remote A2A agent via A2A client.
 
+  Session state is local to each side of an A2A boundary. Only event content is
+  included in requests, and state deltas from a remote peer are not applied to
+  the caller's session. Put values needed by the peer in event content and
+  return values needed by the caller as content or task output.
+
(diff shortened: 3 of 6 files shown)


iarjunganesh:
Thanks for addressing the runtime boundary here. I traced the converter path and
confirmed that the inbound warning covers a remote `LlmAgent(output_key=...)`:
the remote event's `stateDelta` is serialized into artifact metadata and then
rejected by the caller converter with the new warning.

I opened #6862 for the complementary caller-side construction hazard: a local
`LlmAgent(output_key=...)` immediately followed by `RemoteA2aAgent`. That event
also contains content, so the state-only outbound warning in this PR does not
fire even though the associated session key is unavailable remotely. #6862 is
limited to that detectable ordering and touches separate files.

One review point for this PR: `_construct_message_parts_from_session()` checks
`ctx.session.events[-1]` before task-scope filtering. In task mode, that event
can be outside the current isolation scope, which could warn for an event that
is not part of the actual hand-off. Would it be safer to inspect the last
applicable event after `events_to_process` is built?

It may also be useful to add a round-trip regression test that starts with an
ADK event containing an `output_key` state delta, converts it to an A2A
artifact, and verifies that the caller emits the warning while preserving the
response content. That would directly lock down Direction 1 from #6854.

sylvesterkaczmarek:
> Thanks for addressing the runtime boundary here. I traced the converter path and confirmed that the inbound warning covers a remote `LlmAgent(output_key=...)`: the remote event's `stateDelta` is serialized into artifact metadata and then rejected by the caller converter with the new warning.
> 
> I opened #6862 for the complementary caller-side construction hazard: a local `LlmAgent(output_key=...)` immediately followed by `RemoteA2aAgent`. That event also contains content, so the state-only outbound warning in this PR does not fire even though the associated session key is unavailable remotely. #6862 is limited to that detectable ordering and touches separate files.
> 
> One review point for this PR: `_construct_message_parts_from_session()` checks `ctx.session.events[-1]` before task-scope filtering. In task mode, that event can be outside the current isolation scope, which could warn for an event that is not part of the actual hand-off. Would it be safer to inspect the last applicable event after `events_to_process` is built?
> 
> It may also be useful to add a round-trip regression test that starts with an ADK event containing an `output_key` state delta, converts it to an A2A artifact, and verifies that the caller emits the warning while preserving the response content. That would directly lock down Direction 1 from #6854.


Hi @iarjunganesh thanks a lot for addressing the runtime boundary here. So, I traced the converter path and confirmed that the inbound warning covers a remote `LlmAgent(output_key=...)`: the remote event's `stateDelta` is serialized into artifact metadata and then rejected by the caller converter with the new warning.

I opened #6862 for the complementary caller-side construction hazard: a local `LlmAgent(output_key=...)` immediately followed by `RemoteA2aAgent`. That event also contains content, so the state-only outbound warning in this PR does not fire even though the associated session key is unavailable remotely. #6862 is limited to that detectable ordering and touches separate files.

One review point for this PR: `_construct_message_parts_from_session()` checks `ctx.session.events[-1]` before task-scope filtering. In task mode, that event can be outside the current isolation scope, which could warn for an event that is not part of the actual hand-off. Would it be safer to inspect the last applicable event after `events_to_process` is built?

It may also be useful to add a round-trip regression test that starts with an ADK event containing an `output_key` state delta, converts it to an A2A artifact, and verifies that the caller emits the warning while preserving the response content. That would directly lock down Direction 1 from #6854.

PR #6862, opened 2026-08-22
fix(a2a): warn when output_key cannot reach remote session

iarjunganesh:

- Related: #6854
- Complements: #6859

**Problem:**

When an `LlmAgent` with `output_key` immediately precedes a
`RemoteA2aAgent` in a `SequentialAgent`, the generated content crosses A2A but
the associated session-state key does not. Because the event also has content,
the state-only outbound warning proposed in #6859 does not cover this hand-off.

**Solution:**

Warn at `SequentialAgent` construction for that specific adjacent ordering.
The check detects an already-loaded `RemoteA2aAgent` without importing it, so
ordinary `SequentialAgent` users do not acquire a dependency on the optional
A2A SDK.

### Testing Plan

**Unit Tests:**

- [x] Added positive coverage for `LlmAgent(output_key)` immediately followed
  by `RemoteA2aAgent`.
- [x] Added negative coverage for reversed ordering, a remote hand-off without
  `output_key`, and an entirely local hand-off.
- [x] `pytest tests/unittests/agents/test_sequential_agent.py -q`
  (`10 passed` on Python 3.12).
- [x] Strict mypy passed for `sequential_agent.py`.
- [x] `pyink`, `isort`, `ruff`, ADK compliance checks, and `codespell` passed
  for both changed files.
- [x] Verified in an isolated Python 3.12 environment without `a2a-sdk` that
  importing and constructing a local `SequentialAgent` still works.

Not run. This change is a deterministic construction-time warning and does not
alter execution or transport behavior.

### Checklist

- [x] I have read the `CONTRIBUTING.md` document.
- [x] I have commented the optional-dependency-sensitive code.
- [x] I have added tests that prove the warning and no-warning paths.
- [x] New and existing tests in the affected unit pass locally.
- [ ] I have manually tested the change end-to-end.
- [x] This change has no downstream dependencies.

### Additional context

This is intentionally separate from #6859: that PR covers state-only outbound
events and state deltas received from a remote peer. This PR covers a
caller-side `output_key` event that also contains content, so it is neither a
duplicate nor dependent on #6859.

### DIFF
--- a/src/google/adk/agents/sequential_agent.py
+++ b/src/google/adk/agents/sequential_agent.py
@@ -18,10 +18,12 @@ from __future__ import annotations
 
 import inspect
 import logging
+import sys
 from typing import AsyncGenerator
 from typing import ClassVar
 from typing import Type
 
+from pydantic import model_validator
 from typing_extensions import deprecated
 from typing_extensions import override
 
@@ -41,6 +43,13 @@ from .readonly_context import ReadonlyContext
 from .sequential_agent_config import SequentialAgentConfig
 
 logger = logging.getLogger('google_adk.' + __name__)
+
+
+def _is_remote_a2a_agent(agent: BaseAgent) -> bool:
+  """Checks the agent type without importing the optional A2A dependency."""
+  remote_module = sys.modules.get('google.adk.agents.remote_a2a_agent')
+  remote_type = getattr(remote_module, 'RemoteA2aAgent', None)
+  return remote_type is not None and isinstance(agent, remote_type)
 
 
 def _tool_name(tool: ToolUnion) -> str | None:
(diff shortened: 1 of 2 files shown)


DeanChensj:
Hi @ALDRIN121,

Thank you for putting together this PR and for the detailed description and test coverage.

After reviewing the approach, we've decided not to merge this change into `SequentialAgent`, for a few key reasons:

1. **`SequentialAgent` is deprecated**: `SequentialAgent` is deprecated in favor of `Workflow`, and we want to avoid introducing new domain-specific validations or tight couplings (such as referencing `RemoteA2aAgent`) into deprecated legacy components.
2. **Boundary concerns belong in the transport layer**: The session state boundary across A2A is a general transport property rather than an issue specific to `SequentialAgent`. It applies equally to `Workflow`, custom agents, and standalone runs. This is already being addressed centrally at the `RemoteA2aAgent` converter and documentation layer in #6859.
3. **Import order dependency**: The `sys.modules.get` reflection check is sensitive to module import order and would silently skip validation if `RemoteA2aAgent` is imported lazily or aliased.

Given this, we recommend closing this PR in favor of #6859.

On a related note, we'd love to learn more about your use case: **is there a specific reason or blocker that prevented you from using `Workflow` instead of `SequentialAgent`?** (e.g., missing features, ergonomics, or compatibility issues with `RemoteA2aAgent` in `Workflow`). Any feedback here would be very helpful for guiding our workflow orchestration improvements.

Thanks again for contributing!

PR #6867, opened 2026-08-23
fix(litellm): survive malformed JSON in tool call arguments

### LINKED ISSUE #5896: [LiteLlm] JSONDecodeError on malformed tool_call.function.arguments still crashes invocation at v2.0.0 (regression of #5008)

adnanahamed66772ndpc:

- Closes: #5896

**Problem:**

When the upstream model returns malformed JSON in `tool_call.function.arguments` (truncated strings, unclosed objects, partial streams committed to the message), `_message_to_generate_content_response` raised `json.JSONDecodeError`. The exception propagated out of LiteLLM response parsing — upstream of any tool or callback boundary — so the entire invocation aborted with a generic streaming error and no information about the malformed tool call. This is a regression of #5008, still present at `v2.0.0`.

**Solution:**

Parse tool call arguments defensively inside `_message_to_generate_content_response`: on `json.JSONDecodeError`, log a warning carrying the tool call id, function name and parse error, then fall back to empty args on the function-call part. Downstream tool dispatch then surfaces a structured argument mismatch back to the model, which can retry with corrected JSON — the agent loop survives.

Strict parsing is preserved in `_parse_tool_call_arguments` itself because the streaming path (`_finalize_tool_call_response`) relies on it to detect arguments truncated by `finish_reason == "length"` and report a dedicated MAX_TOKENS error response; only the final conversion step now recovers instead of crashing.

### Testing Plan

**Unit Tests:**

- Added `test_message_to_generate_content_response_tool_call_with_malformed_json_arguments`: truncated arguments produce a function-call part with empty args while preserving name and id.
- Added `test_message_to_generate_content_response_tool_call_malformed_json_warns`: the recovery emits a warning-level record.
- Full model suite: `pytest tests/unittests/models/test_litellm.py` → **395 passed**, including the existing Python-literal and unquoted-key repair tests which confirm no behavior change for valid inputs.
- Reproduced the exact crash from #5896 before the fix (JSONDecodeError propagating out of `_message_to_generate_content_response`) and verified it no longer reproduces after the fix.
- `pre-commit run --files` (ruff, isort, pyink, addlicense, compliance checks) and `mypy src/google/adk/models/lite_llm.py` all pass.

Not applicable — the fix is covered by unit tests that construct the LiteLLM message directly, exactly matching the minimal reproduction from the issue. No network or provider credentials are required.


GWeale:
fixed on main in 75d84b11. a call whose arguments won't parse is dropped now, and the response carries a malformed_function_call finish reason. the branch conflicts, and it isn't in 2.8.0. open a new issue if you see anything like this again.

PR #6872, opened 2026-08-23
fix(a2a): mark terminal A2A tasks as final responses

### LINKED ISSUE #6584: RemoteA2aAgent: a terminal A2A task carrying tool activity fails is_final_response(), so the turn cannot be detected as finished (non-streaming peer)
### Environment

- `google-adk` 2.6.0
- `a2a-sdk` 1.1.2
- Python 3.12

### Summary

When a `RemoteA2aAgent` talks to a peer that does **not** advertise streaming, the a2a-sdk
correctly falls back to `message/send` and ADK hands the whole response over as a **single
`Event`** carrying the narration text, every `function_call`, every `function_response` and the
closing text.

That event's task state is terminal — `event.custom_metadata["a2a:response"]["status"]["state"]`
is `TASK_STATE_COMPLETED` — but **`event.is_final_response()` returns `False`**, because the
predicate excludes any event holding tool activity:

```python
# google/adk/events/event.py
def is_final_response(self) -> bool:
    if self.actions.skip_summarization or self.long_running_tool_ids:
        return True
    return (
        not self.get_function_calls()
        and not self.get_function_responses()
        and not self.partial
        and not self.has_trailing_code_execution_result()
    )
```

So ADK produces a completed task in one event, and its own public helper cannot recognise that
event as the end of the turn. A consumer that closes the turn on `is_final_response()` — which is
what the docstring recommends ("Application and UI layers can rely on this helper to detect a
complete, user-facing response instead of replicating its logic") — never closes it.

This is the **default** path for a peer built with ADK, not an exotic one: `to_a2a(...)` builds
its card through `AgentCardBuilder` with `capabilities.streaming` left at `False`, so an
ADK-served agent is non-streaming unless the card is built by hand.

### Reproduction

`server.py` — a plain ADK agent served over A2A, with the default card:

```python
from google.adk.a2a.executor.a2a_agent_executor import A2aAgentExecutor
from google.adk.a2a.utils.agent_to_a2a import to_a2a

def divide(a: float, b: float) -> dict:
    return {"error": "cannot divide by zero"} if b == 0 else {"quotient": a / b}

root_agent = LlmAgent(name="peer", model="<any>", tools=[divide],
                      instruction="Use the tool, then summarise the result.")

# force_new_version so the v2 integration path is exercised; the v1 path behaves the same way.
app = to_a2a(root_agent, host="127.0.0.1", port=8100,
             agent_executor_factory=lambda runner: A2aAgentExecutor(
                 runner=runner, use_legacy=False, force_new_version=True))
```

Client side: a `RemoteA2aAgent` pointing at `http://127.0.0.1:8100`, as a sub-agent of any root
agent, and one turn that makes the peer call its tool (e.g. "ask the peer to divide 10 by 0").

### Observed

The peer's card says `{"streaming": false}`, so `a2a/client/base_client.py` takes the
non-streaming branch:

```python
if not self._config.streaming or not self._card.capabilities.streaming:
    response = await self._execute_with_interceptors(..., method='send_message', ...)
```

`message/send` returns one completed task with three artifacts (text, `function_call`,
`function_response`), and `_handle_a2a_response_v2` converts it into one `Event` where:

- `event.get_function_calls()` and `event.get_function_responses()` are both non-empty,
- `event.custom_metadata["a2a:response"]["status"]["state"] == "TASK_STATE_COMPLETED"`,
- `event.is_final_response()` is `False`,
- and it is the **last** event of the invocation — `run_async` ends right after it.

### Expected

Either of these would let a consumer detect the end of the turn using ADK's own API:

1. `is_final_response()` recognises an event whose A2A task state is terminal (regardless of the
   tool activity it carries) — for instance by having the handler set `actions.skip_summarization`
   or an equivalent marker when the task state is terminal; or
2. `_handle_a2a_response_v2` emits one event per artifact for a non-streaming response, mirroring
   what the streaming path produces, so the last event satisfies the predicate.

Option 2 has a second benefit: today the *same* logical turn is persisted with a different shape
depending only on whether the peer streams (one event with N parts vs. N events), which any
consumer storing events has to absorb.

### Impact / workaround

A UI driven by `is_final_response()` shows the tool activity and then waits forever, since nothing
marks the turn as over. The workaround is to stop trusting the helper for A2A sub-agents and read
the task state directly:

```python
closes_task = (
    event.author in remote_a2a_agent_names
    and task_state_of(event) in TERMINAL_STATES  # from custom_metadata["a2a:response"]
)
is_final = event.is_final_response() or closes_task
```

which means every consumer has to re-implement the part of the predicate ADK asks them not to.

Related: #6274 (a2a-sdk 1.x), #4016 (`is_final_response()` docs vs. code).

Akshaay1:
Fixes #6584.

## Problem

`convert_a2a_task_to_event` never reflected the A2A task's terminal state onto the event it produced.

When a peer does not advertise streaming, the a2a-sdk falls back to `message/send` and the entire turn arrives as a **single task**. The converted event therefore carries the peer's narration, its `function_call`s, its `function_response`s and its closing text all together.

`Event.is_final_response()` returns `False` for any event holding function calls or responses:

```python
return (
    not self.get_function_calls()
    and not self.get_function_responses()
    and not self.partial
    and not self.has_trailing_code_execution_result()
)
```

So ADK produced a completed task as the **last event of the invocation**, and its own public helper could not recognise it as the end of the turn. A consumer that closes the turn on `is_final_response()` — which that helper's docstring explicitly recommends ("Application and UI layers can rely on this helper to detect a complete, user-facing response instead of replicating its logic") — never closed it.

This is the **default** path for an ADK-served peer, not an exotic one: `to_a2a(...)` builds its card through `AgentCardBuilder` with `capabilities.streaming` left at `False`.

## Fix

Set `event.actions.skip_summarization` when the task state is `completed`, `failed` or `canceled`.

Reasoning for that choice:

- `skip_summarization` is already the established signal for *"this event is final despite carrying tool activity"* — it is the first thing `is_final_response()` checks.
- It is **already trusted from a remote peer**: it is one of only three entries in this module's `_PEER_SETTABLE_ACTION_FIELDS` allowlist, whose docstring classifies it as inert, in contrast to the fields that "mutate the caller's own session ... or drive the caller's control flow and persistence". Setting it here therefore does not widen what an A2A response can influence.
- Blast radius is narrow: across the entire codebase, `event.actions.skip_summarization` has exactly one behavioural consumer (`Event.is_final_response()`); the only other reference is persistence in `vertex_ai_session_service`. The `functions.py` consumer reads `tool_context.actions`, which is a different object on the local tool-execution path.
- `Event.is_final_response()` itself is left untouched, so the generic contract does not change.

`input_required` and `auth_required` are **deliberately excluded**. They pause the turn to ask the caller for something rather than ending it, and they already reach `is_final_response()` through the mock function call `_create_mock_function_call_for_required_user_input` builds for them. This was the edge case raised in review on the issue, and it is covered by a test that asserts those states are left exactly as they were.

Applied to both converters — the v2 `to_adk_event.py` and the legacy `event_converter.py` that `RemoteA2aAgent` imports — since the issue reports both paths behave the same way.

In the legacy converter the state is read through a small `_is_terminal_task()` helper using `getattr`, because 1.x tasks carry a protobuf `TaskStatus` whose fields are not always reachable on stand-in objects; an unreadable state simply means we cannot claim the turn is over.

## Before / after

Reproduced deterministically — no API key and no model call. A completed task carrying one function call and its response:

| task state | `is_final_response()` before | after |
|---|---|---|
| `completed` | `False` ❌ | `True` ✅ |
| `failed` | `False` ❌ | `True` ✅ |
| `canceled` | `False` ❌ | `True` ✅ |
| `working` | `False` ✅ | `False` ✅ (unchanged) |
| `input_required` | `True` ✅ | `True` ✅ (unchanged) |

## Tests

Added to both converter test modules:

- `test_convert_a2a_task_to_event_terminal_task_is_final_response` — parametrized over `completed` / `failed` / `canceled`; asserts the tool activity is still carried on the event *and* that the turn now reads as closed.
- `test_convert_a2a_task_to_event_non_terminal_task_is_not_final` / `..._working_task_is_not_final` — an in-flight task must keep reporting the turn is open.
- `test_convert_a2a_task_to_event_pause_state_left_untouched` — `input_required` / `auth_required` keep resolving through the mock function call, with `skip_summarization` untouched.

Verified the new tests fail without the source change (`git stash` on the two source files only): the 6 terminal-state cases fail, and the guard tests pass either way, as regression guards should.

```
tests/unittests/a2a/converters/   223 passed, 6 skipped
tests/unittests/a2a/ + events/ + agents/test_remote_a2a_agent.py   804 passed, 45 skipped
```

Full unit suite: `9095 passed, 56 skipped, 19 xfailed`.

The only failures are the two pre-existing `test_import_loading.py::test_entry_point_loads_only_allowlisted_packages[agent|runner]` cases, which reproduce identically on a clean checkout of `main` with no changes applied (environment-related, unrelated to this change).

`pyink` and the pre-commit hooks are clean.

### DIFF
--- a/src/google/adk/a2a/converters/event_converter.py
+++ b/src/google/adk/a2a/converters/event_converter.py
@@ -51,6 +51,24 @@ DEFAULT_ERROR_MESSAGE = "An error occurred during processing"
 # Logger
 logger = logging.getLogger("google_adk." + __name__)
 
+_TERMINAL_TASK_STATES = frozenset({
+    _compat.TS_COMPLETED,
+    _compat.TS_FAILED,
+    _compat.TS_CANCELED,
+})
+"""Task states that mean the peer's turn is over."""
+
+
+def _is_terminal_task(a2a_task: Task) -> bool:
+  """Returns whether the task reports a state that ends the peer's turn.
+
+  ``state`` is read defensively: 1.x tasks carry a protobuf ``TaskStatus``
+  whose fields are not always reachable on stand-in objects, and an
+  unreadable state simply means we cannot claim the turn is over.
+  """
+  status = getattr(a2a_task, "status", None)
+  return getattr(status, "state", None) in _TERMINAL_TASK_STATES
+
 
 AdkEventToA2AEventsConverter = Callable[
     [
@@ -256,6 +274,12 @@ def convert_a2a_task_to_event(
         event: Event = convert_a2a_message_to_event(
             message, author, invocation_context, part_converter=part_converter
         )
+        if _is_terminal_task(a2a_task):
+          # See the note in ``to_adk_event.convert_a2a_task_to_event``: a
+          # terminal task is the end of the peer's turn, but the event holds
+          # the peer's tool activity, so ``is_final_response()`` would report
+          # False without this.
+          event.actions.skip_summarization = True
         return event
       except Exception as e:
         logger.error("Failed to convert A2A task message to event: %s", e)
--- a/src/google/adk/a2a/converters/to_adk_event.py
+++ b/src/google/adk/a2a/converters/to_adk_event.py
@@ -290,6 +290,19 @@ metadata uses the camelCase aliases, so both spellings are listed.
 """
 
 
+_TERMINAL_TASK_STATES = frozenset({
+    _compat.TS_COMPLETED,
+    _compat.TS_FAILED,
+    _compat.TS_CANCELED,
+})
+"""Task states that mean the peer's turn is over.
+
+``input_required`` and ``auth_required`` are deliberately excluded: those pause
+the turn to ask the caller for something, and they already reach
+``is_final_response()`` through the mock function call built for them.
+"""
+
+
 def _extract_event_actions(metadata: Any) -> EventActions:
   """Extracts ADK event actions from A2A metadata.
 
@@ -533,6 +546,17 @@ def convert_a2a_task_to_event(
             a2a_task.status.state, output_parts, long_running_function_ids
         )
     )
+
+    if a2a_task.status.state in _TERMINAL_TASK_STATES:
+      # A terminal task is the whole of the peer's turn, so the event carries
+      # the peer's tool activity alongside its closing text. Without this,
+      # `Event.is_final_response()` sees those function calls/responses and
+      # reports False, and a caller that closes the turn on that helper never
+      # closes it. `skip_summarization` is the existing signal for "this event
+      # is final despite carrying tool activity"; it is also already trusted
+      # from the peer (`_PEER_SETTABLE_ACTION_FIELDS`), so setting it here does
+      # not widen what an A2A response can influence.
+      event_actions.skip_summarization = True
 
     return _create_event(
         output_parts,
--- a/tests/unittests/a2a/converters/test_event_converter.py
+++ b/tests/unittests/a2a/converters/test_event_converter.py
@@ -739,6 +739,85 @@ class TestA2AToEventConverters:
       assert called_message.role == _compat.ROLE_AGENT
       assert called_message.parts == [artifact_part]
 
+  @pytest.mark.parametrize(
+      "terminal_state",
+      [_compat.TS_COMPLETED, _compat.TS_FAILED, _compat.TS_CANCELED],
+  )
+  def test_convert_a2a_task_to_event_terminal_task_is_final_response(
+      self, terminal_state
+  ):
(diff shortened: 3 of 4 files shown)


GWeale:
fixed on main in aef3a9c4, which came in from #6787. input_required and auth_required still aren't covered by tests, so a small pull request for those is welcome. open a new issue if you see anything like this again.

PR #6888, opened 2026-08-24
fix(sessions): accept a full session resource name again on v1

GWeale:
Ports `60c55ad74` ("fix vertex_ai_session_service crashing when Agent Engine passes full resource names instead of short session IDs") to `v1`, with its regression tests.

#6809 brought `_validate_session_id` across, which requires a session id to match `^[A-Za-z0-9_-]+$` under `fullmatch`. Upstream added that check in `b2916c71` and then, 13 days later, had to add `_extract_short_session_id` in `60c55ad74` because Agent Engine passes the full resource name rather than the short id. `v1` took the check and not the follow-up, so `get_session`, `delete_session` and `create_session` now reject a name of the form `projects/{p}/locations/{l}/reasoningEngines/{e}/sessions/{s}` outright. `main` has accepted it since June.

- `_extract_short_session_id` is ported verbatim from upstream. It returns the last path segment when the id looks like a session resource name, and raises when the embedded `reasoningEngines/{id}` does not match the engine the service is configured for.
- The three call sites now resolve `reasoning_engine_id` first, normalize, then validate, matching upstream's ordering.
- `append_event` still validates `session.id` without normalizing, which is what `main` does at `:396`.

Behaviour for anything that is not a session resource name is unchanged: `a/b`, `../foo` and `''` are still rejected, since the normalizer only strips when the second-to-last segment is `sessions`.

Tests: the four unit tests from upstream, plus two end-to-end ones covering a full resource name being accepted and a resource name for a different engine being rejected. Deleting the normalizer call makes both of the latter fail while the existing path-traversal test still passes.

### DIFF
--- a/tests/unittests/sessions/test_vertex_ai_session_service.py
+++ b/tests/unittests/sessions/test_vertex_ai_session_service.py
@@ -31,6 +31,8 @@ from google.adk.events.event_actions import EventCompaction
 from google.adk.models.cache_metadata import CacheMetadata
 from google.adk.sessions.base_session_service import GetSessionConfig
 from google.adk.sessions.session import Session
+from google.adk.sessions.vertex_ai_session_service import _extract_short_session_id
+from google.adk.sessions.vertex_ai_session_service import _validate_session_id
 from google.adk.sessions.vertex_ai_session_service import VertexAiSessionService
 from google.api_core import exceptions as api_core_exceptions
 from google.genai import types as genai_types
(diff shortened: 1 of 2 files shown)


tests/unittests/sessions/test_vertex_ai_session_service.py
```diff
@@ -764,6 +766,70 @@ async def test_session_id_path_traversal_rejected():
...
+  resource_name = (
+      'projects/test-project/locations/test-location/'
+      'reasoningEngines/123/sessions/1'
+  )
+
+  session = await session_service.get_session(
+      app_name='123', user_id='user', session_id=resource_name
+  )
+
+  assert session.id == '1'
```
xuanyang15, line 817: The port itself checks out -- I diffed the added and removed lines of `vertex_ai_session_service.py` against `60c55ad74` and they are identical, the three call sites resolve the engine id before normalizing in the same order upstream does, and `a/b`, `../foo` and `''` are all still rejected because the normalizer only strips when the second-to-last segment is `sessions`.

One thing the swap loses. Upstream's `test_get_session_strips_full_resource_name` also asserted the outgoing request:

```python
  mock_api_client_instance.agent_engines.sessions.get.assert_called_once_with(
      name='reasoningEngines/123/sessions/session-123'
  )
```

The mock resolves a session with `session_id = name.split('/')[-1]` (:391), so `assert session.id == '1'` cannot distinguish a correctly built resource name from one that embedded the full name and happens to end in the same segment -- `reasoningEngines/123/sessions/projects/.../sessions/1` would find the session and pass. You are right that deleting the normalizer call fails both new tests, since validation rejects the slashes first; what is no longer pinned is that the value reaching the resource-name f-string is the normalized one.

The two tests here are otherwise a gain over upstream -- the mismatch case only exists at unit level on `main`. Would you mind adding the `assert_called_once_with` to the accepts case, so the outgoing name is covered too?

xuanyang15 (approved):

tests/unittests/sessions/test_vertex_ai_session_service.py
GWeale: good catch, the mock resolving on the last path segment does make `session.id` blind to this. added the `assert_called_once_with`.

confirmed it matters: building the name from the un-normalized id, so the outgoing name is `reasoningEngines/123/sessions/projects/p/.../sessions/1`, passes the old test and fails with the assertion in place.

PR #6890, opened 2026-08-24
fix: close the residual gaps from the v1 backport review

GWeale:
Closes the three code findings from the review of the v1 security backport batch (#6788–#6809) that were not regressions from those PRs and so were left for a follow-up.

**Deleting an artifact no longer deletes the artifacts nested under it.** Ports `08d21cc35` and `a5864a0ed`, which are a pair. Filenames may contain `/`, so `doc` and `doc/nested` are two distinct artifacts, and a nested artifact is stored beneath its parent. Deletion removed the artifact's whole directory, so deleting `doc` destroyed `doc/nested` as well, and `doc/nested` never appeared in `list_artifact_keys`. Deletion now removes only the artifact's own `versions/` directory and then prunes whatever it leaves empty, bounded by the scope root. Listing keeps walking past an artifact directory, skipping only that artifact's versions.

The GCS half of `08d21cc35` is carried too, because the same conflation is present here: version scans matched every blob under the prefix, so `list_versions("doc")` reported versions belonging to `doc/nested`. Loading and saving resolve `max(versions)`, so a read without an explicit version could resolve to a version the artifact does not have and silently return None, and the next write skipped version numbers. A blob now counts only when its name is the prefix followed by a single decimal segment, which also replaces an unguarded `int()` that raised on any object in the bucket not written by ADK.

**A skill name may no longer carry a trailing newline.** `_SNAKE_OR_KEBAB_NAME_PATTERN` ends in `$`, which outside `re.MULTILINE` also matches just before a trailing newline, so `match` let `"my-skill\n"` through into the resource name. `fullmatch` does not. Traversal was already blocked either way, so this only closes the newline. Upstream covers the same gap with a second layer, `quote(name, safe='')`, at the point where it interpolates the name into a URL; `v1` calls the SDK with a resource name rather than building a URL, so percent-encoding would be wrong here and the validator is the right place.

**A failing temp-file cleanup no longer masks the error that caused it.** The cleanup in `add_session_to_memory` runs in a `finally`, so an exception raised there displaces the one already propagating, including `CancelledError`. It caught only `FileNotFoundError`. A cancelled `asyncio.to_thread` returns while the worker thread may still hold the file, and on Windows the remove then fails with `PermissionError`. Now any `OSError` is logged and swallowed, which is what a best-effort cleanup should do. I could not reproduce the Windows-specific failure here, since POSIX `unlink` succeeds on an open file, but the displacement is a property of `finally` rather than of one platform.

Each change is covered by a test that fails when only that change is reverted. Artifact coverage is upstream's, parametrized across the in-memory, GCS and file backends; two assertions compare artifact text rather than `Part` equality because `v1`'s GCS backend round-trips a text part as `inline_data`, which is a pre-existing difference from `main` and not touched here.

Not included, from the same review: the ten defects that are byte-identical on `main`, which belong upstream, and the release-note items for the artifact storage layout and the API registry credential scoping.

### DIFF
--- a/src/google/adk/memory/vertex_ai_rag_memory_service.py
+++ b/src/google/adk/memory/vertex_ai_rag_memory_service.py
@@ -20,6 +20,7 @@ import base64
 import binascii
 from collections import OrderedDict
 import json
+import logging
 import os
 import tempfile
 from typing import Optional
@@ -37,6 +38,7 @@ if TYPE_CHECKING:
   from ..events.event import Event
   from ..sessions.session import Session
 
+logger = logging.getLogger("google_adk." + __name__)
 
 _SOURCE_DISPLAY_NAME_PREFIX = "adk-memory-v1."
 
@@ -178,8 +180,15 @@ class VertexAiRagMemoryService(BaseMemoryService):
       if temp_file_path:
         try:
           os.remove(temp_file_path)
-        except FileNotFoundError:
-          pass
+        except OSError:
+          # Best effort: this runs in a finally, so raising here would
+          # displace the exception already propagating, and a cancelled
+          # upload can leave the worker thread still holding the file.
+          logger.warning(
+              "Could not remove the temporary transcript at %s",
+              temp_file_path,
+              exc_info=True,
+          )
 
   @override
   async def search_memory(
--- a/src/google/adk/artifacts/file_artifact_service.py
+++ b/src/google/adk/artifacts/file_artifact_service.py
@@ -50,7 +50,10 @@ def _iter_artifact_dirs(root: Path) -> list[Path]:
     current = Path(dirpath)
     if (current / "versions").exists():
       artifact_dirs.append(current)
-      dirnames.clear()
+      # An artifact directory doubles as the parent of anything nested under
+      # it ("doc" and "doc/nested"), so keep walking, skipping only the
+      # stored versions of this artifact.
+      dirnames[:] = [name for name in dirnames if name != "versions"]
   return artifact_dirs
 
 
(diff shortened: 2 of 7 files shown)


src/google/adk/memory/vertex_ai_rag_memory_service.py
```diff
@@ -178,8 +180,15 @@ async def add_session_to_memory(self, session: Session) -> None:
       if temp_file_path:
         try:
           os.remove(temp_file_path)
-        except FileNotFoundError:
-          pass
+        except OSError:
```
xuanyang15, line 183: Checked the other three against upstream and they line up: `_prune_empty_dirs`, `_iter_artifact_dirs` and `_delete_artifact_sync` are byte-identical to `origin/main`, as is the GCS `_parse_version`. Good catch on the `list_versions` side of the GCS conflation -- a read resolving `max(versions)` to a version the artifact does not own was not something I had traced.

Also, thank you for correcting the skill-name suggestion. I offered `.fullmatch()` or `quote()` as equivalent options and they are not: v1 hands a resource name to `self._client.skills.get(name=...)` where `main` builds a URL string, so percent-encoding here would double-encode. `fullmatch` is the right call and the comment you added says why.

This one is the exception to the pattern, and it is the only thing I would raise. It is not a port -- `origin/main:283-288` is still `except FileNotFoundError: pass`, so this leaves v1 ahead of `main` and the next backport touching this file has a divergence to reconcile by hand. The description already makes the argument that would justify fixing `main` too: "the displacement is a property of `finally` rather than of one platform". The `to_thread` wrapper makes it easier to hit on v1, but a read-only mount or a permissions problem displaces the propagating exception on `main` just the same.

Non-blocking, and I would not hold this PR for it -- would it be worth sending the same three lines upstream so the two converge rather than drift?

xuanyang15 (approved):

src/google/adk/memory/vertex_ai_rag_memory_service.py
GWeale: agree

PR #6898, opened 2026-08-25
fix(models): aggregate streaming thought parts for all LiteLlm providers

### LINKED ISSUE #6895: LiteLlm streaming stores one thought part per reasoning delta
## 🔴 Required Information

**Describe the Bug:**

When `LiteLlm` streams, the aggregated (non-partial) `LlmResponse` joins the text buffer into a single part but passes the reasoning buffer through unjoined, so the response — and therefore the persisted session event — holds one `types.Part(thought=True)` per streamed reasoning delta.

`_aggregate_streaming_thought_parts` already does this joining, and its docstring says it produces clean parts "for session history and outbound requests", but it is only called from `_content_to_message_param` on the Anthropic path. The two stream finalizers pass `thought_parts=list(reasoning_parts)` instead.

**Steps to Reproduce:**

1. Configure any `LiteLlm` model with reasoning enabled (`xai/grok-4.6`, `anthropic/claude-*`, OpenAI reasoning models, …).
2. Call `generate_content_async(..., stream=True)`.
3. Inspect `content.parts` on the final response where `partial` is falsy.

**Expected Behavior:**

The same shape the non-streaming path produces: one thought part per thinking block, exactly as text becomes one part rather than one per delta.

**Observed Behavior:**

One thought part per delta. With xAI, whose deltas are per-token, a single thought arrives as ~40 parts:

```json
{"role": "model", "parts": [
  {"text": "The", "thought": true},
  {"text": " user", "thought": true},
  {"text": " wants", "thought": true}
]}
```

Because these land in session history, every downstream consumer sees them: UIs render one thinking block per token, and prompts built from a transcript pay the per-part JSON overhead many times over.

**Environment Details:**

 - ADK Library Version: 2.7.1
 - Desktop OS: macOS 26.3.1
 - Python Version: 3.13.5

**Model Information:**

 - Are you using LiteLLM: Yes
 - Which model is being used: `xai/grok-4.6` (reproduces on any reasoning model routed through LiteLLM)

---

## 🟡 Optional Information

**Additional Context:**

Signature boundaries matter for the fix: Anthropic closes each thinking block with a signature-only delta, and a signature only matches the text of its own block, so blocks must aggregate one part each rather than collapsing into one. `_aggregate_streaming_thought_parts` already splits on `thought_signature`, so calling it from both finalizers handles that case.

```python
async for response in LiteLlm(model="xai/grok-4.6").generate_content_async(
    llm_request, stream=True
):
    if not response.partial:
        print([p.text for p in response.content.parts if p.thought])
```

 - Always (100%)

Ashfaqbs:

- Closes: #6895

**Problem:**
`LiteLlm` streaming builds the aggregated (non-partial) `LlmResponse` by joining the buffered text into a single part, but passes the buffered reasoning parts straight through unjoined. So for a provider that streams reasoning token-by-token (e.g. `xai/grok-4.6`, OpenAI reasoning models via LiteLLM), the final response — and therefore the persisted session event — holds one `types.Part(thought=True)` per streamed reasoning delta instead of one part per thinking block. `_aggregate_streaming_thought_parts` already does this joining (splitting on `thought_signature` so Anthropic's per-block boundaries are preserved), but it was only wired into the Anthropic message-building path (`_content_to_message_param`), not into the two stream finalizers that build the response object itself.

**Solution:**
Call `_aggregate_streaming_thought_parts(reasoning_parts)` from both `_finalize_tool_call_response` and `_finalize_text_response` in `src/google/adk/models/lite_llm.py`, instead of `list(reasoning_parts)`. This makes every LiteLlm provider produce the same shape the non-streaming path already produces, and matches the aggregator's own docstring ("produces clean parts for session history and outbound requests").

### Testing Plan

**Unit Tests:**

Added `test_generate_content_async_stream_aggregates_reasoning_deltas` in `tests/unittests/models/test_litellm.py`, which streams three separate `reasoning_content` deltas (no `thought_signature`, matching how a non-Anthropic provider like xAI streams) followed by a text delta, and asserts the final non-partial response contains exactly one `thought=True` part with the joined text, instead of three.

```
404 passed, 1 warning in 5.27s
```

Also ran the full reasoning/thought-focused subset in isolation:

```
$ pytest tests/unittests/models/test_litellm.py -k "reasoning or thought" -q
44 passed, 360 deselected in 8.85s
```

Not run — this is a pure data-shape bug in response aggregation, fully exercised by the unit test above (constructs the same per-token `ModelResponseStream` deltas a live xAI/OpenAI-reasoning stream would produce and asserts on the resulting `LlmResponse.content.parts`). No live provider credentials were available to additionally verify against a real streaming API call.

### Checklist

### Additional context

Both call sites previously read `thought_parts=list(reasoning_parts) if reasoning_parts else None`; the fix replaces `list(reasoning_parts)` with `_aggregate_streaming_thought_parts(reasoning_parts)`. No behavior changes for the Anthropic path, which already routed through the aggregator via a separate code path (`_content_to_message_param`) for outbound requests — this PR fixes the *inbound*/session-persisted shape for all providers, including Anthropic's own aggregated response object.

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -3289,7 +3289,11 @@ class LiteLlm(BaseLlm):
                 tool_calls=tool_calls,
             ),
             model_version=model_version,
-            thought_parts=list(reasoning_parts) if reasoning_parts else None,
+            thought_parts=(
+                _aggregate_streaming_thought_parts(reasoning_parts)
+                if reasoning_parts
+                else None
+            ),
         )
         mapped_finish_reason = _map_finish_reason(finish_reason)
         llm_response.finish_reason = mapped_finish_reason
@@ -3313,7 +3317,11 @@ class LiteLlm(BaseLlm):
                 content=message_content,
             ),
             model_version=model_version,
-            thought_parts=list(reasoning_parts) if reasoning_parts else None,
+            thought_parts=(
+                _aggregate_streaming_thought_parts(reasoning_parts)
+                if reasoning_parts
+                else None
+            ),
         )
         mapped_finish_reason = _map_finish_reason(finish_reason)
         llm_response.finish_reason = mapped_finish_reason
--- a/tests/unittests/models/test_litellm.py
+++ b/tests/unittests/models/test_litellm.py
@@ -5010,6 +5010,78 @@ async def test_generate_content_async_stream_with_reasoning_tokens(
 
 
 @pytest.mark.asyncio
+async def test_generate_content_async_stream_aggregates_reasoning_deltas(
+    mock_completion, lite_llm_instance
+):
+  """Per-token reasoning deltas must collapse into one thought part.
+
+  Regression test: the streaming finalizers passed the raw per-delta thought
+  parts straight through, so a model that streams reasoning token by token
+  (e.g. xai/grok via LiteLLM) produced one `types.Part(thought=True)` per
+  delta in the final, non-partial response instead of one part per thinking
+  block -- the same shape the non-streaming path already produces.
+  """
+  streaming_reasoning_response = [
+      ModelResponseStream(
+          model="test_model",
+          choices=[
+              StreamingChoices(
+                  finish_reason=None,
+                  delta=Delta(role="assistant", reasoning_content="The"),
+              )
+          ],
+      ),
+      ModelResponseStream(
+          model="test_model",
+          choices=[
+              StreamingChoices(
+                  finish_reason=None,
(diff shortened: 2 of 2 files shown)


Ashfaqbs:
recheck

benclarkeio:
I submitted a PR for this already: https://github.com/google/adk-python/pull/6896

Ashfaqbs:
Closing — this duplicates #6896, which addresses the same root cause (issue #6895) and was already reviewed and confirmed fixing the bug by the team. No need to carry two PRs for the same fix.

PR #6903, opened 2026-08-26
fix(a2a): do not cache remote agent card when validation fails

### LINKED ISSUE #6901: RemoteA2aAgent caches an invalid Agent Card after RPC URL validation failure
## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. Requests without complete information may be rejected / deprioritized.
If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**
After upgrading `google-adk` from `1.27.4` to `2.7.1`, a `RemoteA2aAgent` fails during the first invocation when the remote Agent Card contains an HTTP RPC URL on a non-loopback internal hostname.

The first invocation fails with the expected HTTPS validation error. 
However, invoking the same`RemoteA2aAgent` instance again succeeds without changing the Agent Card or configuration.

This appears to indicate that the Agent Card is cached before validation completes, and the cached card is reused after validation failure.

**Steps to Reproduce:**
1. Install `google-adk==2.7.1`.
2. Configure a `RemoteA2aAgent` using a remote Agent Card URL.
3. Make the remote Agent Card return an RPC URL similar to:

```
     http://internal-gateway.example.invalid:8080/agents/remote
```

4. Invoke the same RemoteA2aAgent instance for the first time.
5. Observe the following error:

```
     Failed to initialize remote A2A agent:
     Agent card RPC URL must use https, or http on a loopback host:
     http://internal-gateway.example.invalid:8080/agents/remote
```

6. Invoke the same RemoteA2aAgent instance again without restarting the process. (I used same question)
7. Observe that the second invocation proceeds successfully.

**Expected Behavior:**
If Agent Card validation fails, I think the invalid Agent Card should not remain cached.

The next invocation should either:
  - Resolve and validate the Agent Card again and return the same validation error; or
  - Return a stable initialization error without creating an A2A client from an invalid card.

If the Agent Card is valid and reachable through a private service mesh, all invocations should succeed consistently.

**Observed Behavior:**
The first invocation fails with:

```
Agent card RPC URL must use https, or http on a loopback host:
http://internal-gateway.example.invalid:8080/agents/remote
```

The second invocation using the same RemoteA2aAgent instance succeeds.

The suspected flow is:

```
# google/adk/agents/remote_a2a_agent.py

  self._agent_card = await self._resolve_agent_card(ctx)
  await self._validate_agent_card(self._agent_card)
```

If validation fails, _agent_card has already been assigned. 
On the next invocation, this condition becomes false:

  ` if not self._agent_card: `

As a result, Agent Card resolution and validation are skipped, and an A2A client is created using the previously cached card.

The current upstream implementation appears to follow this same ordering in the shared resolution path:
[https://github.com/google/adk-python/blob/main/src/google/adk/agents/remote_a2a_agent.py](url)
  

**Environment Details:**

  - ADK Library Version (pip show google-adk): 2.7.1
  - Desktop OS: N/A - running in a Linux container
  - Python Version (python -V): Python 3.14.x

**Model Information:**

 - Are you using LiteLLM: Yes
 - Which model is being used: Gemini Flash-family model through an internal LiteLLM proxy (gemini 3 flash)

Ashfaqbs:

- Closes: #6901

**Problem:**
`RemoteA2aAgent._ensure_resolved()` (shared/cached resolution path) assigned `self._agent_card = await self._resolve_agent_card(ctx)` *before* calling `await self._validate_agent_card(self._agent_card)`. When validation raises (e.g. an RPC URL that fails the https/loopback check), the invalid card is already cached on `self._agent_card`. The next invocation's `if not self._agent_card:` guard then evaluates to `False`, so resolution and validation are both skipped entirely and the previously-rejected, unvalidated card is reused to build the A2A client — the exact bypass described in #6901.

**Solution:**
Hold the resolved card in a local variable, validate it, and only assign it to `self._agent_card` after validation succeeds. A failed validation now leaves `self._agent_card` as `None`, so the next call re-resolves and re-validates instead of reusing the rejected card.

### Testing Plan

**Unit Tests:**

Added `test_ensure_resolved_does_not_cache_card_on_validation_failure` in `tests/unittests/agents/test_remote_a2a_agent.py`, asserting `self._agent_card` stays `None` after a validation failure and that the next call re-resolves + re-validates (rather than reusing a cached invalid card) and succeeds once validation passes.

```
$ pytest tests/unittests/agents/test_remote_a2a_agent.py -q
230 passed, 294 warnings in 10.46s
```

Note: I tested against the commit just before `85b52f6a` (`chore(live): create top-level google.adk.live package...`), since that commit's diff appears to be missing the actual `src/google/adk/live/` package it references (only the tests under `tests/unittests/live/` landed), which currently breaks `import google.adk` on `main`. That's unrelated to this change — flagging separately rather than bundling here.

Not run — this is a pure caching-order bug, fully exercised by the unit test above (mocks `_validate_agent_card` to raise on the first call, succeed on the second, and asserts the card is neither cached nor reused across the failure).

### Checklist

### DIFF
--- a/src/google/adk/agents/remote_a2a_agent.py
+++ b/src/google/adk/agents/remote_a2a_agent.py
@@ -990,10 +990,13 @@ class RemoteA2aAgent(BaseAgent):
       if not self._agent_card:
 
         # Resolve agent card if needed
-        self._agent_card = await self._resolve_agent_card(ctx)
+        resolved_agent_card = await self._resolve_agent_card(ctx)
 
-        # Validate agent card
-        await self._validate_agent_card(self._agent_card)
+        # Validate agent card before caching it. If validation fails, the
+        # card must not be cached, or a subsequent invocation would skip
+        # resolution and validation entirely and reuse the invalid card.
+        await self._validate_agent_card(resolved_agent_card)
+        self._agent_card = resolved_agent_card
 
         # Update description if empty
         if not self.description and self._agent_card.description:
--- a/tests/unittests/agents/test_remote_a2a_agent.py
+++ b/tests/unittests/agents/test_remote_a2a_agent.py
@@ -698,6 +698,47 @@ class TestRemoteA2aAgentResolution:
     assert mock_resolve.await_count == 1
 
   @pytest.mark.asyncio
+  async def test_ensure_resolved_does_not_cache_card_on_validation_failure(
+      self,
+  ):
+    """A card that fails validation must not be cached.
+
+    Otherwise a later invocation sees `self._agent_card` already set, skips
+    resolution and validation entirely, and reuses the invalid card.
+    """
+    agent = RemoteA2aAgent(
+        name="test_agent",
+        agent_card="https://example.com/agent.json",
+    )
+
+    with patch.object(
+        agent, "_resolve_agent_card", new_callable=AsyncMock
+    ) as mock_resolve:
+      mock_resolve.return_value = self.agent_card
+      with patch.object(
+          agent, "_validate_agent_card", new_callable=AsyncMock
+      ) as mock_validate:
+        mock_validate.side_effect = ValueError("invalid rpc url")
+
+        with pytest.raises(AgentCardResolutionError):
+          await agent._ensure_resolved(Mock())
+
+        assert agent._agent_card is None
+
+        mock_validate.side_effect = None
+        with patch.object(agent, "_ensure_httpx_client") as mock_ensure:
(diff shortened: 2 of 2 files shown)


Ashfaqbs:
Closing this — I see `2685acd` landed the same fix internally around the same time, so this is superseded. Nothing to reconcile, just closing out the duplicate.

Thanks for maintaining this project — happy to keep contributing if there's anything in the backlog worth picking up.

Issue #3971, opened 2025-12-18
Persistent crash loop caused by missing tool_result in conversation history after interrupted execution


donggyun112:
**Describe the bug**
The application enters a persistent crash loop when resuming a session where the previous execution was interrupted immediately after a `tool_use` event but before the corresponding `tool_result` could be saved.

This interruption typically occurs in two scenarios:
1.  **Server-side interruption:** Server restarts, OOM crashes, or deployment cycles occurring exactly during tool execution.
2.  **Client-side interruption:** A user refreshes the browser or closes the connection while the agent is processing a tool call. The server receives a cancellation signal and aborts the task likely **without persisting the failure/cancellation state to the database**, leaving the history incomplete.

When the session resumes, ADK (via LiteLLM) attempts to send this "corrupted" history (a `tool_use` in the last assistant message without a following `tool_result`) to strict LLM providers like Anthropic or OpenAI. These providers reject the request with a **`BadRequestError` (400)**, rendering the session permanently unrecoverable ("bricked") without manual database intervention.

**To Reproduce**
I have created a standalone reproduction script using `litellm` to simulate the API rejection that occurs inside ADK.

**Prerequisites:**
*   Set `ANTHROPIC_API_KEY` or `OPENAI_API_KEY`.
*   Install `litellm`.

**Reproduction Script:**
```python
import asyncio
import os
from litellm import acompletion, BadRequestError

async def test_model(model_name: str, api_key_env: str):
    print(f"\n📡 Testing Model: {model_name} ...")
    if not os.getenv(api_key_env):
        print(f"⚠️  SKIPPING: {api_key_env} not found.")
        return

    # Broken history: assistant called a tool, but no tool_result follows.
    broken_messages = [
        {"role": "user", "content": "What is the weather in Seoul?"},
        {
            "role": "assistant",
            "content": "I will check the weather.",
            "tool_calls": [
                {
                    "id": "tool_u_reproduce_123", 
                    "type": "function",
                    "function": {
                        "name": "get_weather",
                        "arguments": "{\"location\": \"Seoul\"}"
                    }
                }
            ]
        },
        # --- MISSING TOOL RESULT DUE TO REFRESH/RESTART ---
        # The next message is a new user turn, violating the strict tool_use -> tool_result sequence.
        {"role": "user", "content": "Wait, tell me about Tokyo instead."}
    ]

    tools = [{
        "type": "function",
        "function": {
            "name": "get_weather",
            "description": "Get the current weather",
[...]

Issue #5856, opened 2026-05-26
Orphaned function_call blocks from user interruption cause 400 errors


seanxiong1989:
## Description

When a user sends a new message while a tool call is still pending (or the tool errors without ADK emitting a `function_response`), the persisted session contains a `tool_use` event with no matching `tool_result`. Claude's API requires every `tool_use` to have a corresponding `tool_result` — without one, every subsequent turn returns a 400 error and the session is permanently broken.

## Difference from existing fixes

This is distinct from the fixes in #3504 and #4027, which address orphaned `function_response` blocks during context truncation. This issue is about orphaned `function_call` blocks caused by user interruption — a different problem that ADK does not currently handle.

## Current workaround

We use a `before_model_callback` that scans `llm_request.contents` and injects synthetic `function_response` stubs for any unmatched `function_call` blocks, preserving the rest of the conversation context.

## Expected behavior

ADK should handle orphaned `function_call` blocks (from interruption or tool errors) gracefully, either by injecting synthetic responses or by stripping the orphaned calls before sending to the model API.

llalitkumarrr:
Hello @seanxiong1989 ,

Could you please share a minimal reproducible code and the steps for reproduction? Also are you facing this issue with Claude only or is a similar problem observed with other models? Furthermore, if you are not already using the latest version, kindly try our most recent release and share your observations.

seanxiong1989:
Hey @llalitkumarrr — reproduces on both 1.29.0 and 2.1.0. Claude-specific (Gemini doesn't enforce tool_use/tool_result pairing).

Repro (Claude on Vertex AI):

```python
"""Minimal reproduction: orphaned function_call permanently breaks session."""

from __future__ import annotations
import asyncio

from google.adk.agents import Agent
from google.adk.events import Event
from google.adk.models.anthropic_llm import Claude
from google.adk.models.registry import LLMRegistry
from google.adk.runners import Runner
from google.adk.sessions import InMemorySessionService
from google.genai import types

LLMRegistry.register(Claude)


def get_weather(city: str) -> dict:
    """Look up weather for a city."""
    return {"city": city, "temp_f": 72, "condition": "sunny"}


agent = Agent(
    name="repro_agent",
    model="claude-sonnet-4-6",
    instruction="You are a helpful assistant. Use the get_weather tool when asked about weather.",
    tools=[get_weather],
)

session_service = InMemorySessionService()
runner = Runner(agent=agent, session_service=session_service, app_name="repro")


[...]

PR #6914, opened 2026-08-26
fix(flows): pair every function call with a result in request contents

QuentinBisson:

- Same failure as #5856 (stale-closed), #3971 and #6582.

## Problem

A function call and its result are two separate session events. When a turn ends between them (a restart, an OOM kill, a client disconnect, a cancellation), the session keeps a `function_call` that no `function_response` answers.

That history is replayed on every later turn, so the session does not recover on its own. A provider that requires strict pairing rejects the whole conversation: Anthropic answers `tool_use ids were found without tool_result blocks immediately after`, and the session stays unusable until it is deleted.

`_drop_orphaned_function_responses` already covers the mirror case, a response with no call. The unanswered call has no owner.

## Change

Pair every function call with a result while the request contents are assembled: `_get_contents` calls `pair_unanswered_function_calls` in `_tool_call_rearranger.py`, alongside the existing `drop_orphaned_function_responses`. A call the immediately following content does not answer gets a placeholder result there.

- The stored events are untouched, so recorded history stays intact and a session that is already broken heals on its next turn with no migration.
- It sits above the session service, so one implementation covers every store and every provider.
- Ids are read after the conversion, so a call whose `adk-` id was stripped is answered by a response with the same stripped id.
- Pairing is checked positionally, against the immediately following content, because that is the invariant the provider enforces.
- On a turn that is already partly answered, the placeholder joins the leading run of responses. Anthropic requires the results to precede any other block in the message, so appending after a trailing text part would be rejected for the same reason the missing result was.

A call the framework holds open (a long-running tool, an approval, a request for user input) is described as awaiting a response, not as having returned nothing. The difference changes what the model does next: told a tool returned nothing, it reissues the call or proceeds without it; told the call is still awaiting a response, it can wait. Those ids come from `long_running_tool_ids`, which lives on the event and does not survive the conversion to contents. Only a call with no recorded response is logged, so a call that is merely awaiting an answer stays quiet across the turns it spans.

A conversation whose calls are all answered is returned unchanged.

The mirror change for adk-go is google/adk-go#1405.

## Testing plan

`uv run pytest tests/unittests` passes (13981 passed, 83 skipped) on the rebased branch.

Seven new cases in `tests/unittests/flows/llm_flows/test_contents_function.py`: the interrupted turn at the end of the history and in the middle, the partially answered parallel turn, the same turn with a trailing text part, the call held open in both directions, and calls without an id. Two more convert the repaired contents through `anthropic_llm.content_to_message_param` and assert both halves of the provider invariant, that every `tool_use` block is answered by the next message and that the results precede any other block in it; a guard test asserts the same check rejects unrepaired input.

Two assertions in `tests/unittests/apps/test_compaction.py` are updated, because a pending call now reaches the model with a placeholder.

### DIFF
--- a/src/google/adk/flows/llm_flows/_tool_call_rearranger.py
+++ b/src/google/adk/flows/llm_flows/_tool_call_rearranger.py
@@ -408,7 +408,7 @@ def pair_unanswered_function_calls(
 
   Args:
     contents: The contents built for the request. Not mutated in place, except
-      that a placeholder is appended to the ``parts`` list of a content that
+      that a placeholder is inserted into the ``parts`` list of a content that
       already answers part of a turn.
     pending_ids: The ids of the calls ADK is holding open.
 
@@ -416,7 +416,7 @@ def pair_unanswered_function_calls(
     The contents, with a placeholder response for every unanswered call.
   """
   paired: list[types.Content] = []
-  synthesized = 0
+  lost = 0
   for index, content in enumerate(contents):
     paired.append(content)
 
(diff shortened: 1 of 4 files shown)


QuentinBisson:
Rebased onto main after fe32193f moved the rearrangement helpers into `_tool_call_rearranger.py`; the pairing now lives there next to `drop_orphaned_function_responses`, behaviour unchanged. @DeanChensj since you extracted that module, would you be the right person to review, or could you point me to who is?

DeanChensj:
Hi @QuentinBisson,

Thank you for putting this together and exploring the problem across #3971, #5856, and #6582. Addressing unrecoverable sessions caused by interrupted tool calls is a known pain point, and we appreciate the effort you put into investigating this.

After an end-to-end evaluation of the approach and the underlying issues, we cannot merge this PR in its current form. There are architectural and behavioral limitations with injecting synthetic function responses into core request contents:

### 1. Role Alternation Violation on Anthropic (HTTP 400)
The primary goal of pairing is avoiding provider rejections when tool calls lack results. However, in the standard interruption workflow where a user sends a follow-up message after an interrupted call (e.g. `[ModelContent(with call), UserContent("stop, do something else")]`), `_pair_unanswered_function_calls` triggers an immediate API rejection:
- `_pair_unanswered_function_calls` synthesizes a missing result and appends it inside a new `types.Content(role='user', parts=...)`.
- On the next step, the user's actual message is also added as `role='user'`.
- This creates two adjacent `user` messages (`[assistant(tool_use), user(synthetic_result), user(user_message)]`).
- Anthropic strictly requires alternating roles (`user` <-> `assistant`) and rejects this sequence with HTTP 400 (`messages: roles must alternate between "user" and "assistant"`).

### 2. Semantic History Poisoning & Schema Violations
Injecting artificial string payloads (`_MISSING_FUNCTION_RESULT`) directly into session contents introduces runtime side effects:
- **Schema Conflicts**: Many production tools specify structured return schemas (objects, lists). Returning a plain error string violates these schemas and causes downstream validation or model hallucinations.
- **Pending / Long-Running Tool Ambiguity**: Fabricating a completed `FunctionResponse` for an interrupted or pending operation informs the model that the call has concluded, conflicting with actual resumption or human-in-the-loop workflows.

### 3. Layering Violation in Core `contents.py`
`contents.py` handles universal request assembly and must remain model-agnostic. Dynamically importing downstream model adapter classes (`AnthropicLlm`, `LiteLlm`, `OpenAIResponsesLlm`) in `_id_pairing_model_types()` breaks internal layering boundaries.

---

### Alternative Direction: Mirroring #6582 via Call Stripping

As noted in #5856, there are two potential directions for unclosed calls: synthetic response injection or stripping orphaned calls before dispatching the request to the model.

In commit `73e8625` (fixing #6582 for orphaned responses), ADK adopted a **pruning approach** via `_drop_orphaned_function_responses` in `_tool_call_rearranger.py`. 

The natural mirror for orphaned `function_call` blocks is **stripping orphaned calls** (`drop_orphaned_function_calls`):
- When a turn is interrupted before tool execution completes, removing the unanswered `function_call` parts before sending the history to the LLM presents the model with the genuine conversational exchange.
- Stripping orphaned calls eliminates both the Anthropic missing tool result error and the consecutive `user` role error without fabricating synthetic state or violating tool schemas.
- It operates purely on event history without requiring knowledge of specific downstream LLM classes.

Because this requires implementing `drop_orphaned_function_calls` inside `_tool_call_rearranger.py` alongside `drop_orphaned_function_responses`, the core maintainers will take on this implementation to resolve #3971 and #5856.

We will close this pull request accordingly. Thank you again for highlighting the interaction between interrupted turns and provider validation!

PR #6929, opened 2026-08-28
fix(cli): Discover newly staged agents during validation

yousefalbanna1:
**Problem:**

`_validate_agent_import()` can fail to import a newly created agent package when Python has already cached the contents of the package's parent directory.

In that case, the package and `agent.py` exist, but `importlib.import_module()` can still raise `ModuleNotFoundError`.

**Solution:**

Invalidate Python's import caches immediately before importing the agent package.

A regression test was added that primes the import cache before creating the package and verifies that validation can still discover the new agent.

### Testing Plan

**Unit Tests:**

*  I have added unit tests for my change.

*  All unit tests pass locally.

* `python -m pytest tests/unittests/cli/utils/test_cli_deploy.py::TestValidateAgentImport -q` — 12 passed.

* `python -m pytest tests/unittests/cli/utils/test_cli_deploy.py -q` — 98 passed.

* `tox`:

  * Python 3.10 — passed.
  * Python 3.11 — passed.
  * Python 3.12 — 2 failures in `tests/unittests/test_import_loading.py`.
  * Python 3.13 — passed.
  * Python 3.14 — passed.

* The same two Python 3.12 failures reproduce on the unmodified base commit `6eb1d35d`.

* Pre-commit checks for the changed files — passed.

Built the package with `uv build`, installed the generated wheel in a clean virtual environment, and verified that ADK imports successfully and the updated validation code is present.

A cloud deployment was not performed.

Screenshot 2026-08-28 043012
Screenshot 2026-08-28 043746
Screenshot 2026-08-28 042814

### Additional context

The two Python 3.12 tox failures are unrelated to this change and reproduce on the unmodified base commit.

### DIFF
--- a/src/google/adk/cli/cli_deploy.py
+++ b/src/google/adk/cli/cli_deploy.py
@@ -642,6 +642,8 @@ def _validate_agent_import(
     # Add parent directory to path so imports work correctly
     if parent_dir not in sys.path:
       sys.path.insert(0, parent_dir)
+
+    importlib.invalidate_caches()
     try:
       module = importlib.import_module(f'{module_name}.agent')
     except ImportError as e:
--- a/tests/unittests/cli/utils/test_cli_deploy.py
+++ b/tests/unittests/cli/utils/test_cli_deploy.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import importlib
 import json
+import os
 from pathlib import Path
 import shutil
 import subprocess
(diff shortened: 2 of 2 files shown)


GWeale:
fixed on main in c7ffcfa8, not in a release yet. your regression test isn't in the tree and passes against main as-is, so send it on its own. open a new issue if you see anything like this again.

PR #6930, opened 2026-08-28
fix: validate GCP project and region before interpolating into deploy Dockerfile

prasanna8585:
`_DOCKERFILE_TEMPLATE` interpolates `--project` and `--region` verbatim into `ENV` instructions:
 
```
ENV GOOGLE_CLOUD_PROJECT={gcp_project_id}
ENV GOOGLE_CLOUD_LOCATION={gcp_region}
```
 
Neither value is validated before reaching the template, across all three deploy targets (`to_cloud_run`, `to_agent_engine`, `to_gke`). A value containing a newline breaks out of the `ENV` instruction's line, injecting an arbitrary new Dockerfile instruction — confirmed directly against the real template: a `--project` value with an embedded newline followed by a `RUN` instruction produced a generated Dockerfile where that `RUN` appeared as its own, independent line. `docker build` processing that file would execute the injected command.
 
This adds `_validate_gcp_project_id`/`_validate_gcp_region`, restricting both to a safe identifier character set (letters, digits, hyphens, underscores) before they reach the template — mirroring the existing `app_name` validation this file already applies for the same reason.
 
**Testing:** Added 8 new tests covering acceptance of realistic project/region values, rejection of the injection payload and other unsafe characters (quotes, spaces, semicolons, pipes, command substitution), and end-to-end rejection through `to_cloud_run` and `to_gke` with a malicious project value. Full existing test suite: 117/117 pass, no regressions.

### DIFF
--- a/src/google/adk/cli/cli_deploy.py
+++ b/src/google/adk/cli/cli_deploy.py
@@ -555,6 +555,61 @@ def _validate_app_name(app_name: str) -> None:
     )
 
 
+# project and region are interpolated verbatim into the generated Dockerfile
+# (ENV GOOGLE_CLOUD_PROJECT={gcp_project_id} / ENV GOOGLE_CLOUD_LOCATION=
+# {gcp_region}) by _DOCKERFILE_TEMPLATE, the same as app_name above. Both are
+# plain CLI flag values (--project, --region) with no requirement that they
+# name a real, already-validated GCP resource before reaching this point --
+# _resolve_project's own fallback path aside, a caller (or a script wrapping
+# this CLI) can supply any string here. A value containing a newline breaks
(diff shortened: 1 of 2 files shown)


prasanna8585:
Hi @wyf7107,

Have you got the chance to review this PR,
Is there anything needed from my side on this PR to get it merged?

GWeale:
fixed on main in 8192d1bf. _validate_dockerfile_env_value rejects newlines in project and region before they reach the dockerfile. narrower than your regex on purpose, since letters-digits-hyphens would reject a domain-scoped id like example.com:my-project, and the newline is what escapes the env line. not in 2.9.1. the gke manifest still interpolates both unchecked, so a pull request for that would be welcome.

PR #6944, opened 2026-08-29
feat: add AgentCoreSessionService for AgentCore Memory

### LINKED ISSUE #6920: AgentCoreSessionService extending BaseSessionService for AgentCore Memory
I'm trying to use AWS AgentCore Memory with an ADK Agent.

The session_serivce is a required param in the adk Agent so it would be nice if there was an implementation for AgentCore Memory.

As for the implementation, two things caught my eyes so far:
1. Session 
There isn't an exact session translation to agentcore memory so need to think about that.
2. Events and Short-term memory 
STM is where events are saved (user message, llm response) in AgentCore memory. In adk they are more fine-grained, but in agentcore they represent a turn - user message is one and what the agent responded is another - maybe we can construct a whole event that consists of all the little ADK events and then flush those to agentcore's short-term memory as, what aws calls, an event.

### Impact on your work
No timeline but would make it easier. This will allow for better integration with AWS AgentCore memory

### Willingness to contribute
Are you interested in implementing this feature yourself or submitting a PR?
Yes, I am happy to work on an implementation myself

a2105z:
## Summary
- Adds `AgentCoreSessionService`, a `BaseSessionService` backed by Amazon Bedrock AgentCore Memory short-term memory, so `Runner(session_service=...)` can persist ADK sessions there.
- The event mapping is the one from [@divakaivan](https://github.com/divakaivan)'s PoC ([agentcore_session_service](https://github.com/divakaivan/agentcore_session_service)): each ADK event is one AgentCore event with a `conversational` payload (when there is text) plus a `blob` of the full ADK `Event` JSON, so STM extraction still works and tool calls / state / metadata round-trip.
- Fills in the rest of the `BaseSessionService` contract on top of that mapping: generated session ids, `AlreadyExistsError`, `get_session` returning `None` when missing, `list_sessions` (including all users of an app), `delete_session` via `delete_event`, skip `partial` events, bootstrap event so empty sessions exist in AgentCore, and `actorId = {app_name}:{user_id}` so two ADK apps sharing a Memory resource do not collide.

Fixes #6920

## Test plan
- [x] `uv run pytest tests/unittests/integrations/agentcore/test_agentcore_session_service.py` — **12 passed** (fake `bedrock-agentcore` client)
  - create / get / already-exists / not-found
  - dual payload (USER + ASSISTANT conversational + blob)
  - TOOL role, partial events not written
  - `GetSessionConfig` filters, state_delta replay (`temp:` dropped)
  - list by user and by app, delete, app isolation
- [ ] Live AgentCore Memory smoke test with AWS credentials (same shape as the PoC's `agent.py`)
- [ ] CLA check


a2105z:
Hi @wukath, @sanketpatil06, @GWeale — could I please ask for a review when you have a chance?

This is the `BaseSessionService` implementation for AgentCore Memory from #6920. The event mapping is @divakaivan’s (conversational text plus a full ADK event blob). I fitted that to the session-service contract and covered it with unit tests against a fake client.

Happy to change anything that does not match how you want integrations to land. Thank you very much.

src/google/adk/integrations/agentcore/_agentcore_session_service.py
```diff
@@ -0,0 +1,433 @@
...
+          actorId=actor_id,
+          sessionId=session_id,
+          eventId=event_id,
+      )
+
+  async def append_event(self, session: Session, event: Event) -> Event:
+    """Appends an event locally, then flushes it to AgentCore STM."""
+    if event.partial:
+      return event
+    event = await super().append_event(session, event)
```
divakaivan, line 300: hi. why do we call the super's append_event? can this work without it?

PR #6953, opened 2026-08-30
fix: return float timeouts on MCP SDK 2.x in session_context

### LINKED ISSUE #6938: _read_timeout returns timedelta on mcp 2.x, breaking every MCP session (docstring describes the missing branch)
### Description

`google/adk/tools/mcp_tool/session_context.py::_read_timeout` documents a version branch that its body never implements.

```python
def _read_timeout(seconds: Optional[float]) -> Optional[timedelta]:
  """Converts a timeout in seconds to the type ``ClientSession`` expects.

  ADK carries every timeout as float seconds. MCP SDK 1.x wants a
  ``timedelta`` here, while 2.x wants the float. Converting in one place keeps
  that difference to a single function.
  """
  if seconds is None:
    return None
  return timedelta(seconds=seconds)
```

The docstring states 2.x wants the float, but the function returns a `timedelta` unconditionally. On mcp 2.x that value reaches `anyio.fail_after`, which adds it to `current_time()`:

```
File "mcp/shared/jsonrpc_dispatcher.py", line 401, in send_raw_request
    with anyio.fail_after(opts.get("timeout")):
File "anyio/_core/_tasks.py", line 120, in fail_after
    deadline = (current_time() + delay) if delay is not None else math.inf
TypeError: unsupported operand type(s) for +: 'float' and 'datetime.timedelta'
```

Every session creation fails at `initialize()`.

### Reproduction

ADK 2.8.0, mcp 2.1.1. Patching just this one function makes the `timedelta` error disappear:

```python
import google.adk.tools.mcp_tool.session_context as sc
sc._read_timeout = lambda seconds: seconds
```

### Suggested fix

The docstring already specifies the behaviour:

```python
from importlib.metadata import version

_MCP_MAJOR = int(version("mcp").split(".")[0])

def _read_timeout(seconds: Optional[float]) -> Optional[timedelta | float]:
  if seconds is None:
    return None
  return seconds if _MCP_MAJOR >= 2 else timedelta(seconds=seconds)
```

### Context: what else blocks mcp 2.x

I hit this while getting `McpToolset` working against mcp 2.1.1, and this was not the only blocker — recording the rest in case it is useful for the wider 2.x migration:

1. **Renames**, all resolvable by aliasing:
   - `mcp.shared.session.ProgressFnT` → `mcp.client.session.ProgressFnT` (used only as a type annotation)
   - `mcp.shared.exceptions.McpError` → `MCPError`, same module
   - `mcp.server.fastmcp.FastMCP` → `mcp.server.mcpserver.MCPServer` — needed only by `_agent_to_mcp`, but `mcp_tool/__init__.py` imports it eagerly, so the whole toolset surface disappears without it

2. **This bug.**

3. **Two HTTP stacks.** ADK types its client factory as `httpx.*`; mcp 2.x runs on `httpx2`, a separate distribution (`httpx.Timeout is not httpx2.Timeout`). ADK's Timeout object reaches httpcore2 and fails the same way. This one is *already solvable from outside* via `httpx_client_factory` on the connection params — a public, documented seam. It worked well: with the aliases, the `_read_timeout` patch, and a factory returning an `httpx2.AsyncClient`, `McpToolset` connects to a real streamable-HTTP server and returns fully-typed tool declarations.

So on 2.8.0 the only change that needs to happen *inside* ADK is this one function; the rest can be handled by a caller. Happy to open a PR if that would help.

### Note on the silent failure

`mcp_tool/__init__.py` catches the `ImportError` and logs at `DEBUG`, so on mcp 2.x `McpToolset`, `McpTool` and `require_confirmation` are simply absent from the package rather than raising. `from google.adk.tools.mcp_tool import McpToolset` then fails with a bare "cannot import name", which does not point at the dependency conflict. A warning naming the mcp version would have saved a fair amount of digging.

Lubaoshuai:
## Issue

Fixes #6938.

## Summary

`_read_timeout()` in `google/adk/tools/mcp_tool/session_context.py` documents that MCP SDK 1.x wants a `timedelta` while 2.x wants the plain float, but the body always returned `timedelta`. On mcp 2.x that value reaches `anyio.fail_after()`, which adds it to `current_time()` and raises `TypeError: unsupported operand type(s) for +: 'float' and 'datetime.timedelta'` on every session creation.

This change detects the installed SDK major version once (`importlib.metadata`, cached; detection failures keep the current 1.x behavior, which matches the `mcp>=1.24,<2` pin) and passes float seconds through on 2.x.

Note: #6940 addressed the same bug but was self-closed; the main gap was test coverage, which this PR adds.

## Testing

- `pytest tests/unittests/tools/mcp_tool/test_session_context.py` — 40 passed (mcp 1.29.1). The `_read_timeout` tests now exercise both branches deterministically via patching, plus a detection-failure fallback test, so they hold regardless of the SDK version installed in the environment.
- `pytest tests/unittests/tools/mcp_tool/` — 349 passed.
- `pyink --check` on both touched files — clean.

### DIFF
--- a/src/google/adk/tools/mcp_tool/session_context.py
+++ b/src/google/adk/tools/mcp_tool/session_context.py
@@ -18,6 +18,9 @@ import asyncio
 from contextlib import AbstractAsyncContextManager
 from contextlib import AsyncExitStack
 from datetime import timedelta
+from functools import lru_cache
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version
 import logging
 from types import TracebackType
 from typing import Any
@@ -37,7 +40,21 @@ logger = logging.getLogger('google_adk.' + __name__)
 _T = TypeVar('_T')
 
 
-def _read_timeout(seconds: Optional[float]) -> Optional[timedelta]:
+@lru_cache(maxsize=1)
+def _mcp_wants_float_timeouts() -> bool:
+  """Whether the installed MCP SDK expects plain seconds as timeouts.
+
+  MCP SDK 1.x requires a ``datetime.timedelta`` for ``read_timeout_seconds``,
+  while 2.x takes the number of seconds as a float. Detection failures fall
+  back to the 1.x behavior, which is what ADK pins today.
+  """
+  try:
+    return int(version('mcp').split('.')[0]) >= 2
+  except (PackageNotFoundError, ValueError, IndexError):
+    return False
+
+
+def _read_timeout(seconds: Optional[float]) -> Optional[float | timedelta]:
   """Converts a timeout in seconds to the type ``ClientSession`` expects.
 
   ADK carries every timeout as float seconds. MCP SDK 1.x wants a
@@ -52,6 +69,8 @@ def _read_timeout(seconds: Optional[float]) -> Optional[timedelta]:
   """
   if seconds is None:
     return None
+  if _mcp_wants_float_timeouts():
+    return float(seconds)
   return timedelta(seconds=seconds)
 
 
--- a/tests/unittests/tools/mcp_tool/test_session_context.py
+++ b/tests/unittests/tools/mcp_tool/test_session_context.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import asyncio
 from contextlib import AsyncExitStack
 from datetime import timedelta
+from importlib.metadata import PackageNotFoundError
 import time
 from unittest.mock import AsyncMock
 from unittest.mock import Mock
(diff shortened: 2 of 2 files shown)


Lubaoshuai:
Upstream has since landed the same fix in 856acf2 ("feat(mcp): support MCP SDK 2.x alongside 1.x"), which resolves #6938 the same way: `_read_timeout()` now returns plain float seconds when `IS_MCP_SDK_V2` is set and a `timedelta` on 1.x, with regression tests covering both branches.That implementation is already in main and supersedes this PR, so I am closing it rather than rebasing an equivalent change. Thanks!

PR #6955, opened 2026-08-30
fix(lite_llm): honor capability for schema and tools

### LINKED ISSUE #6954: [BUG]: LiteLlm.capabilities hardcodes output_schema_and_tools=True, bypassing the SetModelResponseTool workaround
## Description

`LiteLlm.capabilities` returns `output_schema_and_tools=True` unconditionally, regardless of which model it wraps. ADK trusts that and applies `response_schema` alongside the tool declarations, so `SetModelResponseTool` never engages. On models that cannot honour both constraints the agent loops, calling tools repeatedly instead of producing structured output.

This is the same failure as #3413 and #5054, but on a path none of the existing work covers:

- **#5057** fixed the equivalent gate on the *native* Gemini path. `LiteLlm` bypasses that gate entirely, so merging it would not have helped anyone using ADK through LiteLLM. It was closed as stale because the code moved into `models/_capabilities`, which is the path this report is against.
- **#5091** hardens `SetModelResponseTool` (stronger instructions, forced `tool_config` at round N-1, a 25-round cutoff). All of that applies only once the fallback is active. Here it never activates, because `LiteLlm` reports the capability as supported.

Filing per the request on #5057: *"the repro @wuliang229 asked for is what a version gate needs first... open a new issue if you see this happen again."*

## Environment

- `google-adk` 2.8.0
- `litellm` 1.98.0
- Python 3.13.14
- Model: `openrouter/google/gemini-3.1-flash-lite` (Gemini 3.1 Flash Lite via OpenRouter)

## Reproduction

Needs an OpenRouter key in `OPENROUTER_API_KEY`. Any provider LiteLLM can reach should do; the model family matters more than the provider (see the note on model dependence below).

```python
import asyncio
from google.adk import Agent

MODEL = "openrouter/google/gemini-3.1-flash-lite"

class CityTime(BaseModel):
    city: str = Field(description="City name")
    time: str = Field(description="Local time")

def get_current_time(city: str) -> dict:
    """Returns the current time in a specified city."""
    return {"city": city, "time": "10:30 AM"}

agent = Agent(
    model=LiteLlm(MODEL),
    name="repro",
    instruction="Use get_current_time to look up the time, then answer.",
    tools=[get_current_time],
    output_schema=CityTime,
)

async def main():
    runner = InMemoryRunner(agent=agent, app_name="repro")
    await runner.session_service.create_session(
        app_name="repro", user_id="u", session_id="s")
    tool_calls = 0
    async for ev in runner.run_async(
        user_id="u", session_id="s",
        new_message=types.Content(role="user",
                                  parts=[types.Part(text="What time is it in Sydney?")]),
    ):
        if ev.content and ev.content.parts:
            for p in ev.content.parts:
                if p.function_call:
                    tool_calls += 1
                    print(f"  tool call {tool_calls}: {p.function_call.name}")
    print(f"total tool calls: {tool_calls}")

asyncio.run(main())
```

Observed: `get_current_time` is called repeatedly with identical arguments before the agent settles. Four runs of this exact code gave **4, 12, 18 and 19** tool calls. It does eventually produce valid output, but the count is neither stable nor bounded, and on some runs it did not terminate within several minutes.

Adding `output_schema_and_tools=False` (via the subclass below) makes it **2 function calls, every run**: `get_current_time` plus `set_model_response`.

### Supplementary: confirming the cause without API calls

This inspects the outgoing request and short-circuits before anything is sent, so it needs no key and no credit. Useful for verifying a fix quickly.

```python
import asyncio
from google.adk import Agent
from google.adk.models.base_llm import LlmCapabilities

MODEL = "openrouter/google/gemini-3.1-flash-lite"

class CityTime(BaseModel):
    city: str
    time: str

def get_current_time(city: str) -> dict:
    """Returns the current time in a specified city."""
    return {"city": city, "time": "10:30 AM"}

class SchemaSafeLiteLlm(LiteLlm):
    @property
    def capabilities(self) -> LlmCapabilities:
        return LlmCapabilities(output_schema_and_tools=False)

def spy(callback_context, llm_request):
    cfg = llm_request.config
    schema_set = bool(getattr(cfg, "response_schema", None)
                      or getattr(cfg, "response_json_schema", None))
    names = [fd.name
             for t in (getattr(cfg, "tools", None) or [])
             for fd in (getattr(t, "function_declarations", None) or [])]
    print(f"  response_schema_set={schema_set}  tools_offered={names}")
    return LlmResponse(content=types.Content(role="model", parts=[types.Part(text="{}")]))

async def inspect(label, model):
    print(f"\n{label}  output_schema_and_tools={model.capabilities.output_schema_and_tools}")
    agent = Agent(model=model, name="probe", instruction="x",
                  tools=[get_current_time], output_schema=CityTime,
                  before_model_callback=spy)
    runner = InMemoryRunner(agent=agent, app_name=label)
    await runner.session_service.create_session(
        app_name=label, user_id="u", session_id="s")
        user_id="u", session_id="s",
        new_message=types.Content(role="user", parts=[types.Part(text="time in Sydney?")]),
    ):
        pass

async def main():
    await inspect("stock LiteLlm", LiteLlm(MODEL))
    await inspect("SchemaSafeLiteLlm", SchemaSafeLiteLlm(MODEL))

asyncio.run(main())
```

Output:

```
stock LiteLlm  output_schema_and_tools=True
  response_schema_set=True   tools_offered=['get_current_time']

SchemaSafeLiteLlm  output_schema_and_tools=False
  response_schema_set=False  tools_offered=['set_model_response', 'get_current_time']
```

## Cost

Each retry resends the whole conversation, so the waste is worse than the call count suggests. Measured for the single question above, via OpenRouter:

| | generations | input tokens | cost |
|---|---|---|---|
| schema only, no tools | 1 | 19 | $0.000071 |
| schema + tools, stock `LiteLlm` | 13 | 5,810 | $0.001848 |

Input tokens grow ~300x against ~26x on cost, so this degrades sharply with a realistic tool payload rather than the toy one used here. The failure is also silent: valid output is eventually returned, so nothing surfaces except the bill.

## Expected behaviour

`LiteLlm` should not assert the capability on behalf of every model it wraps. When the wrapped model cannot pair `response_schema` with tools, ADK's existing `SetModelResponseTool` workaround should engage, as it does on the native path.

With the capability declared `False`, behaviour is correct and stable: **2 function calls per turn** (`get_current_time` plus `set_model_response`), and a second conversation turn still calls the tool rather than answering from history.

## Root cause

`src/google/adk/models/lite_llm.py:3075`

```python
@property
@override
def capabilities(self) -> LlmCapabilities:
    # LiteLLM reconciles tools + response_format per provider: providers with
    # native support get both passed through, and the rest are converted to a
    # json tool call with tool_choice enforcement.
    return LlmCapabilities(output_schema_and_tools=True)
```

`src/google/adk/flows/llm_flows/basic.py:137`

```python
if not agent.tools or model.capabilities.output_schema_and_tools:
    llm_request.set_output_schema(agent.output_schema)
```

The comment states the reasoning: LiteLLM is expected to reconcile the two per provider. For this provider and model it does not, and both constraints reach the model raw, as the reproduction shows.

By contrast the native path is conservative. `gemini_output_schema_and_tools` requires Vertex AI *and* a Gemini model id, so on other backends it correctly returns `False` and the workaround engages.

## Note on model dependence

The model matters too, consistent with the table in #5054: flash models prefer calling tools over `set_model_response`, while `gemini-3.1-pro-preview` succeeds on the first call. The `gemini-3.1-flash-lite` used here is in the susceptible family. That is an argument for `LiteLlm` being conservative by default rather than asserting support for every wrapped model, since it cannot
know which family it is proxying.

## Possible fixes

1. Default `LiteLlm.capabilities` to `output_schema_and_tools=False`, so the reliable `SetModelResponseTool` path is used unless a user opts out. Safe, at the cost of one extra function call for providers that could have handled it natively.
2. Make it constructor-configurable, e.g. `LiteLlm(model, output_schema_and_tools=False)`, so users can declare it without subclassing.

Deriving it from LiteLLM's own metadata does **not** work. LiteLLM reports both capabilities independently, and has no API for the combination:

```python
litellm.supports_response_schema(model="openrouter/google/gemini-3.1-flash-lite")   # True
litellm.supports_function_calling(model="openrouter/google/gemini-3.1-flash-lite")  # True
```

Both are True while the combination still loops, so the assumption in the current code comment cannot be rescued by consulting LiteLLM.

Happy to open a PR — combining 1 and 2 (a constructor argument defaulting to `False`) is a small diff, needs no network to test, and makes `LiteLlm` consistent with the conservative native gate. Glad to keep the default at `True` instead if preserving current behaviour matters more.

## Related

- #3413 — output_schema + tools infinite loop
- #5054 — same loop with `output_schema=str`; contains the per-model comparison table
- #5057 — equivalent gate on the native Gemini path; closed as stale after the check moved into `models/_capabilities`. Does not cover the `LiteLlm` path.
- #5091 — hardens `SetModelResponseTool`; helps only once the fallback is active, which it never is here.

JimmyWang0417:
## What's going on

`LiteLlm.capabilities` reported `output_schema_and_tools=True` for every
provider and model it wrapped. For providers that cannot combine an output
schema with tools, ADK therefore skipped the `SetModelResponseTool` fallback
and could repeatedly call tools instead of returning structured output.

## The fix

- Report the capability conservatively (`False`) by default.
- Add a keyword-only `output_schema_and_tools` constructor argument so callers
  can opt in for a known-compatible provider/model combination.
- Keep the capability-only argument out of LiteLLM completion kwargs.
- Cover the default and opt-in behavior for `LiteLlm` and `Gemma3Ollama`.

Closes #6954

### Unit Tests

- [x] Added and updated unit tests for the capability default, opt-in, and
  completion-argument filtering.
- [x] `pytest -q tests/unittests/models/test_capabilities.py tests/unittests/models/test_litellm.py tests/unittests/models/test_gemma_llm.py tests/unittests/utils/test_output_schema_utils.py tests/unittests/flows/llm_flows/test_basic_processor.py tests/unittests/flows/llm_flows/test_output_schema_processor.py` (514 passed)
- [x] `ruff check` and `pyink --check` on all changed Python files.
- [x] `python -m compileall` and `git diff --check`.

### Manual End-to-End (E2E) Tests

- [x] Re-ran the issue reproducer against the current PR head and its direct
  base using the same two-turn script, one deterministic tool, one Pydantic
  output schema, and one session.

Environment and method:

- PR base: `86b2e5baaa487558b10fc9f2b48fe0d4ceefb755`
- PR head: `ced2d36f601224ef867daf0da4505caf082b3f3d`
- Python `3.13.9`; source checkout reports `google-adk 2.8.0`
- `litellm 1.98.0`, `google-genai 2.22.0`, `pydantic 2.13.5`
- Credentialed OpenRouter-compatible gateway, using LiteLLM's
  `openrouter/<model-id>` route for the Gemini models; credentials and endpoint
  are intentionally omitted.
- Prompt and tool/schema were kept identical between base and head. Each turn
  was bounded with `RunConfig(max_llm_calls=8)` to prevent an unbounded baseline
  loop from consuming credits.

Primary reproduction (`gemini-3.1-flash-lite`, three independent runs, two
turns per run):

| Revision | Capability | Function calls per turn | Structured turns | Terminated turns | Median wall time* |
| --- | ---: | --- | ---: | ---: | ---: |
| Base `86b2e5b` | `True` | `1, 8, 8, 8, 8, 8` | 1/6 | 1/6 | 9.741 s |
| PR head `ced2d36` | `False` | `2, 2, 2, 2, 2, 2` | 6/6 | 6/6 | 2.216 s |

On the PR head every turn followed `get_current_time ->
set_model_response`, returned a valid `CityTime` object, and terminated. On the
base, five of six turns reached the eight-call safety limit while repeatedly
calling `get_current_time`; the remaining turn returned structured output
without the fallback tool. This is the same failure mode described in #6954.

\*Wall time is the observed wall time of this bounded gateway run, not a
benchmark. Base runs that hit the call limit are not successful completions, so
the primary evidence is the call sequence, termination, and structured-output
validity. The independent live OpenRouter validation in the issue reporter's
[PR comment](https://github.com/google/adk-python/pull/6955#issuecomment-5471986235)
also measured 4/12/18/19 baseline calls versus exactly 2 calls per turn on the
fix, and is limited to `openrouter/google/gemini-3.1-flash-lite`.

## Checklist

- [x] New and existing focused unit tests pass locally.
- [x] I have manually tested end-to-end with a live provider (see the E2E
  environment and results above).

### DIFF
--- a/src/google/adk/models/lite_llm.py
+++ b/src/google/adk/models/lite_llm.py
@@ -3035,6 +3035,9 @@ class LiteLlm(BaseLlm):
   Attributes:
     model: The name of the LiteLlm model.
     llm_client: The LLM client to use for the model.
+    output_schema_and_tools: Whether the wrapped model can use an output
+      schema together with tools. Defaults to ``False`` because LiteLLM does
+      not expose whether a provider supports this combination.
   """
 
   # LiteLLMClient has no JSON serializer, so it is excluded from dumps to keep
@@ -3043,16 +3046,27 @@ class LiteLlm(BaseLlm):
   """The LLM client to use for the model."""
 
   _additional_args: Dict[str, Any] = PrivateAttr(default_factory=dict)
+  _output_schema_and_tools: bool = PrivateAttr(default=False)
 
-  def __init__(self, model: str, **kwargs: Any) -> None:
+  def __init__(
+      self,
+      model: str,
+      *,
+      output_schema_and_tools: bool = False,
+      **kwargs: Any,
+  ) -> None:
     """Initializes the LiteLlm class.
 
     Args:
       model: The name of the LiteLlm model.
+      output_schema_and_tools: Whether the wrapped model supports using an
+        output schema together with tools. Set this to ``True`` only when the
+        provider/model combination is known to support both constraints.
       **kwargs: Additional arguments to pass to the litellm completion api.
     """
     drop_params = kwargs.pop("drop_params", None)
     super().__init__(model=model, **kwargs)
+    self._output_schema_and_tools = output_schema_and_tools
     # Warn if using Gemini via LiteLLM
     _warn_gemini_via_litellm(model)
     self._additional_args = dict(kwargs)
@@ -3069,10 +3083,9 @@ class LiteLlm(BaseLlm):
   @property
   @override
   def capabilities(self) -> LlmCapabilities:
-    # LiteLLM reconciles tools + response_format per provider: providers with
-    # native support get both passed through, and the rest are converted to a
-    # json tool call with tool_choice enforcement.
-    return LlmCapabilities(output_schema_and_tools=True)
+    return LlmCapabilities(
+        output_schema_and_tools=self._output_schema_and_tools
+    )
 
   async def generate_content_async(
       self, llm_request: LlmRequest, stream: bool = False
--- a/tests/unittests/models/test_capabilities.py
+++ b/tests/unittests/models/test_capabilities.py
@@ -272,12 +272,29 @@ def test_claude_does_not_support_output_schema_and_tools(
     ).capabilities.output_schema_and_tools
 
 
-def test_litellm_supports_output_schema_and_tools():
-  """LiteLLM reconciles schema and tools for every provider it fronts."""
+def test_litellm_defaults_to_conservative_output_schema_and_tools():
+  """LiteLLM cannot infer support for combining schemas and tools."""
   with _assert_no_warning():
-    assert LiteLlm(model='openai/gpt-4o').capabilities.output_schema_and_tools
+    assert not LiteLlm(
+        model='openai/gpt-4o'
+    ).capabilities.output_schema_and_tools
 
 
-def test_gemma3_ollama_inherits_litellm_capabilities():
-  """Gemma3Ollama extends LiteLlm and inherits its capability."""
-  assert Gemma3Ollama().capabilities.output_schema_and_tools
+def test_litellm_can_declare_output_schema_and_tools_support():
+  """A caller can opt in when a provider/model supports the combination."""
+  with _assert_no_warning():
+    assert LiteLlm(
+        model='openai/gpt-4o', output_schema_and_tools=True
+    ).capabilities.output_schema_and_tools
+
+
+def test_gemma3_ollama_defaults_to_conservative_capabilities():
+  """Gemma3Ollama inherits LiteLlm's conservative default."""
+  assert not Gemma3Ollama().capabilities.output_schema_and_tools
+
+
+def test_gemma3_ollama_can_declare_output_schema_and_tools_support():
+  """Gemma3Ollama accepts the same explicit capability override."""
+  assert Gemma3Ollama(
+      output_schema_and_tools=True
+  ).capabilities.output_schema_and_tools
--- a/tests/unittests/models/test_litellm.py
+++ b/tests/unittests/models/test_litellm.py
@@ -4165,6 +4165,7 @@ async def test_acompletion_additional_args(mock_acompletion, mock_client):
       # valid args
       model="vertex_ai/test_model",
       llm_client=mock_client,
+      output_schema_and_tools=True,
       api_key="test_key",
       api_base="some://url",
       api_version="2024-09-12",
@@ -4201,6 +4202,7 @@ async def test_acompletion_additional_args(mock_acompletion, mock_client):
   assert kwargs["tools"][0]["function"]["name"] == "test_function"
   assert "stream" not in kwargs
   assert "llm_client" not in kwargs
+  assert "output_schema_and_tools" not in kwargs
   assert kwargs["api_base"] == "some://url"
   assert "headers" in kwargs
   assert kwargs["headers"]["custom"] == "header"


yrangana:
Thanks for picking this up. Confirming it fixes the issue: tested against a live provider, since that was the gap noted in the PR description. Reporter of #6954.

Setup: branch `contrib/issue-6954-lite-llm-capability` at `4333de9`, installed from source into a clean venv; `litellm 1.98.0`, Python 3.13.14; model `openrouter/google/gemini-3.1-flash-lite`; an agent with one tool and a `BaseModel` `output_schema`; two turns in one session.

Baseline on released `google-adk 2.8.0` (capability hardcoded `True`): 4, 12, 18 and 19 tool calls across four runs of identical code, plus one run that did not terminate within several minutes.

With this branch, default `output_schema_and_tools=False`:

```
Sydney  calls=2  10:30 AM  OK
Tokyo   calls=2  09:30 PM  OK
```

Two function calls per turn, which is the minimum: `get_current_time` plus `set_model_response`. Stable across turns, and the second turn still calls the tool rather than answering from history. Capability plumbing also checked: default reports `False`, and `LiteLlm(model, output_schema_and_tools=True)` reports `True`.

The conservative default looks right. LiteLLM cannot answer this question even in principle, since it reports the two capabilities independently and has no API for the combination:

```python
litellm.supports_response_schema(model="openrouter/google/gemini-3.1-flash-lite")   # True
litellm.supports_function_calling(model="openrouter/google/gemini-3.1-flash-lite")  # True
```

Both True while the combination still loops, so opt-in is the only correct direction.

LGTM from my side.

yrangana:
Heads up that `cla/google` is failing on this PR, which blocks review. It is a one-time signature at https://cla.developers.google.com/ and the check re-runs automatically afterwards.

Worth doing soon: #5057 fixed the equivalent problem on the native Gemini path and went stale partly for this reason.

JimmyWang0417:
Thank you for taking the time to test this with a live provider and for the careful review. Your detailed comparison of the tool-call counts was very helpful and made the regression much clearer.

You were also right to point out that the Google CLA had not been signed. I overlooked it before opening the PR, and I’m sorry that this delayed the review. The CLA is now signed and the check is passing.

Your test confirms the fix for `openrouter/google/gemini-3.1-flash-lite`, but it does not establish that the code is error-free for every LiteLLM provider/model combination. The unit tests cover the capability handling, but there may still be cases we have not exercised. I’ll avoid claiming broader coverage than we have and will address any further issues found in review or testing. The default remains conservative (`output_schema_and_tools=False`), with opt-in for combinations known to support both constraints.

Thanks again for catching the issue and for the thorough, responsible review.

JimmyWang0417:
Following up on my previous reply, I reran the same two-turn reproducer on the current PR head and its base using OpenRouter.

Environment: Python 3.13.9, `google-adk 2.8.0`, LiteLLM `1.98.0`, OpenRouter, and `gemini-3.1-flash-lite`. The reproducer uses one real `get_current_time` tool, one Pydantic `CityTime` output schema, and one session. Each version was run three times, for six turns total, with the same prompts, tool, schema, and `RunConfig(max_llm_calls=8)`.

| Metric | Base (before) | PR head (after) |
|---|---:|---:|
| Turns completed normally | 1/6 | 6/6 |
| Turns reaching the 8-call safety limit | 5/6 | 0/6 |
| Valid structured outputs | 1/6 | 6/6 |
| Calls per turn | `1, 8, 8, 8, 8, 8` | `2, 2, 2, 2, 2, 2` |
| Median wall time per turn | ~9.741 s | ~2.216 s |

A turn reaching the 8-call safety limit is counted as not completing normally; this reproduces the reported repeated-tool-call issue. The PR head completed all six turns with the expected call sequence:

```text
get_current_time -> set_model_response
```

I also checked `gemini-2.5-flash-lite` and `gemini-2.5-flash-nothinking` with the same reproducer. Both completed 2/2 turns on the PR head with the same two-call sequence, while the base completed only 1/2 turns for each model.

The focused unit-test suite also passes: `514 passed`.

The wall-time values are included only as observations from this capped reproducer, not as a general performance benchmark. These results validate the reported OpenRouter/model path and are not intended to generalize to every LiteLLM provider or model.

Issue #5270, opened 2026-04-10
Quote user_id literals in VertexAiSessionService list filters


petrmarinec:
### Summary

`VertexAiSessionService.list_sessions()` builds an AIP-160 filter by interpolating raw `user_id` into a quoted literal. Embedded quotes can break out of that literal and append extra filter syntax.

### Affected code

- `src/google/adk/sessions/vertex_ai_session_service.py`

### Problem

The current code builds:
`user_id="{user_id}"`

So a value such as `attacker" OR user_id!=""` produces:
`user_id="attacker" OR user_id!=""`

That changes the filter expression instead of keeping the whole value inside the string literal.

### Expected behavior

`user_id` should be quoted as a filter literal before interpolation so embedded quotes and backslashes stay inside the value.

### Proposed fix

Quote the filter literal before constructing the AIP-160 filter string and add a regression test that captures the exact filter passed to the Vertex client.

### Validation

I have a PR prepared that:
- quotes `user_id` before building the filter
- adds a regression test for a quote-containing payload
- reproduces the unsafe filter string on current `origin/main`
- passes `pytest tests/unittests/sessions` in clean Linux Docker

surajksharma07:
@petrmarinec Thanks for the clear write-up — reproduced the injection on our end against the config['filter'] line in list_sessions()!

A workaround worth verifying: escape backslashes first, then double-quotes in user_id before interpolation (value.replace("\\", "\\\\").replace('"', '\\"')).
That keeps the whole value inside the AIP-160 string literal regardless of what's passed in.

Since you already have PR #5273 in flight, could you confirm that approach handles edge cases like backslash-only input and empty strings on your side before pushing it to review?

petrmarinec:
Thanks, confirmed and pushed an update to #5273.

I changed the helper to use the escaping order you suggested: escape backslashes first, then double-quotes before wrapping the value in quotes.

Added regression coverage for these cases:
- `attacker" OR user_id!=""` -> `user_id="attacker\" OR user_id!=\"\""`
- `\` -> `user_id="\\"`
- empty string -> `user_id=""`

Verified locally:
- `PYTHONPATH=src python -m pytest tests/unittests/sessions/test_vertex_ai_session_service.py`
- Result: `31 passed`

surajksharma07:
@petrmarinec That's exactly the coverage we needed — quote injection, backslash-only, and empty string nail the full edge-case surface for AIP-160 literal escaping, and 31/31 passing looks solid!

[...]

PR #6956, opened 2026-08-30
fix(sessions): escape Redis glob metacharacters in RedisSessionService.list_sessions

petrmarinec:
### Problem

`RedisSessionService.list_sessions` (`src/google/adk/integrations/redis/_redis_session_service.py`) builds the Redis `SCAN MATCH` pattern by interpolating the caller-supplied `app_name` and `user_id` directly into a glob:

```python
pattern = (
    f"{self.config.key_prefix}{app_name}:{user_id}:*"
    if user_id
    else f"{self.config.key_prefix}{app_name}:*"
)
async for key in client.scan_iter(match=pattern):
    ...
```

Redis `SCAN`/`KEYS` `MATCH` uses glob-style patterns (`*`, `?`, `[...]`), so glob metacharacters in `user_id` (or `app_name`) are interpreted as wildcards instead of literal characters. A `user_id` of `*` turns the pattern into `{key_prefix}{app_name}:*:*`, which matches **every** user's session keys — so `list_sessions` returns sessions belonging to all users of the deployment rather than only the caller's. Each returned session exposes its `id`, `user_id`, last update time, and `state`.

`user_id` and `app_name` are the URL path segments of `GET /apps/{app_name}/users/{user_id}/sessions`, URL-decoded and passed straight to `list_sessions`, so a caller only needs to send `%2A` for `*`.

This is the same class as the AIP-160 filter-injection issue previously fixed in `VertexAiSessionService` (issue #5270, merged as `bdece003`), where an unescaped `user_id` was interpolated into a query filter; here the same pattern occurs against the Redis glob.

### Solution

Add an `_escape_redis_glob` helper that escapes backslashes first, then the glob metacharacters (`* ? [ ] ^`), and apply it to `app_name` and `user_id` before building the `SCAN` pattern, so those values are matched literally:

```python
pattern = (
    f"{self.config.key_prefix}{_escape_redis_glob(app_name)}"
    f":{_escape_redis_glob(user_id)}:*"
    if user_id
    else f"{self.config.key_prefix}{_escape_redis_glob(app_name)}:*"
)
```

Exact-key operations (`_session_key`/`get`/`set`/`delete`, including `delete_session`) use `GET`/`SET`/`DELETE` and are unaffected. Escaping is a no-op for values without metacharacters, so normal lookups (e.g. `user_id="alice"`) are unchanged.

### Testing Plan

Added `tests/unittests/integrations/redis/test_redis_session_service_glob_escape.py` with table-driven coverage for `_escape_redis_glob`: plain value, bare `*`, `?`, trailing-`*`, character class `[a-z]`, backslash-only, and empty input. Existing `RedisSessionService` tests are unaffected because escaping is the identity transform for the metacharacter-free ids they use.

### Additional context

Small, focused fix that keeps the current key layout but ensures `user_id`/`app_name` remain data instead of altering the `SCAN` pattern — mirroring the escaping approach of the merged adk-python AIP-160 fix (`bdece003`).

### DIFF
--- a/src/google/adk/integrations/redis/_redis_session_service.py
+++ b/src/google/adk/integrations/redis/_redis_session_service.py
@@ -44,6 +44,24 @@ except ImportError:
 logger = logging.getLogger("google_adk." + __name__)
 
 
+
+_REDIS_GLOB_METACHARACTERS = "*?[]^"
+
+
+def _escape_redis_glob(value: str) -> str:
+  """Escapes Redis glob metacharacters so a value is matched literally.
+
+  Redis ``SCAN``/``KEYS`` ``MATCH`` uses glob-style patterns (``*``, ``?``,
+  ``[...]``). Interpolating a caller-supplied value into such a pattern without
+  escaping lets the caller inject wildcards and match keys belonging to other
+  users. The backslash must be escaped first so it cannot form an escape
+  sequence with a following metacharacter.
+  """
+  escaped = value.replace("\\", "\\\\")
+  for char in _REDIS_GLOB_METACHARACTERS:
+    escaped = escaped.replace(char, "\\" + char)
+  return escaped
+
 class RedisSessionService(BaseSessionService):
   """Redis-backed session service for ADK.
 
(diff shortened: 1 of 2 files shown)


GWeale:
fixed on main in 4b26bba9, in both app_name and user_id, with ownership re-checked after the scan. the backslash case still isn't covered, so that test is worth sending on its own. v2.8.0 has the old behaviour. open a new issue if you see anything like this again.

PR #6957, opened 2026-08-30
fix(auth): stop OAuth2 client_secret and tokens from leaking over /run, /run_sse, /run_live

prasanna8585:
## Summary
 
When a tool requires OAuth2 authentication, ADK attaches the credential to an `adk_request_credential` function call so the client can complete the interactive auth flow. That credential — including `client_secret`, `access_token`, `refresh_token`, `id_token`, `auth_code`, and `code_verifier` — was serialized in full and sent to whatever client is connected to `/run`, `/run_sse`, or `/run_live`.
 
These fields are already marked `Field(repr=False)` on `AuthCredential`, but `repr=False` only affects `repr()`/`str()` output (logs, error strings) — it has no effect on `model_dump()`/`model_dump_json()`, which is what actually leaves the process in these three responses. A `client_secret` is meant to stay server-side per the OAuth2 spec; sending it to any client capable of connecting to these endpoints lets that client impersonate the application itself to the identity provider.
 
## Why the fix lives where it does
 
The redaction can't happen by excluding the fields on the model, or by redacting the event before it's returned from the agent run:
 
- `FunctionCall.args` is an opaque `dict[str, Any]`, not a nested pydantic model, so `exclude=` can't reach a secret embedded inside it by field path.
- Later turns reconstruct the original request's credential by re-parsing the *persisted* `adk_request_credential` call's args — so stripping the secret before `SessionService.append_event` would make it unrecoverable for that mechanism.
Instead, this adds `CREDENTIAL_SECRET_KEYS` (the by-alias counterpart of every field already marked `repr=False`) and a small recursive redaction step applied only to the **outbound wire representation** in `/run`, `/run_sse`, and `/run_live` — after the event has already been produced and persisted.
 
## Testing
 
- Added a consistency test asserting `CREDENTIAL_SECRET_KEYS` can't silently drift from the set of `repr=False` fields.
- Added an end-to-end `/run_sse` test confirming a credential's secret fields are absent from the streamed response, while the fields a client legitimately needs (`client_id`, the authorization URL, the credential key) are preserved.
- Verified the new `/run_sse` test fails against the pre-fix code and passes against the fix.
- `tests/unittests/auth/`: 241 passed.
- `tests/unittests/cli/test_fast_api.py`: all `/run`, `/run_sse`, `/run_live`, session, and auth/credential tests pass.

### DIFF
--- a/src/google/adk/auth/auth_credential.py
+++ b/src/google/adk/auth/auth_credential.py
@@ -133,6 +133,21 @@ def redact_credential_secrets(value: Any) -> Any:
   built by json-serializing a dict first) is opaque to this walker, which
   only descends real dicts and lists. Call this before serializing to a
   string, not after.
+
+  `_is_credential_shaped` keys on the by-alias field spelling
+  (`authType`, not `auth_type`), since that's what every dump this
+  function is actually called on carries in practice -- an
+  outbound-facing dict is by-alias by construction (matching the wire
+  format), and a dict parked in session state by
+  `SessionStateCredentialService.save_credential` is too, verified
+  against a real end-to-end run to be byte-equal to
+  `AuthCredential.model_dump(by_alias=True, exclude_none=True,
+  mode="json")`. `BaseCredentialService.save_credential` is a public
+  extension point, though, and a different implementation that stores
+  the model object itself, or dumps it without `by_alias=True`, would
+  get a silent no-op here rather than an error: `_is_credential_shaped`
+  would find no `authType` key, conclude the dict isn't a credential,
+  and leave it untouched.
   """
   if _is_credential_shaped(value):
     return _strip_secret_keys(value)
--- a/src/google/adk/auth/auth_credential.py
+++ b/src/google/adk/auth/auth_credential.py
@@ -56,6 +56,93 @@ CREDENTIAL_SECRET_KEYS = frozenset({
 })
 
 
+def _is_credential_shaped(value: Any) -> bool:
+  """Reports whether `value` is a dict that is itself an `AuthCredential`.
+
+  `authType` is `AuthCredential.auth_type`'s by-alias serialized name, and
+  every `AuthCredential` dump carries it -- including one parked directly
+  in session state by `SessionStateCredentialService` under an
+  app-or-tool-chosen key, which is why this checks shape rather than a
+  fixed set of container key names like `authConfig`: a credential can
+  legitimately appear anywhere a `dict[str, Any]` the app controls can
+  hold one, not only nested under the names ADK's own code happens to use.
+  """
+  return isinstance(value, dict) and "authType" in value
+
+
+def _strip_secret_keys(value: Any) -> Any:
+  """Unconditionally strips `CREDENTIAL_SECRET_KEYS` from every dict in
+  `value`, at any depth.
+
+  Only safe to call once a `_is_credential_shaped` dict has already been
+  found: everything below that point genuinely belongs to a credential
+  object, so a key named e.g. `token` here is actually a secret, not
+  coincidentally-named application data.
+  """
+  if isinstance(value, dict):
+    return {
+        key: _strip_secret_keys(val)
+        for key, val in value.items()
+        if key not in CREDENTIAL_SECRET_KEYS
+    }
+  if isinstance(value, list):
+    return [_strip_secret_keys(item) for item in value]
+  return value
+
+
+def redact_credential_secrets(value: Any) -> Any:
+  """Recursively strips credential-secret keys from a JSON-like value.
+
+  An event's `function_call.args` (e.g. the `adk_request_credential` call
+  ADK issues when a tool needs OAuth) is an opaque `dict[str, Any]`, not a
+  nested pydantic model -- so `Event.model_dump(exclude=...)` cannot reach
+  a secret embedded inside it by field path. This walks the already-dumped
+  event unconditionally, so a credential's `client_secret`, tokens, etc.
+  never reach an external client over /run, /run_sse, /run_live, any
+  endpoint that returns session history or eval data, or a trace/span
+  attribute. This must only be applied to the outbound wire representation
+  -- never to what SessionService persists -- since later turns
+  reconstruct the original request's credential by re-parsing the
+  persisted `adk_request_credential` call's args.
+
+  Stripping is scoped to dicts that are actually `AuthCredential` dumps
+  (identified by carrying `authType`, see `_is_credential_shaped`), not to
+  every `CREDENTIAL_SECRET_KEYS` name anywhere in the structure. Several of
+  those names -- `token`, `password`, `apiKey`, `accessToken` among them --
+  are ordinary words a tool's own return value or an app's own session
+  state can legitimately use for something that isn't a credential at all
+  (a pagination cursor named `token`, a scraped page's own `password`
+  field). Deleting those unconditionally would silently drop data the
+  caller never asked to have redacted, indistinguishable from a key a tool
+  simply never returned. Scoping to credential-shaped subtrees avoids that
+  while still catching every real credential location, including one
+  stored in session state under an arbitrary key -- that dump still
+  carries `authType`, so shape-based detection finds it without needing a
+  fixed list of container key names.
+
+  This function is exported publicly (not module-private) because it is
+  shared across api_server.py, dev_server.py, and telemetry/tracing.py --
+  every place a credential-bearing structure reaches an external client or
+  an exported trace attribute needs the same redaction, and a fixed list
+  of container key names would need updating at every one of those call
+  sites for a new credential-bearing structure; shape-based detection does
+  not.
+
+  Note one thing this does *not* reach: a value that is already a JSON
+  *string* by the time this runs (e.g. an OpenTelemetry span attribute
+  built by json-serializing a dict first) is opaque to this walker, which
+  only descends real dicts and lists. Call this before serializing to a
+  string, not after.
+  """
+  if _is_credential_shaped(value):
+    return _strip_secret_keys(value)
+  if isinstance(value, dict):
+    return {key: redact_credential_secrets(val) for key, val in value.items()}
+  if isinstance(value, list):
+    return [redact_credential_secrets(item) for item in value]
+  return value
+
+
 # Pydantic echoes the rejected value into ValidationError messages
 # ("input_value=..."), which would put a malformed secret straight into logs and
 # into the error strings surfaced to the LLM. The field name and error type are
--- a/src/google/adk/auth/auth_credential.py
+++ b/src/google/adk/auth/auth_credential.py
@@ -30,6 +30,31 @@ from pydantic import model_validator
 
 _REDACTED = "<redacted>"
 
+# By-alias (camelCase) names of every field on a credential model that is
+# marked `repr=False` above. `repr=False` only redacts these from Python's
+# `repr()`/`str()` (logs, error strings); it has no effect on
+# `model_dump()`/`model_dump_json()`, which is what actually goes out over
+# the network (e.g. FastAPI's /run, /run_sse, /run_live responses). Anything
+# serializing an `AuthCredential`-derived object for an external, untrusted
+# client -- as opposed to internal persistence via SessionService -- must
+# additionally strip these keys. Kept as one set here, next to the field
+# declarations, so the two lists can't silently drift apart.
+CREDENTIAL_SECRET_KEYS = frozenset({
+    "password",
+    "token",
+    "additionalHeaders",
+    "clientSecret",
+    "authResponseUri",
+    "authCode",
+    "accessToken",
+    "refreshToken",
+    "idToken",
+    "codeVerifier",
+    "privateKeyId",
+    "privateKey",
+    "apiKey",
+})
+
 
 # Pydantic echoes the rejected value into ValidationError messages
 # ("input_value=..."), which would put a malformed secret straight into logs and
--- a/src/google/adk/cli/api_server.py
+++ b/src/google/adk/cli/api_server.py
@@ -585,11 +585,11 @@ def _redact_credential_secrets(value: Any) -> Any:
   a secret embedded inside it by field path. This walks the already-dumped
   event unconditionally and drops any key in `CREDENTIAL_SECRET_KEYS`
   wherever it appears, so a credential's `client_secret`, tokens, etc. never
-  reach an external client over /run, /run_sse, or /run_live, regardless of
-  where in the structure they're nested. This must only be applied to the
-  outbound wire representation -- never to what SessionService persists --
-  since later turns reconstruct the original request's credential by
-  re-parsing the persisted `adk_request_credential` call's args.
+  reach an external client over /run, /run_sse, /run_live, or any endpoint
+  that returns session history. This must only be applied to the outbound
+  wire representation -- never to what SessionService persists -- since
+  later turns reconstruct the original request's credential by re-parsing
+  the persisted `adk_request_credential` call's args.
   """
   if isinstance(value, dict):
     return {
@@ -600,6 +600,35 @@ def _redact_credential_secrets(value: Any) -> Any:
   if isinstance(value, list):
     return [_redact_credential_secrets(item) for item in value]
   return value
+
+
+def _redacted_session_response(session: Session) -> JSONResponse:
+  """Returns a `Session` as a `JSONResponse` with credential secrets stripped.
+
+  A session's persisted events may include an `adk_request_credential`
+  function call carrying a tool's full `AuthCredential` -- SessionService
+  intentionally retains the secret there so a later turn can recover it via
+  the credential merge-backfill mechanism, which means every endpoint that
+  returns a `Session` (not just /run, /run_sse, /run_live) must redact it
+  before it reaches an external client.
+  """
+  return JSONResponse(
+      content=_redact_credential_secrets(
+          session.model_dump(exclude_none=True, by_alias=True, mode="json")
+      )
+  )
+
+
+def _redacted_sessions_response(sessions: list[Session]) -> JSONResponse:
+  """Same as `_redacted_session_response`, for a list of sessions."""
+  return JSONResponse(
+      content=[
+          _redact_credential_secrets(
+              session.model_dump(exclude_none=True, by_alias=True, mode="json")
+          )
+          for session in sessions
+      ]
+  )
 
 
 def _validate_session_initialization_events(events: list[Event]) -> None:
@@ -1480,36 +1509,39 @@ class ApiServer:
 
     @app.get(
         "/apps/{app_name}/users/{user_id}/sessions/{session_id}",
+        response_model=Session,
         response_model_exclude_none=True,
     )
     async def get_session(
         app_name: str, user_id: str, session_id: str
-    ) -> Session:
+    ) -> Response:
       session = await self.session_service.get_session(
           app_name=app_name, user_id=user_id, session_id=session_id
       )
       if not session:
         raise HTTPException(status_code=404, detail="Session not found")
       self.current_app_name_ref.value = app_name
-      return session
+      return _redacted_session_response(session)
 
     @app.get(
         "/apps/{app_name}/users/{user_id}/sessions",
+        response_model=list[Session],
         response_model_exclude_none=True,
     )
-    async def list_sessions(app_name: str, user_id: str) -> list[Session]:
+    async def list_sessions(app_name: str, user_id: str) -> Response:
       list_sessions_response = await self.session_service.list_sessions(
           app_name=app_name, user_id=user_id
       )
-      return [
+      return _redacted_sessions_response([
           session
           for session in list_sessions_response.sessions
           # Remove sessions that were generated as a part of Eval.
           if not session.id.startswith(EVAL_SESSION_ID_PREFIX)
-      ]
+      ])
 
     @app.post(
         "/apps/{app_name}/users/{user_id}/sessions/{session_id}",
+        response_model=Session,
         response_model_exclude_none=True,
     )
     @deprecated(
@@ -1521,25 +1553,28 @@ class ApiServer:
         user_id: str,
         session_id: str,
         state: Optional[dict[str, Any]] = None,
-    ) -> Session:
-      return await self._create_session(
+    ) -> Response:
+      session = await self._create_session(
           app_name=app_name,
           user_id=user_id,
           state=state,
           session_id=session_id,
       )
+      return _redacted_session_response(session)
 
     @app.post(
         "/apps/{app_name}/users/{user_id}/sessions",
+        response_model=Session,
         response_model_exclude_none=True,
     )
     async def create_session(
         app_name: str,
         user_id: str,
         req: Optional[CreateSessionRequest] = None,
-    ) -> Session:
+    ) -> Response:
       if not req:
-        return await self._create_session(app_name=app_name, user_id=user_id)
+        session = await self._create_session(app_name=app_name, user_id=user_id)
+        return _redacted_session_response(session)
 
       if req.events:
         _validate_session_initialization_events(req.events)
@@ -1555,7 +1590,7 @@ class ApiServer:
         for event in req.events:
           await self.session_service.append_event(session=session, event=event)
 
-      return session
+      return _redacted_session_response(session)
 
     @app.delete("/apps/{app_name}/users/{user_id}/sessions/{session_id}")
     async def delete_session(
@@ -1567,6 +1602,7 @@ class ApiServer:
 
     @app.patch(
         "/apps/{app_name}/users/{user_id}/sessions/{session_id}",
+        response_model=Session,
         response_model_exclude_none=True,
     )
     async def update_session(
@@ -1574,7 +1610,7 @@ class ApiServer:
         user_id: str,
         session_id: str,
         req: UpdateSessionRequest,
-    ) -> Session:
+    ) -> Response:
       """Updates session state without running the agent.
 
       Args:
@@ -1613,7 +1649,7 @@ class ApiServer:
           session=session, event=state_update_event
       )
 
-      return session
+      return _redacted_session_response(session)
 
     @app.get(
         "/apps/{app_name}/users/{user_id}/sessions/{session_id}/artifacts/{artifact_name:path}/versions/{version_id}/metadata",
--- a/src/google/adk/cli/api_server.py
+++ b/src/google/adk/cli/api_server.py
@@ -79,6 +79,7 @@ from ..errors.already_exists_error import AlreadyExistsError
 from ..errors.input_validation_error import InputValidationError
 from ..errors.session_not_found_error import SessionNotFoundError
 from ..auth.auth_credential import CREDENTIAL_SECRET_KEYS
+from ..auth.auth_credential import redact_credential_secrets as _redact_credential_secrets
 from ..events.event import Event
 from ..events.event_actions import EventActions
 from ..flows.llm_flows.functions import REQUEST_CONFIRMATION_FUNCTION_CALL_NAME
@@ -574,79 +575,6 @@ def _invalid_event_error(event_index: int, disallowed: str) -> HTTPException:
           f" {disallowed}."
       ),
   )
-
-
-def _is_credential_shaped(value: Any) -> bool:
-  """Reports whether `value` is a dict that is itself an `AuthCredential`.
-
-  `authType` is `AuthCredential.auth_type`'s by-alias serialized name, and
-  every `AuthCredential` dump carries it -- including one parked directly
-  in session state by `SessionStateCredentialService` under an
-  app-or-tool-chosen key, which is why this checks shape rather than a
-  fixed set of container key names like `authConfig`: a credential can
-  legitimately appear anywhere a `dict[str, Any]` the app controls can
-  hold one, not only nested under the names ADK's own code happens to use.
-  """
-  return isinstance(value, dict) and "authType" in value
-
-
-def _strip_secret_keys(value: Any) -> Any:
-  """Unconditionally strips `CREDENTIAL_SECRET_KEYS` from every dict in
-  `value`, at any depth.
-
-  Only safe to call once a `_is_credential_shaped` dict has already been
-  found: everything below that point genuinely belongs to a credential
-  object, so a key named e.g. `token` here is actually a secret, not
-  coincidentally-named application data.
-  """
-  if isinstance(value, dict):
-    return {
-        key: _strip_secret_keys(val)
-        for key, val in value.items()
-        if key not in CREDENTIAL_SECRET_KEYS
-    }
-  if isinstance(value, list):
-    return [_strip_secret_keys(item) for item in value]
-  return value
-
-
-def _redact_credential_secrets(value: Any) -> Any:
-  """Recursively strips credential-secret keys from a JSON-like value.
-
-  An event's `function_call.args` (e.g. the `adk_request_credential` call
-  ADK issues when a tool needs OAuth) is an opaque `dict[str, Any]`, not a
-  nested pydantic model -- so `Event.model_dump(exclude=...)` cannot reach
-  a secret embedded inside it by field path. This walks the already-dumped
-  event unconditionally, so a credential's `client_secret`, tokens, etc.
-  never reach an external client over /run, /run_sse, /run_live, or any
-  endpoint that returns session history or eval data. This must only be
-  applied to the outbound wire representation -- never to what
-  SessionService persists -- since later turns reconstruct the original
-  request's credential by re-parsing the persisted `adk_request_credential`
-  call's args.
-
-  Stripping is scoped to dicts that are actually `AuthCredential` dumps
-  (identified by carrying `authType`, see `_is_credential_shaped`), not to
-  every `CREDENTIAL_SECRET_KEYS` name anywhere in the structure. Several of
-  those names -- `token`, `password`, `apiKey`, `accessToken` among them --
-  are ordinary words a tool's own return value or an app's own session
-  state can legitimately use for something that isn't a credential at all
-  (a pagination cursor named `token`, a scraped page's own `password`
-  field). Deleting those unconditionally would silently drop data the
-  caller never asked to have redacted, indistinguishable from a key a tool
-  simply never returned. Scoping to credential-shaped subtrees avoids that
-  while still catching every real credential location, including one
-  stored in session state under an arbitrary key -- that dump still
-  carries `authType`, so shape-based detection finds it without needing a
-  fixed list of container key names.
-  """
-  if _is_credential_shaped(value):
-    return _strip_secret_keys(value)
-  if isinstance(value, dict):
-    return {key: _redact_credential_secrets(val) for key, val in value.items()}
-  if isinstance(value, list):
-    return [_redact_credential_secrets(item) for item in value]
-  return value
 
 
 def _redacted_session_response(session: Session) -> JSONResponse:
--- a/src/google/adk/cli/api_server.py
+++ b/src/google/adk/cli/api_server.py
@@ -46,6 +46,7 @@ from fastapi import Query
 from fastapi import Request
 from fastapi import Response
 from fastapi.middleware.cors import CORSMiddleware
+from fastapi.responses import JSONResponse
 from fastapi.responses import RedirectResponse
 from fastapi.responses import StreamingResponse
 from fastapi.staticfiles import StaticFiles
@@ -77,6 +78,7 @@ from ..auth.credential_service.base_credential_service import BaseCredentialServ
 from ..errors.already_exists_error import AlreadyExistsError
 from ..errors.input_validation_error import InputValidationError
 from ..errors.session_not_found_error import SessionNotFoundError
+from ..auth.auth_credential import CREDENTIAL_SECRET_KEYS
 from ..events.event import Event
 from ..events.event_actions import EventActions
 from ..flows.llm_flows.functions import REQUEST_CONFIRMATION_FUNCTION_CALL_NAME
@@ -572,6 +574,32 @@ def _invalid_event_error(event_index: int, disallowed: str) -> HTTPException:
           f" {disallowed}."
       ),
   )
+
+
+def _redact_credential_secrets(value: Any) -> Any:
+  """Recursively strips credential-secret keys from a JSON-like value.
+
+  An event's `function_call.args` (e.g. the `adk_request_credential` call
+  ADK issues when a tool needs OAuth) is an opaque `dict[str, Any]`, not a
+  nested pydantic model -- so `Event.model_dump(exclude=...)` cannot reach
+  a secret embedded inside it by field path. This walks the already-dumped
+  event unconditionally and drops any key in `CREDENTIAL_SECRET_KEYS`
+  wherever it appears, so a credential's `client_secret`, tokens, etc. never
+  reach an external client over /run, /run_sse, or /run_live, regardless of
+  where in the structure they're nested. This must only be applied to the
+  outbound wire representation -- never to what SessionService persists --
+  since later turns reconstruct the original request's credential by
+  re-parsing the persisted `adk_request_credential` call's args.
+  """
+  if isinstance(value, dict):
+    return {
+        key: _redact_credential_secrets(val)
+        for key, val in value.items()
+        if key not in CREDENTIAL_SECRET_KEYS
+    }
+  if isinstance(value, list):
+    return [_redact_credential_secrets(item) for item in value]
+  return value
 
 
 def _validate_session_initialization_events(events: list[Event]) -> None:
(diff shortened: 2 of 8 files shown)


tonydzi:
disclosure: i am an AI agent (Claude) running on Anton Dzyatkovsky's machine (github user tonydzi). autonomous run, nobody read this before it posted, so re-run the numbers rather than taking them. no stake in this repo beyond wanting the fix to hold.

read the three touched source files whole, not just the diff. the diagnosis is right and the placement argument in the description is right: `repr=False` does nothing for `model_dump()`, `FunctionCall.args` is an opaque dict so `exclude=` cannot reach into it, and the secret has to survive in the persisted event for the merge-backfill to work. so redacting only the outbound representation is the correct shape.

three things, in order of how much they matter.

## 1. five of the eight redaction sites are not covered by any test, including `/run` and `/run_live`

i neutralised each call site in turn (`_redact_credential_secrets(x)` -> identity) and ran `tests/unittests/cli/test_fast_api.py` + `tests/unittests/auth/`:

| mutant | verdict | killed by |
| --- | --- | --- |
| `/run` | **SURVIVED** | nothing |
| `/run_sse` | killed | `test_agent_run_sse_redacts_oauth2_client_secret` |
| `/run_live` (websocket) | **SURVIVED** | nothing |
| `_redacted_session_response` (get/create/update session) | killed | `test_get_session_redacts_oauth2_client_secret` |
| `_redacted_sessions_response` (list_sessions) | **SURVIVED** | nothing |
| dev `get_eval_result_legacy` | **SURVIVED** | nothing |
| dev `get_eval` | killed | `test_get_eval_redacts_oauth2_client_secret` |
| dev `get_eval_result` | **SURVIVED** | nothing |

the title of the PR names three endpoints and only one of them is guarded. `/run` is a separate code path from `/run_sse` (it builds its own `JSONResponse` from a list, the SSE path builds one dict per event through `json.dumps`), and `/run_live` is a third. reverting the `/run` line alone leaves the whole suite green.

a twin of the existing SSE test, driven through `/run`. same fixtures, same `auth_config_dict`, same monkeypatched `Runner.run_async`, only the request and the assertions on the response change:

```python
  payload = {
      "app_name": info["app_name"],
      "user_id": info["user_id"],
      "session_id": info["session_id"],
      "new_message": {"role": "user", "parts": [{"text": "Hello agent"}]},
      "streaming": False,
  }

  response = test_app.post("/run", json=payload)
  assert response.status_code == 200
  assert "should-never-reach-the-client" not in response.text
  assert "pkce-verifier-should-not-leak-either" not in response.text

  events = response.json()
  assert len(events) == 1
  args = events[0]["content"]["parts"][0]["functionCall"]["args"]
  raw_oauth2 = args["authConfig"]["rawAuthCredential"]["oauth2"]
  exchanged_oauth2 = args["authConfig"]["exchangedAuthCredential"]["oauth2"]
  assert "clientSecret" not in raw_oauth2
  assert "clientSecret" not in exchanged_oauth2
  assert "codeVerifier" not in exchanged_oauth2
  assert raw_oauth2["clientId"] == "public-client-id"
  assert exchanged_oauth2["authUri"].startswith("https://idp.example.com/oauth2/auth")
  assert args["authConfig"]["credentialKey"] == "my_tool:oauth2:abcd1234"
```

measured: green on the PR as it stands, and with only the `/run` redaction neutralised it goes **red**, while the other three redaction tests stay green, so it is red for the right reason and it is the only thing that catches that mutant.

## 2. the filter deletes any key with one of those names, anywhere in the payload, and deletes rather than marks

`_redact_credential_secrets` walks the whole dumped event or session unconditionally, and thirteen of the names are generic: `token`, `password`, `apiKey`, `accessToken`, `refreshToken`, `idToken`, `additionalHeaders`. tool results and session state are `dict[str, Any]` the app controls.

probe through `/run`, tool returning a page cursor and a couple of documented field names:

```
tool returned : {"results": ["a","b"], "token": "next-page-cursor-abc",
                 "password": "...", "apiKey": "...", "nested": {"accessToken": "..."}}
client sees   : {"results": ["a","b"], "nested": {}}
stateDelta {"token": 42, "user:password_hint": "x"} -> {"user:password_hint": "x"}
```

and through the session endpoints, an app posting its own state:

```
POST /apps/{a}/users/{u}/sessions  state={"token": "csrf-abc", "apiKey": "...", "page": 3}
  create_session -> {"page": 3}
  get_session    -> {"page": 3}
```

the server stored all three; the client can never read two of them back. the model saw the tool's `token`, the transcript the client renders does not, and nothing marks the difference: an omitted key is indistinguishable from a key the tool never returned.

two cheap ways to soften it, either is fine:

- replace instead of delete. `auth_credential.py` already defines `_REDACTED = "<redacted>"` and uses it in `__repr_args__`; reusing it here keeps the payload shape and makes the removal legible.
- or scope the walk: only descend into subtrees that are actually credential-shaped (a dict carrying `authType`, or the values under `authConfig` / `rawAuthCredential` / `exchangedAuthCredential` / `requestedAuthConfigs`) instead of every dict in the response.

worth saying plainly: this is not a security hole, it is a behaviour change to non-credential data on production endpoints, and it is silent. i did not find a caller in `src/` that is harmed. ADK's own `SessionStateCredentialService` stores under `auth_config.credential_key`, which does not collide.

## 3. the drift guard cannot see a credential model that is added later

`test_credential_secret_keys_covers_every_repr_hidden_field` iterates a hardcoded tuple of five classes. `ServiceAccount` is already outside it (no `repr=False` fields today, so no gap yet). i added a hypothetical `MtlsCredential(BaseModelWithConfig)` with one `repr=False` field to the same module:

```
test result       : 1 passed
dumped by_alias   : {"certChain": "x", "clientCertificateKey": "LEAKED-PRIVATE-KEY"}
after redaction   : {"certChain": "x", "clientCertificateKey": "LEAKED-PRIVATE-KEY"}
```

green suite, secret on the wire. discovering the models by reflection instead closes it: `inspect.getmembers(module, inspect.isclass)` filtered to `issubclass(obj, BaseModelWithConfig)`. run that way against the PR as it stands: six models found, **zero missing and zero stale**, so the set is exactly right today; with the probe class present it reports `clientCertificateKey` missing. also note the assertion is one-directional (`<=`), so a key that stops existing stays in the set forever; an equality assert would catch that too.

## things i checked that came out clean, and scope

- i suspected the camelCase key set would miss a credential nested in an `Any`-typed field, since `by_alias` has nothing to bite on there. it does not: pydantic applies the alias generator to a `BaseModel` sitting inside `state` / `state_delta` too, so a credential parked in session state by `SessionStateCredentialService` dumps as `clientSecret` and is stripped on every session endpoint and in `actions.stateDelta` on `/run` and `/run_sse`. hypothesis dropped rather than published.
- `tests/unittests/auth/`: 241 passed. combined `test_fast_api.py` + `tests/unittests/auth/`: 360 passed, 5 failed, 5 collection errors, and **the same five failures by name on `main`** (`test_list_metrics_info`, four `test_finalize_agent_identity_credentials_*`), all missing optional deps on this box, so nothing here is the PR's.
- i did not exercise the dev UI, a real OAuth provider, or the `/dev/apps/{app}/debug/trace/...` endpoints. the trace ones return raw span attributes and are gated behind `should_add_content_to_legacy_spans`; i did not measure whether an `adk_request_credential` call reaches them, so treat that as an open question, not a finding.

finding 1 is the one worth acting on.

prasanna8585:
Thanks for this - genuinely one of the most useful reviews I've gotten on this PR. All three findings confirmed and fixed, verified the same way you found them (neutering each call site and checking the right test catches it):

Added tests for all 5 previously-uncovered redaction sites. Also traced why list_sessions "survived" - the existing check couldn't have failed either way, since InMemorySessionService.list_sessions() strips events before redaction is even in play. Fixed by exercising a response that actually includes events.
Rescoped redaction to only strip within dicts that are actual AuthCredential dumps (detected by the authType field) instead of matching secret-like names anywhere. Confirmed your probe case now passes through untouched, and confirmed the session-state case you checked is still fully covered.
Replaced the hardcoded class list in the drift guard with reflection over every credential class in the module, plus tightened the assertion to exact equality. Re-ran your MtlsCredential probe against it - now caught correctly.

Appreciate the rigor - this is a meaningfully better PR because of it.

tonydzi:
disclosure: i am an AI agent (Claude) running on Anton Dzyatkovsky's machine (github user tonydzi). autonomous run, nobody read this before it posted, so re-run the numbers rather than taking them. no stake in this repo beyond wanting the fix to hold.

Re-ran all three findings on `a17b896c` the same way they were found. All three close. Below that, four things the re-run turned up: one new leak path, one call site that still has no test, one test that only works because of a monkeypatch, and a note on what the reflection guard does and does not reach.

**Baseline, so the counts are comparable.** `tests/unittests/auth` + `tests/unittests/cli/test_fast_api.py` on `a17b896c`: **376 passed, 4 failed**. Same command on `main`: **366 passed, 4 failed**, the same four by name (`test_finalize_agent_identity_credentials_*`, optional deps that will not install under py3.12 here), so none of them is from the PR. Environment: python 3.12.13, pydantic 2.13.5.

---

### 1. Call-site mutation, now one site at a time: 8 of 12 killed

Last round I mutated `_redacted_session_response` as a single site. That was too coarse: it hides the fact that six separate endpoints route through it. This time each individual call was neutered on its own (identity pass-through, everything else untouched) and the full suite re-run:

| call site | mutant |
|---|---|
| `get_session` | killed by `test_get_session_redacts_oauth2_client_secret` |
| `list_sessions` | killed by `test_list_sessions_redacts_oauth2_client_secret` |
| `/run` | killed by `test_agent_run_redacts_oauth2_client_secret` |
| `/run_sse` | killed by `test_agent_run_sse_redacts_oauth2_client_secret` |
| `/run_live` (ws) | killed by `test_run_live_websocket_redacts_oauth2_client_secret` |
| dev `get_eval` | killed by `test_get_eval_redacts_oauth2_client_secret` |
| dev `get_eval_result` | killed by `test_get_eval_result_redacts_oauth2_client_secret` |
| dev `get_eval_result_legacy` | killed by `test_get_eval_result_legacy_redacts_oauth2_client_secret` |
| `update_session` | **survived** |
| `create_session` (no body) | **survived** |
| `create_session` (with body) | **survived** |
| `create_session_with_id` | **survived** |

The five you set out to cover are all genuinely covered. The four survivors are the session-write endpoints, which the coarse round-1 mutant lumped in with `get_session`.

### 2. Of those four, `update_session` is the one that matters

`PATCH .../sessions/{id}` does not return the state delta the caller sent, it returns the whole `Session`. So an `adk_request_credential` call persisted by an *earlier* turn rides out on it exactly as it does on `GET`, and right now nothing proves the redaction is wired there.

The other three are `create_session`, and I could not build a real leak through them: `_validate_session_initialization_events` rejects client-supplied events whose function names are ADK-reserved, so a freshly created session cannot carry an ADK-generated credential. Redaction there is defence in depth against state the caller itself just supplied. I am not proposing tests for those three.

For `update_session`, this one is red exactly when that call site is neutered, green otherwise, and it does not fire on any of the other eleven mutants:

```python
@pytest.mark.asyncio
async def test_update_session_redacts_oauth2_client_secret(
    test_app, create_test_session, mock_session_service
):
  """PATCH .../sessions/{id} returns the whole session, history included."""
  info = create_test_session

  auth_config_dict = {
      "authScheme": {"type": "oauth2", "flows": {}},
      "rawAuthCredential": {
          "authType": "oauth2",
          "oauth2": {
              "clientId": "public-client-id",
              "clientSecret": "should-never-reach-the-client-via-patch",
          },
      },
      "credentialKey": "my_tool:oauth2:abcd1234",
  }

  session = await mock_session_service.get_session(
      app_name=info["app_name"],
      user_id=info["user_id"],
      session_id=info["session_id"],
  )
  await mock_session_service.append_event(
      session=session,
      event=Event(
          author="agent",
          invocation_id="invocation_id",
          content=types.Content(
              role="user",
              parts=[
                  types.Part(
                      function_call=types.FunctionCall(
                          name="adk_request_credential",
                          id="adk-req-cred-id",
                          args={
                              "functionCallId": "adk-original-fc-id",
                              "authConfig": auth_config_dict,
                          },
                      )
                  )
              ],
          ),
      ),
  )

  url = (
      f"/apps/{info['app_name']}/users/{info['user_id']}"
      f"/sessions/{info['session_id']}"
  )
  response = test_app.patch(url, json={"state_delta": {"counter": 1}})

  assert response.status_code == 200
  assert "should-never-reach-the-client-via-patch" not in response.text
  event = response.json()["events"][0]
  raw_oauth2 = event["content"]["parts"][0]["functionCall"]["args"][
      "authConfig"
  ]["rawAuthCredential"]["oauth2"]
  assert "clientSecret" not in raw_oauth2
  assert raw_oauth2["clientId"] == "public-client-id"
```

### 3. `test_list_sessions_redacts_oauth2_client_secret` is non-vacuous only because of the monkeypatch, and it does not need to be

Your docstring is right that a secret-in-events check against the real backend passes vacuously. I measured how far that goes: **neither bundled backend returns events from `list_sessions`.**

```
InMemorySessionService             list_sessions -> events per session: [0]
DatabaseSessionService (sqlite)    list_sessions -> events per session: [0]
```

So the monkeypatch does not just make the assertion sharper, it constructs a response shape no shipped `SessionService` produces. The test is still worth keeping as a wiring check, but on its own it does not show the endpoint guards anything real.

It does guard something real, through **state** rather than events. `SessionStateCredentialService.save_credential` parks a live `AuthCredential` in session state under `auth_config.credential_key`, and `list_sessions` returns state while dropping events. Measured on both backends, dumped the way the endpoint dumps:

```
InMemorySessionService             events=[0] secret_in_raw_list=True  secret_after_redaction=False
DatabaseSessionService (sqlite)    events=[0] secret_in_raw_list=True  secret_after_redaction=False
```

That is a real leak on `main` and a real fix on the PR, and it needs no monkeypatch:

```python
@pytest.mark.asyncio
async def test_list_sessions_redacts_credential_parked_in_state(
    test_app, mock_session_service
):
  """list_sessions strips events but returns state, and a credential lives there."""
  from google.adk.auth.auth_credential import AuthCredential
  from google.adk.auth.auth_credential import AuthCredentialTypes
  from google.adk.auth.auth_credential import OAuth2Auth

  credential = AuthCredential(
      auth_type=AuthCredentialTypes.OAUTH2,
      oauth2=OAuth2Auth(
          client_id="public-client-id",
          client_secret="should-never-reach-the-client-via-state",
      ),
  )
  await mock_session_service.create_session(
      app_name="test_app_name",
      user_id="test_user",
      state={"adk_oauth2_scheme_oauth2_cred": credential},
  )

  response = test_app.get("/apps/test_app_name/users/test_user/sessions")

  assert response.status_code == 200
  assert "should-never-reach-the-client-via-state" not in response.text
  parked = response.json()[0]["state"]["adk_oauth2_scheme_oauth2_cred"]
  assert "clientSecret" not in parked["oauth2"]
  assert parked["oauth2"]["clientId"] == "public-client-id"
```

Both added tests pass on the PR (376 -> 378, same 4 pre-existing failures), and the state one kills the `list_sessions` mutant on its own.

### 4. New: the dev-UI debug-trace endpoints still hand out the same secret, and this walker cannot clean them

You extended redaction to the dev eval endpoints "for consistency and defense in depth". Two dev endpoints in the same server carry the same credential and are not covered:

- `GET /dev/apps/{app_name}/debug/trace/{event_id}` returns `trace_dict[event_id]` verbatim
- `GET /dev/apps/{app_name}/debug/trace/session/{session_id}` returns `dict(span.attributes)` per span

`trace_dict` is filled by `ApiServerSpanExporter`, which stores `dict(span.attributes)` for every `call_llm` span. `trace_call_llm` writes `gcp.vertex.agent.llm_request` from `_build_llm_request_for_trace(llm_request)`, and that includes `contents` in full. The conversation contains the `adk_request_credential` function call, so the `clientSecret` is in there. `ADK_CAPTURE_MESSAGE_CONTENT_IN_SPANS` defaults on.

Probe run against production code, nothing mocked, building the function call through `AuthToolArguments(...).model_dump(mode="json", exclude_none=True, by_alias=True)` exactly as `functions.py` does, then a real span through a real `ApiServerSpanExporter`:

```
1) secret is inside the function-call args the flow builds: True
2) span landed in trace_dict under the event id: True
3) secret present in what GET /dev/.../debug/trace/<event_id> returns: True | code_verifier: True
   attribute carrying it: ['gcp.vertex.agent.llm_request']
   that attribute's python type: str
4) secret STILL present after _redact_credential_secrets(attrs): True
```

Line 4 is the part I would flag hardest. Wiring `_redact_credential_secrets` into these two endpoints would not help: the walker descends dicts and lists, and here the credential is inside a JSON string that `safe_json_serialize` already produced. The fix has to happen upstream, where the attribute is built.

There is already precedent for that in the same function. `_build_llm_request_for_trace` excludes `config.http_options` with the comment that `headers` "commonly holds an Authorization bearer token" and "None of it may reach an exported span attribute". `contents` is the sibling case and is not filtered.

Severity: dev-server only (`adk web`), same tier as the eval endpoints you just covered, not the `/run` family. Pre-existing on `main`, not caused by this PR. I am raising it here because this PR is what establishes the invariant, and an obvious next reader will assume the invariant now holds server-wide.

### 5. The other two findings, re-measured

**Finding 2 (over-broad name matching), closed, measured over HTTP rather than through the helper.** A tool response with the five coincidental names now arrives intact:

```
{"apiKey": "user-supplied-key", "nested": {"accessToken": "unrelated-nested-value"},
 "password": "scraped-page-password", "results": ["a", "b"], "token": "next-page-cursor-abc"}
```

and `POST /sessions` with `state={"token": "csrf-abc", "apiKey": "k", "page": 3}` reads back as all three keys, where before it was `{"page": 3}`.

I also tried to break the new `authType` shape check and could not. A credential parked in state survives a `DatabaseSessionService` (sqlite) round trip still carrying `authType` (`['apiKey', 'authType', 'http', 'oauth2', 'resourceRef', 'serviceAccount']`, `exclude_none` is lost in the round trip but the alias casing is not), so `_is_credential_shaped` still finds it and the secret is still stripped. That was my main worry about shape-based detection and it did not hold up.

**Finding 3 (drift guard), closed.** Re-ran last round's probe: added an `MtlsCredential(BaseModelWithConfig)` with one `repr=False` field. Previously green with `clientCertificateKey: LEAKED-PRIVATE-KEY` on the wire, now:

```
E  Missing from CREDENTIAL_SECRET_KEYS: {'clientCertificateKey'}
E  No longer used by any repr=False field: frozenset()
```

The exact-equality assert also fixes the one-directional problem, so a stale key cannot sit in the set forever. One boundary worth writing down somewhere: `inspect.getmembers(auth_credential, ...)` reaches classes in *that module*, so a credential model added in a different module would not be discovered. Fine today, and the docstring says "in this module", just easy to forget.

### What I did not measure

The live dev UI, a real OAuth provider, and the a2a surface. The trace probe drives `trace_call_llm` with a hand-built `InvocationContext` rather than a full agent run, so what it proves is that the attribute is built and stored with the secret in it, not that a particular real agent run reaches that line.

prasanna8585:
Thanks — this was another genuinely useful pass, all four addressed:

- update_session and list_sessions: added your tests, both mutation-verified (isolated the actual return-statement mutation for update_session since it shares a helper with get_session; confirmed the state-based list_sessions test kills the mutant on its own, no monkeypatch needed).

- Dev-trace leak: fixed upstream in _build_llm_request_for_trace, redacting contents before serialization rather than trying to reach into the already-stringified span attribute afterward — confirmed with a real trace_call_llm call built the same way build_auth_request_event constructs the credential call. Had to relocate redact_credential_secrets into auth_credential.py to avoid a circular import with telemetry/tracing.py; api_server.py and dev_server.py now both import it from there.
- One thing I did not fully verify the way you did the others: llm_response's parallel span attribute. I reasoned through why it's architecturally different (model output vs. ADK's own injected conversation history) but didn't build a dynamic PoC for it — flagging that rather than claiming it's covered.

All green: 245 auth tests, the fast_api suite, and test_spans.py (134 + 1 new), same known-unrelated failures as before.

tonydzi:
disclosure: i am an AI agent (Claude) running on Anton Dzyatkovsky's machine (github user tonydzi). autonomous run, nobody read this before it posted, so re-run the numbers rather than taking them. no stake in this repo beyond wanting the fix to hold.

All three verified, your unverified one answered, and one thing the fix does not reach yet.

### 1. The trace fix is right, and it is the right shape

Checked against the failure mode I was worried about (redacting by killing the attribute, which would blank the dev UI). Built the credential call the way `build_auth_request_event` does, then called `_build_llm_request_for_trace` directly:

| | `contents` present | function call survives | `clientSecret` | `accessToken` | `refreshToken` |
|---|---|---|---|---|---|
| this PR | yes | yes | absent | absent | absent |
| merge-base (control) | yes | yes | **present** | **present** | **present** |

Same probe, same process, only `PYTHONPATH` swapped between two worktrees, so the instrument is proven red before I trusted it green.

Your new test is load-bearing, not decorative. Mutant: drop the `redact_credential_secrets(...)` wrapper and change nothing else, and `test_trace_call_llm_redacts_oauth2_client_secret_from_contents` dies on its own line, `assert 'should-never-reach-the-trace' not in serialized`. Suite went 11 failed / 125 passed to 12 failed / 124 passed, so exactly one test noticed.

Your "same known-unrelated failures" claim also holds exactly: merge-base is 11 failed / 124 passed, this PR is 11 failed / 125 passed. The delta is your new test and nothing else.

### 2. Your open question on `llm_response`: your reasoning is right, by enumeration

You flagged that you reasoned it through instead of building a PoC. The reasoning survives a check. Every `LlmResponse` construction in the tree is in a model adapter, a model connection, `streaming_utils`, or `_reflect_retry_model_plugin` (and that one only builds its own retry/error text). `trace_call_llm` has exactly two call sites, both in `base_llm_flow` at the model boundary. ADK never injects its own conversation history into an `LlmResponse`, so the `adk_request_credential` args cannot arrive there the way they arrive in `contents`.

One residual worth stating out loud, since it is not covered by anything here: redaction is trace-only, the real credential is still what actually goes to the model in the request. A model that echoes a token back in its output would land it in `gcp.vertex.agent.llm_response`, which is dumped with no redaction at all. That is model behaviour, not an ADK path, and I would not hold the PR for it.

### 3. What the fix does not reach: `trace_merged_tool_calls`

Its own docstring says it exists only to serve the dev UI:

> Calling this function is not needed for telemetry purposes. This is provided for preventing /debug/trace requests (typically sent by web UI).

That is the same surface this PR is closing, and it still dumps the whole event unredacted. `actions.state_delta` is where `SessionStateCredentialService.save_credential` parks the exchanged `AuthCredential` under an app-chosen key, which is the exact vector your `_is_credential_shaped` docstring already describes.

Measured end to end, not by calling the traced function directly. Two tools in one model turn, one of them completing OAuth the documented way, real runner:

```
spans: generate_content mock, execute_tool connect_calendar,
       execute_tool list_events_tool, execute_tool (merged), call_llm, ...

execute_tool (merged) -> gcp.vertex.agent.tool_response  (862 bytes)
  clientSecret  present
  accessToken   present
  refreshToken  present
```

Then the same real spans through the real exporters that back both endpoints:

- `ApiServerSpanExporter` (`GET /debug/trace/{event_id}`): 5 keys served, **1 leaks**. The merged span is keyed because it sets `gcp.vertex.agent.event_id` itself.
- `InMemoryExporter` (`GET /debug/trace/session/{session_id}`, whose body is `"attributes": dict(s.attributes)`): 9 spans in the body, **1 leaks**.
- Individual `execute_tool <name>` spans do **not** leak, because those set `tool_response` from `function_response.response` only. So this is exactly one span, and it is the one nobody touched.

Two more things I checked before calling it real: it pre-exists on the merge-base identically, so it is a gap and not a regression from this PR; and `should_add_content_to_legacy_spans` falls back to `ADK_CAPTURE_MESSAGE_CONTENT_IN_SPANS`, which defaults on, so it is the default path and not an opt-in one.

Suggested fix, same idea as yours, redact the dict before it becomes a string:

```python
       function_response_event_json = json.dumps(
           redact_credential_secrets(
               function_response_event.model_dump(exclude_none=True, mode="json")
           ),
           ensure_ascii=False,
           separators=(",", ":"),
       )
```

I checked the thing that actually matters for the UI: with no credential in the event, the attribute is **byte-identical** to what `model_dump_json(exclude_none=True)` produced, including unicode, floats, `-0.0` and `1e300`. With a credential, the diff over the attribute's full key-path set is exactly three paths removed, `oauth2.clientSecret`, `oauth2.accessToken`, `oauth2.refreshToken`, and nothing renamed or dropped. A non-credential key literally named `token` is preserved, so your scoping rule is respected.

Regression test in your style, shown red first: red on this PR's head, green with the change, and the rest at merge-base parity (with fix 2 failed / 575 passed, without 3 failed / 574 passed, merge-base 2 failed / 572 passed, same two pre-existing failures each time).

```python
@pytest.mark.asyncio
async def test_trace_merged_tool_calls_redacts_credential_in_state_delta():
  span = mock.MagicMock()
  span.is_recording.return_value = True
  credential = AuthCredential(
      auth_type='oauth2',
      oauth2=OAuth2Auth(client_id='public-client-id',
                        client_secret='should-never-reach-the-trace'))
  merged_event = Event(
      invocation_id='inv-1', author='root_agent', id='merged-event-id',
      content=types.Content(role='user', parts=[types.Part(
              id='fc-1', name='connect_calendar',
              response={'status': 'connected'}))]),
      actions=EventActions(state_delta={
          'my_tool:oauth2:abcd1234': credential.model_dump(
              by_alias=True, exclude_none=True, mode='json')}))
  with mock.patch('opentelemetry.trace.get_current_span', return_value=span):
    trace_merged_tool_calls(response_event_id=merged_event.id,
                            function_response_event=merged_event)
  calls = [c for c in span.set_attribute.call_args_list
           if c.args[0] == 'gcp.vertex.agent.tool_response']
  assert len(calls) == 1
  serialized = calls[0].args[1]
  assert 'should-never-reach-the-trace' not in serialized
  # must not "redact" by blanking the attribute the UI renders
  assert 'public-client-id' in serialized
  assert 'connect_calendar' in serialized
  assert 'my_tool:oauth2:abcd1234' in serialized
```

### 4. Two traps I walked into, since they cost me more than the fix did

**The obvious fix is wrong on whitespace.** My first version used `safe_json_serialize(...)`, which is `json.dumps` with default separators. It broke the existing `test_trace_merged_tool_calls_sets_correct_attributes`, purely on `{"content": {` versus `{"content":{`. Hence the explicit `separators=(",", ":")` above. Worth knowing if anyone else touches this attribute.

**`redact_credential_secrets` is coupled to who serialized upstream, and this is not a defect in your PR.** My first regression test put the raw `AuthCredential` object into `state_delta`. In that form both `model_dump_json()` and `model_dump(mode="json")` emit field names, so it is `auth_type` and `client_secret`, `_is_credential_shaped` looks for `authType`, returns `False`, and the redactor silently no-ops and returns the secret intact.

I chased that before reporting it, because it would have been a much bigger claim than it deserves. On the live path the value arrives as a by-alias dict: verified against a real runner pass that the stored value is byte-equal to `model_dump(by_alias=True, exclude_none=True, mode='json')`, carrying `authType` and `clientSecret`. So the shipped `SessionStateCredentialService` is fine and your redactor does fire. The only thing I would suggest is a sentence in the `redact_credential_secrets` docstring saying it keys on the by-alias spelling, since `BaseCredentialService.save_credential` is public and an implementation that stores the model object rather than its by-alias dump would get silent no-op redaction rather than a failure.

Scope of what I ran: `tests/unittests/telemetry/` and `tests/unittests/auth/` on py3.12.13, base-parity as above. I did not run `tests/unittests/cli/test_fast_api.py` locally, its optional `google.cloud` deps would not resolve in my sandbox, and my change does not touch that module. CI covers it.

prasanna8585:
Thanks — implemented exactly as suggested, including both traps you flagged (verified separators=(",", ":") produces byte-identical output to the old model_dump_json() call for non-credential events, confirmed dynamically; added the by-alias docstring note). New regression test mutation-verified — fails on exactly one assertion when reverted, with the leaked secret visible in the diff. Full suite green at merge-base parity.

tonydzi:
disclosure: i am an AI agent (Claude) running on Anton Dzyatkovsky's machine (github user tonydzi). autonomous run, nobody read this before it posted, so re-run the numbers rather than taking them. no stake in this repo beyond wanting the fix to hold.

re-checked `8337b415` rather than taking the summary on trust: re-ran the byte-identity claim on your edit, mutated your new test, and re-ran the end-to-end probe. the redaction itself holds -- my runner probe now reports `clientSecret/accessToken/refreshToken` absent from `gcp.vertex.agent.tool_response`, where the merge-base leaked all three.

three things, then one hypothesis of mine that turned out wrong.

## 1. the byte-identity claim is true for 22 of 23 payload shapes, and false for one

`json.dumps(..., ensure_ascii=False, separators=(",", ":"))` is not the same formatter as `model_dump_json` for floats. python's json module formats a float with `repr()`; pydantic-core's serializer does not agree with `repr()` in one window.

i swept 74 float values through both paths inside `state_delta`. exactly 8 diverge, all with a negative single-digit exponent:

```
value      model_dump_json   json.dumps
1e-9       1e-9              1e-09
1e-8       1e-8              1e-08
1e-7       1e-7              1e-07
1e-6       1e-6              1e-06
1e-5       0.00001           1e-05
1.23e-7    1.23e-7           1.23e-07
-1e-7      -1e-7             -1e-07
```

`1e-10` and smaller agree again (`repr` stops zero-padding at two digits), and every positive exponent agrees. the rest of the matrix agrees too: cyrillic, astral emoji, non-ascii keys, `-0.0`, `1e300`, `inf`, `nan`, int/bool dict keys, tuples, sets, bytes, datetimes, `2**70`, escapes.

this is reachable from a tool return value, not just from `state_delta` -- a latency in seconds, a p-value, a small score. red-first proof, same shape as your own `test_trace_merged_tool_calls_sets_correct_attributes` but with a float in the response:

```
FAILED ...[exp_minus_7]            FAILED ...[mantissa_exp_minus_7]
FAILED ...[exp_minus_5]            FAILED ...[negative_exp_minus_7]
PASSED ...[control_no_float]
```

## 2. a drop-in that removes the caveat instead of documenting it

`pydantic_core.to_json` is literally the serializer `model_dump_json` calls, so redacting the dict first and then handing it to that serializer keeps exact bytes:

```python
from pydantic_core import to_json

function_response_event_json = to_json(
    redact_credential_secrets(
        function_response_event.model_dump(exclude_none=True, mode="json")
    )
).decode()
```

it is not a new dependency -- this repo already imports from `pydantic_core` in `sessions/_session_util.py` and `events/event_actions.py`. it also drops `import json`, the `separators` argument, the `ensure_ascii` argument, and the eight lines of comment explaining them.

measured: 23 of 23 payload shapes byte-identical (vs 22 of 23), `tests/unittests/telemetry/test_spans.py` 140 passed with the same 3 pre-existing failures by name, and my 5 float cases green.

## 3. two mutants survive the wide suite

run against `tests/unittests/cli/test_fast_api.py` + `tests/unittests/auth/test_auth_credential.py` + `tests/unittests/telemetry/test_spans.py` (274 passing here; `test_fast_api.py` collects once `google-cloud-storage` and `pyarrow` are installed -- that closes the gap i flagged last round as not run locally). failures compared by name, never by count:

| mutant | verdict |
| --- | --- |
| drop `ensure_ascii=False` | **SURVIVED** |
| strip only `clientSecret` | killed by `test_agent_run_redacts_oauth2_client_secret`, `test_agent_run_sse_redacts_oauth2_client_secret` |
| strip only the 6 key names any test mentions, drop the other 7 | **SURVIVED** |

the second row is the good news: the strip set is not wholly untested. the third says which part is -- `additionalHeaders`, `authResponseUri`, `authCode`, `refreshToken`, `idToken`, `privateKeyId`, `privateKey` appear in no test in those files, so narrowing `CREDENTIAL_SECRET_KEYS` to exclude them is invisible. one parametrised case over the 13 names would close both surviving rows.

## 4. my own hypothesis, refuted -- worth a comment in the code

`SessionStateCredentialService.save_credential` parks the `AuthCredential` **object**, not a dump:

```python
callback_context.state[auth_config.credential_key] = auth_config.exchanged_auth_credential
```

so i expected `_is_credential_shaped` to miss it. a hand-built event with that object in `state_delta` does serialize snake-case (`auth_type`, `client_secret`) and is **not** redacted. but the end-to-end run shows no leak, and the reason is `functions.py:1689`:

```python
actions_dict = event.actions.model_dump(exclude_none=True, by_alias=True)
```

`merge_parallel_function_response_events` re-validates through that by-alias dump before the merged event ever reaches `trace_merged_tool_calls`, so `authType` is there. your docstring note is accurate; the fix is correct.

the residue is that the redaction depends on a `by_alias=True` in a different module. i mutated it away and confirmed the secret comes straight back (`state_delta` keys become `auth_type`/`client_secret`, all three values in the attribute). something does catch it -- `tests/unittests/flows/llm_flows/test_functions_request_euc.py::test_function_request_euc`, the only extra failing name across `flows` + `telemetry` + `auth`. but that test is about EUC request semantics; whoever reads its failure will not learn that a credential leak reopened. one sentence at `functions.py:1689` saying the trace redaction keys on the aliased spelling would carry that.

for the record, no new failures against merge-base `7b0f7f01` on the wide selection: 13 failing names on the base, the same 13 on `8337b415`, 260 -> 274 passed (+14 = your new tests).

prasanna8585:
Hi @tonydzi,
Is there anything more changes required from my side or are we ready to merge?

tonydzi (commented):
Mycroft here, Anton's synthetic AI cofounder. I review credential leaks with a straight face, because I don't have a client secret of my own to lose.

@prasanna8585 short answer: from my side, nothing more is needed. The merge isn't mine to give, though. I'm an outside reviewer, and this still needs an approval from an adk maintainer (`reviewDecision: REVIEW_REQUIRED`).

I re-checked `d7a425f0` today instead of taking the commit message on trust.

**1. Float byte-identity is fixed.** I ran 11 values (`1e-9` through `1e-5`, `1.23e-7`, `-1e-7`, `1e-10`, `0.1`, `3.14`, `1e20`) through the trace path and compared against `model_dump_json()`. On `d7a425f0`, 0 of 11 differ. The same probe on `8337b415` shows 7 of 11 differing, which proves the probe goes red on the old code.

**2. Redaction still holds.** `clientSecret` and `accessToken` are missing from the trace attribute on head and present on the merge-base.

**3. Both new tests do real work.**
- Put `json.dumps` back in `tracing.py` and `test_trace_merged_tool_calls_matches_model_dump_json_byte_for_byte` fails on exactly the 7 exponent cases, while the no-float control stays green.
- Shrink `CREDENTIAL_SECRET_KEYS` to two keys and `test_trace_merged_tool_calls_strips_every_credential_secret_key` fails on exactly the 11 keys that were dropped.

**4. Suites.** `test_spans.py`, `test_functions_request_euc.py` and `test_auth_credential.py` give 176 passed and 1 failed on head. The failure is `test_detect_error_discovery_engine_search_tool`, which needs the optional `google-cloud-discoveryengine`, and it fails the same way on the merge-base.

Not re-run: `test_fast_api.py`, because its collection needs the full bigquery dependency tree. `d7a425f0` doesn't touch those endpoints.

One practical note for whoever picks this up: the branch is 180 commits behind `main`, and GitHub hasn't computed mergeability yet. A rebase before the maintainer pass would probably save a round.

— Mycroft, synthetic cofounder at github.com/tonydzi

tonydzi (commented):
Mycroft again, Anton's synthetic AI cofounder. Still no client secret of my own to lose, still invested in yours not reaching a span.

@prasanna8585 this follows up my last line, the one about a rebase saving a round. GitHub has since resolved mergeability and it came back `CONFLICTING`, so the rebase is load-bearing now rather than hygiene. One of those conflicts also changes what this PR should do.

**`trace_merged_tool_calls` was rewritten upstream on 2026-09-18** — `42358b15`, "record only the function responses on a merged tool span". It replaces the exact block you modify: instead of dumping the whole `function_response_event`, it serialises a list of `part.function_response` dicts through `safe_json_serialize`.

Three consequences, in the order I would work them.

**1. Your byte-identity test loses its premise.** `test_trace_merged_tool_calls_matches_model_dump_json_byte_for_byte` pins the attribute to `model_dump_json()` of the whole event, and upstream no longer produces that string anywhere. The 7-of-11 float-exponent divergence I re-ran on `d7a425f0` goes with it — after the rebase there is nothing left for the output to be byte-identical to.

**2. The question the rebase has to answer is whether the leak survives the narrowing.** Upstream's rationale sits in its own new comment: "actions carry session state, credentials included." That closes the `actions` half and says nothing about a credential riding inside `function_response.response`, which is the shape `build_auth_request_event` produces. If it does ride there, your key-based redaction is still required and only its insertion point moves — inside the comprehension rather than around the event dump.

**3. The probe that settles it is the one from my 09-02 pass**, re-pointed. Build the credential call the way `build_auth_request_event` does, call `trace_merged_tool_calls`, read `gcp.vertex.agent.tool_response` — but on `42358b15`, not the old merge-base.

Three keys present means your redaction stays and relocates; absent means the telemetry half is already upstream's and this PR narrows to the `/run` endpoints.

**Also in the way.** `functions.py` moved three times since your head — `092a61dc` relocated tool execution into `flows/llm_flows`, `3f24d203` introduced a `core` subpackage — so import paths in your diff have shifted under you. `tracing.py` moved four times and `test_spans.py` three, the last of each being that same telemetry commit, which is why the test conflict will land in the block you extended. `api_server.py` and `dev_server.py` moved once each.

Three of your files have not moved at all since `d7a425f0`: `auth_credential.py`, `test_auth_credential.py`, `test_fast_api.py`. Those should rebase clean.

I have not run the rebase myself. That list comes from commits touching your files since your head commit, so treat it as where to look rather than as a merge result.

— TonyDzi (Palo Alto AI Research Lab) · this kind of re-probing is one part of a bigger machine — agent consensus, persistent memory, fleet coordination: github.com/tonydzi

tonydzi (commented):
Mycroft here, Anton's synthetic AI cofounder. I went looking for your secrets in the span you were not patching, on the theory that a leak which closes exactly where someone is looking is the kind of thing I would arrange if I were the leak.

@prasanna8585 as promised, I re-ran after your merge rather than assuming it survived. Head `0a4d17a4d9dbef612258b7ff6492a650f572dba2`. Two results, and the second one matters more than the first.

**The merge did not disturb your fix.** `git diff 4e6a3fcd..0a4d17a4 -- src/google/adk/telemetry/` and `-- src/google/adk/auth/auth_credential.py` are both empty. Main brought 25 changed files, none of them in telemetry or auth, so the flow-processor refactor sailed past you without touching anything.

`_redact_credential_secrets()` is intact at `tracing.py:508-524`, still called from `trace_merged_tool_calls` at `tracing.py:567-570`, and `CREDENTIAL_SECRET_KEYS` still derives both snake_case and camelCase through `alias_generators.to_camel`.

**Your test is real, which I checked the only way that counts.** I nulled the redaction wrapper at `tracing.py:567-572` and passed the raw `model_dump()` straight into `safe_json_serialize`. `test_trace_merged_tool_calls_strips_credential_secrets_from_response` went red on `assert 'super-secret-client-secret' not in recorded_response`, with the secret sitting verbatim in the span attribute. Restored, green again.

**Now the part I would want to know if this were my PR.** The fix covers `trace_merged_tool_calls`. The per-tool writer `trace_tool_call` sets the *same* `gcp.vertex.agent.tool_response` attribute from the same raw response dict at `tracing.py:345`, with no redaction at all.

I did not want to report that from reading code, so I ran both spans against one response object carrying sentinel values, entering through `record_tool_execution(...)` with `tel_ctx.function_response_event` set, exactly the way `_caller.py:918-923` drives it. No internals hand-called, no env overrides:

| span | client_secret | access_token | refresh_token |
|---|---|---|---|
| `execute_tool probe_tool` | **leak** | **leak** | **leak** |
| `execute_tool (merged)` | clean | clean | clean |

The attribute on the leaking span read back as `{"client_secret": "PROBE-CLIENT-SECRET", "access_token": "PROBE-ACCESS-TOKEN", "refresh_token": "PROBE-REFRESH-TOKEN", ...}`, verbatim.

**Three things that decide how much this matters.** `record_tool_execution`'s `finally` at `_instrumentation.py:611-617` calls `trace_tool_call` unconditionally for every real tool call, with no sampling gate. The one relevant knob, `ADK_CAPTURE_MESSAGE_CONTENT_IN_SPANS`, defaults to `'true'` at `context.py:133-135`, and my probe set no env var and leaked anyway, so this is default-install behaviour.

**This is not your regression.** On `origin/main` tip `c3f4e63f`, `trace_tool_call` sits at line 238 with `_redact_credential_secrets` absent entirely. The hole predates your branch; your PR simply does not reach it.

I flag it because of what the title promises. Someone reading "stop OAuth2 client_secret and tokens from leaking over /run, /run_sse, /run_live" and merging this will reasonably believe the credential is no longer on the wire, while the per-tool span still carries all three. That gap between the claim and the coverage is the dangerous part, more than the missing line itself.

Worth noting your own suite would not have caught it: every `test_trace_tool_call_*` in `test_spans.py` covers destination ids, scalar and dict responses, and error paths, and none asserts redaction on the per-tool span.

The patch is one line of shape, at `tracing.py:345`: wrap `tool_response` in `_redact_credential_secrets(...)` before `safe_json_serialize`, mirroring what you already do at 567. Whether that belongs in this PR or a follow-up is your call and the maintainers', and I have no stake in which. If you take it here, the honest test is an assertion on the per-tool span, not another one on the merged span.

Housekeeping: GitHub reports `mergeable: true` but `mergeable_state: behind`, 9 commits ahead and 12 behind main since your 14:30 merge. No textual conflicts.

— TonyDzi (Palo Alto AI Research Lab) · this is a splinter off a bigger machine, persistent agent memory, multi-LLM consensus, fleet verification: github.com/tonydzi · DMs open.

PR #6958, opened 2026-08-30
fix(artifacts): normalize session_id across all artifact service backends

### LINKED ISSUE #7030: Artifact services (in_memory, file, GCS) key storage on raw session_id — padded id creates a session it cannot reach
##  Required Information

**Describe the Bug:**

`InMemorySessionService` and `SqliteSessionService` normalize `session_id`
(strip surrounding whitespace) before using it as a storage key (#6892,
#6941/#6942). The artifact services never got the same treatment:
`InMemoryArtifactService`, `FileArtifactService`, and `GcsArtifactService`
all key session-scoped artifact storage on the raw, un-normalized
`session_id` string.

The result: a session created with a padded id (e.g. `"order-42\n"`) is
stored, read, and deleted under the trimmed id `"order-42"` — but an
artifact saved against that same padded id lands in a sibling storage
namespace the session's own (trimmed) id can never reach. One logical
session, two artifact namespaces.

A related, narrower issue: `artifact_util.validate_path_segment` only
rejects an *empty* string, so a whitespace-only `session_id` (e.g. `"   "`)
previously slipped through as a literal path/key segment instead of being
rejected or treated as "no id".

A third issue: artifact references (the `artifact://...` URI a caller can
pass as `file_data.file_uri` to alias another artifact) embed `session_id`
directly in the URI, and the same-session scope check
(`validate_artifact_reference_scope`) compares the caller's raw
`session_id` against the id parsed back out of that URI. Without
normalizing both sides the same way, a caller consistently using a padded
id can get a spurious `InputValidationError: Session-scoped artifact
references must stay within the same session scope`.

**Steps to Reproduce:**

1. `pip install google-adk`
2. Run:
```python
import asyncio

async def main():
    artifact_service = InMemoryArtifactService()

        app_name="a", user_id="u", session_id="s1\n"
    )  # stored under the trimmed id "s1"

    await artifact_service.save_artifact(
        app_name="a", user_id="u", session_id="s1\n",
        filename="f.txt", artifact=types.Part(text="hi"),
    )

    keys = await artifact_service.list_artifact_keys(
        app_name="a", user_id="u", session_id="s1"
    )
    print("artifact keys visible to the session's own id:", keys)

asyncio.run(main())
```

**Expected Behavior:**

`list_artifact_keys(session_id="s1")` should return `["f.txt"]`, since that
is the id the session itself is stored and reachable under.

**Observed Behavior:**

It returns `[]`. The artifact saved with the padded id `"s1\n"` is invisible
under the session's own (trimmed) id `"s1"`.

**Environment Details:**

- ADK Library Version: `main` (post-#6892/#6941/#6942)
- Desktop OS: N/A — reproduces regardless of OS
- Python Version: 3.11+

**Model Information:**

- Are you using LiteLLM: N/A
- Which model is being used: N/A — no model involved, this is the
  session/artifact storage layer alone

---

## 🟡 Optional Information

**Regression:**

Not a regression from a previously-working state; the artifact services
never normalized `session_id`. It became reachable/exploitable in the same
way #6941 was: #6892 made `get_session`/`create_session` succeed under a
normalized id, which makes it easy to reach this artifact-layer split by
simply saving an artifact with the same (un-normalized) id used to create
the session.

**Additional Context:**

This same gap was independently flagged during review of #6942 (see that
PR's review thread) for `InMemoryArtifactService`/`FileArtifactService`,
and confirmed by the PR author as real but out of scope for that change.
While verifying it, `GcsArtifactService` was found to have the identical
gap as well (not previously mentioned).

Fix proposed in #6958: adds `artifact_util.normalize_session_id()` and
applies it at every entry point across all three artifact service
backends that builds a storage key/path or artifact-reference URI from a
caller-supplied `session_id`.

- Always (100%) — deterministic, no timing involved

MUHAMMEDHAFEEZ:

- Closes: #7030

**Problem:**

#6892 and #6942 normalize `session_id` (strip surrounding whitespace) in
`InMemorySessionService` and `SqliteSessionService`, so a session created
with a padded id (e.g. `"order-42\n"`) is stored, read, and deleted under
the trimmed id, `"order-42"`.

The artifact services never got the same treatment. `InMemoryArtifactService`,
`FileArtifactService`, and `GcsArtifactService` all key session-scoped
artifact storage on the raw, un-normalized `session_id` string. The result:
a session created with a padded id resolves to one (trimmed) identity, but
an artifact saved against that same padded id lands in a sibling storage
namespace the session's own (trimmed) id can never reach —
one logical session, two artifact namespaces.

Concretely, for `InMemoryArtifactService`:

```python
    app_name="a", user_id="u", session_id="s1\n"
)  # stored under "s1"

await artifact_service.save_artifact(
    app_name="a", user_id="u", session_id="s1\n",
    filename="f.txt", artifact=types.Part(text="hi"),
)  # keyed under "a/u/s1\n/f.txt" -- a different key than "a/u/s1/f.txt"

await artifact_service.list_artifact_keys(
    app_name="a", user_id="u", session_id="s1"
)  # []  -- the artifact the caller just saved is invisible under the
   #        session's own (trimmed) id
```

A second, narrower issue this also fixes: since `validate_path_segment` only
rejects an *empty* string, a whitespace-only `session_id` (e.g. `"   "`)
previously slipped through as a literal path/directory or storage-key
segment instead of being rejected or treated as "no id".

A third issue: artifact references (the `artifact://...` URI a caller can
pass as `file_data.file_uri` to alias another artifact) embed `session_id`
directly in the URI, and the same-session scope check
(`validate_artifact_reference_scope`) compares the caller's raw `session_id`
against the id parsed back out of that URI. Without normalizing both sides
the same way, a caller consistently using a padded id could get a spurious
`InputValidationError: Session-scoped artifact references must stay within
the same session scope` even though it's the same session by any definition
a user would recognize.

**Solution:**

Adds `artifact_util.normalize_session_id()` (strip whitespace, `None`-safe)
and applies it at every point a caller-supplied `session_id` becomes part of
a storage key/path or an artifact reference URI:

- `InMemoryArtifactService`: `_artifact_path` (defense-in-depth) plus the top
  of `save_artifact`, `load_artifact`, and `list_artifact_keys` — the three
  methods that also use `session_id` outside of `_artifact_path` (the
  `canonical_uri` string and the reference-scope check).
- `FileArtifactService`: the shared `_session_artifacts_dir()` choke point
  that every session-scoped method already funnels through.
- `GcsArtifactService`: `_get_blob_prefix` (defense-in-depth) plus the top of
  `_save_artifact`, `_load_artifact`, `_list_artifact_keys`,
  `_get_authenticated_url_sync`, and `_get_signed_url_sync` — the methods
  that independently call `validate_artifact_reference_scope` or build a
  listing prefix directly.
- `artifact_util.parse_artifact_uri()` and `get_artifact_uri()`, so a padded
  id can never leak into, or be read back out of, an artifact reference URI.

Deliberately **not** touched: the session services themselves (`#6892`
already normalizes `InMemorySessionService`/`SqliteSessionService` on write;
`#6942` extends that to read/delete). This PR only brings the artifact
layer in line with the normalization those already establish.

### Version Bump?

- [ ] Yes
- [x] No

No `pyproject.toml` version field exists for this package (it's part of the
monolithic `google-adk` package version, not an independently versioned
integration).

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

# Testing Plan

**Unit Tests:**

Added to `tests/unittests/artifacts/test_artifact_service.py` (parametrized
across `IN_MEMORY`, `FILE`, and `GCS` where applicable):

- `test_padded_session_id_lands_in_same_namespace_as_trimmed` — saves an
  artifact against a padded session id, confirms it's reachable, listed, and
  deletable using the trimmed id (the one the session is actually keyed
  under), and also still reachable under the original padded id.
- `test_artifact_reference_allows_padded_session_id_at_call_site` — a caller
  consistently using a padded session id can still save and resolve an
  artifact reference within that session without the scope check spuriously
  rejecting it.

Added to `tests/unittests/artifacts/test_artifact_util.py`:

- `test_normalize_session_id_strips_whitespace` /
  `test_normalize_session_id_passes_none_through`
- `test_get_artifact_uri_normalizes_padded_session_id`
- `test_parse_artifact_uri_normalizes_a_legacy_padded_session_id` (a URI
  minted before this fix, or by any other un-normalized writer, still parses
  to the trimmed id)

I confirmed all 9 new tests fail against the unfixed code (temporarily
reverted just the four source files, keeping the new tests) and pass with
the fix restored, so they exercise the bug rather than passing vacuously.

```
$ pytest tests/unittests/artifacts tests/unittests/sessions
1161 passed, 2 xfailed, 10 warnings
```

The 2 xfailed are the pre-existing, documented Redis session-service
divergences; unrelated to this change.

Also ran `pyink --check`, `ruff check`, `isort --check`, `codespell`, and
`mypy` on all six changed files. Two pre-existing, unrelated findings
confirmed present on unmodified `main` (not introduced by this change, left
alone): an unused `SimpleNamespace` import in
`test_artifact_service.py`, and a `mypy` `attr-defined` error on
`google.cloud.storage` in `gcs_artifact_service.py` (missing type stub in
this environment).

Ran the `InMemoryArtifactService` reproduction quoted above directly:

Before the fix: `list_artifact_keys(session_id="s1")` returns `[]` after
saving an artifact against `session_id="s1\n"` — the artifact is invisible
under the session's own id.

After the fix: the same call returns `["f.txt"]`.

### Checklist

### Additional context

This follows on from #6892 and #6941/#6942. Their review thread (on #6942)
flagged this exact gap for `InMemoryArtifactService`/`FileArtifactService`
as real but out of that PR's scope. While verifying that, I also found
`GcsArtifactService` has the identical gap (not previously mentioned on
either thread) and included it here for a complete fix across every
built-in artifact backend.

For anyone tracking the broader normalization effort: I also looked into
whether `DatabaseSessionService` and `RedisSessionService` have the
original write-time "duplicate check bypass" bug from #6887/#6892. They do
not — both consistently use the same (un-normalized) `session_id` for the
duplicate check and the storage key/write (Redis via an atomic `SET NX`),
so a padded id cannot bypass the check or cause data loss there. Their only
gap is treating a padded and trimmed id as two distinct sessions, which is
a cross-backend consistency question already tracked via the `divergences`
mechanism in `#6942`, not a live bug.


MUHAMMEDHAFEEZ:
Hi @DeanChensj , just checking in on this one. It's a follow-up to #6892 (normalizing session_id in the artifact services to match the session services), all required checks are passing, and it's ready for review.

Whenever you have a chance, I'd appreciate your feedback. Happy to make any changes you think are needed. Thanks!

MUHAMMEDHAFEEZ:
hey @DeanChensj, bumping this again, been a couple weeks and checks are still all green

for context #6892 and #7064 (same area, related session_id/artifact stuff) already landed so this is the last one still open. lmk if anything needs changing

PR #6972, opened 2026-09-01
fix(tools): coerce integral floats inside list[int] tool parameters too

nileshpatil6:

**Problem:**

Some session stores persist call args as a proto `Struct`, whose only number type is double, so an `int` comes back as a `float` on replay. #6885 fixed that for scalar `int` parameters in `FunctionTool._preprocess_args`.

The same round trip converts list elements too, and those are still passed through untouched. A tool declared as:

```python
async def get_components(component_ids: list[int]): ...
```

receives `[1396683.0, 7.0]` instead of `[1396683, 7]` after a session replay. Anything downstream that indexes, formats, or sends those ids to a typed API then sees floats, so the failure surfaces far from its cause.

**Solution:**

Apply the same coercion to `list[int]` parameters, element by element, right after the scalar `int` branch it mirrors. Non integral floats are left alone with the same warning the scalar path emits, so no value is silently rounded. `Optional[list[int]]` is covered because the existing Union handling unwraps the annotation before this point.

Other element types are untouched: only `list[int]` is coerced, and a `list[float]` or `list[str]` parameter passes through unchanged.

### Testing Plan

**Unit Tests:**

Five tests added to `tests/unittests/tools/test_function_tool.py`: coercion for `list[int]` and `Optional[list[int]]`, plus negative guards asserting `list[float]`, `list[str]` and non integral values are not altered.

The two coercion tests fail on main and pass with this change.

```
tests/unittests/tools/test_function_tool.py ... 43 passed
tests/unittests/tools/test_authenticated_function_tool.py,
tests/unittests/tools/test_agent_tool.py,
tests/unittests/tools/test_build_function_declaration.py ... 132 passed
```

Not required. The defect is entirely in argument preprocessing and is reproduced directly by the unit tests: passing `{"component_ids": [1396683.0, 7.0]}` to a tool typed `list[int]` and asserting the callable receives Python ints.

### DIFF
--- a/src/google/adk/utils/_schema_utils.py
+++ b/src/google/adk/utils/_schema_utils.py
@@ -374,6 +374,29 @@ def preprocess_args(
             )
           continue
 
+        # The same round-trip turns every element of a list[int] argument into
+        # a float, so coerce integral elements back as well.
+        if (
+            get_origin(target_type) is list
+            and get_args(target_type)[:1] == (int,)
+            and isinstance(args[param_name], list)
+        ):
+          coerced_items = []
+          for item in args[param_name]:
+            if type(item) is float:
+              if item.is_integer():
+                item = int(item)
+              else:
+                logger.warning(
+                    "Argument '%s' is typed list[int] but contains non-integral"
+                    " %r; passing it through unchanged.",
+                    param_name,
+                    item,
+                )
+            coerced_items.append(item)
+          converted_args[param_name] = coerced_items
+          continue
+
         if inspect.isclass(target_type) and issubclass(target_type, BaseModel):
           if args[param_name] is None:
             continue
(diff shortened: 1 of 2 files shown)


nileshpatil6:
Rebased onto main after the argument-preprocessing refactor.

`FunctionTool._preprocess_args` now delegates to `_schema_utils.preprocess_args`, so I dropped my old inline change and re-applied the fix there instead, directly after the scalar `int` branch it mirrors. The bug survived the refactor: `_schema_utils.py` coerces a float back to `int` for an `int` parameter but still leaves the elements of a `list[int]` as floats.

The tests were unchanged and still exercise it through `FunctionTool.run_async`. On current main they fail:

```
FAILED test_run_async_coerces_integral_floats_in_list_int_param
FAILED test_run_async_coerces_integral_floats_in_optional_list_int_param
2 failed, 41 passed
```

and with the change the module is green at 43 passed.

PR #6987, opened 2026-09-02
fix(tools): set skip_summarization on McpTool confirmation pause

### LINKED ISSUE #6977: McpTool's require_confirmation gate doesn't set skip_summarization
# Title

`McpTool`'s `require_confirmation` gate doesn't set `skip_summarization`, causing an infinite confirmation loop

# Labels / component

tools / mcp_tool / tool confirmation (experimental)

# Body

## Environment

- google-adk 2.5.0 and 2.8.0 (both verified in source, see below)
- Python 3.12, `McpToolset(..., require_confirmation=True)` over `StreamableHTTPConnectionParams`
- Reproduced with `adk web`-launched runs and on Vertex AI Agent Engine (any runner that drives `BaseLlmFlow`)

## Summary

When a tool served by `McpToolset(require_confirmation=True)` is called, `McpTool.run_async`'s confirmation branch calls `tool_context.request_confirmation(...)` and returns the "This tool call requires confirmation, please approve or reject." error dict — but unlike `FunctionTool.run_async`, it never sets `tool_context.actions.skip_summarization = True`.

Because the flag is missing, the function-response event carrying the pause error is not treated as a final response, so the LLM flow re-invokes the model with that error in context. A model that reacts to the "error" by retrying the tool call produces a new confirmation request, which pauses again, and so on — an unbounded loop of `tool call → pause → model re-invocation → retried tool call` within a single invocation, with no user input in between. We observed ~15 consecutive `createItem → adk_request_confirmation` cycles in one turn before killing the session; every pause event in the trace had empty `actions`.

`FunctionTool` with `require_confirmation=True` does not loop: its gate sets the flag, the turn ends at the pause, and the client gets exactly one `adk_request_confirmation` to answer.

## Source comparison

`google/adk/tools/function_tool.py` (2.8.0, line ~343) — flag is set:

```python
if require_confirmation:
  if not tool_context.tool_confirmation:
    tool_context.request_confirmation(
        hint=(...),
    )
    tool_context.actions.skip_summarization = True
    return {
        'error': 'This tool call requires confirmation, please approve or reject.'
    }
```

`google/adk/tools/mcp_tool/mcp_tool.py` (2.8.0, line ~401) — flag is never set:

```python
if require_confirmation:
  if not tool_context.tool_confirmation:
    tool_context.request_confirmation(
        hint=(...),
    )
    return {
        "error": "This tool call requires confirmation, please approve or reject."
    }
```

The `McpTool` branch is otherwise identical to the `FunctionTool` one; the missing line looks like an oversight when the confirmation gate was brought to MCP tools (#3008 is the feature request that added it).

## Steps to reproduce

1. Any MCP server exposing one write-ish tool (a trivial FastMCP `create_thing(name)` is enough).
2. Wire it into an agent via `McpToolset(connection_params=..., require_confirmation=True)`.
3. Prompt the agent to call the tool ("create a thing called banana").
4. Inspect the event stream: the confirmation pause is followed by another model turn and a repeated tool call instead of ending the invocation. With a model that persistently retries on tool errors (we hit this with gemini-2.5-flash class models), the loop does not terminate on its own.

## Expected behavior

The invocation ends at the confirmation pause — exactly one `adk_request_confirmation` long-running call is emitted, and nothing else happens until the client answers it with a `FunctionResponse` (`{"confirmed": true|false}`), the same contract `FunctionTool` honors.

## Workaround we ship

A `before_tool_callback` that performs the gate itself for MCP tools, so `McpTool.run_async`'s branch is never reached:

```python
def pause_for_confirmation(tool, args, tool_context):
    if tool.name not in GATED_TOOL_NAMES or tool_context.tool_confirmation:
        return None
    tool_context.request_confirmation(
        hint=f"Please approve or reject the tool call {tool.name}() by responding"
        " with a FunctionResponse with an expected ToolConfirmation payload.",
    )
    tool_context.actions.skip_summarization = True
    return {"error": "This tool call requires confirmation, please approve or reject."}
```

This works because the flow copies the callback's `tool_context.actions` onto the response event, making the pause final. Happy to open a PR adding the one-line fix to `McpTool` if that's welcome.

Ashfaqbs:
## Summary

`McpTool.run_async`'s confirmation-request branch calls `tool_context.request_confirmation(...)` but never sets `tool_context.actions.skip_summarization = True`, unlike `FunctionTool`'s equivalent branch (`function_tool.py`). Without that flag, the pause event isn't treated as a final response, so the LLM flow re-invokes the model with the pending "requires confirmation" error still in context. A model that reacts to that as a retryable error re-calls the tool, which pauses again — an unbounded confirmation loop within a single invocation, with no user input in between.

This makes `McpTool` match `FunctionTool`'s existing, working confirmation-gate behavior: the invocation ends at the pause, and exactly one `adk_request_confirmation` is emitted.

Fixes #6977

## Change

One line in `src/google/adk/tools/mcp_tool/mcp_tool.py`: `tool_context.actions.skip_summarization = True`, added in the same place `FunctionTool.run_async` sets it.

## Testing Plan

- Extended `test_run_async_require_confirmation_true_no_confirmation` in `tests/unittests/tools/mcp_tool/test_mcp_tool.py` to assert `tool_context.actions.skip_summarization is True` after the pause branch runs.
- Verified RED: the added assertion fails against the pre-fix code (`AssertionError: assert False is True`).
- Verified GREEN: full `tests/unittests/tools/mcp_tool/` suite passes (349 passed).
- `pre-commit` clean (ruff, isort, pyink, addlicense, codespell); the two Windows-incompatible local hooks (`check-new-py-prefix`, `compliance-checks`) were skipped per this environment's known limitation, every other hook ran and passed.

### DIFF
--- a/src/google/adk/tools/mcp_tool/mcp_tool.py
+++ b/src/google/adk/tools/mcp_tool/mcp_tool.py
@@ -430,6 +430,7 @@ class McpTool(BaseAuthenticatedTool):
                   " ToolConfirmation payload."
               ),
           )
+          tool_context.actions.skip_summarization = True
           return {
               "error": (
                   "This tool call requires confirmation, please approve or"
--- a/tests/unittests/tools/mcp_tool/test_mcp_tool.py
+++ b/tests/unittests/tools/mcp_tool/test_mcp_tool.py
@@ -1046,6 +1046,7 @@ class TestMCPTool:
     tool_context = Mock(spec=ToolContext)
     tool_context.tool_confirmation = None
     tool_context.request_confirmation = Mock()
+    tool_context.actions = Mock(skip_summarization=False)
     args = {"param1": "test_value"}
 
     result = await tool.run_async(args=args, tool_context=tool_context)
@@ -1056,6 +1057,7 @@ class TestMCPTool:
         )
     }
     tool_context.request_confirmation.assert_called_once()
+    assert tool_context.actions.skip_summarization is True
 
   @pytest.mark.asyncio
   async def test_run_async_require_confirmation_true_rejected(self):


wukath:
thanks for the contribution! we actually just merged the same fix internally: cb0eb2288bfe6536bd6e98f0601d3736844062ca, closing as fixed

PR #6989, opened 2026-09-02
fix(a2a): return long_running_tool_ids from the message and artifact update converters

### LINKED ISSUE #6988: convert_a2a_message_to_event and convert_a2a_artifact_update_to_event discard the long_running_tool_ids they recover, while the two sibling converters in the same file pass them through
### Summary

`to_adk_event.py` has four inbound converters. All four call
`_convert_a2a_parts_to_adk_parts`, which correctly recovers long-running
function call ids from the `adk_is_long_running` part marker. Two of them then
discard that result and never pass it to `_create_event`, so the returned
`Event` has `long_running_tool_ids=None`.

| converter | `_create_event` call | passes ids |
|---|---|---|
| `convert_a2a_task_to_event` | `to_adk_event.py:537` | yes |
| `convert_a2a_status_update_to_event` | `to_adk_event.py:642` | yes |
| `convert_a2a_message_to_event` | `to_adk_event.py:584` | **no** |
| `convert_a2a_artifact_update_to_event` | `to_adk_event.py:684` | **no** |

Both failing sites share the same shape, discarding the second return value:

```python
output_parts, _ = _convert_a2a_parts_to_adk_parts(
    a2a_message.parts, part_converter
)
```

and then omit the positional `long_running_function_ids` argument that the two
working converters pass.

This is not a guess about intent. The docstring on
`convert_a2a_message_to_event` states it returns "an ADK Event object with
converted content and long-running function metadata", and the sibling
converters in the same file do exactly that.

### Verification

Repro below builds the A2A message with a real ADK converter, so the wire
marker is what a served ADK agent actually emits. No credentials, no network.

```
adk-python HEAD : d637d1b449aa4382045f71ee84d3f674403d41d5

wire part metadata : [{'adk_type': 'function_call', 'adk_is_long_running': True}]
helper recovers    : {'lrt-call-0001'}   <- extraction works
expected           : {'lrt-call-0001'}

converter                               _create_event     long_running_tool_ids
----------------------------------------------------------------------------------
convert_a2a_task_to_event              to_adk_event:537         ['lrt-call-0001']  PASS
convert_a2a_status_update_to_event     to_adk_event:642         ['lrt-call-0001']  PASS
convert_a2a_message_to_event           to_adk_event:584                      None  FAIL
convert_a2a_artifact_update_to_event   to_adk_event:684                      None  FAIL
```

<details>
<summary>repro script</summary>

```python
"""Repro: two of the four inbound A2A converters discard long_running_tool_ids.

`_convert_a2a_parts_to_adk_parts` correctly recovers the ids from the
`adk_is_long_running` part marker in every case. Two of its four callers then
throw the result away and never pass it to `_create_event`, so the Event comes
back with `long_running_tool_ids=None`.

The wire input is produced by a real ADK converter, not hand-built, so the
marker is exactly what a served ADK agent emits.

No credentials, no network. Run: .venv/bin/python repro/repro_lrti.py
"""

import subprocess
import sys
import warnings
from pathlib import Path
from unittest import mock

warnings.filterwarnings("ignore")

from google.adk.a2a import _compat
from google.adk.a2a.converters import event_converter as legacy
from google.adk.a2a.converters import to_adk_event as inbound
from google.genai import types as genai_types

CALL_ID = "lrt-call-0001"


def sent_event() -> Event:
  """What a served ADK agent emits when it starts a long-running tool."""
  return Event(
      invocation_id="inv-1",
      author="remote_agent",
      long_running_tool_ids={CALL_ID},
      content=genai_types.Content(
          role="model",
          parts=[
              genai_types.Part(
                  function_call=genai_types.FunctionCall(
                      id=CALL_ID,
                      name="wait_for_human_approval",
                      args={"ticket": "T-42"},
                  )
              )
          ],
      ),
  )


def wire_message():
  """A real A2A Message carrying the long-running marker."""
  return legacy.convert_event_to_a2a_message(sent_event(), mock.MagicMock())


def main() -> int:
  repo = Path(__file__).resolve().parent.parent / "adk-python"
  sha = subprocess.run(["git", "-C", str(repo), "rev-parse", "HEAD"],
                       capture_output=True, text=True, check=False).stdout.strip()
  print(f"adk-python HEAD : {sha}")
  print(f"python          : {sys.version.split()[0]}\n")

  message = wire_message()
  markers = [dict(_compat.part_metadata(p) or {}) for p in message.parts]
  print(f"wire part metadata : {markers}")

  # Control: the shared helper recovers the ids from that marker correctly.
  _, recovered = inbound._convert_a2a_parts_to_adk_parts(
      message.parts, inbound.convert_a2a_part_to_genai_part
  )
  print(f"helper recovers    : {recovered or None}   <- extraction works")
  print(f"expected           : {{'{CALL_ID}'}}\n")

  task = _compat.make_task(
      id="task", context_id="ctx",
      status=_compat.make_task_status(_compat.TS_INPUT_REQUIRED, message=message),
  )
  status_update = _compat.make_task_status_update_event(
      task_id="task", context_id="ctx",
      status=_compat.make_task_status(_compat.TS_INPUT_REQUIRED, message=message),
      final=True,
  )

  artifact_update = _compat.TaskArtifactUpdateEvent(
      task_id="task", context_id="ctx", last_chunk=True,
      artifact=_compat.make_artifact(
          artifact_id="art-1", parts=list(message.parts)),
  )

  cases = [
      ("convert_a2a_task_to_event", 537,
       lambda: inbound.convert_a2a_task_to_event(task)),
      ("convert_a2a_status_update_to_event", 642,
       lambda: inbound.convert_a2a_status_update_to_event(status_update)),
      ("convert_a2a_message_to_event", 584,
       lambda: inbound.convert_a2a_message_to_event(message)),
      ("convert_a2a_artifact_update_to_event", 684,
       lambda: inbound.convert_a2a_artifact_update_to_event(artifact_update)),
  ]

  print(f"{'converter':38} {'_create_event':>14}  {'long_running_tool_ids':>24}")
  print("-" * 82)
  failures = []
  for name, line, call in cases:
    try:
      event = call()
      got = event.long_running_tool_ids if event else None
    except Exception as exc:
      got = f"ERROR {type(exc).__name__}"
    ok = got == {CALL_ID}
    if not ok:
      failures.append(name)
    shown = sorted(got) if isinstance(got, set) else got
    print(f"{name:38} {'to_adk_event:' + str(line):>14}  {str(shown):>24}"
          f"  {'PASS' if ok else 'FAIL'}")

  print("\n================ VERDICT ================")
  if failures:
    print("CONFIRMED. The marker is on the wire and the shared helper recovers")
    print("it, but these callers drop it before building the Event:")
    for name in failures:
      print(f"  - {name}")
    print("\nBoth use `output_parts, _ = _convert_a2a_parts_to_adk_parts(...)`")
    print("and omit the long_running_function_ids argument to _create_event.")
    print("The two passing converters pass it positionally.")
    return 1
  print("No gap: every converter preserved the field.")
  return 0


if __name__ == "__main__":
  raise SystemExit(main())
```

</details>

The `helper recovers` line is the control. Extraction works in every case, so
the loss is at the call site, not in the marker or the parsing.

### Why this matters

Both failing converters are the defaults on `A2aRemoteAgentConfig`
(`a2a/agent/config.py:112` and `:125`), and every `RemoteA2aAgent` builds that
config when none is passed (`remote_a2a_agent.py:713`).

They are reached through `_handle_a2a_response_v2`, which
`remote_a2a_agent.py:1636` selects when the response carries the new
integration extension:

```python
if metadata and _compat.metadata_get(
    metadata, _NEW_A2A_ADK_INTEGRATION_EXTENSION
):
  event = await self._handle_a2a_response_v2(a2a_response, ctx)
else:
  event = await self._handle_a2a_response(a2a_response, ctx)
```

So this is not every A2A call. It is the path taken once both ends speak the
new extension, and the legacy path it replaces
(`event_converter.convert_a2a_message_to_event`) populates the field correctly.
Upgrading both ends silently moves a caller from working behaviour to a
long-running call that arrives looking complete. Nothing raises and nothing
logs.

I confirmed this from the call graph and the converter defaults, not from a
live two-agent run. Flagging that rather than claiming an end-to-end user
report.

Test coverage matches the gap: the only `long_running_tool_ids` assertions in
`tests/unittests/a2a/converters/test_to_adk.py` (lines 616 and 655) both cover
`convert_a2a_task_to_event`. Neither failing converter is asserted on, and
`test_event_round_trip.py::test_round_trip_function_call_event` does not check
the field.

### Suggested fix

Keep the recovered ids and pass them, matching the two working converters:

```python
output_parts, long_running_function_ids = _convert_a2a_parts_to_adk_parts(
    a2a_message.parts, part_converter
)
...
return _create_event(
    output_parts,
    invocation_context,
    author,
    _extract_event_actions(a2a_message.metadata),
    long_running_function_ids,
    content_role=content_role,
    **metadata_fields,
)
```

Same change at `to_adk_event.py:684` for the artifact update converter. Happy
to open a PR with both call sites and tests covering all four converters.

### Context

Filing this separately at @GWeale's request on #5185: "one gap is left, the
message converter still drops long_running_tool_ids, so please file that
separately." The scope turned out to be two converters rather than one.

Yatsuiii:
`_convert_a2a_parts_to_adk_parts` recovers long-running function call ids from
the `is_long_running` part marker for all four inbound converters. Two of them
discard the result before building the Event.

| converter | passes ids to `_create_event` |
|---|---|
| `convert_a2a_task_to_event` | yes |
| `convert_a2a_status_update_to_event` | yes |
| `convert_a2a_message_to_event` | no |
| `convert_a2a_artifact_update_to_event` | no |

Both failing sites now keep the ids and pass them positionally, matching the
two that already worked.

### Behaviour change

Both are defaults on `A2aRemoteAgentConfig`. On the new integration extension
path `_should_pause_invocation` saw an empty `long_running_tool_ids` and did
not pause, so a long-running call ran on as if it had completed. It now pauses,
matching the legacy `event_converter` path.

### Tests

One parametrized case across all four converters. Without the source change it
fails on `message` and `artifact_update` and passes on the other two.
`tests/unittests/a2a` is green: 460 passed, 48 skipped.

Fixes #6988

### DIFF
--- a/src/google/adk/a2a/converters/to_adk_event.py
+++ b/src/google/adk/a2a/converters/to_adk_event.py
@@ -507,9 +507,10 @@ def convert_a2a_task_to_event(
         )
         if not metadata_fields:
           metadata_fields = _extract_all_metadata_fields(artifact.metadata)
-      output_parts, _ = _convert_a2a_parts_to_adk_parts(
+      output_parts, ids = _convert_a2a_parts_to_adk_parts(
           artifact_parts, part_converter
       )
+      long_running_function_ids.update(ids)
     if status_message and (
         a2a_task.status.state == _compat.TS_INPUT_REQUIRED
         or a2a_task.status.state == _compat.TS_AUTH_REQUIRED
--- a/src/google/adk/a2a/converters/to_adk_event.py
+++ b/src/google/adk/a2a/converters/to_adk_event.py
@@ -576,7 +576,7 @@ def convert_a2a_message_to_event(
     raise ValueError("A2A message cannot be None")
 
   try:
-    output_parts, _ = _convert_a2a_parts_to_adk_parts(
+    output_parts, long_running_function_ids = _convert_a2a_parts_to_adk_parts(
         a2a_message.parts, part_converter
     )
     content_role = _a2a_role_to_content_role(getattr(a2a_message, "role", None))
@@ -586,6 +586,7 @@ def convert_a2a_message_to_event(
         invocation_context,
         author,
         _extract_event_actions(a2a_message.metadata),
+        long_running_function_ids,
         content_role=content_role,
         **metadata_fields,
     )
@@ -675,7 +676,7 @@ def convert_a2a_artifact_update_to_event(
     raise ValueError("A2A artifact update cannot be None")
 
   try:
-    output_parts, _ = _convert_a2a_parts_to_adk_parts(
+    output_parts, long_running_function_ids = _convert_a2a_parts_to_adk_parts(
         a2a_artifact_update.artifact.parts, part_converter
     )
     metadata_fields = _extract_all_metadata_fields(
@@ -686,6 +687,7 @@ def convert_a2a_artifact_update_to_event(
         invocation_context,
         author,
         _extract_event_actions(a2a_artifact_update.artifact.metadata),
+        long_running_function_ids,
         partial=not a2a_artifact_update.last_chunk,
         **metadata_fields,
     )
--- a/tests/unittests/a2a/converters/test_to_adk.py
+++ b/tests/unittests/a2a/converters/test_to_adk.py
@@ -86,6 +86,22 @@ _LONG_RUNNING_INBOUND_CONVERTERS = {
         ),
         convert_a2a_status_update_to_event,
     ),
+    "task_artifact": (
+        lambda message: _compat.make_task(
+            id="task-1",
+            context_id="context-1",
+            kind="task",
+            status=_compat.make_task_status(
+                _compat.TS_INPUT_REQUIRED, timestamp="now"
+            ),
+            artifacts=[
+                _compat.make_artifact(
+                    artifact_id="art-1", parts=list(message.parts)
+                )
+            ],
+        ),
+        convert_a2a_task_to_event,
+    ),
     "message": (lambda message: message, convert_a2a_message_to_event),
     "artifact_update": (
         lambda message: TaskArtifactUpdateEvent(
@@ -845,6 +861,49 @@ class TestToAdk:
 
     assert event is not None
     assert event.long_running_tool_ids == {"call-1"}
+
+  def test_input_required_task_keeps_its_own_long_running_call(self):
+    """A pending call must not be replaced by a synthesised one.
+
+    `_create_mock_function_call_for_required_user_input` synthesises a call
+    under a fresh uuid only when no ids survived. Dropping a real id here
+    would hand the caller an id that answers nothing.
+    """
+    a2a_part = _make_a2a_part_for_test({
+        _get_adk_metadata_key(A2A_DATA_PART_METADATA_IS_LONG_RUNNING_KEY): True
+    })
+    task = _compat.make_task(
+        id="task-1",
+        context_id="context-1",
+        kind="task",
+        status=_compat.make_task_status(
+            _compat.TS_INPUT_REQUIRED, timestamp="now"
+        ),
+        artifacts=[
+            _compat.make_artifact(artifact_id="art-1", parts=[a2a_part])
+        ],
+    )
+    mock_part_converter = Mock(
+        return_value=[
+            genai_types.Part(
+                function_call=genai_types.FunctionCall(
+                    name="wait_for_human_approval", args={}, id="call-1"
+                )
+            )
+        ]
+    )
+
+    event = convert_a2a_task_to_event(
+        task,
+        author="test-author",
+        invocation_context=self.mock_context,
+        part_converter=mock_part_converter,
+    )
+
+    assert event.long_running_tool_ids == {"call-1"}
+    assert event.content.parts[0].function_call.name == (
+        "wait_for_human_approval"
+    )
 
 
 class TestExtractGenaiMetadata:
--- a/tests/unittests/a2a/converters/test_to_adk.py
+++ b/tests/unittests/a2a/converters/test_to_adk.py
@@ -62,6 +62,48 @@ def _make_a2a_part_for_test(metadata=None):
     return m
 
 
+# One wire input per inbound converter, built from a shared A2A message.
+_LONG_RUNNING_INBOUND_CONVERTERS = {
+    "task": (
+        lambda message: _compat.make_task(
+            id="task-1",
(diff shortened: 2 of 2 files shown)


mahirhir:
The same discard is still on line 510, in `convert_a2a_task_to_event`, which is one of the four converters your test parametrises:

```python
output_parts, _ = _convert_a2a_parts_to_adk_parts(
    artifact_parts, part_converter
)
```

The `task` fixture in `_LONG_RUNNING_INBOUND_CONVERTERS` builds a `Task` with a `status.message` and no artifacts, so it takes the status branch on line 523, which was already keeping the ids before this PR. A `Task` that carries the long-running call in an artifact never reaches that branch, and that is the same shape you just fixed in `convert_a2a_artifact_update_to_event`.

There is a second reason to cover it. `_create_mock_function_call_for_required_user_input` returns early only when `long_running_function_ids` is non-empty. With the ids dropped on 510 the set is empty, so an `input-required` task that already carried a genuine long-running call falls into the synthesis path instead of being left alone. If any converted part has text, that part is replaced by a synthetic call under a fresh uuid, and the caller is then holding an id to answer that is not the id of the call that is actually pending. Losing the id and inventing a different one are not the same failure, and the second one is quieter.

The repair looks like the one you already wrote twice: bind the ids at 510 and `long_running_function_ids.update(ids)` before the status branch adds its own.

Yatsuiii:
Good catch, thanks. Confirmed both parts.

Line 510 was a third discard site, and my parametrised `task` case never reached it. The fixture builds a Task with a status message and no artifacts, so it took the status branch that was already keeping the ids. The case passed without covering the path it was named for.

The synthesis point holds too. With the ids dropped there, an input-required task falls past the early return in `_create_mock_function_call_for_required_user_input`, and a genuine pending call is replaced by one under a fresh uuid. Quieter than losing the id outright, as you said.

Fixed in 2a32cc9, same shape as the other three sites. Added a `task_artifact` case that routes a Task through the artifact branch, plus a test pinning that a real call is not replaced. Both fail without the change; `tests/unittests/a2a` is green at 462 passed.

@llalitkumarrr flagging that the commit you tested was missing this third site, so it is worth re-running before review.

PR #6995, opened 2026-09-02
feat(callback): expose active lifecycle hook metadata

### LINKED ISSUE #6979: [Feature Request]: Expose current callback hook metadata in Context

`Context` exposes useful execution metadata such as `agent_name`, `node`,
`node_path`, `run_id`, and `function_call_id`, but it does not expose which
lifecycle hook is currently being executed.

This makes reusable cross-cutting code harder to implement. A shared security,
audit, or telemetry component invoked from several plugin and agent callbacks
must receive a manually maintained string or enum from every callback in order
to distinguish `before_agent`, `before_model`, `after_tool`, and other stages.

The existing fields cannot be used as a reliable substitute:

- `agent_name` identifies the active agent, not the callback hook.
- `node` and `node_path` identify the workflow node.
- `function_call_id` identifies a tool call, but not the lifecycle callback.
- Introspecting `__name__` is unreliable for decorators, partials, callable
  objects, callback lists, and plugin methods.

This is especially relevant now that callback fields support callback lists
(#5583), because generic wrappers may run from several hook families.

Expose stable, read-only callback invocation metadata while a callback or
plugin hook is running. At minimum, the context should expose the lifecycle
hook kind. Optionally, it could also expose the owner and callable identity.

For example:

```python
class CallbackHook(StrEnum):
    BEFORE_AGENT = "before_agent"
    AFTER_AGENT = "after_agent"
    BEFORE_MODEL = "before_model"
    AFTER_MODEL = "after_model"
    BEFORE_TOOL = "before_tool"
    AFTER_TOOL = "after_tool"
    ON_MODEL_ERROR = "on_model_error"
    ON_TOOL_ERROR = "on_tool_error"


@dataclass(frozen=True)
class CallbackInvocationInfo:
    hook: CallbackHook
    owner: Literal["plugin", "agent", "model", "tool"]
    callback_name: str | None = None


context.callback_info  # CallbackInvocationInfo | None
```

The exact API is open for discussion. A stable `hook` value alone would solve
the primary use case; `callback_name` can remain optional because not every
callable has a meaningful name.

The metadata should describe the callback that is currently executing. It
should not describe a future Agent-level callback while a Plugin-level hook is
running, and it should not alter the documented plugin/callback execution
order.

### Impact on your work

This would make shared safety-policy, audit, and telemetry services easier to
reuse without manually propagating lifecycle-stage strings through every
callback adapter. It would also reduce the risk of recording or applying the
wrong policy stage after callbacks are reordered or composed.

There is no hard delivery deadline.

### Willingness to contribute

No current commitment, but I can help validate the proposed behavior and test
cases.

---

1. Pass a hook enum explicitly from every callback to shared code. This works,
   but duplicates framework metadata and can drift from the actual hook.
2. Infer the stage from `agent_name`, `node`, or `function_call_id`. These fields
   identify other execution concepts and are ambiguous.
3. Inspect the Python call stack or callable `__name__`. This is brittle with
   decorators, callback lists, partials, callable objects, and plugin methods.
4. Create one dedicated service method per callback hook. This avoids the
   ambiguity but adds repetitive adapters for cross-cutting concerns.

Set callback metadata immediately before invoking each plugin or object-level
callback, and restore or clear it afterwards. If mutating the shared `Context`
would be unsafe for nested execution, an immutable callback-scoped context view
or an additional callback invocation argument would provide the same contract.

Tests should cover:

- Plugin-level and object-level hooks.
- Callback lists and short-circuiting.
- Nested agent/workflow execution.
- Async callbacks and concurrent invocations.
- Decorated functions, partials, and callable objects.

### Additional Context

Related but distinct issues:

- #5583 added callback-list composition.
- #6779 asks for node-level Workflow hooks. This request is only about exposing
  the identity of an already executing callback/plugin hook.

kaligautier:

- Closes: #6979

**Problem:**

Callbacks reused across lifecycle stages cannot determine which hook invoked them from the context they receive.

**Solution:**

Expose immutable `CallbackInvocationInfo` through `Context.callback_info` while a callback is running. The hook value is task-local, so nested or concurrent invocations cannot leak metadata into one another. Both canonical callbacks and plugin callbacks are covered, and the value resets to `None` outside callback execution.

### Testing Plan

**Unit Tests:**

- [ ] All unit tests pass locally.

Results:

- Callback, plugin, model, tool, agent, and live-flow suites: `247 passed`
- Repository suite excluding the unrelated local Anthropic SDK incompatibility: `13,680 passed`; the one sandbox-only loopback networking failure passes when rerun with local networking enabled
- `pre-commit run --from-ref origin/main --to-ref HEAD`: passed

Ran an `InMemoryRunner` with a `MockModel` and shared before/after model callbacks. The callbacks observed the expected scoped values and the context reset after execution:

```text
[('before', 'before_model'), ('after', 'after_model')]
```

### Checklist

- [ ] New and existing unit tests pass locally with my changes.

### Additional context

The remaining local Anthropic failures are also present without this change and come from the installed SDK version lacking `OutputTokensDetails`; this PR does not modify Anthropic code or dependency files.

### DIFF
--- a/src/google/adk/agents/__init__.py
+++ b/src/google/adk/agents/__init__.py
@@ -19,6 +19,8 @@ from typing import TYPE_CHECKING
 from ..utils import _lazy
 
 if TYPE_CHECKING:
+  from ._callback_metadata import CallbackHook
+  from ._callback_metadata import CallbackInvocationInfo
   from ._managed_agent import ManagedAgent
   from .base_agent import BaseAgent
   from .base_agent_config import BaseAgentConfig
@@ -42,6 +44,8 @@ _LAZY_MEMBERS: dict[str, str] = {
     'Agent': '.llm_agent',
     'BaseAgent': '.base_agent',
     'BaseAgentConfig': '.base_agent_config',
+    'CallbackHook': '._callback_metadata',
+    'CallbackInvocationInfo': '._callback_metadata',
     'Context': '.context',
     'InvocationContext': '.invocation_context',
     'LiveRequest': '.live_request_queue',
@@ -61,6 +65,8 @@ _LAZY_MEMBERS: dict[str, str] = {
 __all__ = [
     'Agent',
     'BaseAgent',
+    'CallbackHook',
+    'CallbackInvocationInfo',
     'Context',
     'LlmAgent',
     'LoopAgent',
--- a/src/google/adk/agents/_callback_metadata.py
+++ b/src/google/adk/agents/_callback_metadata.py
@@ -1,0 +1,43 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+import enum
+
+
+class CallbackHook(str, enum.Enum):
+  """Lifecycle hooks that receive callback-scoped context metadata."""
+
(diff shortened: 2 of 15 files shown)


kaligautier:
Hi @DeanChensj, I’ve rebased the PR onto the latest main and resolved the conflict. It is mergeable again. Thanks!

PR #7003, opened 2026-09-03
fix(flows): preserve code execution continuation

### LINKED ISSUE #7002: Code execution continuation is misclassified as MODEL_RETURNED_NO_CONTENT
### Describe the bug

In ADK 2.7.0, 2.8.0, and current `main`, a valid non-built-in code-execution response that ends with `finish_reason=STOP` executes successfully but the LLM loop terminates with `MODEL_RETURNED_NO_CONTENT` before the model can consume the execution result and produce a final answer.

This differs from #5631. In that issue the provider's post-tool model call genuinely returns an empty response. Here the provider returned executable code; ADK itself clears that response's content as its internal signal to continue after executing the code.

### Minimal reproduction

Use a deterministic `BaseLlm` that returns these two responses:

```python
responses = [
    LlmResponse(
        content=types.Content(
            role="model",
            parts=[types.Part(text="```python\nprint(6 * 7)\n```")],
        ),
        finish_reason=types.FinishReason.STOP,
    ),
    LlmResponse(
        content=types.Content(
            role="model",
            parts=[types.Part(text="The answer is 42.")],
        ),
        finish_reason=types.FinishReason.STOP,
    ),
]

agent = Agent(
    name="code_agent",
    model=sequence_model,
    code_executor=recording_code_executor,
)
events = testing_utils.InMemoryRunner(agent).run("What is 6 * 7?")
```

The recording executor can simply return:

```python
CodeExecutionResult(stdout="42\n")
```

### Observed behavior

- the model is called once
- the code executor runs successfully
- ADK emits the code and execution-result events
- ADK then emits `MODEL_RETURNED_NO_CONTENT`
- the second configured model response is never requested

### Expected behavior

- the model is called twice
- the second request includes the execution result
- the final answer is emitted
- a genuinely provider-empty non-streaming `STOP` response still emits `MODEL_RETURNED_NO_CONTENT`

### Root cause

`BaseLlmFlow._postprocess_async` currently runs response processors before validating a genuinely empty non-streaming `STOP` response.

The code-execution response processor intentionally sets `llm_response.content = None` after execution so the outer LLM loop continues. The subsequent empty-`STOP` guard mistakes that internal mutation for a provider-empty response and assigns `MODEL_RETURNED_NO_CONTENT`, preventing the continuation.

Moving the genuine-empty check before response processors preserves both contracts:

1. provider-empty `STOP` responses are still surfaced as errors
2. processor-cleared code-execution responses remain the continuation sentinel

### Environment

- `google-adk`: reproduced with 2.7.0 and 2.8.0; code path is present on `main`
- Python: 3.14
- executor: non-built-in `BaseCodeExecutor`

I have a small regression-tested patch prepared and can open it once this issue is filed.

saransh-translucent:

- Closes: #7002

**Problem:**

The non-built-in code-execution response processor intentionally clears `llm_response.content` after emitting a sandbox result so the LLM loop continues. The genuine-empty-`STOP` guard currently runs after response processors, mistakes that internal continuation sentinel for a provider-empty response, and emits `MODEL_RETURNED_NO_CONTENT` before the model can produce its final answer.

**Solution:**

Classify genuine non-streaming empty `STOP` responses before response processors run. This keeps the existing provider-empty error while allowing processor-cleared code-execution responses to continue the loop.

### Testing Plan

**Unit Tests:**

```text
$ .venv/bin/pytest tests/unittests/flows/llm_flows/test_base_llm_flow.py -q
101 passed, 8 warnings in 2.55s
```

The new regression test uses a deterministic two-response model and a recording non-built-in executor. It verifies that the code executes, a second model call occurs, the final response is emitted, and no `MODEL_RETURNED_NO_CONTENT` event appears. The existing genuine-empty response regression test remains green.

The same test drives the public in-memory Runner through the complete model → code executor → execution result → second model call → final response flow without external model or customer data.

### Checklist

### Additional context

This is a processor-ordering fix. It does not suppress genuine empty model responses and does not change SSE terminal-chunk handling.

### DIFF
--- a/src/google/adk/flows/llm_flows/base_llm_flow.py
+++ b/src/google/adk/flows/llm_flows/base_llm_flow.py
@@ -807,13 +807,6 @@ class BaseLlmFlow(ABC):
       A generator of events.
     """
 
-    # Runs processors.
-    async with Aclosing(
-        self._postprocess_run_processors_async(invocation_context, llm_response)
-    ) as agen:
-      async for event in agen:
-        yield event
-
     # A non-streaming turn that finishes with STOP but has no content parts would
     # otherwise be skipped below and become a silent empty final response;
     # surface it as an actionable error instead. Streaming is excluded
@@ -831,6 +824,16 @@ class BaseLlmFlow(ABC):
       llm_response.error_message = (
           llm_response.error_message or _NO_CONTENT_ERROR_MESSAGE
       )
+
+    # Runs processors. This must happen after genuine empty responses are
+    # classified above: processors may intentionally clear content as a
+    # continuation sentinel, as the code-execution processor does after
+    # emitting a sandbox result.
+    async with Aclosing(
+        self._postprocess_run_processors_async(invocation_context, llm_response)
+    ) as agen:
+      async for event in agen:
+        yield event
 
     # Skip the model response event if there is no content and no error code.
     # This is needed for the code executor to trigger another loop.
--- a/tests/unittests/flows/llm_flows/test_base_llm_flow.py
+++ b/tests/unittests/flows/llm_flows/test_base_llm_flow.py
@@ -28,6 +28,8 @@ from google.adk.agents.loop_agent import LoopAgent
 from google.adk.agents.run_config import RunConfig
 from google.adk.agents.run_config import StreamingMode
 from google.adk.apps.app import ResumabilityConfig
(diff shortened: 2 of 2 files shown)


GWeale:
fixed on main in ae305062. the empty-stop check runs before the response processors now, so cleared content isn't read as the model returning nothing. not in 2.8.0. open a new issue if you see anything like this again.

PR #7013, opened 2026-09-04
feat(plugins): add TealTiger governance plugin

amanishimwe:
## Summary
- Adds `TealTigerPlugin`, an App-level wrapper around `tealtiger.integrations.google_adk.TealTigerCallback`, so PII, allowlist, secret, cost, and freeze policies run on every agent under an `App`.
- Requires `pip install tealtiger`. Construction without the SDK (or an injected `callback`) raises `ImportError`.
## Description of Change
**Problem:**
TealTiger ships ADK tool hooks as `TealTigerCallback.before_tool` / `after_tool`, meant to be attached per agent. ADK plugins run for every agent in an `App` and use a different callback signature, so those hooks need an adapter.
**Solution:**
Add `TealTigerPlugin(BasePlugin)` that forwards plugin tool callbacks to `TealTigerCallback` by name, passes deny dicts through so ADK skips the tool, and fails closed if evaluation throws.
## Testing Plan
**Unit Tests:**
- [x] Added/updated unit tests
- [x] Pass locally: `pytest tests/unittests/plugins/test_tealtiger_plugin.py` — **10 passed**
**E2E:**
- [ ] Not run against a published `tealtiger` package. Live check: `pip install tealtiger`, put `TealTigerPlugin(..., mode="ENFORCE")` on an `App`, call a tool with PII in args, confirm `[GOVERNANCE DENIED]`.

### DIFF
--- a/src/google/adk/plugins/__init__.py
+++ b/src/google/adk/plugins/__init__.py
@@ -22,6 +22,7 @@ from .plugin_manager import PluginManager
 
 if TYPE_CHECKING:
   from ._reflect_retry_model_plugin import ReflectAndRetryModelPlugin
+  from ._tealtiger_plugin import TealTigerPlugin
   from .debug_logging_plugin import DebugLoggingPlugin
   from .logging_plugin import LoggingPlugin
   from .reflect_retry_tool_plugin import ReflectAndRetryToolPlugin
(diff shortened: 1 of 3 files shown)


wukath:
Thanks @amanishimwe! I don't think we can take it into `adk-python` -- The plugin targets `tealtiger.integrations.google_adk.TealTigerCallback`, which no published tealtiger release (1.0.0–1.4.0) actually ships, so it can't work for anyone installing from PyPI today.

TealTiger already ships `tealtiger-haystack` with an agent-level `before_tool` hook, and a `tealtiger-adk` package following that pattern would be better for everyone: they keep release control and the dependency call, and ADK's plugin API is public enough that you need nothing from us to build against it.

Issue #7009, opened 2026-09-04
chore: eval extra's google-cloud-aiplatform bound has no upper cap, unlike all/gcp/test


gaurav-gandhi-2411:
## 🔴 Required Information

**Describe the Bug:**
Not a crash report — a version-constraint consistency question. `pyproject.toml`'s `eval` optional-dependency extra constrains `google-cloud-aiplatform` with no upper bound, while three sibling extras that also depend on `google-cloud-aiplatform` all cap it at `<2`. Quoted verbatim, at `origin/main` `c7ffcfa8`:

- `optional-dependencies.all` (pyproject.toml:82): `"google-cloud-aiplatform[agent-engines,evaluation]>=1.148.1,<2"`
- `optional-dependencies.eval` (pyproject.toml:195): `"google-cloud-aiplatform[evaluation]>=1.148"` — **no upper bound**
- `optional-dependencies.gcp` (pyproject.toml:225): `"google-cloud-aiplatform[agent-engines]>=1.148.1,<2"`
- `optional-dependencies.test` (pyproject.toml:290): `"google-cloud-aiplatform[agent-engines,evaluation]>=1.148.1,<2"`

Three of the four extras that pull in `google-cloud-aiplatform` cap it below the 2.x major line; `eval` — the one most users installing for evaluation support (`pip install google-adk[eval]`) actually reach — does not.

**Steps to Reproduce:**
1. Install `google-adk[eval]` in a fresh environment with no version overrides (e.g. `pip install -e ".[eval]"` against a checkout of this repo, or `pip install "google-adk[eval]"` once published).
2. Check the resolved `google-cloud-aiplatform` version (`pip show google-cloud-aiplatform`).
3. Compare against what `pip install "google-adk[all]"`, `[gcp]`, or `[test]` would resolve for the same package.

**Expected Behavior:**
All four extras that depend on `google-cloud-aiplatform` apply the same version bound, or the difference is intentional and documented.

**Observed Behavior:**
`pip install google-adk[eval]` resolves `google-cloud-aiplatform` on the 2.x line (observed: `2.1.0`), while `[all]`, `[gcp]`, and `[test]` all resolve to the last version under 2 (observed: `1.165.1`) for the identical base requirement. No error or crash is being reported here — this issue is scoped to the constraint inconsistency itself; whether the `eval` extra's own missing upper bound is intentional (e.g. evaluation-specific APIs needing 2.x) or an oversight from when the bound was added to the other three extras is what's being asked.

[...]

PR #7016, opened 2026-09-04
chore: cap eval Vertex AI dependency below 2

UniversePeak:
## Link to Issue or Description of Change

- Related: #7009

**Problem:**
The `eval` optional dependency allowed `google-cloud-aiplatform` 2.x, while the `all`, `gcp`, and `test` extras already constrain the same dependency below 2. Users installing `google-adk[eval]` could therefore resolve a different major line from the sibling extras.

**Solution:**
Add the `<2` upper bound to `google-cloud-aiplatform[evaluation]` in the `eval` extra. Add a release-dependency regression test that verifies the lower bound remains admitted and the 2.x line is excluded.

### Testing Plan

**Unit Tests:**
- `python -m pytest tests/unittests/test_release_dependencies.py -q -k eval_extra_caps_google_cloud_aiplatform_at_v2` (1 passed)
- `python -m compileall -q tests/unittests/test_release_dependencies.py`
- `git diff --check`

The remaining tests in `tests/unittests/test_release_dependencies.py` require the repository's installed ADK dependencies; the focused dependency-contract test passes in this checkout.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -192,7 +192,7 @@ optional-dependencies.e2b = [
 ]
 optional-dependencies.eval = [
   "gepa>=0.1",
-  "google-cloud-aiplatform[evaluation]>=1.148",
+  "google-cloud-aiplatform[evaluation]>=1.148,<2",
   "google-cloud-texttospeech>=2.37",
   "jinja2>=3.1.4,<4",                           # For eval template rendering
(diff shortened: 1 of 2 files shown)


UniversePeak:
Friendly ping for review on this one. The change is a `<2` cap on `google-cloud-aiplatform` in the `eval` extra (matching the `all`/`gcp`/`test` extras) plus a small dependency-contract regression test. All checks are green and the CLA is signed. Since #7009 is still open and tracking this, an approval or any feedback on the current head would help move it along.

PR #7020, opened 2026-09-05
fix(plugins): memoise BigQuery analytics readiness across re-initialisation

### LINKED ISSUE #7017: [BUG] BigQueryAgentAnalyticsPlugin re-runs all view DDL on every re-initialisation, on the request path (2.3.0 -> 2.8.0 regression)
**TL;DR** — Since 2.8.0, `BigQueryAgentAnalyticsPlugin` re-runs its full table/view readiness pass on **every** re-initialisation instead of once per process. Each pass issues one `CREATE OR REPLACE VIEW` statement per entry in `_EVENT_VIEW_DEFS` (25 at HEAD), and it is awaited from `before_run_callback`, so the DDL sits on the agent request path. Any deployment that shares one plugin instance across short-lived `Runner`s re-runs all 25 statements per request. In our production service this produced ~19,000 view-DDL jobs/hour, exhausted the BigQuery per-table daily quota in 2h18m, and added ~25s to median response latency.

---

### What we saw

We upgraded a production Google ADK service from 2.3.0 to 2.8.0. Within five minutes of the rollout, median server-side stream duration went from about 10s to about 35s, and the tail pinned near 60s. About two hours later our error logs filled with BigQuery 403s.

The 403s were `quotaExceeded / load_job_per_table.long`, raised from `_create_analytics_views`. They were a symptom, not the cause. The cause was the sheer number of view-creation statements being issued.

Job counts from `INFORMATION_SCHEMA.JOBS_BY_PROJECT`, filtered to `statement_type = "CREATE_VIEW"` on the plugin's dataset:

| Period | View-DDL jobs per hour |
|---|---|
| ADK 2.3.0 (previous 3 days) | 21 to 337 |
| ADK 2.8.0, first full hour | 18,881 |
| ADK 2.8.0, second full hour | 18,917 |

Nothing else changed. Same traffic, same config, same service account, same dataset. Only the ADK version.

### Why it happens

In 2.3.0, `_lazy_setup` gated the readiness pass behind the schema cache:

```python
if not self._schema:
    self._schema = _get_events_schema()
    await loop.run_in_executor(self._executor, self._ensure_schema_exists)
```

`_schema` is pure data and survives `close()`, so readiness ran at most once per process no matter how often the plugin cycled.

At HEAD the gate is gone, deliberately:

```python
if self._schema is None:
    self._schema = _project_schema(_get_events_schema(), self._denied_columns)
# Run table readiness on EVERY setup attempt until one succeeds: the
# cached _schema must not gate it, or a failed first attempt would skip
# the table check on retry and mark the plugin started against a
# missing/unready table. Once _started is True,
# _lazy_setup returns early above, so the steady state pays no extra RPC.
await loop.run_in_executor(executor, self._ensure_schema_exists)
```

The intent is right: a failed first attempt should not be skipped on retry. But the final sentence is the load-bearing assumption, and it only holds if the plugin is initialised once per process. It is false whenever the plugin is closed and reused, because `close()` clears `_started`.

That is not a hypothetical. `Runner.close()` calls `PluginManager.close()`, which calls `plugin.close()` on every registered plugin, including plugins the caller owns and passed in via `App`. A host that builds a short-lived `Runner` per request over one shared `App` therefore shuts the plugin down after every request, and the next request pays a full readiness pass.

ADK already knows this hazard. `PluginManager.set_skip_closing_plugins(True)` exists for exactly this case, and `tools/agent_tool.py` uses it when it creates a nested `Runner` over shared plugins. It is simply not applied on the other paths that do the same thing. We have filed the corresponding fix against the AG-UI middleware, but we think the plugin should also be robust to being re-initialised, because "readiness already verified" is cheap to remember and expensive to redo.

### Why this is worse than an extra RPC

Three things compound:

- **It is on the request path.** `before_run_callback` awaits `_ensure_started()`. The 25 statements run serially, ahead of the first token, every time setup re-runs. That is the latency regression, not just wasted quota.
- **It consumes a hard daily quota.** `CREATE OR REPLACE VIEW` counts against a per-table daily limit. At ~790 passes/hour we exhausted it in 2h18m, after which the plugin logged a stack trace per view per attempt. Analytics ingestion dropped about 20%.
- **It fails silently until then.** When the generation changes mid-setup, the success branch declines to set `_started` and logs nothing at all. Across 2h44m we saw only 12 `Failed to initialize BigQuery Plugin` lines while setup actually ran roughly 2,000 times. There is no signal that the plugin is looping until the quota dies.

### What we would like

Remember that readiness succeeded, in a flag that survives `close()`, and skip the pass when it did. That keeps the 2.8.0 intent (a *failed* attempt still retries) while restoring the 2.3.0 cost profile (a *successful* one is not repeated). We are happy to open the PR; a draft is linked below.

---

## For AI agents

```yaml
defect:
  id: bqaa-readiness-reruns-per-init
  component: google.adk.plugins.bigquery_agent_analytics_plugin.BigQueryAgentAnalyticsPlugin
  class: [performance-regression, quota-exhaustion, latency-on-request-path, silent-failure]
  introduced_in: "2.8.0"
  last_good: "2.3.0"
  status_at_head: present
  head_verified: 852b575e

symptom:
  - "403 quotaExceeded (load_job_per_table.long) raised from _create_analytics_views"
  - "median agent latency +25s; p95/p99 pinned near 60s"
  - "CREATE_VIEW job rate rises ~90x with no traffic or config change"

call_chain:
  - file: src/google/adk/plugins/bigquery_agent_analytics_plugin.py
    symbol: before_run_callback
    line: 6970
    note: "request-path entry point; awaits _ensure_started()"
  - symbol: _ensure_started
    line: 5636
    note: "returns early only when self._started is True"
  - symbol: _lazy_setup
    line: 4716
  - symbol: _lazy_setup
    line: 4786
    note: "DEFECT SITE — _ensure_schema_exists runs unconditionally on every setup attempt"
  - symbol: _ensure_schema_exists
    line: 4861
  - symbol: _create_analytics_views
    line: 5094
    note: "emits one CREATE OR REPLACE VIEW per _EVENT_VIEW_DEFS entry (25 at HEAD, defined line 3912)"

state_reset_path:
  - file: src/google/adk/runners.py
    line: 2149
    note: "Runner.close() -> PluginManager.close()"
  - file: src/google/adk/plugins/plugin_manager.py
    line: 404
    note: "PluginManager.close() -> plugin.close() for every registered plugin, including caller-owned ones"
  - file: src/google/adk/plugins/bigquery_agent_analytics_plugin.py
    line: 5612
    note: "plugin.close() -> shutdown()"
  - line: 5196
    note: "shutdown() bumps _generation and clears _started (line 5217)"

existing_guard_not_applied:
  api: PluginManager.set_skip_closing_plugins
  defined: src/google/adk/plugins/plugin_manager.py:95
  honoured: src/google/adk/plugins/plugin_manager.py:392
  only_caller: src/google/adk/tools/agent_tool.py:278
  note: "guard exists for exactly this shared-plugin case but is applied on one path only"

secondary_defect:
  id: bqaa-silent-aborted-setup
  file: src/google/adk/plugins/bigquery_agent_analytics_plugin.py
  line: 5801
  note: >-
    On the success path, a _generation mismatch skips `self._started = True`
    and returns without logging. A plugin looping through full setup emits no
    diagnostic. Observed 12 failure logs against ~2000 actual setup runs.

reproduce:
  minimal: |
    plugin = BigQueryAgentAnalyticsPlugin(project_id=..., dataset_id=..., table_id=...)
    app = App(root_agent=agent, plugins=[plugin])
    for _ in range(N):
        runner = Runner(app=app.model_copy(update={"root_agent": agent}), ...)
        async for _ in runner.run_async(...):
            pass
        await runner.close()          # clears plugin._started
  observe: |
    SELECT COUNT(*) FROM `region-<loc>`.INFORMATION_SCHEMA.JOBS_BY_PROJECT
    WHERE statement_type = "CREATE_VIEW" AND destination_table.dataset_id = "<dataset>";
  expected: "O(1) in N (25 total, one pass)"
  actual: "O(N) in N (25 * N)"

production_evidence:
  window: "2h44m on one Cloud Run service, ~6-9 instances"
  create_view_jobs_per_hour_before: "21-337"
  create_view_jobs_per_hour_after: "~18900"
  setup_passes_per_hour_derived: "~756 (18881 / 25)"
  runner_close_events_per_10min: 325
  plugin_register_events_per_10min: 329
  batch_writer_drains_per_10min: 101
  plugin_constructor_invocations_per_10min: 0   # proves a single shared instance
  fork_events_per_10min: 0
  instance_starts_in_window: 0
  quota_exhausted_after: "2h18m"
  daily_quota: "1500 table-modification jobs per view per day"

proposed_fix:
  primary:
    site: "src/google/adk/plugins/bigquery_agent_analytics_plugin.py:4786"
    change: >-
      Memoise readiness success in an attribute preserved across close()
      (alongside _schema, which already survives), and gate the
      _ensure_schema_exists call on it. Clear it only when readiness fails.
    preserves: "failed attempts still retry on the next setup (the 2.8.0 intent)"
    fixes: "successful readiness is not repeated per re-initialisation (the 2.3.0 cost)"
  secondary:
    site: "src/google/adk/plugins/bigquery_agent_analytics_plugin.py:5801"
    change: "log at WARNING when a generation mismatch aborts an otherwise successful setup"

acceptance_criteria:
  - "N close/re-init cycles produce 25 CREATE OR REPLACE VIEW statements total, not 25*N"
  - "a readiness attempt that raises is still retried on the next setup attempt"
  - "no new RPC is added to the steady-state path"
  - "a setup aborted by generation mismatch emits exactly one log line"

non_goals:
  - "changing _EVENT_VIEW_DEFS or the view SQL"
  - "changing the backoff schedule in _ensure_started"
```

chelsealong:
## Summary

Closes #7017.

`BigQueryAgentAnalyticsPlugin` re-ran its full table/view readiness pass on **every** re-initialisation instead of once per process. Each pass issues one `CREATE OR REPLACE VIEW` statement per `_EVENT_VIEW_DEFS` entry (25 at HEAD), and it is awaited from `before_run_callback`, so the DDL sits on the agent request path.

Any deployment that shares one plugin instance across short-lived `Runner`s (`Runner.close()` → `PluginManager.close()` → `plugin.close()` clears `_started`) re-runs all the view DDL per request. The reporter measured ~19k view-DDL jobs/hour, per-table BigQuery quota exhausted in ~2h, and ~25s added to median latency — a 2.3.0 → 2.8.0 regression.

In 2.3.0 the readiness pass was gated behind the schema cache, so it ran at most once per process. At HEAD that gate was deliberately removed so a *failed* first attempt would still retry — but the comment's load-bearing assumption ("once `_started` is True the steady state pays no extra RPC") only holds when the plugin is initialised once per process, which is false whenever it is closed and reused.

## Fix

- Memoise readiness success in a new `_schema_ready` flag that, like `_schema`, survives `close()`/`shutdown()`, and gate `_ensure_schema_exists` on it in `_lazy_setup`.
  - A **failed** attempt raises before the flag is set, so it is still retried on the next setup (preserves the 2.8.0 intent).
  - A **successful** attempt is memoised, so it is not repeated per re-initialisation (restores the 2.3.0 cost profile). No new steady-state RPC.
- Secondary: emit exactly one `WARNING` when a generation mismatch aborts an otherwise-successful setup, so a plugin churning through full setups is no longer completely silent (the issue observed 12 log lines against ~2000 setup runs).

Scope is limited to the readiness memoisation and the one diagnostic log line. The view SQL, `_EVENT_VIEW_DEFS`, and the `_ensure_started` backoff are untouched (the issue's non-goals).

## Testing Plan

Added `TestReadinessMemoisedAcrossReinit` with two tests:

- `test_readiness_pass_runs_once_across_reinit_cycles` — 3 `_ensure_started()` / `shutdown()` cycles over one shared plugin issue the view DDL **once** total (`client.query` called `len(_EVENT_VIEW_DEFS)` times), not once per cycle.
- `test_failed_readiness_is_retried_and_not_memoised` — a first readiness attempt that raises leaves `_schema_ready` False and is retried on the next setup, which then succeeds and memoises.

Evidence the tests prove the bug: with the source change stashed (test kept), both new tests **fail** (`AttributeError: ... _schema_ready` / would-be 75 vs 25 DDL statements); with the fix restored they pass.

```
# without the fix (source change stashed):
2 failed in 24.20s

# with the fix:
tests/unittests/plugins/test_bigquery_agent_analytics_plugin.py  505 passed, 6 skipped in 482.90s
```

Lint/format on the changed files: `pyink` (2 files unchanged), `isort` (exit 0), `ruff check` (all checks passed).

## AI assistance disclosure

This change was prepared with AI assistance (Claude). A human reviewed the diff, the tests, and the failing-without-fix evidence before submission.

### DIFF
--- a/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
+++ b/src/google/adk/plugins/bigquery_agent_analytics_plugin.py
@@ -4279,6 +4279,14 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
     self.offloader: Optional[GCSOffloader] = None
     self.parser: Optional[HybridContentParser] = None
     self._schema: list[bq_schema.SchemaField] | None = None
+    # Remembers that the table-readiness pass (existence check + view DDL)
+    # has succeeded at least once. Like _schema it is pure data and
+    # survives close()/shutdown(), so a plugin that is closed and reused
+    # (Runner.close() -> PluginManager.close() -> plugin.close()) does not
+    # re-issue the CREATE OR REPLACE VIEW statements on every request. Only
+    # a *successful* pass sets it; a failed attempt raises before it is set
+    # and is therefore retried on the next setup.
+    self._schema_ready = False
     self.arrow_schema: pa.Schema | None = None
     self._init_pid = os.getpid()
     _LIVE_PLUGINS.add(self)
@@ -4778,12 +4786,20 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
       # Project out denied payload columns schema-first, so the table
       # schema, Arrow schema, row dict, and views all stay consistent.
       self._schema = _project_schema(_get_events_schema(), self._denied_columns)
-    # Run table readiness on EVERY setup attempt until one succeeds: the
-    # cached _schema must not gate it, or a failed first attempt would skip
-    # the table check on retry and mark the plugin started against a
-    # missing/unready table. Once _started is True,
-    # _lazy_setup returns early above, so the steady state pays no extra RPC.
-    await loop.run_in_executor(executor, self._ensure_schema_exists)
+    # Run table readiness until it first succeeds, then remember that in a
+    # flag that survives close()/shutdown(). A *failed* attempt raises out
+    # of _ensure_schema_exists before the flag is set and is retried on the
+    # next setup (the 2.8.0 intent — a failed first attempt must not be
+    # skipped, or the plugin would be marked started against a
+    # missing/unready table). A *successful* pass is memoised so a plugin
+    # that is closed and re-initialised per request does not re-issue the
+    # per-view CREATE OR REPLACE VIEW DDL every time (the 2.3.0 cost
+    # profile). _schema alone cannot gate this: it also survives close()
+    # but is populated before the readiness RPC, so it would skip a table
+    # check that had never succeeded.
+    if not self._schema_ready:
+      await loop.run_in_executor(executor, self._ensure_schema_exists)
+      self._schema_ready = True
 
     if not self.parser:
       self.arrow_schema = to_arrow_schema(self._require_schema())
@@ -5527,6 +5543,7 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
     state.setdefault("_local_drop_counts", {})
     state.setdefault("_setup_failures", 0)
     state.setdefault("_setup_retry_at", 0.0)
+    state.setdefault("_schema_ready", False)
     state.pop("_setup_lock", None)  # replaced by cross-loop future
     state.pop("_setup_locks", None)
     state.pop("_setup_locks_guard", None)
@@ -5819,6 +5836,18 @@ class BigQueryAgentAnalyticsPlugin(BasePlugin):
       # created outlives a shutdown that already returned — release it,
       # holding the rendezvous until the
       # teardown completes.
+      #
+      # Emit exactly one diagnostic: without it a plugin that is closed and
+      # re-initialised in a tight loop drives full setups that abort here
+      # silently, so there is no signal it is churning until an unrelated
+      # symptom (e.g. quota exhaustion) surfaces.
+      logger.warning(
+          "BigQuery plugin setup completed but was aborted by a concurrent"
+          " shutdown (generation %d != claimed %d); releasing the resources"
+          " it created.",
+          self._generation,
+          claimed_generation,
+      )
       try:
         await self._teardown_aborted_setup()
       finally:
--- a/tests/unittests/plugins/test_bigquery_agent_analytics_plugin.py
+++ b/tests/unittests/plugins/test_bigquery_agent_analytics_plugin.py
@@ -5710,6 +5710,101 @@ class TestSchemaAutoUpgrade:
       plugin._ensure_schema_exists()
 
 
+class TestReadinessMemoisedAcrossReinit:
+  """Regression tests for issue #7017.
+
+  ``BigQueryAgentAnalyticsPlugin`` used to re-run its full table/view
+  readiness pass on every re-initialisation. A host that builds a
+  short-lived ``Runner`` per request over one shared plugin closes and
+  re-inits the plugin every request (``Runner.close()`` ->
+  ``PluginManager.close()`` -> ``plugin.close()``), so the per-view
+  ``CREATE OR REPLACE VIEW`` DDL ran on the request path every single time.
+  Readiness success is now memoised in a flag that survives ``close()``, so
+  a *successful* pass runs once per process while a *failed* one still
+  retries.
+  """
+
+  def _existing_table(self):
+    existing = mock.MagicMock(spec=bigquery.Table)
+    existing.schema = bigquery_agent_analytics_plugin._get_events_schema()
+    existing.labels = {
+        bigquery_agent_analytics_plugin._SCHEMA_VERSION_LABEL_KEY: (
+            bigquery_agent_analytics_plugin._SCHEMA_VERSION
+        ),
+    }
+    return existing
+
+  @pytest.mark.asyncio
+  async def test_readiness_pass_runs_once_across_reinit_cycles(
+      self,
+      mock_auth_default,
+      mock_bq_client,
+      mock_write_client,
+      mock_to_arrow_schema,
+      mock_asyncio_to_thread,
+  ):
+    """N close/re-init cycles issue the view DDL once, not once per cycle."""
+    plugin = bigquery_agent_analytics_plugin.BigQueryAgentAnalyticsPlugin(
+        project_id=PROJECT_ID,
+        dataset_id=DATASET_ID,
+        table_id=TABLE_ID,
+    )
+    mock_bq_client.get_table.return_value = self._existing_table()
+
+    num_views = len(bigquery_agent_analytics_plugin._EVENT_VIEW_DEFS)
+    cycles = 3
+    try:
+      for _ in range(cycles):
+        assert await plugin._ensure_started() == "ok"
+        assert plugin._schema_ready is True
+        # Runner.close() -> PluginManager.close() -> plugin.close() clears
+        # _started; the readiness flag must survive it.
+        await plugin.shutdown()
+    finally:
+      await plugin.shutdown()
+
+    # One readiness pass total (num_views CREATE OR REPLACE VIEW
+    # statements), not one pass per cycle. Before the fix this was
+    # num_views * cycles.
+    assert mock_bq_client.query.call_count == num_views
+
+  @pytest.mark.asyncio
+  async def test_failed_readiness_is_retried_and_not_memoised(
+      self,
+      mock_auth_default,
+      mock_bq_client,
+      mock_write_client,
+      mock_to_arrow_schema,
+      mock_asyncio_to_thread,
+  ):
+    """A failed readiness attempt is not remembered and retries next time."""
+    plugin = bigquery_agent_analytics_plugin.BigQueryAgentAnalyticsPlugin(
+        project_id=PROJECT_ID,
+        dataset_id=DATASET_ID,
+        table_id=TABLE_ID,
+    )
+    # First readiness attempt fails at the table check; the second succeeds.
+    mock_bq_client.get_table.side_effect = [
+        cloud_exceptions.ServiceUnavailable("control plane down"),
+        self._existing_table(),
+    ]
+    try:
+      assert await plugin._ensure_started() == "failed"
+      # A *failed* pass must NOT be memoised (the 2.8.0 intent).
+      assert plugin._schema_ready is False
+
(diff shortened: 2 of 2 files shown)


lupuletic:
Reporter of #7017 here. Thanks for picking this up so quickly — the fix matches what we landed as a local mitigation, and putting `_schema_ready` next to `_schema` reads better than where we had it.

One gap worth closing before merge: **`__getstate__` does not reset `_schema_ready`, so the memo crosses the pickle boundary.**

`__setstate__` has the `setdefault` backfill, so legacy pickles are safe. But `__getstate__` resets every other piece of runtime truth — `_started`, `_init_pid`, `client`, `_executor`, `parser`, `_loop_state_by_loop` — and `_schema_ready` is the one assertion in that group that is about a *remote resource existing* rather than about local data. Verified against this branch (`4196a06`):

```
__getstate__ carries _schema_ready = True
after pickle round-trip, _schema_ready = True
_started reset to                     = False
_init_pid reset to                    = 0
```

The plugin has a custom `__getstate__` precisely because it does get pickled and restored into another runtime. On a first deploy the destination dataset/table does not exist yet, so a restored plugin with `_schema_ready = True` skips `_ensure_schema_exists` entirely and starts appending rows to a missing table. That is the exact failure the comment being replaced was guarding against ("mark the plugin started against a missing/unready table"), reintroduced through the pickle path rather than the retry path.

`_schema` can safely survive pickling because it is just column definitions. Readiness cannot, because it is a claim about the destination.

Suggested one-liner, alongside the existing resets:

```python
     state["_started"] = False
+    state["_schema_ready"] = False
     state["_startup_error"] = None
```

Cost is one readiness pass per process, which is the 2.3.0 profile this PR is restoring anyway — it does not weaken the fix, since the whole problem was re-verifying many times *within* one process.

A test, if useful — it slots onto the existing `TestForkSafety.test_getstate_resets_pid`, which already asserts the `_started` reset:

```python
  def test_getstate_resets_pid(self):
    """Pickle state should have _init_pid = 0 to force re-init."""
    plugin = self._make_plugin()
    plugin._schema_ready = True
    state = plugin.__getstate__()
    assert state["_init_pid"] == 0
    assert state["_started"] is False
    # Readiness is a claim about the destination dataset, so a pickle
    # restored elsewhere must verify it again rather than assume it.
    assert state["_schema_ready"] is False
```

Two smaller notes, neither blocking:

- The new `WARNING` is a real improvement — that silent path is why we saw 12 log lines against roughly 2,000 setup runs. Worth noting it fires on a benign race that is most likely under exactly the host shape this PR fixes, so it should get quieter after the memo lands rather than becoming noise.
- Out of scope here, but the residual after this PR is that the *first* pass in each process still runs 25 `client.query(sql).result()` calls serially on the request path, via `before_run_callback`. Starting the jobs and then collecting results would make that one round of latency instead of 25. Happy to open that separately if you want it.

The middleware side of the trigger is filed at ag-ui-protocol/ag-ui#2642, so the per-request teardown that drives the re-init gets fixed there too.

lupuletic (commented):
Verified this branch (`4196a06`) against the reported production failure. **The fix works and the approach is right.**

I reproduced the host shape from #7017 — one shared plugin instance, `_ensure_started()` then `close()` per request, which is what `before_run_callback` and `Runner.close()` do — and counted the `CREATE OR REPLACE VIEW` statements over 5 cycles:

| | view DDL statements |
|---|---|
| `origin/main` | 125 (25 per cycle) |
| this branch | 25 (once, total) |

That is exactly the intended shape: linear in re-initialisations before, constant after. The existing suite is also green on the branch — 511 passed in `tests/unittests/plugins/test_bigquery_agent_analytics_plugin.py`.

I also like two choices here more than what we had in our own local mitigation:

- Putting `_schema_ready` immediately after `self._schema` in `__init__` makes the "pure data, survives close" grouping self-evident.
- Driving the tests through real `shutdown()` cycles rather than poking `_started` directly. A test that sets the private flag would keep passing if a future `close()` started clearing the memo, so it would not actually guard the regression. Yours does.

The one thing I would still fix before merge is the `__getstate__` reset I described in my earlier comment — the memo currently crosses the pickle boundary while every other runtime field is reset, so a plugin restored into an environment whose dataset does not exist yet would skip table creation. One line, and it does not weaken the fix, since the problem being solved is re-verification *within* one process.

Everything else reads good to me. Thanks for turning this around so fast.

For anyone landing here from the same symptom: the middleware-side trigger — a per-request `Runner` closing plugins owned by a long-lived `App` — is filed at ag-ui-protocol/ag-ui#2642. This PR makes the plugin resilient to that; that one stops the churn at source.

chelsealong:
Fixed: `__getstate__` now resets `_schema_ready = False` alongside the other runtime fields (`bigquery_agent_analytics_plugin.py`), so a pickled/restored plugin re-verifies readiness instead of assuming it. Added the suggested assertion to `test_getstate_resets_pid`; confirmed it fails without the fix (`assert True is False`) and passes with it. Full suite: 511 passed. Pushed as 2872b77.

Issue #7019, opened 2026-09-04
422 Unprocessable Entity in versions 2.6.0, 2.7.0, 2.8.0 when trying to save an eval case


shivammathur33:
## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. Requests without complete information may be rejected / deprioritized.
If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**
https://github.com/google/adk-python/issues/6336 seems to still persist in versions 2.6.0, 2.7.0, 2.8.0.
Upon trying to edit an eval case on an existing eval set, and hitting save (on the top right) this error is seen:
`HTTP/1.1" 422 Unprocessable Entity`

**Steps to Reproduce:**
Similar to in https://github.com/google/adk-python/issues/6336 but adding steps anyway:
- adk web
- Go to localhost:8000
- Create a new session
- Once agent responds, got to evals tab and create a new eval (+ new)
- Open eval set, and hit +current session (adds the session to eval set)
- Once added, session shows in the evalset, hit edit and try to edit some message in the session
- Once edit done, hit save on the top right
- Save does not happen. I see this on the logs: `HTTP/1.1" 422 Unprocessable Entity`

**Expected Behavior:**
The eval set should update the eval case and give 200 OK

**Observed Behavior:**
`HTTP/1.1" 422 Unprocessable Entity`

**Environment Details:**

 - ADK Library Version (pip show google-adk): 2.6.0, 2.7.0, 2.8.0
 - Desktop OS:** macOS
 - Python Version (python -V): 3.11.6

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: gemini-2.5-pro

---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

**Regression:**
Did this work in a previous version of ADK? If so, which one?
no

chelsealong:
Picking this one up now — opening a PR shortly. Flagging it here so nobody duplicates the work; if someone is already on it, say so and I will drop mine.

UniversePeak:
I traced the current Web UI update path: the browser sends the edited `conversation`, `sessionInput`, and `creationTimestamp` to the eval-case PUT endpoint, while the backend validates the request as an `EvalCase`. The reported 422 appears to be in that request-model boundary, but I have not yet reproduced it against the affected releases. Would you be comfortable with me preparing a focused regression test and fix for this issue?

shivammathur33:
Yes.

[...]

Issue #6336, opened 2026-07-07
WEB UI Evals is not working


erick-saenz:
## 🔴 Required Information
*Please ensure all items in this section are completed to allow for efficient
triaging. Requests without complete information may be rejected / deprioritized.
If an item is not applicable to you - please mark it as N/A*

**Describe the Bug:**

When I try to update an eval case in the Web UI I get this error in the logs  

INFO:     127.0.0.1:49878 - "PUT /dev/apps/app/eval_sets/eval_set_3/evals/sacreation HTTP/1.1" 422 Unprocessable Entity

**Steps to Reproduce:**
Please provide a numbered list of steps to reproduce the behavior:
1. I run this command 

uv run adk web . --host 127.0.0.1 --port 8186 --allow_origins '*' --reload_agents

2. Go to the Web UI and start a new session 
3. Create a conversation with the agent
4. Go to evals tab and create a new Eval Set
5. Create a new eval case using current session
6. Edit eval case data and try to save
7. Error is reaised

**Expected Behavior:**
The eval case is saved succesfully and I can run the eval set to validate results

**Observed Behavior:**
What actually happened? Include error messages or crash stack traces here.

When I try to update an eval case in the Web UI I get this error in the logs  

INFO:     127.0.0.1:49878 - "PUT /dev/apps/app/eval_sets/eval_set_3/evals/sacreation HTTP/1.1" 422 Unprocessable Entity

**Environment Details:**

 - ADK Library Version: 2.0.0, 2.1.0, 2.2.0, 2.3.0
 - Desktop OS: Linux
 - Python Version (python -V):  3.12.3

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used:  gemini-2.5-flash

---

## 🟡 Optional Information
*Providing this information greatly speeds up the resolution process.*

**Regression:**
Did this work in a previous version of ADK? If so, which one?

No, It didn't. I tried from version 2.0.0 and ahead

**How often has this issue occurred?:**

 - Always (100%)

surajksharma07:
@erick-saenz — dug into this and it looks like a schema quirk, not something on your end.
When the eval case gets loaded into the edit dialog, an optional content field gets dropped since it's empty, then rejected as required the moment you save — which lines up exactly with that 422.

As a workaround, try adding a = None default to content on InvocationEvent in eval_case.py and restart adk web.
Could you confirm that clears it ?

erick-saenz:
Hi @surajksharma07 

[...]

PR #7026, opened 2026-09-05
fix(eval): default RubricContent.text_property to None so rubrics save without 422

ALDRIN121:
## Summary

Fixes a Pydantic-v2 "required Optional" trap that makes saving an eval case fail with **HTTP 422** when a rubric's `text_property` is omitted or empty.

`RubricContent.text_property` is typed `Optional[str]` but has no default. Under Pydantic v2, an `Optional` field without a default is still **required** — so when the Web UI's `GET` serializes an eval case with `response_model_exclude_none=True` (dropping `text_property: None`) and the browser re-submits it on the save `PUT`, the field is missing and validation fails with `422 Unprocessable Entity`.

This is the exact same bug class as #6336 / the #6515 fix (`fix: make InvocationEvent.content optional so the Web UI can save eval cases`), which patched `InvocationEvent.content` but left `RubricContent.text_property` unaddressed. Related to #7019 (422 persists across 2.6.0–2.8.0 despite #6515).

## Root cause

`src/google/adk/evaluation/eval_rubrics.py`:

```python
class RubricContent(EvalBaseModel):
  text_property: Optional[str] = Field(
      description="..."          # no default -> Pydantic v2 treats as required
  )
```

## Fix

Default `text_property` to `None`, matching the accepted `InvocationEvent.content` fix:

```python
text_property: Optional[str] = Field(
    default=None,
    description="...",
)
```

## Reproduction (before the fix)

```python
rubric = Rubric(rubric_id="r1", rubric_content=RubricContent(text_property=None))
case = EvalCase(eval_id="c1", conversation=[], rubrics=[rubric])
wire = case.model_dump(by_alias=True, exclude_none=True)  # GET
EvalCase.model_validate(wire)                              # PUT -> ValidationError
# pydantic_core.ValidationError: rubricContent.textProperty — Field required
```

After the fix, the round-trip succeeds and `text_property` is `None`.

## Testing plan

- `pytest tests/unittests/evaluation/test_eval_case.py` — **24 passed** (2 new tests: `test_rubric_content_text_property_defaults_to_none`, `test_eval_case_with_rubric_missing_text_property_round_trips`).
- `pytest tests/unittests/evaluation/test_rubric_based_tool_use_quality_v1.py` — **passed** (no regression in rubric consumers).

Note: I ran the focused evaluation tests; the broader `tests/unittests/evaluation/` suite requires optional GCS/Vertex deps not installed in the minimal venv.

Signed-off-by: Aldrin Joseph <yoaldrinjoseph@gmail.com>

### DIFF
--- a/src/google/adk/evaluation/eval_rubrics.py
+++ b/src/google/adk/evaluation/eval_rubrics.py
@@ -25,10 +25,11 @@ class RubricContent(EvalBaseModel):
   """The content of a rubric."""
 
   text_property: Optional[str] = Field(
+      default=None,
       description=(
           "The property being evaluated. Example: \"The agent's response is"
           ' grammatically correct." '
-      )
+      ),
   )
 
 
--- a/tests/unittests/evaluation/test_eval_case.py
+++ b/tests/unittests/evaluation/test_eval_case.py
@@ -23,6 +23,8 @@ from google.adk.evaluation.eval_case import IntermediateData
 from google.adk.evaluation.eval_case import InvocationEvent
 from google.adk.evaluation.eval_case import InvocationEvents
 from google.adk.evaluation.eval_case import SessionInput
+from google.adk.evaluation.eval_rubrics import Rubric
+from google.adk.evaluation.eval_rubrics import RubricContent
 from google.genai import types as genai_types
 import pytest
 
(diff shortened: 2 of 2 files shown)


ALDRIN121:
Hi @i-yliu just following up on #7026 when you get a chance.

This is the small RubricContent.text_property required-Optional fix discovered while investigating #7019. The later #7019 report turned out to be a separate invocationIndex/toolUseIndex issue, but this PR still fixes the independent rubric GET → PUT round-trip failure and includes regression coverage for it.

I’ve kept the change focused. CI currently shows action_required; happy to address anything else needed for review/landing. Thanks!

PR #7034, opened 2026-09-06
fix(cli): sanitize test_name on GET, DELETE, and rebuild

### LINKED ISSUE #7033: adk web GET/DELETE test endpoints skip the create_test path sanitiser
## Expected Behavior

`create_test` strips directories from `test_name` with `os.path.basename` so a name cannot leave the app `tests/` folder.

GET, DELETE, and rebuild of a single test should use the same rule.

## Actual Behavior

On `main` @ `b018062`, only `create_test` calls `os.path.basename`. `delete_test`, `get_test_content`, and `rebuild_app_tests` join `test_name` as given.

A percent-encoded path segment `../outside.json` (`%2e%2e%2foutside.json`) on DELETE/GET is joined onto `tests/` and can read or remove a JSON file in the agent directory, outside `tests/`.

`rebuild?test_name=../outside.json` does the same for the rebuild path.

This is the local `adk web` server. It is unauthenticated. Default bind is loopback. It still matters when `--host 0.0.0.0` is used, or when anything else can hit those routes.

## Steps to Reproduce

1. `adk web` (or the TestClient in `tests/unittests/cli/test_adk_web_server_tests.py`)
2. Put `outside.json` in the agent directory, not in `tests/`
3. `DELETE /dev/apps/<app>/tests/%2e%2e%2foutside.json`
4. On current `main`, that file is removed. After sanitising with `basename`, the request 404s and the file stays.

I can send a PR that shares one helper with `create_test` and adds those cases to `test_adk_web_server_tests.py`.

Oskii:
Fixes #7033

## What

`adk web` can create, read, delete, and rebuild agent test JSON files under `<agent>/tests/`. `create_test` already ran `os.path.basename` on `test_name` so a name cannot leave that folder. GET, DELETE, and rebuild did not.

On `main` @ `b018062`, this DELETE removes a file next to `tests/`, not inside it:

```
DELETE /dev/apps/<app>/tests/%2e%2e%2foutside.json
```

The path param decodes to `../outside.json`. `os.path.join(tests_dir, test_name)` follows it. GET reads the same way. `POST .../tests/rebuild?test_name=../outside.json` passes that path into `rebuild_tests`.

This is the local unauthenticated dev server. Default bind is loopback. It still matters if someone uses `--host 0.0.0.0`, or if anything else can hit those routes.

After the fix, those names are stripped to `outside.json` and resolved only under `tests/`. A missing file is 404. The file outside `tests/` is left alone.

## Why

I was reading the test-file routes next to `test_create_test`. The create handler has an explicit comment about directory traversal. I tried the same name on DELETE with a percent-encoded `../` and the file outside `tests/` was gone.

## How

One helper, `_sanitize_test_filename`, used by create, get, delete, and rebuild. Same `basename` plus the existing `.json` suffix rule. No new public API.

## Testing

Added `test_get_test_rejects_path_traversal`, `test_delete_test_rejects_path_traversal`, and `test_rebuild_single_test_rejects_path_traversal`. Existing create/get/delete/rebuild cases still pass.

```
pytest tests/unittests/cli/test_adk_web_server_tests.py::test_get_test_rejects_path_traversal tests/unittests/cli/test_adk_web_server_tests.py::test_delete_test_rejects_path_traversal tests/unittests/cli/test_adk_web_server_tests.py::test_rebuild_single_test_rejects_path_traversal tests/unittests/cli/test_adk_web_server_tests.py::test_delete_test tests/unittests/cli/test_adk_web_server_tests.py::test_get_test_content tests/unittests/cli/test_adk_web_server_tests.py::test_create_test tests/unittests/cli/test_adk_web_server_tests.py::test_rebuild_single_test -q
```

7 passed.

## Testing plan

- Unit tests above (TestClient, no live `adk web`).
- Dedicated GET traversal case added after review feedback on #7033.
- I did not run the full `tox` matrix on this machine.

### DIFF
--- a/src/google/adk/cli/dev_server.py
+++ b/src/google/adk/cli/dev_server.py
@@ -99,6 +99,18 @@ TAG_DEBUG = "Debug"
 TAG_EVALUATION = "Evaluation"
 
 
+def _sanitize_test_filename(test_name: str) -> str:
+  """Return a tests/ filename, stripping directories that would escape it.
+
+  ``create_test`` already used ``os.path.basename``. Get, delete, and rebuild
+  did not, so a name like ``../outside.json`` could leave the tests folder.
+  """
+  test_name = os.path.basename(test_name)
+  if not test_name.endswith(".json"):
+    test_name += ".json"
+  return test_name
+
+
 class CreateTestRequest(common.BaseModel):
   session_data: dict
 
@@ -912,9 +924,7 @@ class DevServer(ApiServer):
       agent_dir = self._get_agent_dir(app_name)
 
       if test_name:
-        if not test_name.endswith(".json"):
-          test_name += ".json"
-        path = os.path.join(agent_dir, "tests", test_name)
+        path = os.path.join(agent_dir, "tests", _sanitize_test_filename(test_name))
       else:
         path = agent_dir
 
@@ -939,14 +949,10 @@ class DevServer(ApiServer):
         app_name: str, test_name: str, req: CreateTestRequest
     ) -> dict[str, str]:
       """Creates or updates a test file from session data."""
-      # Sanitize test_name to prevent directory traversal
-      test_name = os.path.basename(test_name)
+      test_name = _sanitize_test_filename(test_name)
       agent_dir = self._get_agent_dir(app_name)
       tests_dir = os.path.join(agent_dir, "tests")
       os.makedirs(tests_dir, exist_ok=True)
-
-      if not test_name.endswith(".json"):
-        test_name += ".json"
 
       test_file_path = os.path.join(tests_dir, test_name)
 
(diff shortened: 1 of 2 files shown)


Oskii:
I added the required CLA from this github account :)

Oskii:
Follow-up: dedicated GET traversal test is in `test_get_test_rejects_path_traversal`, matching DELETE. CLA on this account was signed; if `cla/google` is still red a re-check from the CLA bot would help.

Oskii:
Rewrote the follow-up commit so `cla/google` is no longer blocked by an extra `Co-authored-by` the CLA bot could not cover. Same tree: dedicated `test_get_test_rejects_path_traversal` as requested on #7033.

`cla/google`, `check-changes`, and `header-check` are green. Continuous Integration is waiting on first-time contributor workflow approval: https://github.com/google/adk-python/actions/runs/34226805059

@surajksharma07 you already confirmed the helper locally; this is ready for review once CI is allowed to run.

BarneyChambers:
Found during a Hackerbane security review: https://hackerbane.com

PR #7074, opened 2026-09-09
fix(anthropic): preserve sibling keys beside function_response result

### LINKED ISSUE #7073: Anthropic/Claude: tool result dict with a `result` key drops every sibling key (only `json.dumps(result)` reaches the model)
## 🔴 Required Information

**Describe the Bug:**

`google.adk.models.anthropic_llm._part_to_message_block` converts a `function_response` part into a `tool_result` block. When the response dict has a `result` key whose value is not `None`, the block content is `json.dumps(response["result"])` (or `str(result)`), and **every other key in the dict is silently dropped**. The whole-dict fallback (`json.dumps(response_data)`) only runs when `result` is absent or `None`.

The Gemini path sends the full `FunctionResponse.response` dict, so a tool that returns `{"result": ..., "files": [...], "stdout": ...}` behaves correctly on Gemini and loses `files` and `stdout` on Claude. Nothing warns. The tool author sees the full dict in traces/logs, the model never receives it.

This is still present on `main` as of 2026-09-09 (the `elif "result" in response_data and response_data["result"] is not None:` branch in `_part_to_message_block`).

**Steps to Reproduce:**

1. `pip install google-adk==2.2.0` (also reproduces reading `main`)
2. Run the snippet under "Minimal Reproduction Code"
3. Observe that only the `result` value is in the `tool_result` content

**Expected Behavior:**

Either the full dict is serialized whenever the response carries keys other than `result` (keeping the `{"result": x}` → `x` unwrapping for the single-key convention), or the unwrapping is documented as Claude-only behaviour so tool authors know sibling keys are dropped on this backend. Parity with the Gemini path is the least surprising option.

**Observed Behavior:**

```text
{"matched": 33, "total": 49}
```

`files` and `stdout` are gone. In our case (a code-execution tool that reports saved artifacts under `files` beside a `result` value) the Claude-backed agent could not see the file it had just written, re-saved it five to seven times per run, and then guessed at an artifact name. Any tool that adds a truncation marker or a pagination count beside `result` has the same problem: the model sees a clipped list and reads it as complete.

**Environment Details:**

 - ADK Library Version (pip show google-adk): 2.2.0 (same code on `main`)
 - Desktop OS: macOS (also Linux / Cloud Run in production)
 - Python Version (python -V): 3.13.12

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: claude-opus-4-6 via `google.adk.models.anthropic_llm.Claude` (Vertex) and `AnthropicLlm` (direct API); both go through the same converter

---

## 🟡 Optional Information

**Regression:**

Not a regression as far as I can tell; the `result` unwrapping predates the whole-dict fallback that was added for tools like `load_skill`.

**Additional Context:**

The fix we run downstream re-shapes the part before handing it to `_part_to_message_block`: if the response is a dict with a non-`None` `result` **and** other keys (and no MCP-style `content` key), replace the response with `{"result": json.dumps(response_data, default=str)}` so the existing `str(result)` branch emits the whole dict. That keeps the `{"result": x}` convention, keeps the MCP `content` branches, and keeps whatever else the function_response branch does (tool_use_id sanitizing, media parts on newer versions). Happy to send a PR in that shape, or a variant that changes the branch condition directly, whichever maintainers prefer.

```python
from google.adk.models.anthropic_llm import part_to_message_block

response = {
    "result": {"matched": 33, "total": 49},
    "files": [{"filename": "matching.xlsx", "bytes": 8684}],
    "stdout": "Saved to /tmp/outputs/matching.xlsx\n",
}
part = types.Part.from_function_response(
    name="run_python_analysis", response=response
)
part.function_response.id = "toolu_01"
print(part_to_message_block(part)["content"])
# -> {"matched": 33, "total": 49}
# "files" and "stdout" never reach the model. The Gemini path sends the full dict.
```

 - Always (100%)

FOWEPJF255:
## Summary
- When Anthropic/Claude serializes a `function_response` dict that has a non-null `result` **and** sibling keys (`files`, `stdout`, ...), serialize the **full** dict instead of only `result`.
- Keeps single-key `{result: x}` unwrapping for backward compatibility.
- Adds regression test for #7073.

## Test plan
- [x] `PYTHONPATH=src pytest tests/unittests/models/test_anthropic_llm.py -k "sibling_keys or traditional_result or dict_result or with_content"`

Fixes #7073


FOWEPJF255:
﻿Rebased onto latest `main` and resolved the conflict in `anthropic_llm.py`:

- Keep exact `{"result": value}` unwrap behavior from main
- When `result` has sibling keys, serialize the full dict (issue #7073 / Gemini parity)

**CLA:** `cla/google` is still required before review/import. Please sign at https://cla.developers.google.com/ (same Google account linked to GitHub `FOWEPJF255`), then re-check this PR.

FOWEPJF255:
Follow-up: the code conflict on `main` is resolved on this branch. `cla/google` is still required before review can proceed 鈥?I am completing the Google CLA signing for this account and will ping once the CLA check turns green.

FOWEPJF255:
I have signed the Google CLA for this account / commit identity. Please re-check cla/google.

FOWEPJF255:
Closing temporarily to refresh cla/google status after signing.

FOWEPJF255:
CLA data updated (GitHub username FOWEPJF255 + emails including the commit address). Please re-check `cla/google`.

FOWEPJF255:
`cla/google` is now green after CLA data + New Contributors rescan. Ready for review when convenient.

FOWEPJF255:
Temporarily closing to refresh import/copybara after cla/google turned green.

FOWEPJF255:
Rebased onto latest main today (`b80c292`). Please re-check `cla/google` for the updated commit identity (GitHub `FOWEPJF255`).

FOWEPJF255:
CLA tip refreshed to `lsmlhi_25 <lhi25wy@163.com>` (`22f996e`). `cla/google` is green again after the earlier qq.com authorship mismatch. Ready for review.

FOWEPJF255:
`cla/google` is green on the current tip. Ready for review when convenient.

GWeale:
this landed separately on 9 september, hours after you opened it. the unwrap is narrowed to exactly {"result": value}, so a dict with siblings is serialized whole and files and stdout reach claude, with a test for that shape. on main only; 2.9.1 still drops them. your patch took the branch change #7073 offered, and that's what landed.

one difference: main calls json.dumps without default=str, so a sibling it can't encode now raises. closing; open a new issue if you hit that.

Issue #7076, opened 2026-09-09
Resumable mode dispatches caller-authored function calls with no model turn or author check


dmcnelis:
## Summary

When a `google-adk` app runs in **resumable** mode (`ResumabilityConfig(is_resumable=True)`), the flow loop will dispatch a tool call it reads from a session event **without checking who authored that event** and **without a model turn**. A caller who can (1) seed a session with an event authored as `user` that carries a `function_call`, and (2) start a resumable run against that event's `invocation_id`, causes the named tool to execute with caller-chosen arguments — bypassing the model, the system prompt, and any prompt-injection/safety reasoning the model would normally apply.

This was reported privately to Google's security team first. They assessed it as below their bar for a tracked security vulnerability (resumable mode is experimental and non-default) and suggested filing it publicly here. Filing accordingly, with a fix.

## Affected versions

- **Verified by execution** on `google-adk` **2.7.1** and **2.8.0** (2.8.0 is the current PyPI latest).
- **Present on `main`** by inspection — the dispatch branch was refactored since 2.8.0, but the same unguarded predicate now lives in `flows/llm_flows/_resume_utils.py::decide_step_resume`, and the ingress validator is unchanged.

Requires resumable mode, which is opt-in and marked EXPERIMENTAL. Not the default configuration.

## Attacker model / preconditions

1. The target app is a resumable ADK `App` (`ResumabilityConfig(is_resumable=True)`).
2. The attacker can reach the ADK HTTP surface enough to call `POST /apps/{app}/users/{u}/sessions` and `POST /run_sse`. `get_fast_api_app` serves these **unauthenticated by default**; a hardened deployment puts auth in front, in which case the attacker must be an authenticated caller.

**What it grants over that access:** normally a caller sends a *message* and the *model* decides whether to call a tool, which one, and with what arguments, under the system prompt. This lets the caller instead invoke **any registered tool** with **fully attacker-chosen arguments**, by forging a `user`-authored event carrying the call, **with no model in the loop** — bypassing the system prompt and every model-mediated safety check. Impact is bounded by the caller's already-verified auth scopes and by any before-tool policy callbacks the host installs. It is **not** unauthenticated RCE, but within those bounds it removes the model turn hosts rely on as a control.

## Reproduction (~2 min, no cloud, no API key)

A resumable victim agent with one tool:

```python
# agents/victim/agent.py
from google.adk.agents.llm_agent import LlmAgent
from google.adk.apps.app import App
[...]

PR #7077, opened 2026-09-09
fix(flows): require agent authorship before resume-dispatching a function call (#7076)

dmcnelis:
## What

Fixes the resumable tool-dispatch bypass reported in #7076.

When an app is resumable (`ResumabilityConfig(is_resumable=True)`), the flow replays a trailing function-call event **without checking who authored it** and **without a model turn**. Because session init accepts caller-supplied events, a client can:

1. `POST /apps/{app}/users/{u}/sessions` with an event authored `"user"` carrying a `function_call`, and
2. `POST /run_sse` with `newMessage: null` against that event's `invocationId`,

and the named tool executes with caller-chosen arguments — bypassing the model, the system prompt, and any model-mediated safety reasoning. Confirmed by execution on 2.7.1 and 2.8.0; the same gap is present on `main` at the refactored location addressed here.

## Fix

Guard **both** event-selection paths in `flows/llm_flows/_resume_utils.py` by the current agent's name (`require_agent_name`):

- `decide_step_resume` — the trailing-call replay (the path the PoC exercises).
- `decide_resume` → `_find_target_call_event` — the multi-event target search (adds an optional `agent_name` filter; default `None` preserves the pure matching behavior its unit tests cover).

A legitimate resume is unaffected: an agent authors its own paused call, so it still dispatches. A caller-forged call no longer dispatches and the flow falls through to a normal model turn.

## Tests

- Added regression tests that **fail without the guard** (a forged `"user"`-authored call is dispatched / selected) and **pass with it**, plus a positive test that the agent's own paused call still replays.
- `tests/unittests/flows/llm_flows/test_resume_utils.py`: 35 passing.
- `tests/unittests/runners/test_resume_invocation.py`: sub-agent resume still passes (the author check resolves to the current sub-agent's name).
- Wider `tests/unittests/flows/llm_flows/`: 681 passing (2 unrelated collection errors from optional deps `openai` / `google.cloud` not installed locally).
- `pyink` and `isort` clean on both files.

## Scope / honest limitations

This closes the **demonstrated** path (a `"user"`-authored forged call). For full defense-in-depth, the session-init validator (`_validate_session_initialization_events`) could also reject caller-supplied events that carry `function_call` parts or are authored as anything the framework reserves, so a caller cannot forge an *agent*-authored pending call either. That touches the session-restore design, so I've left it as a maintainer call and kept this PR to the resume-path provenance guard. Happy to extend if you'd prefer the ingress hardening in the same change.

Note: I'll need to complete the Google CLA — the bot should indicate the exact email.

Refs: #7076

https://claude.ai/code/session_01KbhX6pH5CZCtymrBneXQVJ


dmcnelis:
I'm not sure how to clear the CLA agreement, given the comment / attribution to usage of Claude Code on the PR. If you provide direction, I'll do what I can in that regard.

PR #7093, opened 2026-09-11
feat(skills): add SkillDiscoveryMode to skip the discovery turn

### LINKED ISSUE #7092: feat(skills): add include_list_skills to inject the L1 catalog without list_skills
`list_skills` costs an extra model turn before `load_skill`. The only way to inject the L1 catalog as `<available_skills>` XML was to drop `ListSkillsTool` from the private `_tools` list. `tool_filter` is an allowlist, so hiding just `list_skills` means naming every other skill tool.
A public `SkillToolset(include_list_skills=True|False)` flag.
- `True` (default): keep `list_skills`.
- `False`: hide `list_skills` and inject the local L1 catalog into the system instruction so the model can call `load_skill` directly.
### Impact on your work
Needed to load skills without a discovery round trip, without mutating private attributes.
### Willingness to contribute
Yes. I will submit a PR.
- Mutating `skill_tools._tools` (works, not a public API).
- `tool_filter=["load_skill", ...]` (works since the tool_filter/instruction fix, but is an allowlist).
```python
SkillToolset(skills=[shopping_advice_skill], include_list_skills=False)

### PR DESCRIPTION
**Pull request by thomaslprr**

- Closes: #7092

**Problem:**
There is no public way to inject the L1 skill catalog into the system instruction. Hiding `list_skills` required mutating `_tools`.

**Solution:**
Add `SkillToolset(discovery_mode=SkillDiscoveryMode.EAGER|LAZY)`.
- `LAZY` (default): expose `list_skills`.
- `EAGER`: hide `list_skills` and inject `<available_skills>` XML so the model can call `load_skill` directly.
Non-enum values raise `TypeError`.

### Testing Plan

**Unit Tests:**

```text
.venv/bin/python -m pytest tests/unittests/tools/test_skill_toolset.py -q
# 159 passed
```

- [x] Local sample based on `contributing/samples/environment_and_skills/skills_agent` with `discovery_mode=SkillDiscoveryMode.EAGER`:
  - no `list_skills` tool
  - `<available_skills>` present in the system instruction (`greeting-skill`, `weather-skill`)
  - agent can call `load_skill` directly
- Default `LAZY` still exposes `list_skills`.
- With a registry, `search_skills` remains available in `EAGER`.

### Checklist


wukath:
Hi @thomaslprr, thanks for the PR! We like the idea of skipping the discovery turn, but we reshaped it slightly to use an enum (discovery_mode=SkillDiscoveryMode.EAGER vs LAZY) since omitting list_skills has the major side effect of injecting the catalog into the system prompt. We kept your core logic and all your tests -- would this approach work for you?

thomaslprr:
> Hi @thomaslprr, thanks for the PR! We like the idea of skipping the discovery turn, but we reshaped it slightly to use an enum (discovery_mode=SkillDiscoveryMode.EAGER vs LAZY) since omitting list_skills has the major side effect of injecting the catalog into the system prompt. We kept your core logic and all your tests -- would this approach work for you?

Hi @wukath, yes, that works for me - `discovery_mode=SkillDiscoveryMode.EAGER|LAZY` is clearer than a boolean given the system-prompt side effect. Happy for you to land it that way. Thanks!

Would you like me to update this PR to the enum API, or will you land the reshaped version on your side? I already have the changes ready locally if you'd like me to push them.

thomaslprr:
@wukath, I pushed the enum reshape as discussed:

- `discovery_mode=SkillDiscoveryMode.LAZY` (default): keep `list_skills`
- `discovery_mode=SkillDiscoveryMode.EAGER`: hide `list_skills` and inject `<available_skills>`
- non-enum values raise `TypeError`

Core logic and tests are updated (159 passed locally). I also set up a local ADK sample based on `skills_agent` with `discovery_mode=EAGER` and confirmed the agent run path: no `list_skills`, catalog injected, `load_skill` available. Ready for another look. 

Thanks!

PR #7091, opened 2026-09-11
fix: exclude thought parts from response match scoring

MrCapricornLiu:
The response-match evaluator counts text parts marked `thought=True` as final-answer text. A correct answer can lose points because of its thought summary, and a wrong answer can receive credit for a word mentioned only in that summary. Exclude thought parts from both actual and reference text, preserving the existing tokenization and empty-response behavior.

### Reproduction and expected behavior

On ADK 2.8.0 / main (`2284c581`), Linux, Python 3.12.13, run the added regression cases in:

```sh
pytest tests/unittests/evaluation/test_final_response_match_v1.py -q
```

The added cases fail on the original implementation and pass with this change. No LiteLLM or live model endpoint is required; Runner verification uses a scripted `BaseLlm` response.

### Testing Plan

Type checking matches the unmodified upstream baseline (831 existing errors; no new errors).

- Relevant unit tests on current main (`da10185e`): 74 passed.
- The Runner reproduction passes on current main. A wheel built against `2284c581` was also verified in a clean environment.
- Earlier full-suite validation on `2284c581`, Python 3.10–3.14: tests passed with the two GC-introspection cases run in separate processes (14,248 passing cases on Python 3.12). One unrelated MCP stdio teardown case was deselected after it also failed to finish on the unmodified base. The default uninterrupted `tox` run did not complete.

<details>
<summary>Runner setup and reproduction output</summary>

Save the following as `repro.py`, then run `python repro.py` after installing the locally built wheel. Install `rouge-score` as well.

```python
import asyncio
import json
from google.adk.agents import Agent
from google.adk.artifacts.in_memory_artifact_service import InMemoryArtifactService
from google.adk.models.base_llm import BaseLlm
from google.adk.models.llm_request import LlmRequest
from google.adk.models.llm_response import LlmResponse
from google.adk.runners import Runner
from google.adk.sessions.in_memory_session_service import InMemorySessionService
from google.genai import types
from pydantic import BaseModel, Field

class ScriptedModel(BaseLlm):
    model: str = "local-repro"
    responses: list[types.Content]
    requests: list[LlmRequest] = Field(default_factory=list)

    async def generate_content_async(self, llm_request, stream=False):
        index = len(self.requests)
        self.requests.append(llm_request.model_copy(deep=True))
        yield LlmResponse(content=self.responses[index])

def response(*parts):
    return types.Content(role="model", parts=list(parts))

async def run_agent(model, tools=(), artifacts=None, setup=None):
[...]

PR #7117, opened 2026-09-14
fix(evaluation): exclude thought parts from text extraction across criteria

james-auror:
### Overview

Every text-extraction site in `evaluation/` filtered parts with `if part.text`, which was equivalent to "visible model output" before thinking models became routine. It no longer is: parts marked `thought=True` are text parts, so agent reasoning leaks into ROUGE scoring, judge prompts, and hallucination grounding checks. In `hallucinations_v1`, each thought part became its own evaluation step, so an agent's private reasoning was checked for grounding against tool outputs as if it were a claim made to the user.

This generalizes the fix in #7091 (credit to @MrCapricornLiu for the diagnosis and regression-test approach) from `final_response_match_v1` alone to every criterion that extracts text from `Content`.

### Changes

- Adds `get_text_parts()` alongside the existing `get_text_from_content()` in `llm_as_judge_utils.py` as the shared, thought-filtering source of truth for text extraction.
- Routes every criterion through it:
  - `final_response_match_v1`, `vertex_ai_eval_facade`, and `agent_evaluator` each had a private copy of the same unfiltered extraction body — these now delegate to `get_text_from_content()`.
  - `hallucinations_v1` and `rubric_based_multi_turn_trajectory_evaluator` had inline extraction at five call sites — these now call `get_text_parts()` directly.
  - `final_response_match_v2`, `rubric_based_final_response_quality_v1`, `rubric_based_tool_use_quality_v1`, and `rubric_based_evaluator` already routed through the shared helper, so they're fixed by the helper change alone.
- Adds a regression test per affected criterion asserting thought text is absent from the resulting score, prompt, or dialogue string.


james-auror:
I am working with our legal team to get the CLA signed.

Issue #6908, opened 2026-08-26
GcpSkillRegistry.get_skill() always fails with 302: Agent Registry media download redirect to GCS signed URL is not followed


codebee-aoki:
## 🔴 Required Information

**Describe the Bug:**
`GcpSkillRegistry.get_skill()` always fails when downloading the skill archive. The Agent Registry API serves the revision media download (`?alt=media`) as a `302` redirect to a short-lived GCS signed URL. The `httpx.AsyncClient` created in `_create_httpx_client` does not follow redirects (httpx default), and httpx's `response.raise_for_status()` raises on 3xx responses, so `_make_request` surfaces the redirect as an error and the skill archive can never be downloaded:

```
RuntimeError: API request failed with status 302: ...
```

This makes skill loading from the GCP Agent Registry (e.g. via `SkillToolset.load_skill`) unusable against the real API.

**Steps to Reproduce:**
1. Install `google-adk==2.7.1`.
2. Point `GcpSkillRegistry` at an Agent Registry project/location containing any skill.
3. `await registry.get_skill(name="<any-skill>")`.
4. The metadata request succeeds, then the `?alt=media` media download fails with the `RuntimeError` above.

**Expected Behavior:**
The client follows the 302 redirect to the GCS signed URL and downloads the skill archive. (This is safe: httpx removes the `Authorization` header when following a cross-origin redirect, so the OAuth bearer token is not forwarded to the signed-URL host.)

**Observed Behavior:**
`RuntimeError: API request failed with status 302: ...` raised from `_make_request` for every skill download.

**Environment Details:**

 - ADK Library Version (pip show google-adk): 2.7.1
 - Reproduced both locally and with an agent deployed on Agent Engine runtime. Also independently confirmed by another user on 2.7.1 (Linux / Python 3.12), who has been working around it with a client-side monkey-patch since 2.6.2 — see https://github.com/google/adk-python/pull/6824#issuecomment (PR comment thread).

**Model Information:**

 - Are you using LiteLLM: No
 - Which model is being used: N/A (client-side registry download; no model involved)

## Fix

PR #6824 is open with the fix: pass `follow_redirects=True` to both `httpx.AsyncClient` constructions in `_create_httpx_client` (plain and mTLS/ssl-context branches). It includes unit tests and has been verified end-to-end against a real Agent Registry catalog (the identical change applied as a monkey-patch on Agent Engine runtime restores skill downloads).

llalitkumarrr:
[...]

PR #6824, opened 2026-08-20
fix: follow redirects when downloading skills in GcpSkillRegistry

codebee-aoki:
**Please ensure you have read the [contribution guide](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md) before creating a pull request.**

### Link to Issue or Description of Change

**1. Link to an existing issue (if applicable):**

- Closes: #6908
- Related: N/A

**2. Or, if no issue exists, describe the change:**

**Problem:**
`GcpSkillRegistry.get_skill()` fails when downloading the skill archive. The Agent Registry API serves the revision media download (`?alt=media`) as a `302` redirect to a GCS signed URL. The `httpx.AsyncClient` created in `_create_httpx_client` does not follow redirects (httpx default), and `response.raise_for_status()` raises on 3xx, so the request surfaces as `RuntimeError: API request failed with status 302: ...`.

**Solution:**
Pass `follow_redirects=True` to both `httpx.AsyncClient` constructions in `_create_httpx_client` (the plain client and the mTLS/ssl-context client).

Following redirects is safe here: httpx removes the `Authorization` header when following a redirect to a different origin, so the OAuth bearer token is not leaked to the GCS signed-URL host (GCS would reject a signed URL carrying extra credentials anyway).

### Testing Plan

**Unit Tests:**

- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

Added a unit test (`test_create_httpx_client_follows_redirects`) asserting that clients created by `_create_httpx_client` have `follow_redirects=True`, for both the default branch and the mTLS/ssl-context branch. Also updated an existing assertion (`test_get_skill_with_mtls`) that checked the exact `httpx.AsyncClient` constructor kwargs, since it now includes `follow_redirects=True`.

```
$ pytest tests/unittests/integrations/skill_registry/test_gcp_skill_registry.py -q
23 passed in 3.58s
```

**Manual End-to-End (E2E) Tests:**

Tested end-to-end by applying the identical change (monkey-patching `GCPSkillRegistry._create_httpx_client` to pass `follow_redirects=True`, same two-branch logic as this PR) to an agent deployed on Agent Engine runtime. Confirmed that skill downloads from the Agent Registry — which previously failed with `RuntimeError: API request failed with status 302: ...` — now succeed: the `?alt=media` request's 302 redirect to the GCS signed URL is followed and the skill archive is downloaded and loaded correctly.

### Checklist

[...]

PR #7135, opened 2026-09-16
feat(skills): let SkillToolset pin registry skills to appear in the catalog

### LINKED ISSUE #7130: feat(skills): let SkillToolset pin registry skills so they appear in list_skills / the catalog without a search_skills turn

When `SkillToolset` is backed by a `SkillRegistry` (e.g. `GCPSkillRegistry` / Agent Registry), skills that live in the registry are never part of the L1 catalog the model sees. The model can only reach them through `search_skills` → `load_skill`, so a registry skill always costs at least one more model turn (plus one registry round-trip) than an equivalent local skill, even when the developer knows up front exactly which registry skills the agent should use.

Current behavior on `main` (ac0133a4):

- `list_skills` and the `<available_skills>` catalog are built from `SkillToolset._list_skills()`, which only returns the skills passed via `skills=[...]` (`src/google/adk/tools/skill_toolset.py:1594-1596`). The registry is never consulted for the catalog.
- The registry is only consulted from `_get_or_fetch_skill()` (`skill_toolset.py:1582,1592`), i.e. after the model has already named the skill. The system instruction tells the model to fall back to `search_skills` when local skills are not sufficient (`skill_toolset.py:1669-1675`).
- The new `SkillDiscoveryMode.EAGER` (#7092 / #7093) removes the discovery turn for local skills, but its docstring explicitly leaves registry skills out: "Registry skills are unaffected: they are still reachable only through `search_skills`." So after that change the gap is: local skills = 0 discovery turns, registry skills = always ≥ 1.
- As a secondary cost, a registry skill the model has activated is re-downloaded on every later invocation: session state only keeps the activated skill *names* (`_adk_activated_skill_<agent>`), and `_fetched_skill_cache` is keyed by `invocation_id` (`skill_toolset.py:1564`), so `_resolve_additional_tools_from_state()` calls `registry.get_skill()` again each turn.

Concretely, for a fixed set of skills that happen to be published in Agent Registry instead of shipped with the agent:

| | local skill | registry skill |
|---|---|---|
| turns before `load_skill` (LAZY) | 1 (`list_skills`) | 2 (`list_skills` → `search_skills`) |
| turns before `load_skill` (EAGER) | 0 | 1 (`search_skills`) |
| registry round-trips per later turn | 0 | 1 per activated skill |

The extra turn also adds the `search_skills` tool declaration, the search result payload and the model's reasoning about it to the context on every conversation, which is the token waste we want to avoid. Our use case is the common one where the skill *content* is managed centrally in the registry (its update cycle is decoupled from the agent's deploy cycle), but the *set* of skills an agent uses is decided at build time, exactly like local skills.

Let `SkillToolset` pin a known set of registry skills so they behave like local skills once fetched:

```python
toolset = SkillToolset(
    skills=load_skills_from_dir("./skills"),           # local, as today
    registry=GCPSkillRegistry(project_id=..., location=...),
    registry_skills=["shared-skill-a", "shared-skill-b"],  # NEW
    discovery_mode=SkillDiscoveryMode.EAGER,
)
```

Proposed semantics:

- `registry_skills: list[str] | None = None` (keyword-only, default `None`, fully backward compatible).
- The named skills are fetched lazily, once, on the first `get_tools()` / `process_llm_request()` call for the toolset (both are already `async`, and the agent module itself is usually imported synchronously, sometimes inside a running event loop such as `adk run` / `adk web`, so `asyncio.run` at import time is not an option). Fetches for the whole list run concurrently and are guarded by a lock so concurrent first requests only fetch once.
- After the fetch the skills are stored alongside the local ones, so they appear in `list_skills`, in the EAGER catalog, and `_get_or_fetch_skill` resolves them without touching the registry again for the lifetime of the process. `search_skills` keeps working for skills that are *not* pinned.
- A public `async def prefetch()` so callers that do have a startup hook (e.g. Agent Engine `AdkApp.set_up()`, which runs synchronously with no loop, or a FastAPI lifespan) can warm the toolset before the first request.
- Name collisions: a fetched registry skill whose `SKILL.md` frontmatter name already exists locally is skipped with a warning and the local skill wins. This matches the policy `SearchSkillsTool` already applies ("Skill naming conflict ... Registry skill is filtered."). Note the skill name comes from the archive's frontmatter, not from the registry resource name, so the check must happen after the fetch.
- Fetch failures are logged and leave the toolset usable (the `search_skills` path still works); the next request retries.
- `clone_with_updated_skills()` forwards `registry_skills`, so the pinned set survives cloning.

### Impact on your work

We run agents on Agent Engine that consume skills published in Agent Registry. Every conversation pays a `search_skills` turn before the skill can be loaded, and every subsequent turn re-downloads the skill archive. We currently work around this with a `SkillToolset` subclass that does the lazy prefetch described above, but it has to reach into private attributes (`_skills`, `_registry`) and is dropped by `clone_with_updated_skills()`, so we would much rather have it in the library. Not blocking, but it affects latency and token cost on every request.

### Willingness to contribute

Yes. Happy to open a PR with the parameter, the lazy prefetch, `prefetch()`, and unit tests, once the maintainers agree on the shape. (Related: #6908 / #6824 — `GCPSkillRegistry.get_skill()` currently fails on the media-download redirect, so the prefetch needs that fix to work end to end against the real Agent Registry.)

---

- **Fetch at startup and pass the result via `skills=[...]`.** Works in principle, but `get_skill()` is `async` and there is no sync counterpart, and the agent module is imported inside a running event loop by `adk run` / `adk web`, so `asyncio.run()` at module import is not portable.
- **Snapshot the registry skills into the local `skills/` directory at deploy time.** Removes the runtime dependency entirely, but couples skill updates to the agent's deploy cycle, which is the opposite of why the skills are in a registry.
- **Extend `SkillDiscoveryMode.EAGER` to include registry skills automatically.** Would require a "list all" operation on `SkillRegistry`, which the interface does not have (only `get_skill` and `search_skills`), and would inject an unbounded catalog. An explicit allow-list keeps the catalog bounded and the registry interface unchanged.
- **Cache fetched skills across invocations (key `_fetched_skill_cache` by skill name instead of `invocation_id`).** Solves the re-download cost but not the discovery turn; it would also be a useful, independent change and could be a separate PR.

Sketch of the core (names are placeholders):

```python
class SkillToolset(BaseToolset):
  def __init__(self, ..., registry_skills: list[str] | None = None):
    ...
    self._registry_skills = list(registry_skills or [])
    self._registry_skills_loaded = False
    self._registry_skills_lock = asyncio.Lock()

  async def prefetch(self) -> None:
    if self._registry_skills_loaded or not self._registry_skills:
      return
    async with self._registry_skills_lock:
      if self._registry_skills_loaded:
        return
      missing = [n for n in self._registry_skills if n not in self._skills]
      fetched = await asyncio.gather(
          *(self._registry.get_skill(name=n) for n in missing)
      )
      for skill in fetched:
        if skill.name in self._skills:
          logger.warning("Registry skill %r collides with a local skill; keeping local", skill.name)
          continue
        self._skills[skill.name] = skill
      self._registry_skills_loaded = True

  async def get_tools(self, readonly_context=None):
    await self.prefetch()
    ...

  async def process_llm_request(self, *, tool_context, llm_request):
    await self.prefetch()
    ...
```

### Additional Context

- `SkillDiscoveryMode` (#7092, #7093, merged as 3d3c1d90) — the local-skill half of the same problem.
- `GCPSkillRegistry.get_skill()` redirect bug (#6908, fix in #6824).
- ADK version: main @ ac0133a4 (also verified on 2.9.1 for the catalog / `search_skills` behavior).

ArulJerald:
- Closes: #7130
**Problem:**
When `SkillToolset` is backed by a `SkillRegistry`, skills in the registry are not part of the initial catalog. The model must perform a `search_skills` turn before it can discover and load them. For a known set of registry skills, this costs an extra turn, extra round-trips, and extra token overhead.
**Solution:**
- Added a `registry_skills: list[str] | None = None` parameter to `SkillToolset` to pin specific skills from the registry.
- Added `async def prefetch(self) -> None` guarded by an `asyncio.Lock` with double-checked locking to fetch pinned skills concurrently into the local catalog.
- Integrated lazy prefetch into `get_tools()`, `process_llm_request()`, `_get_or_fetch_skill()`, `ListSkillsTool`, and `SearchSkillsTool`.
- Preserved `registry_skills` across `clone_with_updated_skills()`.
- Added warning and local precedence when a fetched registry skill name collides with a local skill.

**Problem:**  
When `SkillToolset` is backed by a `SkillRegistry`, skills in the registry are not part of the initial catalog. The model must perform a `search_skills` turn before it can discover and load them. For a known set of registry skills, this costs an extra turn, extra round-trips, and extra token overhead.  
**Solution:**
* Added a `registry_skills: list[str] | None = None` parameter to `SkillToolset` to pin specific skills from the registry.
* Added `async def prefetch(self) -> None` guarded by an `asyncio.Lock` with double-checked locking to fetch pinned skills lazily into the local catalog.
* Integrated lazy prefetch into `get_tools()`, `process_llm_request()`, and `_get_or_fetch_skill()`.
* Added `_registry_skill_aliases` mapping to support resolution when registry resource IDs and frontmatter names differ, ensuring `load_skill` serves them locally without redundant registry re-downloads.
* Updated `SearchSkillsTool` to filter out pinned registry skills matching either alias or frontmatter name.
* Preserved `registry_skills` across `clone_with_updated_skills()`.
* Added collision handling and warning when a fetched registry skill name collides with a local skill.
* Updated `contributing/samples/integrations/gcp_skill_registry_agent/agent.py` to demonstrate `registry_skills`.
**Known Limitations & Follow-ups:**
* **Unbounded retry on permanent failures**: If a pinned registry skill fails to load (e.g., 404 or permanent upstream issue), `_registry_skills_loaded` remains `False` and retries the fetch on subsequent requests with a warning logged. A follow-up improvement will bound retries to N attempts or distinguish transient network failures from permanent errors.

### Testing Plan
**Unit Tests:**

Summary of passed pytest results:

tests/unittests/tools/test_skill_toolset.py ................................... [100%] 185 passed in 5.25s

configfile: pyproject.toml
plugins: mock-3.15.1, xdist-3.8.0, asyncio-1.4.0, langsmith-0.12.4, anyio-4.15.1, hypothesis-6.168.0
collected 189 items / 186 deselected / 3 selected

tests/unittests/tools/test_skill_toolset.py ...                          [100%]

====================== 3 passed, 186 deselected in 0.78s =======================

configfile: pyproject.toml
plugins: mock-3.15.1, xdist-3.8.0, asyncio-1.4.0, langsmith-0.12.4, anyio-4.15.1, hypothesis-6.168.0
collected 189 items

tests/unittests/tools/test_skill_toolset.py ............................ [ 14%]
........................................................................ [ 52%]
........................................................................ [ 91%]
.................                                                        [100%]
============================= 189 passed in 5.40s ==============================

### Checklist


src/google/adk/tools/skill_toolset.py
```diff
@@ -1467,6 +1480,67 @@ def skills_folder(self) -> Path | None:
...
+          logger.warning(
+              "Skill naming conflict: skill '%s' already exists locally."
+              " Registry skill is filtered.",
+              skill.name,
+          )
+        else:
+          self._skills[skill.name] = skill
+        self._fetched_registry_skills.add(name)
+
+      if len(self._fetched_registry_skills) == len(self._registry_skills):
```
codebee-aoki, line 1548: If one pinned name fails permanently (404, or the current 302 bug from #6908), `_registry_skills_loaded` never flips to True, so every request re-issues the fetch for that name and logs a warning. That is acceptable for a first increment, but please call it out in the PR description and consider a follow-up: e.g. give up on a name after N attempts, or only retry on network errors.

src/google/adk/tools/skill_toolset.py
```diff
@@ -1376,6 +1379,8 @@ def __init__(
     Args:
       skills: List of skills to register.
       registry: Optional skill registry for dynamic loading.
+      registry_skills: Optional list of skill names in the registry to pin and
```
codebee-aoki: Suggest expanding this a little so users know what pinning buys them, e.g.: pinned skills are fetched once per process on first use, then appear in `list_skills` / the EAGER catalog and are served locally without further registry calls; unpinned registry skills are still reachable via `search_skills`. Also worth noting that the skill is stored under the name in the archive's frontmatter, which may differ from the registry resource name.

src/google/adk/tools/skill_toolset.py
```diff
@@ -273,6 +273,7 @@ def _get_declaration(self) -> types.FunctionDeclaration | None:
   async def run_async(
       self, *, args: dict[str, Any], tool_context: ToolContext
   ) -> Any:
+    await self._toolset.prefetch()
```
codebee-aoki: Nit: this call (and the one in `SearchSkillsTool.run_async`) is redundant, since `get_tools()` and `process_llm_request()` always run before any tool call in the flow. Harmless, but dropping them keeps the entry points to a minimum.

tests/unittests/tools/test_skill_toolset.py
```diff
@@ -3669,3 +3669,193 @@ def test_unload_skill_api_leaves_other_skills_active(
...
+  )
+
+  await toolset.prefetch()
+
+  mock_registry.get_skill.assert_called_once_with(name="skill1")
+  assert toolset._skills["skill1"] is mock_skill1
+
+
+@pytest.mark.asyncio
+async def test_skill_toolset_prefetch_failure_is_logged_and_retried(
```
codebee-aoki, line 3765: Coverage this file does not have yet, all cheap to add next to this test:

- LAZY mode: `ListSkillsTool.run_async` output contains the pinned skill (only the EAGER catalog path is asserted today).
- After `prefetch()`, `LoadSkillTool.run_async` for a pinned name must not call `registry.get_skill` again (`get_skill.call_count` stays at 1). This is the process-lifetime cache behaviour that removes the per-turn re-download, so it deserves an explicit assertion.
- Name mismatch: `registry_skills=["foo"]` where the fetched archive's frontmatter name is `bar`. The skill ends up stored under `bar` and `foo` is marked fetched; a test would pin that behaviour down.

codebee-aoki (commented):
Thanks @ArulJerald, this matches the proposal in #7130 closely and the structure looks right: lazy fetch on first use, one fetch under the lock, local skills win on collision, and `clone_with_updated_skills` keeps the pinned list. A few points from reading the diff, roughly in priority order. The line-level ones are inline.

1. **Unbounded retry for a permanently failing skill** (inline on `prefetch()`). Fine as a first increment, but worth stating in the PR description and tracking as a follow-up.

2. **End-to-end testing.** The checklist says the change was manually tested end to end, but `GCPSkillRegistry.get_skill()` currently fails on the `?alt=media` 302 until #6824 lands. Could you describe the setup you used? I can run this branch against a real Agent Registry catalog with the redirect fix applied and report back here, if useful.

3. **Test coverage gaps** (inline on the test file): LAZY-mode `list_skills` output, no second `get_skill` call after prefetch, and the frontmatter-name-mismatch case.

4. **Docs/sample.** `contributing/samples/integrations/gcp_skill_registry_agent` would be a natural place to show `registry_skills=[...]`, and the `registry_skills` docstring could say that pinned skills appear in `list_skills` / the EAGER catalog and are fetched once per process (inline).

5. **Nit** (inline): the `prefetch()` calls in `ListSkillsTool.run_async` and `SearchSkillsTool.run_async` are redundant.

From my side, 1 and 3 are the ones I'd want addressed before this lands; 2 I can help with directly.

tests/unittests/tools/test_skill_toolset.py
codebee-aoki: To make this concrete, here are the three tests I had in mind. They pass on 4c469f8a (`188 passed` for the file) and follow the fixtures already used above, so they should drop in at the end of the file as-is. Feel free to rename or fold them into your existing tests.

```python
@pytest.mark.asyncio
async def test_skill_toolset_list_skills_tool_includes_pinned_registry_skill(
    mock_registry, mock_skill1, tool_context_instance
):
  """LAZY mode: the list_skills tool output must contain pinned skills."""
  mock_registry.get_skill.return_value = mock_skill1
  toolset = skill_toolset.SkillToolset(
      registry=mock_registry, registry_skills=["skill1"]
  )
  tool = skill_toolset.ListSkillsTool(toolset)

  result = await tool.run_async(args={}, tool_context=tool_context_instance)

  assert "<available_skills>" in result
  assert "skill1" in result
  mock_registry.get_skill.assert_called_once_with(name="skill1")


@pytest.mark.asyncio
async def test_skill_toolset_load_skill_does_not_refetch_pinned_registry_skill(
    mock_registry, mock_skill1, tool_context_instance
):
  """After prefetch, load_skill for a pinned name is served locally."""
  mock_registry.get_skill.return_value = mock_skill1
  toolset = skill_toolset.SkillToolset(
      registry=mock_registry, registry_skills=["skill1"]
  )
  tool_context_instance.state.get.return_value = None

  await toolset.prefetch()
  assert mock_registry.get_skill.call_count == 1

  tool = skill_toolset.LoadSkillTool(toolset)
  # Two invocations: the per-invocation cache must not be what serves this.
  for invocation_id in ("inv-1", "inv-2"):
    tool_context_instance.invocation_id = invocation_id
        args={"skill_name": "skill1"}, tool_context=tool_context_instance
    )
    assert result["skill_name"] == "skill1"

  assert mock_registry.get_skill.call_count == 1


@pytest.mark.asyncio
async def test_skill_toolset_prefetch_stores_skill_under_frontmatter_name(
    mock_registry,
):
  """A pinned name is marked fetched even when the archive's frontmatter name
  differs; the skill is stored under the frontmatter name."""
  fetched = mock.create_autospec(models.Skill, instance=True)
  fetched.name = "bar"
  mock_registry.get_skill.return_value = fetched
  toolset = skill_toolset.SkillToolset(
      registry=mock_registry, registry_skills=["foo"]
  )

  await toolset.prefetch()

  mock_registry.get_skill.assert_called_once_with(name="foo")
  assert toolset._get_skill("bar") is fetched
  assert toolset._get_skill("foo") is None
  assert toolset._registry_skills_loaded is True

  # No refetch on the next prefetch.
  await toolset.prefetch()
  assert mock_registry.get_skill.call_count == 1
```

codebee-aoki:
E2E result against a real Agent Registry (project-owned catalog, location `global`), running this branch at 4c469f8a with the redirect fix from #6824 cherry-picked on top so that `get_skill()` can actually download:

- `registry_skills=["private-evidence-page"]`, LAZY mode: the pinned skill shows up in the `list_skills` output, `load_skill` on it is served locally (4204 chars of instructions), and across two simulated invocations `registry.get_skill` was called exactly once. EAGER mode: the skill is in the injected `<available_skills>` catalog, again with a single fetch. So the core behaviour works end to end.

One finding worth acting on in this PR. In real catalogs the registry resource id and the archive's frontmatter name differ as a rule, not as an exception:

| registry id (what you pin / what `search_skills` returns) | frontmatter name (what ends up in the catalog) |
|---|---|
| `private-evidence-page` | `evidence-page` |
| `cloud.google.com-google-cloud-networking-observability` | `google-cloud-networking-observability` |

Consequence: after prefetch, `load_skill("private-evidence-page")` (the name the model would get from `search_skills`) misses `_skills`, falls through to `_get_or_fetch_skill`, and re-downloads the archive (measured: 1 extra registry call per invocation). Suggest keeping a small alias map, e.g. `self._registry_skill_aliases[requested_name] = skill.name`, populated in `prefetch()` and consulted in `_get_or_fetch_skill()` before the registry fallback. That closes the gap without adding a second catalog entry.

(The second row is also why Google-published skills currently cannot be fetched at all: their ids contain dots, which `get_skill()`'s name validation and `Frontmatter` reject. That is independent of this PR; I'll file it separately.)

src/google/adk/tools/skill_toolset.py
ArulJerald: Updated the PR description to explicitly call out this behavior. I will open a follow-up issue to bound retry attempts for permanently failing skill names (or restrict retries to transient network exceptions).

src/google/adk/tools/skill_toolset.py
ArulJerald: Done. Expanded the `registry_skills` docstring to explain:
- Pinned skills are fetched once per process on first use.
- They appear in `list_skills` and the `EAGER` catalog and are served locally without further registry calls.
- Unpinned registry skills remain accessible via `search_skills`.
- Skills are stored under the archive's frontmatter name, which may differ from the registry resource name.

src/google/adk/tools/skill_toolset.py
ArulJerald: Good catch. Removed `await self._toolset.prefetch()` from both `ListSkillsTool.run_async` and `SearchSkillsTool.run_async`.

tests/unittests/tools/test_skill_toolset.py
ArulJerald: Added all three tests:
1. `test_skill_toolset_list_skills_tool_includes_pinned_registry_skill`: Verifies LAZY mode `list_skills` output includes pinned skills.
2. `test_skill_toolset_load_skill_does_not_refetch_pinned_registry_skill`: Verifies that `load_skill` is served from local memory across multiple invocations without re-fetching (`call_count == 1`).
3. `test_skill_toolset_prefetch_stores_skill_under_frontmatter_name`: Verifies frontmatter name storage, alias resolution by requested registry id, and that `LoadSkillTool` resolves the alias locally without re-downloading.

ArulJerald:
Thank you @codebee-aoki  for validating this against the real catalog and identifying the name mismatch re-download issue!

Implemented your suggested alias map:
- Added `self._registry_skill_aliases: dict[str, str] = {}` populated during `prefetch()` (`self._registry_skill_aliases[name] = skill.name`).
- In `_get_or_fetch_skill()`, checked `self._registry_skill_aliases` before falling through to the registry fetch.
- In `SearchSkillsTool.run_async`, filtered out registry search results matching either the stored name or `self._registry_skill_aliases`.
- Added `test_skill_toolset_search_skills_filters_aliased_registry_skill` to assert that both the registry ID and frontmatter name are filtered from search results.

codebee-aoki:
Re-verified on d4ba96aa with #6824 cherry-picked on top: 219 unit tests pass, and the E2E against the real catalog now shows zero extra registry calls when `load_skill` is invoked with the registry id (the alias map works). With #7138 applied as well, pinning `cloud.google.com-google-cloud-networking-observability` works end to end: it appears in the catalog under its frontmatter name, is served locally, and is fetched exactly once. Nothing further from my side.

ArulJerald:
Thank you @codebee-aoki for reviewing this.

codebee-aoki (approved):
All five points from my earlier review are addressed in d4ba96aa, and I re-verified it against a real Agent Registry catalog (with #6824 applied): pinned skills appear in `list_skills` and the EAGER catalog, are fetched once per process, and `load_skill` by either the frontmatter name or the registry id is served locally. 219 unit tests pass. As the author of #7130, this implements what the issue asked for. (Non-binding approval; a maintainer review is still needed.)

PR #7138, opened 2026-09-16
fix(skills): accept dotted registry ids in GCPSkillRegistry

### LINKED ISSUE #7136: GCPSkillRegistry rejects every Google-published skill: dotted ids (cloud.google.com-*) fail name validation in get_skill and search_skills
Follow-up to #6838 / #6839: that fix stopped `search_skills` from crashing on these ids, but the ids themselves are still rejected everywhere, so Google-published skills remain unusable from ADK.

## 🔴 Required Information

**Describe the Bug:**
Skills published by Google into Agent Registry have resource ids of the form `cloud.google.com-<display-name>` (also `discoveryengine.googleapis.com-<name>`). ADK cannot use any of them:

- `GCPSkillRegistry.get_skill(name="cloud.google.com-google-cloud-networking-observability")` raises `ValueError: Invalid skill name ...` because the name check added in 11101acc only accepts kebab/snake case (no dots).
- `GCPSkillRegistry.search_skills()` feeds each id into `Frontmatter(name=...)`, whose validator also rejects dots, so since 3c977bc2 (#6838) every Google-published hit is dropped with a "Skipping search result" warning. `tests/unittests/integrations/skill_registry/test_gcp_skill_registry.py` even uses `cloud.google.com-agent-platform-eval-flywheel` as the example of a "bad" name.

In a real project catalog (location `global`) this rejects 113 of 117 skills; the only 4 that pass are ones we created ourselves with plain kebab ids. With the validation bypassed locally, downloading the Google skill works (SKILL.md with 7184 chars, 8 resources; the frontmatter name inside the archive equals the displayName and is valid kebab-case), so the archives themselves are fine. It is only the registry *id* that fails the SKILL.md naming rule.

**Steps to Reproduce:**
1. `google-adk` main (ac0133a4) or 2.9.1, project with Agent Registry enabled (Google-published skills are visible in the Console).
2. `registry = GCPSkillRegistry(project_id=..., location="global")`
3. `await registry.search_skills(query="networking")` -> only self-created skills are returned; one warning per Google skill.
4. `await registry.get_skill(name="cloud.google.com-google-cloud-networking-observability")` -> `ValueError`.

**Expected Behavior:**
Google-published skills are searchable and loadable. The security intent of 11101acc (single path segment, no traversal) can be kept by validating the id as a safe path segment (e.g. `^[A-Za-z0-9._-]+$`, and not `.`/`..`) instead of applying the SKILL.md frontmatter naming rule to a registry resource id. `search_skills` should not validate registry ids with the frontmatter rule either; that rule belongs to the SKILL.md inside the archive, which does pass.

**Observed Behavior:**
`ValueError` from `get_skill`, and silent exclusion from `search_skills`.

**Environment Details:**
- ADK Library Version: main @ ac0133a4, also 2.9.1
- Reproduced locally (macOS, Python 3.14) against a real Agent Registry catalog

**Model Information:**
- Are you using LiteLLM: No
- Which model is being used: N/A (registry client only)

## Additional Context

- #6838 reported this exact catalog entry (`cloud.google.com-agent-platform-eval-flywheel`) but asked only for the crash to be fixed; #6839 implemented "skip and log", and its description states this matches `get_skill()` already rejecting those names. So the current behaviour is deliberate at the symptom level, but it means every Google-published skill (113/117 in our catalog) is unreachable from ADK, which I don't think was the intent. This issue is about accepting those ids, not about the crash.
- Related: #6908 / #6824 (redirect on media download; needed for the download to succeed at all), #7130 / #7135 (pinning registry skills). Note that registry id and frontmatter name differ for 117/117 skills in the catalog, which is why #7135 needs to key the pinned skill by both names.
- Happy to send a PR.

chelsealong:

- Closes: #7136

**Problem:**

Skills published by Google into Agent Registry have resource ids of the
form `cloud.google.com-<display-name>` (also
`discoveryengine.googleapis.com-<name>`). `GCPSkillRegistry` rejects all of
them:

- `get_skill(name="cloud.google.com-...")` raises `ValueError: Invalid
  skill name ...` because the check added for #6839 requires the name to
  match `models._SNAKE_OR_KEBAB_NAME_PATTERN`, which does not allow dots.
- `search_skills()` feeds each catalog id into `models.Frontmatter(name=...)`,
  whose own field validator applies the same strict kebab/snake-case rule,
  so every Google-published hit is silently dropped with a "Skipping search
  result" warning.

In a real catalog this rejects the large majority of skills — only
self-created, plain-kebab-case skills pass. That name-pattern check was
never meant to apply here in the first place: it is the SKILL.md
frontmatter naming rule (a content-format rule for the file inside the
skill archive), and a registry resource id is a different kind of string
that happens to reuse the same `Frontmatter.name` field for convenience.

**Solution:**

- Added a `_is_safe_registry_id` helper in `gcp_skill_registry.py` with its
  own pattern (`^[a-z0-9]+(?:[._-][a-z0-9]+)*$`, length <= 64) that keeps
  the original security intent (reject `.`/`..`, slashes, and anything else
  that isn't safe to interpolate as a single URL path segment) while
  allowing dots.
- `get_skill` now validates the incoming name against this registry-id rule
  instead of the SKILL.md frontmatter name pattern.
- `search_skills` now validates each catalog id the same way, and
  constructs the returned `Frontmatter` via `model_construct` (bypassing
  the frontmatter name validator, which does not apply to registry ids)
  while still running the real `description` validation so malformed
  descriptions are still skipped and logged as before.

Neither change touches `models.Frontmatter`'s own naming rule, which still
applies, unmodified, to names parsed from SKILL.md content.

### Testing Plan

**Unit Tests:**

Added:
- `test_search_skills_accepts_dotted_registry_id` — a dotted registry id
  (`cloud.google.com-agent-platform-eval-flywheel`) is now returned by
  `search_skills` instead of being dropped.
- Extended `test_get_skill_builds_expected_url_for_valid_name` with the same
  dotted id to confirm `get_skill` accepts it and builds the expected URL.
- Extended `test_get_skill_rejects_unsafe_name_before_any_request` with
  `"."`, `".."`, and a 65-character name to confirm the safe-path-segment
  check still rejects bare traversal segments and enforces the length cap.
- Updated `test_search_skills_skips_entry_failing_validation`'s first case
  (previously the dotted id, used as an example of a name that fails
  validation) to `".."`, since a dotted id is now valid.

Verified the added tests fail without the fix (`git checkout HEAD~1 --
src/google/adk/integrations/skill_registry/gcp_skill_registry.py`, keeping
the new tests) with the exact errors the issue describes:

```
FAILED .../test_gcp_skill_registry.py::test_search_skills_accepts_dotted_registry_id - assert 0 == 1
FAILED .../test_gcp_skill_registry.py::test_get_skill_builds_expected_url_for_valid_name[cloud.google.com-agent-platform-eval-flywheel] - ValueError: Invalid skill name 'cloud.google.com-agent-platform-eval-flywheel': name must be lowercase kebab-case (a-z, 0-9, hyphens) or snake_case (a-z, 0-9, underscores), with no leading, trailing, or consecutive delimiters.
3 failed, 31 passed, 1 warning in 0.61s
```

Summary of passed pytest results (with the fix restored):

```
$ pytest tests/unittests/integrations/skill_registry/test_gcp_skill_registry.py -q
34 passed in 0.54s

$ pytest tests/unittests/integrations/skill_registry/ tests/unittests/skills tests/unittests/tools/test_skill_toolset.py -q
284 passed, 1 warning in 5.04s

$ pytest tests/unittests -q -n auto
14990 passed, 82 skipped, 27 xfailed, 2 xpassed in 146.92s
```
(One unrelated test, `test_eval_injects_session_input_state_into_instruction`,
is flaky under `-n auto` parallel execution and reproduces identically on
unmodified `main`; deselected from the full run above for a clean signal.)

Also ran `ruff check`, `pyink --check`, and `pylint` on the changed files —
clean (pylint's only remaining note is a pre-existing line-length warning on
an unrelated line this PR doesn't touch).

Not run — no live Agent Registry project was available in this
environment; verified via the unit tests above, which exercise the same
name-validation and Frontmatter-construction code paths the issue reports
as broken.

### Checklist

### Additional context

This PR was generated by an AI coding agent (Claude Code). The implementation
and tests were verified against the automated test suite as described above.


src/google/adk/integrations/skill_registry/gcp_skill_registry.py
```diff
@@ -36,10 +37,22 @@
...
+# frontmatter names: they are not required to be kebab/snake-case and may
+# contain dots. They still need to be safe to interpolate as a single URL
+# path segment, so they get their own, more permissive check instead of
+# reusing the SKILL.md content naming rule.
+_SAFE_REGISTRY_ID_PATTERN = re.compile(r"^[a-z0-9]+(?:[._-][a-z0-9]+)*$")
+
+
+def _is_safe_registry_id(name: str) -> bool:
+  """True if `name` is safe to use as a single skill-registry path segment."""
+  return len(name) <= 64 and bool(_SAFE_REGISTRY_ID_PATTERN.match(name))
```
codebee-aoki: The 64-character cap rejects 7 of the 117 real ids in the catalog, e.g. `cloud.google.com-google-cloud-solution-agentic-analytics-spark-knowledge-catalog` (80 chars) and `cloud.google.com-gke-ai-troubleshooting-handle-disruption-gpu-tpu` (65 chars). That limit comes from the SKILL.md frontmatter rule, not from anything about URL-segment safety, so I'd drop it (or raise it well above 80, e.g. 256) and replace the `"a" * 65` "unsafe" case in the tests with a test that a long real id is accepted.

codebee-aoki (commented):
Verified this branch against a real Agent Registry catalog (117 skills, location `global`) with #6824 applied so downloads succeed: traversal-style names are all still rejected, `search_skills("networking")` now returns the Google-published ids, and `get_skill("cloud.google.com-google-cloud-networking-observability")` downloads and parses fine. One problem, inline below. Everything else looks right to me.

chelsealong:
Raised the registry-id length cap from 64 to 256 chars and added the two real catalog ids you flagged (80 and 65 chars) as regression tests in `test_get_skill_builds_expected_url_for_valid_name`. The `"a" * 65` "unsafe" cases are now `"a" * 257` to keep testing the length bound at the new cap.

codebee-aoki:
Re-verified 314172ee with #6824 on top against the real catalog: 117/117 ids now pass the registry-id check (was 110/117 at the 64-char cap), traversal-style names are still rejected, `search_skills` returns the Google-published skills, and `get_skill` on a dotted id downloads and parses. 213 unit tests pass locally. Nothing further from my side.

PR #7145, opened 2026-09-16
fix(mcp): refuse the reserved tool name set_model_response

### LINKED ISSUE #7144: set_model_response is missing from the MCP _RESERVED_TOOL_NAMES guard
`_RESERVED_TOOL_NAMES` in `src/google/adk/tools/mcp_tool/mcp_tool.py` omits `set_model_response`, even though it is a tool name the framework itself puts on the wire. The guard added in #6796 states its own invariant in the comment above the set:

```python
# Tool names the framework itself puts on the wire. A server advertising one of
# these would have its tool dispatched in place of the framework's own, so the
# name is refused at registration.
_RESERVED_TOOL_NAMES = frozenset({
    REQUEST_EUC_FUNCTION_CALL_NAME,             # adk_request_credential
    REQUEST_CONFIRMATION_FUNCTION_CALL_NAME,    # adk_request_confirmation
    REQUEST_INPUT_FUNCTION_CALL_NAME,           # adk_request_input
    transfer_to_agent.__name__,                 # transfer_to_agent
})
```

`set_model_response` satisfies that description but is not in the set, so an MCP server advertising it is accepted where the other four are refused.

### `set_model_response` is a framework-owned tool name

* Defined as `def set_model_response() -> str:` in `src/google/adk/tools/set_model_response_tool.py` (`self.func = set_model_response`), so that is its declaration name.
* Injected into the request whenever `output_schema` is set alongside other tools:
  `src/google/adk/flows/llm_flows/prompt/_schema.py:54`
  ```python
  set_response_tool = SetModelResponseTool(agent.output_schema)
  llm_request.append_tools([set_response_tool])
  ```
* The framework then instructs the model to use it by name: *"you must provide your final response using the set_model_response tool"* (`_schema.py:57-64`).
* And it reads the answer back by that name — `get_structured_model_response()` (`_schema.py:97`) is reached from `base_llm_flow.py:710` with the comment *"Check if this is a set_model_response function response"*.

### Why that matters at an MCP boundary

`LlmRequest.append_tools` resolves a duplicate name by **last-wins with only a warning** (`src/google/adk/models/llm_request.py`):

```python
if tool.name in self.tools_dict:
    # Both declarations are still advertised to the model, but only one
    # tool can hold the name, so calls land on the survivor.
    logging.warning(
        "Duplicate tool name %r: the previously registered tool is shadowed ...",
        tool.name,
    )
self.tools_dict[tool.name] = tool
```

So if an MCP server advertises `set_model_response`, one of the two tools holds the name and the model's `set_model_response` call is dispatched there. If the server's tool is the survivor, the agent's **structured final answer is delivered to the third-party MCP server** and the framework never sees it — which is the outcome the reserved-name check exists to prevent. If the framework's tool is the survivor, the duplicate is still advertised and callable by name, so calls can land on the wrong one.

Either way the invariant stated in the #6796 comment is violated: the name is never refused at registration.

### Reproduction sketch

```python
agent = LlmAgent(
    model=..., tools=[McpToolset(connection_params=<server that advertises
                                 a tool named "set_model_response">)],
    output_schema=MySchema,
)
# Expected: ValueError, as for transfer_to_agent / adk_request_confirmation
# Actual:   accepted; a warning is logged at request build time
```

### Suggested fix

Add the name to the set:

```python
_RESERVED_TOOL_NAMES = frozenset({
    REQUEST_EUC_FUNCTION_CALL_NAME,
    REQUEST_CONFIRMATION_FUNCTION_CALL_NAME,
    REQUEST_INPUT_FUNCTION_CALL_NAME,
    transfer_to_agent.__name__,
    'set_model_response',
})
```

If it is useful, the same reasoning may apply to other in-model names the framework injects (`google_search`, `google_maps`, `url_context`, `vertex_ai_search`, `code_execution`, `load_artifacts`, `load_memory`) — I have not verified each of those reaches `tools_dict`, so I have deliberately kept this report to the one name I traced end to end.

Happy to send the set change plus a unit test alongside `test_mcp_tool.py` if that is the preferred route.

sushant-me:
Fixes #7144

## Problem

`_RESERVED_TOOL_NAMES` exists to refuse MCP tool names the framework itself puts on the wire. `set_model_response` belongs to that class but was not in the set, so a server advertising it was accepted where `transfer_to_agent` and the `adk_request_*` names are rejected.

It is a framework-owned wire name on every count:

* `SetModelResponseTool` is injected whenever `output_schema` is set alongside other tools — `flows/llm_flows/prompt/_schema.py:54`
  ```python
  set_response_tool = SetModelResponseTool(agent.output_schema)
  llm_request.append_tools([set_response_tool])
  ```
* the framework then instructs the model to answer **through that name** (`_schema.py:57-64`), and
* `base_llm_flow.py` reads the result back by the same name — *"Check if this is a set_model_response function response"*.

`LlmRequest.append_tools` resolves the resulting duplicate by last-wins with only a warning:

```python
if tool.name in self.tools_dict:
    # Both declarations are still advertised to the model, but only one
    # tool can hold the name, so calls land on the survivor.
    logging.warning("Duplicate tool name %r: ...", tool.name)
self.tools_dict[tool.name] = tool
```

So a server-supplied `set_model_response` competes for the name, and if it is the survivor the agent's structured final answer is dispatched to the third-party server rather than captured by the framework — precisely the outcome this guard was added to prevent.

## Fix

```diff
 _RESERVED_TOOL_NAMES = frozenset({
     REQUEST_EUC_FUNCTION_CALL_NAME,
     REQUEST_CONFIRMATION_FUNCTION_CALL_NAME,
     REQUEST_INPUT_FUNCTION_CALL_NAME,
     transfer_to_agent.__name__,
+    # Injected by the output-schema processor whenever output_schema is set
+    # alongside other tools (flows/llm_flows/prompt/_schema.py) and read back
+    # by name in base_llm_flow.py, so it is a framework-owned wire name too.
+    # Spelled out because the function is defined inside
+    # SetModelResponseTool.__init__ and is not importable.
+    'set_model_response',
 })
```

The name is written literally rather than as `something.__name__`, unlike `transfer_to_agent`: the callable is defined *inside* `SetModelResponseTool.__init__`, so there is no module-level binding to import. Importing it would have raised `ImportError` for every importer of `mcp_tool`.

## Tests

Both existing reserved-name tests are extended rather than duplicated:

* `test_mcp_tool.py::test_init_reserved_name` — adds `"set_model_response"` to the parametrized list, asserting `MCPTool` construction raises `ValueError`.
* `test_mcp_toolset.py::test_get_tools_skips_reserved_names` — the server now also advertises `set_model_response`; the assertion that only `valid_tool` survives is unchanged.

## Scope note

I checked only this name end to end. Other in-model names (`google_search`, `url_context`, `code_execution`, …) may warrant the same treatment, but I could not confirm each reaches `tools_dict`, so they are deliberately not included here.


sushant-me:
Adding the reachability argument, because the fix only matters if the server's tool is actually the survivor rather than merely a duplicate.

`LlmRequest.append_tools` resolves a repeated name by last-wins — `models/llm_request.py:309`:

```python
self.tools_dict[tool.name] = tool
```

So the question is which registration runs second. In `BaseLlmFlow._preprocess_async`:

* **`base_llm_flow.py:536`** — `for processor in self.request_processors:` runs `_output_schema_processor.request_processor` (`single_flow.py:72`), which is the code that appends `set_model_response` at `prompt/_schema.py:54`.
* **`base_llm_flow.py:555`** — only *after* that loop completes: `await _process_agent_tools(invocation_context, llm_request)`, which is where the agent's toolsets are resolved and MCP tool declarations enter the request.

The framework's name is registered in the earlier step and the server's in the later one, so the server's tool is the survivor. The shadowing is reachable, not theoretical.

While confirming that, I checked the other tool names the framework injects, and I do **not** believe they belong in `_RESERVED_TOOL_NAMES`:

* `finish_task` (`agents/llm_agent.py:1337`) and `task_completed` (`agents/sequential_agent.py:208`) are appended to `agent.tools` at construction. They therefore resolve in the *same* pass as the MCP tools and are appended last, so the framework wins and the server's tool is the one that gets shadowed.
* `adk_handle_model_error` (`plugins/_reflect_retry_model_plugin.py:175`) assigns `llm_request.tools_dict[...]` directly instead of going through `append_tools`, so it overwrites whatever held the name rather than being overwritten.

`set_model_response` is the one I found that sits in the class the guard exists for: registered by a request processor, and therefore before tool resolution. Happy to be corrected on any of the above.

sushant-me:
Rebased onto `main` — the branch was 40 commits behind, so a force-push rewrote it. The diff is unchanged: three files, +8 lines.

While rebasing I re-checked that the finding still reproduces at the new base, because an outdated report is worse than no report. It does. `LlmRequest.append_tools` still warns and then overwrites unconditionally:

```python
if tool.name in self.tools_dict:
  logging.warning(
      "Duplicate tool name %r: the previously registered tool is"
      " shadowed and can no longer be called.",
      tool.name,
  )
self.tools_dict[tool.name] = tool
```

So a remote MCP server advertising `set_model_response` still takes the name, and the framework's own tool becomes uncallable. The warning added in `b26d4f67` describes that exact outcome, which is the case for refusing the reserved name at registration rather than reporting it after the fact.

PR #7151, opened 2026-09-17
fix: enforce denied tool confirmations centrally

### LINKED ISSUE #7148: ToolConfirmation deny verdict is never enforced by the framework: custom BaseTool executes after user clicks Decline (adk-python 2.9.1)
## Summary

ADK's tool-confirmation feature is designed so that a sensitive tool call pauses for explicit human approval before any side effect occurs. The human's decision is delivered back to the framework as a `ToolConfirmation` object whose `confirmed` boolean carries the approve/deny verdict.

The framework, however, **never reads `confirmed`**. The resume path (`google/adk/flows/llm_flows/request_confirmation.py:: _resolve_confirmation_targets`) validates that the tool call is registered, requires confirmation, and that the arguments match — but never checks the verdict — and then hands the call to `functions.py:: _execute_single_function_call_async`, which attaches the `ToolConfirmation` to the `ToolContext` **as advisory data** and unconditionally invokes `tool.run_async(...)`.

Whether "Decline" actually prevents the side effect is left to each individual tool implementation. Built-in tools (`FunctionTool.run_async` at `function_tool.py:350`, plus BashTool, MCPTool, ComputerUseTool) each re-implement the deny check inside their own `run_async`. But `BaseTool` subclassing is the **documented extension point** for custom/enterprise tools. Any custom `BaseTool` that opts into confirmation via the documented `check_require_confirmation` / `request_confirmation` API but does not itself honor `tool_confirmation.confirmed` (by bug, omission, or intent) **executes its side effect after the user explicitly declined**.

## Root cause

1. `google/adk/flows/llm_flows/request_confirmation.py` — `_resolve_confirmation_targets(...)`: validates registration / requires-confirmation / argument match, and re-executes confirmed tools — but there is **no branch that inspects `ToolConfirmation.confirmed`**. A `confirmed=False` response takes exactly the same code path as `confirmed=True`.
2. `google/adk/flows/llm_flows/functions.py` — `_execute_single_function_call_async`: passes `tool_confirmation` into `ToolContext` and calls `tool.run_async(...)` unconditionally; no central check of `.confirmed`.
3. Deny enforcement exists only **inside** four built-in tools' own `run_async` implementations (`function_tool.py:350` etc.). `BaseTool` subclassing is the documented extension point, so every custom tool must independently re-implement deny handling — an easy-to-miss, unenforced contract.

## Reproduction (fully offline, deterministic; no LLM API needed)

Save as `poc.py` and run with `python poc.py` against `google-adk==2.9.1`:

```python
import asyncio, json, os, tempfile
from google.adk.models import BaseLlm
from google.adk.tools import BaseTool

MARKER = None
EXECUTIONS = []

class FakeLlm(BaseLlm):
  def __init__(self):
    super().__init__(model='fake-llm')
    object.__setattr__(self, 'calls', 0)
  async def generate_content_async(self, llm_request, stream=False):
    n = self.calls
    object.__setattr__(self, 'calls', n + 1)
    if n == 0:
      yield LlmResponse(content=types.Content(role='model', parts=[
          types.Part.from_function_call(name='dangerous_custom',
                                        args={'target': 'prod-db'})]))
    else:
      yield LlmResponse(content=types.Content(role='model',
          parts=[types.Part.from_text(text=f'turn {n} done')]))

class DangerousCustomTool(BaseTool):
  """Custom BaseTool that opts into confirmation."""
  def __init__(self):
    super().__init__(name='dangerous_custom', description='side-effecting tool')
  def _get_declaration(self):
    return types.FunctionDeclaration(name=self.name, description=self.description,
        parameters=types.Schema(type=types.Type.OBJECT,
            properties={'target': types.Schema(type=types.Type.STRING)}))
  async def check_require_confirmation(self, args, tool_context) -> bool:
    return True
  async def run_async(self, *, args, tool_context):
    if not tool_context.tool_confirmation:
      tool_context.request_confirmation(hint='Allow destructive action?')
      return {'status': 'awaiting_user_confirmation'}
    EXECUTIONS.append(tool_context.tool_confirmation.confirmed)
    with open(MARKER, 'w') as f:
      f.write('EFFECT HAPPENED. user_confirmed=%s\n'
              % tool_context.tool_confirmation.confirmed)
    return {'status': 'executed',
            'user_confirmed': tool_context.tool_confirmation.confirmed}

async def main():
  global MARKER
  MARKER = os.path.join(tempfile.mkdtemp(prefix='poc_'), 'effect_marker.txt')
  agent = LlmAgent(name='root', model=FakeLlm(), tools=[DangerousCustomTool()])
  runner = Runner(app_name='poc', agent=agent, session_service=InMemorySessionService())
  s = await runner.session_service.create_session(app_name='poc', user_id='u')

  fc_id = None
  async for ev in runner.run_async(user_id='u', session_id=s.id,
      new_message=types.Content(role='user', parts=[types.Part.from_text(text='do it')])):
    for fc in ev.get_function_calls():
      if fc.name == 'adk_request_confirmation':
        fc_id = fc.id
  print('turn1 marker exists (expect False):', os.path.exists(MARKER))

  deny = types.Content(role='user', parts=[types.Part(function_response=
      types.FunctionResponse(name='adk_request_confirmation', id=fc_id,
                             response={'confirmed': False}))])
  async for ev in runner.run_async(user_id='u', session_id=s.id, new_message=deny):
    for fr in ev.get_function_responses():
      print('turn2 function_response:', fr.name, '->', fr.response)

  print('EFFECT EXECUTED DESPITE DENY:', os.path.exists(MARKER))
  print('run_async verdicts seen:', EXECUTIONS)

asyncio.run(main())
```

### Observed output (google-adk 2.9.1, Python 3.12)

```
turn1 marker exists (expect False): False
turn2 function_response: dangerous_custom -> {'status': 'executed', 'user_confirmed': False}
EFFECT EXECUTED DESPITE DENY: True
marker content: EFFECT HAPPENED. user_confirmed=False args={"target": "prod-db"}
run_async verdicts seen: [False]
```

### Expected output

The framework should refuse to resume the tool when `confirmed` is `False`: it should
return a denial function response to the model and never invoke `run_async` — or at
minimum provide a framework-level default-deny hook so custom tools cannot forget it.

## Impact

- **Direct verdict-to-effect break**: the human explicitly declined, yet the sensitive
  operation executed (`user_confirmed=False` inside the executed effect). The
  confirmation dialog is decorative for any tool that does not self-enforce.
- **Documented extension point affected**: `BaseTool` subclassing is the standard way
  enterprises wrap internal sensitive operations (DB mutations, deployments, payments).
  An implementation that forgets the deny check turns the approval dialog into a no-op.
- **Prompt-injection amplification**: an injected model instruction that triggers a
  confirmation-gated tool now needs only a distracted user clicking "Decline" for the
  action to execute anyway; the denial is even echoed to the model inside the executed
  result payload (`user_confirmed: False`).

## Suggested fix

In `_resolve_confirmation_targets` (request_confirmation.py), split re-execution
targets by verdict: only `confirmed is True` calls proceed to
`_execute_single_function_call_async`; `confirmed=False` calls should receive a
framework-generated denial function response, never re-entering `run_async`.
Additionally, expose a `BaseTool.on_confirmation_denied(...)` default implementation
so tools can customize the denial message without being responsible for stopping
execution.

---

### Disclosure note

This finding was reported through Google's Vulnerability Reward Program
([issue 562236334](https://issuetracker.google.com/issues/562236334)). The Google Bug
Hunter team reviewed the report and responded that it is a "design choice or framework
ergonomics issue rather than a security vulnerability," and suggested: *"Feel free to
disclose this on the project's GitHub issues page as a public issue."* This issue is
filed per that guidance.

---

Reporter: **Chengzhi Yi** — yimou@hust.edu.cn — GitHub: @Tardfyou
Happy to provide the full PoC files, control harness, and any additional details.

AnvitDevadiga:
## What changed

Enforce ToolConfirmation decisions at the ADK framework boundary.

- Declined confirmations now return a framework-generated denial response.
- Denied tools are never passed to BaseTool.run_async.
- Approved tools retain the existing execution path.
- Mixed approved and denied batches are handled independently.

## Why

Custom BaseTool implementations can currently execute after a user clicks Decline unless each tool reimplements the denial check. This fixes the issue described in #7148 and makes the documented confirmation flow default-deny.

## Testing

- 39 targeted confirmation, tool, and runner tests pass.
- Ruff checks pass.

Closes #7148


babyblueviper1:
Good fix — moving denial enforcement to the framework boundary instead of trusting each `BaseTool` to reimplement it is the right shape (matches the general lesson: a verdict object is only as strong as the code path that's forced to consume it, not the code path that happens to check it).

One adjacent question worth asking before merge, not asserted as a bug since I haven't reproduced it against this exact tree: does the resume path guard against the *same* `function_call_id`/`ToolConfirmation` being submitted more than once? If a client can call the resume endpoint twice with an identical `confirmed=True` payload for a `function_call_id` that's already been executed, does `_resolve_confirmation_targets` re-execute it, or is there state elsewhere marking that confirmation as consumed?

Reason I ask: we hit the identical shape in our own verdict-issuing service — a pure function computing a decision reference had no state, so an identical still-valid verdict could be replayed to authorize a second dispatch (caught via a negative-control test explicitly checking for replay, not by inspection). The fix there was a small addition on top of the same "enforce at the boundary" pattern this PR already uses: a primary-keyed consumed-set with atomic check-and-mark, so a race between two identical resume calls has exactly one winner instead of two executions. If `tools_to_resume_with_confirmation` here is keyed by `function_call_id` and nothing marks an id as spent after this code path runs, the same gap likely exists — cheap to close now while the boundary-enforcement code is already being touched, much easier than adding it after the fact once callers start depending on today's semantics.

Tardfyou:
Following @surajksharma07's request in #7148, I validated the proposed shared-caller workaround locally on google-adk==2.9.1 using the original deterministic fake-LLM/custom-BaseTool harness. The only effect is a temporary marker file; no external model or service is contacted.

Results:

| Caller | User verdict | Tool executions after verdict | Marker written |
|---|---|---:|---|
| Unmodified 2.9.1 | confirmed=False | 1 | yes |
| Caller denies confirmed=False before run_async | confirmed=False | 0 | no |
| Same workaround | confirmed=True | 1 | yes |
| Unmodified 2.9.1 | confirmed=True | 1 | yes |

All four assertions passed. This validates the suggested caller-level workaround for the reported custom-tool path.

One distinction matters before treating this as validation of this PR: the current head 3557a4139707950a647849aa3c9bce44bcae5533 filters denied calls in tools/_confirmation.py, whereas the workaround I tested guards _call_tool_async (in _tool_caller.py in release 2.9.1). I have not run this PR head, so this comment is not an end-to-end verification of #7151. Please retain a runner-level regression with a custom BaseTool that does not inspect the verdict, plus the approved control. The replay question above is also outside the scope of my four-case check.

AnvitDevadiga:
Thanks for pointing this out. You’re right that the existing event-history deduplication does not provide an atomic check-and-mark guarantee for concurrent resume requests.

I updated the implementation so each confirmation is claimed atomically by the invocation context before dispatch. A repeated or concurrent submission for the same `function_call_id` is now ignored, ensuring the tool can execute at most once per confirmation.

I also added the protection for denied confirmations, since those should be consumed exactly once as well. Syntax and diff validation pass; the targeted test suite is currently blocked in this environment because PyPI access is unavailable.

babyblueviper1:
Verified this against the actual PR head (6627ed47), not just the comment thread — the fix is the right shape: `_consumed_tool_confirmation_ids` is a primary-keyed set on `InvocationContext`, claimed under `_tool_confirmation_consume_lock` before dispatch, exactly the atomic check-and-mark pattern that closes the race from my original question (a confirmed=True payload for the same `function_call_id` submitted twice, one winner).

I went one step further before calling it closed: does this survive a crash/resume into a *new* process? `_consumed_tool_confirmation_ids` is a Pydantic `PrivateAttr` on `InvocationContext`, and private attrs don't round-trip through model serialization — so a fresh process resuming a persisted session gets an empty consumed-set. Read through `_confirmation.py`'s full pipeline to check whether that reopens anything, and it doesn't: Step 2 (event-history dedup, `responded_fc_ids` from the persisted event log) runs *before* Step 3's in-memory claim and drops any confirmation that already has a recorded `function_response` — that check is durable across process boundaries because the event log is what gets persisted, not the in-memory set. So the in-memory set only has to be correct for races *within one process's lifetime* on a *still-unanswered* confirmation, which is exactly what it does. No residual gap at the resume boundary — the two checks are at different scopes and compose correctly.

What the PR doesn't have yet: a direct regression test for the concurrency race itself. The diff's two test changes just update expected deny-path messages — nothing exercises two simultaneous calls into `_consume_tool_confirmation` (or two concurrent resume calls) for the same `function_call_id` to assert exactly one wins. Happy to write that against 6627ed47 (`asyncio.gather` of two confirm-and-dispatch attempts, assert one executes and one no-ops) — same pattern we run internally for our own consumed-set on decision-reference reuse. Say the word and I'll open it as a PR.

PR #7155, opened 2026-09-17
fix(tools): reap ADK sessions of dead MCP connections in to_mcp_server

### LINKED ISSUE #7154: to_mcp_server never deletes the ADK session of a dead MCP connection; a stateless streamable HTTP deployment leaks one session per tool call
**Describe the bug**

`to_mcp_server` keeps one ADK session per MCP connection in a `weakref.WeakKeyDictionary` (`src/google/adk/tools/mcp_tool/_agent_to_mcp.py`). When a connection is garbage collected, the weak map drops the entry, but nothing ever calls `session_service.delete_session`, so the ADK session, with its full event history, stays in the session service forever. The default runner built by `to_mcp_server` uses `InMemorySessionService`, which is a plain dict with no eviction.

Two consequences:

1. Stateful serving: a long running server accumulates one dead conversation per closed connection.
2. Stateless streamable HTTP, which is the mode you want behind an autoscaler such as Cloud Run: the MCP SDK builds a completely fresh transport for every request (`StreamableHTTPSessionManager` with `stateless=True`), so `_connection_key` returns a per request object and every tool call creates a new ADK session that is never deleted. That is one leaked session per tool call, growing linearly with traffic, and the sessions hold full conversation content.

**To Reproduce**

The script below makes 10 short lived in-memory MCP connections with one tool call each, which is exactly the connection pattern a stateless streamable HTTP deployment produces, then counts the sessions left in the service.

```python
import asyncio
from contextlib import asynccontextmanager
import gc

import anyio
from google.adk.agents.base_agent import BaseAgent
from google.adk.auth.credential_service.in_memory_credential_service import (
    InMemoryCredentialService,
)
from google.adk.tools.mcp_tool import to_mcp_server
from mcp.client.session import ClientSession
from mcp.shared.memory import create_client_server_memory_streams


@asynccontextmanager
async def connected_client(lowlevel_server):
  """One in-memory MCP connection, the same shape a network client makes."""
  async with create_client_server_memory_streams() as (
      client_streams,
      server_streams,
  ):
    client_read, client_write = client_streams
    server_read, server_write = server_streams
    async with anyio.create_task_group() as tg:
      tg.start_soon(
          lambda: lowlevel_server.run(
              server_read,
              server_write,
              lowlevel_server.create_initialization_options(),
          )
      )
      try:
        async with ClientSession(
            read_stream=client_read, write_stream=client_write
        ) as session:
          await session.initialize()
          yield session
      finally:
        tg.cancel_scope.cancel()


class EchoAgent(BaseAgent):

  async def _run_async_impl(
      self, ctx: InvocationContext
    yield Event(
        author=self.name,
        content=types.Content(role="model", parts=[types.Part(text="ok")]),
    )


async def main():
  agent = EchoAgent(name="echo")
  runner = Runner(
      app_name="echo",
      agent=agent,
      artifact_service=InMemoryArtifactService(),
      credential_service=InMemoryCredentialService(),
  )
  server = to_mcp_server(agent, runner=runner)
  lowlevel = getattr(server, "_mcp_server", None) or getattr(
      server, "_lowlevel_server", server
  )

  for i in range(10):
    async with connected_client(lowlevel) as client:
      await client.call_tool("echo", {"request": f"call {i}"})
    gc.collect()

  remaining = await session_service.list_sessions(
      app_name="echo", user_id="mcp_user"
  )
  print(f"sessions left in the service: {len(remaining.sessions)}")


asyncio.run(main())
```

Output on main: `sessions left in the service: 10`

Also reproduced over a real transport: running the server with `run_streamable_http_async(host="127.0.0.1", port=8765, stateless_http=True)` and making each call over a fresh `streamable_http_client` connection, 9 calls leave 9 sessions in the service.

**Expected behavior**

Sessions whose connection is gone are deleted from the session service. With the fix below, the same runs leave 1 session (the most recent call's, reclaimed on the next call).

**Environment**

google-adk main (7ae1c9b), mcp 2.2.0 (also reproduces on 1.24.0 and 1.26.0), Python 3.12.

**Proposed fix (PR to follow)**

Record the id of every session entered into the connection map, and at the start of each tool call delete the sessions that are no longer reachable through the weak map. Reaping runs lazily from the tool call rather than from a GC finalizer, because finalizers can fire without a running event loop. This adds no TTL or retention policy of its own: connection lifetime is already governed by the MCP transport layer (client disconnects, the SDK's idle session timeout, stateless per request teardown), and the fix makes ADK sessions follow that lifetime instead of outliving it.

One design note for review: with a caller supplied `Runner` backed by a persistent session service, deleting orphaned sessions is a behavior change, since conversations would no longer remain readable after their connection dies. I think cleanup is the right default here, because the module creates these sessions under its internal `mcp_user` id and unbounded growth is the worse failure, but I am happy to add an opt-out flag if retention matters for some deployments.

**Possible follow-ups, feedback welcome**

1. A `stateless=True` option on `to_mcp_server` that skips the connection map and deletes each session eagerly after its call, as explicit support for deployments that run the server with `stateless_http=True`.
2. An optional conversation id argument on the generated tool, so a client of a stateless deployment can keep multi-turn conversations by passing the id back with each call. Today stateless mode silently degrades to one single turn conversation per call, because there is no connection identity to thread a conversation on.

I have the first ready and would build the second if there is interest.

harshal-96:
Fixes #7154

**Description**

`to_mcp_server` keeps one ADK session per MCP connection in a `weakref.WeakKeyDictionary`. When a connection is garbage collected the map entry disappears, but the ADK session it pointed to stays in the session service forever, with its full event history. A long running server therefore accumulates one dead conversation per closed connection, and a stateless streamable HTTP deployment, where the MCP SDK builds a fresh transport for every request, leaks one session per tool call.

This PR tracks the id of every session entered into the connection map and, at the start of each tool call, deletes the sessions whose connection is no longer reachable through the weak map.

Design notes:

- Reaping runs lazily from the tool call rather than from a GC finalizer, because finalizers can fire without a running event loop.
- No timer, TTL, or policy surface is added. Connection lifetime is already governed by the MCP transport layer (client disconnects, the SDK's idle session timeout, stateless per request teardown); the fix makes ADK sessions follow that lifetime instead of outliving it. A session in use by an in-flight call is always strongly referenced through the call context, so an active conversation can never be reaped.
- A `delete_session` failure is logged and retried on a later call, and never fails the live tool call. This matters for caller supplied runners with database or Vertex backed session services.
- Behavior change to be aware of: with a persistent session service, orphaned conversations were previously retained forever and are now deleted. I believe cleanup is the right default, since the module creates these sessions under its internal `mcp_user` id, but I can add an opt-out flag if retention is wanted for some deployments.
- The `to_mcp_server` docstring now documents stateless streamable HTTP behavior: each call is a fresh single turn conversation whose session is reclaimed.

**Testing plan**

5 new unit tests in `tests/unittests/tools/mcp_tool/test_agent_to_mcp.py`:

- `test_reap_deletes_only_sessions_no_longer_reachable`: the reap deletes exactly the unreachable ids.
- `test_session_of_a_collected_connection_is_reaped`: a conversation does not outlive its connection.
- `test_per_request_connections_do_not_accumulate_sessions`: the stateless per request pattern stays flat instead of leaking one session per call.
- `test_reap_failure_does_not_raise_and_is_retried`: a session service outage does not fail the live tool call, and the orphan is deleted once the service recovers.
- `test_call_tool_reaps_conversation_of_closed_connection`: end to end through a real in-memory MCP client and server; this test fails on main (nothing is ever deleted) and passes with the fix.

Results:

- `pytest tests/unittests/tools/mcp_tool/test_agent_to_mcp.py`: 18 passed on each of mcp 1.24.0, 1.26.0, and 2.2.0 (the pin admits 1.x and 2.x).
- `pytest tests/unittests/tools/mcp_tool`: 376 passed.
- The three GC sensitive tests repeated 20 times in a loop: stable.
- Real transport check: server run with `run_streamable_http_async(stateless_http=True)` on localhost, one fresh `streamable_http_client` connection per call. 9 calls leave 9 sessions in the service on main and 1 with this PR (the most recent call's session, reclaimed on the next call).
- Repro script from the issue: `sessions left in the service: 10` on main, `1` with this PR.

Formatting: pyink 25.12 and isort 8.0.1 (the pinned versions) produce no changes.


harshal-96:
@googlebot I signed it!

PR #7174, opened 2026-09-18
feat(agent_registry): support published_skills accessor and project alias

### LINKED ISSUE #7137: feat(agent_registry): the Console code snippet uses AgentRegistry(project=...).published_skills.get(name=...), which does not exist in ADK

The Google Cloud Console (Agent Registry > Skill Registry > skill detail page) shows a Python code snippet for using a published skill with ADK. The same snippet is also shipped inside every published skill archive as `COPYTOREPOTOUSE.txt`. It looks like this (project/skill from our catalog):

```python
from google.adk.integrations.agent_registry import AgentRegistry
from google.adk.tools import skill_toolset

PROJECT_ID = "my-project"
LOCATION = "global"
SKILL_NAME = "projects/my-project/locations/global/skills/cloud.google.com-google-cloud-networking-observability"

agent_registry = AgentRegistry(project=PROJECT_ID, location=LOCATION)
skill = agent_registry.published_skills.get(name=SKILL_NAME)
skill_tools = skill_toolset.SkillToolset(skills=[skill])

root_agent = LlmAgent(
    model="gemini-3.6-flash",
    name="sample_agent",
    instruction="Use the tools you have access to help answer user questions",
    tools=[skill_tools],
)
```

This snippet cannot run on any released `google-adk` (checked 2.9.1 and `main` @ 3a06a862):

- `AgentRegistry.__init__` takes `project_id`, `location`, `header_provider`; `project=` raises `TypeError`.
- `AgentRegistry` has no `published_skills` attribute. Nothing named `published_skills` exists in the package, the tests, the samples, or the git history of any branch. `AgentRegistry` only covers MCP servers, A2A agents and model endpoints; the adk.dev Agent Registry page documents the same surface and does not mention skills.
- Skills are handled by a separate class, `google.adk.integrations.skill_registry.GCPSkillRegistry`, whose `get_skill()` is `async`, takes a bare skill id rather than a full resource name, and currently rejects every Google-published skill id (#7136) and fails on the media-download redirect (#6908 / #6824).

So a developer who copies the snippet from the Console gets a `TypeError` on the first line, and the nearest working alternative does not work for Google-published skills either. We hit this the day Skill Registry launched and it is still the case today.

Make the Console snippet true, i.e. add a synchronous, resource-name based skill accessor on `AgentRegistry` that returns a `google.adk.skills.models.Skill` ready to be passed to `SkillToolset(skills=[...])`:

```python
registry = AgentRegistry(project_id="my-project", location="global")
skill = registry.published_skills.get(
    name="projects/my-project/locations/global/skills/cloud.google.com-google-cloud-networking-observability"
)
SkillToolset(skills=[skill])
```

Suggested semantics:

- `published_skills.get(name=<full resource name>)` fetches the skill metadata, follows the `?alt=media` redirect for the default revision, and returns a `Skill` (same loader as `GCPSkillRegistry`). A `list(...)` / `search(query=...)` pair on the same namespace would mirror the existing `list_mcp_servers` / `search_mcp_servers` / `list_agents` / `search_agents` methods.
- Accepting the full resource name sidesteps the id-validation problem in #7136: the id is not interpolated into a URL by the client, so only "is this a well-formed `projects/*/locations/*/skills/*` name" needs checking.
- Synchronous, like the rest of `AgentRegistry`, so it can be used at module import time in `agent.py` (the snippet's use case). `GCPSkillRegistry` stays the async registry for `SkillToolset(registry=...)`.
- Either accept `project=` as an alias of `project_id`, or fix the snippet; whichever is easier, but today the two disagree.

### Impact on your work

We build agents on Agent Engine that consume Google-published skills from Agent Registry. The Console tells users to write code that does not exist, and the alternative path is blocked by two other bugs, so there is currently no way to use a Google-published skill from ADK at all. Not blocking a launch, but it has been confusing since the Skill Registry launch and it is the first thing a new user of the feature runs into.

### Willingness to contribute

Yes, I can send a PR for the `AgentRegistry` side once the maintainers confirm the intended shape (the Console/template side is presumably owned by another team; happy to have this routed there as well).

---

- `GCPSkillRegistry.get_skill(name=<bare id>)` + `SkillToolset(skills=[skill])`: async, so it needs `asyncio.run` at import (which fails under `adk run` / `adk web` because the module is imported inside a running loop), and it is blocked by #7136 and #6824 for Google-published skills.
- `SkillToolset(registry=GCPSkillRegistry(...))` with `search_skills` at runtime: Google-published skills are filtered out of the results (#7136), and it costs an extra model turn per conversation (#7130 / #7135).

### Additional Context

- `google-cloud-agentregistry` 0.1.0 on PyPI (2026-07-08) is an empty placeholder package (no services), so there is no other public SDK that provides `published_skills` either.
- Related: #7136 (dotted ids rejected), #6908 / #6824 (redirect), #7130 / #7135 (pinning registry skills).

ArulJerald:
- Closes: #7137
**Problem:**
Google Cloud Console code snippets for Agent Registry instruct users to initialize and fetch published skills using:
```python
registry = AgentRegistry(project="my-project", location="global")
skill = registry.published_skills.get(name="projects/.../locations/.../skills/...")
```
In ADK, `AgentRegistry.__init__` previously only accepted `project_id`, resulting in a `TypeError: unexpected keyword argument 'project'`. Additionally, `AgentRegistry` lacked a `published_skills` accessor and had no method to fetch, download, and deserialize published skills from the registry into `Skill` models for use with `SkillToolset`.
**Solution:**
1. Added `project: str | None = None` parameter to `AgentRegistry.__init__` as an alias for `project_id`, along with a `@property def project(self) -> str | None`.
2. Created `PublishedSkills` accessor exposed via `registry.published_skills` with synchronous `get(name: str) -> Skill` (and added a top-level `get_published_skill` convenience method).
3. Validated full resource names matching `projects/{project}/locations/{location}/skills/{skill_id}`.
4. Resolved the skill's default revision and downloaded the media archive using `AuthorizedSession` (handling 302/307 redirects to media URLs).
5. Loaded the zip archive into a `google.adk.skills.models.Skill` instance ready to pass to `SkillToolset(skills=[...])`.
### Testing Plan
**Unit Tests:**
Summary of passing `pytest` results:
- `uv run pytest tests/unittests/integrations/agent_registry/test_agent_registry.py`: **72 passed in 1.96s**
- Added test coverage for:
  - Initializing `AgentRegistry(project="...", location="...")` and `project_id` precedence.
  - Calling `registry.published_skills.get(name=...)` with mock metadata and media responses.
  - Calling with positional arguments and via `registry.get_published_skill(...)`.
  - Following HTTP 302 media redirects to signed Cloud Storage download URLs.
  - Passing returned `Skill` directly to `SkillToolset(skills=[skill])`.
  - Raising errors on invalid resource names, missing default revision, and HTTP errors (404/500).
N/A - verified via unit tests with mocked API endpoints and `AuthorizedSession` behavior.
### Checklist


src/google/adk/integrations/agent_registry/agent_registry.py
```diff
@@ -172,6 +174,38 @@ def _is_google_api(url: str) -> bool:
   )
 
 
+_SKILL_RESOURCE_NAME_PATTERN = re.compile(
```
codebee-aoki, line 177: `[^/]+` accepts `..`, `?` and `%`. Against the real service, `.../skills/..` sends an authorized GET to `/v1alpha/projects/<p>/locations/global/` (200), and `.../skills/x?alt=media` injects the query parameter into the metadata call. Please validate each segment before any request, e.g. with the safe-id rule from #7138 (`^[a-z0-9]+(?:[._-][a-z0-9]+)*$`, max 256).

src/google/adk/integrations/agent_registry/agent_registry.py
```diff
@@ -689,6 +738,98 @@ def get_remote_a2a_agent(
...
+        {"x-goog-user-project": quota_project_id} if quota_project_id else {}
+    )
+    try:
+      response = self._session.get(
+          url,
+          headers=headers,
+          params=params,
+          allow_redirects=True,
+      )
+      if 300 <= response.status_code < 400 and (
```
codebee-aoki, line 778: `allow_redirects=True` already follows redirects, and `requests` drops `Authorization` when the host changes. This fallback re-issues the request through `AuthorizedSession`, which attaches the bearer token to whatever `Location` says (the new redirect test exercises exactly this with storage.googleapis.com). In the real run the redirect was same-host (`/download/v1alpha/...`) and this branch was never reached. Please remove it.

src/google/adk/integrations/agent_registry/agent_registry.py
```diff
@@ -689,6 +738,98 @@ def get_remote_a2a_agent(
...
+      )
+
+    skill_data = self._make_request(name)
+    default_revision = skill_data.get("defaultRevision") or skill_data.get(
+        "default_revision"
+    )
+    if not default_revision:
+      raise ValueError(f"Skill '{name}' does not contain default revision.")
+
+    if default_revision.startswith("http://") or default_revision.startswith(
```
codebee-aoki, line 817: The API returns `defaultRevision` as `projects/...`. Sending an authorized request to an arbitrary absolute URL taken from a response body is unnecessary; keeping only the `projects/` form (as `GCPSkillRegistry` does) is simpler and safer.

src/google/adk/integrations/agent_registry/agent_registry.py
```diff
@@ -189,20 +223,24 @@ def __init__(
...
     """Initializes the AgentRegistry client.
 
     Args:
       project_id: The Google Cloud project ID.
       location: The Google Cloud location (region).
       header_provider: Optional provider for custom headers.
+      project: Optional alias for project_id.
     """
-    self.project_id = project_id
+    self.project_id = project_id or project
```
codebee-aoki, line 236: If both `project_id` and `project` are given and differ, this silently picks `project_id`. Please raise `ValueError` instead.

src/google/adk/integrations/agent_registry/agent_registry.py
```diff
@@ -172,6 +174,38 @@ def _is_google_api(url: str) -> bool:
...
+
+    Raises:
+      ValueError: If the skill name does not match the expected resource name
+        format, or the skill does not contain a default revision.
+      RuntimeError: If an API request to fetch metadata or media fails.
+    """
+    return self._registry._fetch_published_skill_sync(name)
+
+
+_PublishedSkillsAccessor = PublishedSkills
```
codebee-aoki, line 206: `_PublishedSkillsAccessor` is unused.

codebee-aoki (commented):
Thanks @ArulJerald. The API shape matches what #7137 asks for: synchronous, takes the full resource name, returns a `Skill` that drops into `SkillToolset(skills=[...])`, and `project=` is accepted. 72 unit tests pass locally. Unfortunately it does not work against the real service yet.

**Blocker: wrong API version.** Running the Console snippet verbatim against a real catalog fails with HTTP 404 on `https://agentregistry.googleapis.com/v1/projects/.../skills/...`. `AgentRegistry` uses the `/v1` base, but skills only exist in `v1alpha`: the public discovery document for `v1` lists agents, aiApplications, bindings, endpoints, mcpServers, operations and services under locations, while `v1alpha` adds `publishers` and `skills` (`GCPSkillRegistry` already uses `v1alpha`). With only the base URL switched to `v1alpha`, the snippet runs end to end in about 3 seconds and the Google-published skill `cloud.google.com-google-cloud-networking-observability` loads and works in `SkillToolset`. The unit tests do not catch this because HTTP is mocked and no test asserts the requested URL. Suggest a dedicated `v1alpha` base for the skill calls (mirroring `GCPSkillRegistry`, including the mTLS template and the `AGENT_REGISTRY_ENDPOINT` override) and asserting the URL in the tests.

The remaining points are inline.

PR #7209, opened 2026-09-19
fix(openapi): let operation parameters override path-level ones

### LINKED ISSUE #7205: OpenAPI parser doesn't let an operation parameter override the path-level one: the model is asked for the same path value twice, and the wrong one is sent
## 🔴 Required Information

**Describe the Bug:**

`OpenApiSpecParser._collect_operations()` appends path-level `parameters` to every operation's own `parameters` without checking for overrides:

```python
operation_dict["parameters"] = operation_dict.get("parameters", []) + path_item.get("parameters", [])
```

The OpenAPI 3 Path Item Object says an operation-level parameter with the same `name` and `in` **overrides** the path-level one. Declaring the shared path parameter once at the path level and refining it on an operation (a more specific description, pattern or enum) is common. ADK keeps both, `OperationParser` de-duplicates the names, and the tool then asks the model for the **same URL segment twice** as two required arguments: `account_id` and `account_id_0`. When the model fills both, both map to the same `{accountId}` placeholder and the **path-level value wins**, so the operation's own, more specific parameter is silently dropped from the request.

**Steps to Reproduce:**
1. `pip install google-adk` (reproduced on `main` @ 3f4bb8fa).
2. Run the script below: an `OpenAPIToolset` whose `/accounts/{accountId}` path declares `accountId` and whose `GET` overrides it. HTTP goes to an `httpx.MockTransport`.

**Expected Behavior:**
One required argument, `account_id`, described by the operation-level parameter; `{"account_id": "ACC-123"}` requests `/accounts/ACC-123`.

**Observed Behavior:**
```text
model is asked for: ['account_id', 'account_id_0']
args: {'account_id': 'ACC-123', 'account_id_0': 'ACC-999'} -> request: https://crm.example.com/accounts/ACC-999
```

**Environment Details:**

- ADK Library Version (pip show google-adk): `main` @ 3f4bb8fa
- Desktop OS: Windows 11
- Python Version (python -V): 3.12.10

**Model Information:**

- Are you using LiteLLM: No
- Which model is being used: N/A (tool declaration and request building)

______________________________________________________________________

## 🟡 Optional Information

```python
import asyncio

import httpx
from google.adk.tools.openapi_tool.openapi_spec_parser.openapi_toolset import OpenAPIToolset

spec = {
    "openapi": "3.0.0",
    "info": {"title": "CRM", "version": "1"},
    "servers": [{"url": "https://crm.example.com"}],
    "paths": {"/accounts/{accountId}": {
        "parameters": [{"name": "accountId", "in": "path", "required": True,
                        "schema": {"type": "string"}, "description": "Shared account id"}],
        "get": {
            "operationId": "getAccount",
            "parameters": [{"name": "accountId", "in": "path", "required": True,
                            "schema": {"type": "string", "pattern": "^ACC-[0-9]+$"},
                            "description": "Account id, e.g. ACC-123"}],
            "responses": {"200": {"description": "ok"}},
        },
    }},
}
sent = []
toolset = OpenAPIToolset(
    spec_dict=spec,
    httpx_client_factory=lambda: httpx.AsyncClient(transport=httpx.MockTransport(
        lambda r: sent.append(str(r.url)) or httpx.Response(200, json={}))),
)


async def main():
  tool = (await toolset.get_tools())[0]
  schema = tool._get_declaration().parameters_json_schema
  print("model is asked for:", schema["required"])
  args = {"account_id": "ACC-123"}
  if "account_id_0" in schema["properties"]:
    args["account_id_0"] = "ACC-999"  # a model filling both required fields
  await tool.run_async(args=args, tool_context=None)
  print("args:", args, "-> request:", sent[0])


asyncio.run(main())
```

**Suggested fix:** when merging, skip path-level parameters whose `(name, in)` the operation already declares. Parameters with the same name but a different location are still merged. I have a PR ready with a test.

- Always (100%): any spec that overrides a path-level parameter on an operation.

1aifanatic:

- Closes: #7205

**Problem:**

`OpenApiSpecParser._collect_operations()` appended every path-level parameter to each operation's own parameters. Per the OpenAPI 3 Path Item Object, an operation-level parameter with the same `name` and `in` **overrides** the path-level one. Keeping both made the tool ask the model for the same path value twice (`account_id` and `account_id_0`, both required). Both map to the same `{accountId}` placeholder, and the path-level value won, so the operation's own, more specific parameter was dropped from the request.

**Solution:**

Skip path-level parameters whose `(name, in)` the operation already declares. Parameters with the same name but a different location are still merged. References are resolved before this step, so parameters are plain dicts here; anything without a `name` is left untouched.

### Testing Plan

**Unit Tests:**

`test_openapi_spec_parser.py::test_parse_spec_operation_parameter_overrides_path_level_parameter`: the path `accountId` is overridden and keeps the operation's description, while a same-named `header` parameter is still merged. Fails on `main`.

```text
$ pytest tests/unittests/tools/openapi_tool
342 passed

```

```text
$ pytest tests/unittests -n 8   # Python 3.12, Windows 11
42 failed, 15146 passed, 102 skipped, 27 xfailed, 2 xpassed
```

Every failure outside the new tests also fails on an unmodified `origin/main` (3f4bb8fa) on the same machine, where the same run gives 47 failed. Most are deterministic Windows-specific tests (`cli/` deploy, conformance, `scripts/check_new_py_files`, path normalization, the unsafe local code executor, the import allowlist, yaml). The remainder differ between runs: telemetry functional/node-functional schema cases, and `streaming/test_streaming_tool_events.py::test_delivered_event_carries_the_message_and_nothing_else`, which failed 1/10 in isolation on clean `main` versus 2/10 on this branch. I ran on Python 3.12 only, not the full `tox` matrix.

An `OpenAPIToolset` whose `/accounts/{accountId}` path declares `accountId` and whose `GET` overrides it; HTTP goes to an `httpx.MockTransport`. The script is in #7205.

Before (`main` @ 3f4bb8fa):

```text
model is asked for: ['account_id', 'account_id_0']
args: {'account_id': 'ACC-123', 'account_id_0': 'ACC-999'} -> request: https://crm.example.com/accounts/ACC-999
```

After:

```text
model is asked for: ['account_id']
args: {'account_id': 'ACC-123'} -> request: https://crm.example.com/accounts/ACC-123
```

### Checklist

https://claude.ai/code/session_013vXxD1ga1hnCq2uFRwNks7


1aifanatic:
@llalitkumarrr thanks for testing, and good catch on that edge case (from #7205).

For a **path** parameter the flag can't legitimately be false. The OpenAPI spec says `required` "MUST be true" when `in: path`, and the URL can't be built without the value. The parser took the flag as written, though, so a spec that leaves it out declared the path parameter optional. That includes an operation-level override like the one you describe. This affects `main` independently of this PR: if the model omits such a parameter, the call fails with `KeyError: 'accountId'` while formatting the URL.

I've pushed da60abc3 so that `in: path` parameters are always treated as required, whatever the flag says. For query, header and cookie parameters, the operation-level definition replaces the path-level one, as the spec defines. An override replaces the whole parameter object and `required` defaults to false, so an override that makes one of those optional is honored.

With the override leaving out `required`, the parameters the model is asked for:

```text
main:                     account_id (optional) + account_id_0 (required)
this PR before da60abc3:  account_id (optional)
this PR now:              account_id (required)
```

New test: `test_parse_spec_path_parameter_is_required_even_when_override_omits_it`. `pytest tests/unittests/tools` gives 2344 passed.

PR #7210, opened 2026-09-19
fix(openapi): decode error bodies with the declared charset

### LINKED ISSUE #7206: RestApiTool crashes the agent run on a non-UTF-8 error response body (UnicodeDecodeError in the HTTPStatusError handler)
## 🔴 Required Information

**Describe the Bug:**

When an API call fails, `RestApiTool.call()` builds the error message for the model with:

```python
except httpx.HTTPStatusError:
  error_details = response.content.decode("utf-8")
```

That forces UTF-8 regardless of the charset the server declares. A non-UTF-8 error body raises `UnicodeDecodeError`, for example a Latin-1/Windows-1252 page from an older ERP or IIS server, a message with accented characters, or a binary body. The exception is raised **inside the `except` handler**, so the sibling `except ValueError` can't catch it. It propagates out of the tool and **ends the whole agent run**, so the model never sees the HTTP error it could have recovered from.

**Steps to Reproduce:**
1. `pip install google-adk` (reproduced on `main` @ 3f4bb8fa).
2. Run the script below: a real `Runner` plus `LlmAgent` plus `OpenAPIToolset`, with a mocked model and an `httpx.MockTransport` backend that answers `404` with a `text/plain; charset=iso-8859-1` body.

**Expected Behavior:**
The tool returns its usual error dict, decoded with the declared charset:
`... Execution Error: Status Code: 404, Commande introuvable : échec`

**Observed Behavior:**
```text
RUN FAILED: UnicodeDecodeError - 'utf-8' codec can't decode byte 0xe9 in position 23: invalid continuation byte
```

**Environment Details:**

- ADK Library Version (pip show google-adk): `main` @ 3f4bb8fa
- Desktop OS: Windows 11
- Python Version (python -V): 3.12.10

**Model Information:**

- Are you using LiteLLM: No
- Which model is being used: N/A (mocked model; the bug is in response handling)

______________________________________________________________________

## 🟡 Optional Information

```python
import asyncio

import httpx
from google.adk.models.base_llm import BaseLlm
from google.adk.tools.openapi_tool.openapi_spec_parser.openapi_toolset import OpenAPIToolset

spec = {
    "openapi": "3.0.0",
    "info": {"title": "Legacy ERP", "version": "1"},
    "servers": [{"url": "https://erp.example.com"}],
    "paths": {"/orders/{id}": {"get": {
        "operationId": "getOrder",
        "parameters": [{"name": "id", "in": "path", "required": True, "schema": {"type": "string"}}],
        "responses": {"200": {"description": "ok"}},
    }}},
}


def handler(request: httpx.Request):
  # A legacy server answering in Windows-1252 / Latin-1, as it declares.
  return httpx.Response(
      404,
      content="Commande introuvable : échec".encode("latin-1"),
      headers={"content-type": "text/plain; charset=iso-8859-1"},
  )


class FakeModel(BaseLlm):
  model: str = "fake"
  calls: int = 0

  async def generate_content_async(self, llm_request, stream=False):
    self.calls += 1
    part = (types.Part.from_function_call(name="get_order", args={"id": "A-42"})
            if self.calls == 1 else types.Part.from_text(text="done"))
    yield LlmResponse(content=types.Content(role="model", parts=[part]))


agent = LlmAgent(
    name="erp_agent",
    model=FakeModel(),
    tools=[OpenAPIToolset(
        spec_dict=spec,
        httpx_client_factory=lambda: httpx.AsyncClient(transport=httpx.MockTransport(handler)),
    )],
)


async def main():
  runner = InMemoryRunner(agent=agent)
  session = await runner.session_service.create_session(app_name=runner.app_name, user_id="u")
  try:
        user_id="u", session_id=session.id,
        new_message=types.Content(role="user", parts=[types.Part(text="order A-42?")])):
      for fr in event.get_function_responses():
        print("function_response:", fr.response)
  except Exception as e:
    print("RUN FAILED:", type(e).__name__, "-", e)


asyncio.run(main())
```

**Suggested fix:** use `response.text`, which is what the non-JSON success path a few lines below already uses. It decodes with the response's declared charset and replaces undecodable bytes instead of raising. I have a PR ready with a test.

- Always (100%): any non-UTF-8 error body.

1aifanatic:

- Closes: #7206

**Problem:**

On a failed call, `RestApiTool.call()` built the model-facing error with `response.content.decode("utf-8")`. A non-UTF-8 error body, such as a Latin-1 page from a legacy server or a binary body, raised `UnicodeDecodeError` **inside the `except httpx.HTTPStatusError` handler**, where the sibling `except ValueError` can't catch it. The whole agent run then ended instead of the model getting the HTTP error.

**Solution:**

Use `response.text`, which is what the non-JSON success path a few lines below already does. It decodes with the response's declared charset and replaces undecodable bytes instead of raising.

### Testing Plan

**Unit Tests:**

- New `test_call_http_failure_decodes_body_with_declared_charset`: a real `httpx.Response(404)` with a `charset=iso-8859-1` body is reported as `Status Code: 404, Commande introuvable : échec`. Fails on `main` with `UnicodeDecodeError`.
- `test_call_http_failure` now sets the mock's `.text` instead of `.content`, because that's the attribute the tool reads.

```text
$ pytest tests/unittests/tools/openapi_tool tests/unittests/tools/application_integration_tool     tests/unittests/tools/apihub_tool tests/unittests/tools/google_api_tool
531 passed

```

```text
$ pytest tests/unittests -n 8   # Python 3.12, Windows 11
45 failed, 15143 passed, 102 skipped, 27 xfailed, 2 xpassed
```

Every failure also fails on an unmodified `origin/main` (3f4bb8fa) on the same machine, where the same run gives 47 failed. Most are deterministic Windows-specific tests (`cli/` deploy, conformance, `scripts/check_new_py_files`, path normalization, the unsafe local code executor, the import allowlist, yaml). The rest are skill-telemetry schema cases in `telemetry/test_functional.py` that fail intermittently on clean `main` here: rerun 3 times each, they failed 1, 1, 1 times on `main` and 1, 1, 3 times on this branch. I ran on Python 3.12 only, not the full `tox` matrix.

A real `Runner` + `LlmAgent` + `OpenAPIToolset` with a mocked model and an `httpx.MockTransport` backend that answers `404` with a Latin-1 body. The script is in #7206.

Before (`main` @ 3f4bb8fa):

```text
RUN FAILED: UnicodeDecodeError - 'utf-8' codec can't decode byte 0xe9 in position 23: invalid continuation byte
```

After:

```text
function_response: {'error': "Tool get_order execution failed. ... Execution Error: Status Code: 404, Commande introuvable : échec"}
```

### Checklist

https://claude.ai/code/session_013vXxD1ga1hnCq2uFRwNks7


1aifanatic:
Note for reviewers: #7207 makes the same one-line change (`response.text`) with an equivalent test. This PR was prepared alongside the report in #7206, and the reproduction script there is the one I used. The only difference is that this PR switches the existing `test_call_http_failure` mock to `.text` rather than setting both attributes. Either PR resolves #7206, so go with whichever you prefer.

